PR #15, opened 2018-02-22
Update generated code

font:


marun:
This is essentially a no-op, since the apiserver-builder controller framework isn't useful for created federated controllers.

PR #22, opened 2018-03-05
Add federated replicaset

marun:
cc: @font


marun:
ReplicaSet creation isn't correctly setting the replica field - as per the following gist:

https://gist.github.com/marun/5fc3473d817d62b41d4149fc57832ca0

PR #27, opened 2018-03-22
Add support for federated kubernetes Namespace

font:


test/e2e/crud.go
```diff
@@ -26,38 +26,47 @@ import (
...
-				clusterClients := f.ClusterClients(userAgent)
-				crudTester := common.NewFederatedTypeCrudTester(framework.NewE2ELogger(), adapter, clusterClients, framework.PollInterval, framework.SingleCallTimeout)
-				clusterNames := []string{}
-				for name, _ := range clusterClients {
-					clusterNames = append(clusterNames, name)
-				}
-				template, placement, override := federatedtypes.NewTestObjects(fedTypeConfig.Kind, f.TestNamespaceName(), clusterNames)
-
-				crudTester.CheckLifecycle(template, placement, override)
+func addE2ECrudTests() {
```
marun: I think it would be preferable to skip within the It() clause.  Likely worth porting over the [skip code](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/util.go#L293) from k/k (to test/e2e/framework/util.go).

test/common/crudtester.go
```diff
@@ -207,11 +207,20 @@ func (c *FederatedTypeCrudTester) CheckPlacementChange(template, placement, over
 
 	clusterNames := adapter.ClusterNames(placement)
 
+	if kind == federatedtypes.FederatedNamespacePlacementKind && len(clusterNames) == 1 {
```
marun: I don't think it is safe to assume that the primary cluster will be the host cluster.  A target deployment may well not join the host cluster.  I'm not saying you have to fix it in this PR, but I think a TODO comment is warranted so that determination of whether a member cluster is the host cluster is eventually implemented.

test/e2e/crud.go
font: You're right, this makes it much cleaner! But first I had to fix a stale bind issue with the `kind` closure variable.

marun:
Nice work!

marun:
Oh, looks like a rebase is required (due to merging #31).

marun:
oops, sorry.  merged #30.  Rebase again.  

Looking forward to not having to commit generated code.

font:
@marun rebased.

PR #28, opened 2018-03-22
minor fix in federated configmap resource

sdminonne:
@marun @font 
Not sure whether or not I've to re-generate something...


marun:
@sdminonne Good catch!  After making changes to an API type, it is necessary to run:

    apiserver-boot build generated

Please isolate the results of code generation in a separate commit (e.g. 'Updated generated code').

sdminonne:
going to do

marun:
lgtm, thank you!

PR #36, opened 2018-03-30
Document E2E tests

font:
cc @marun @onyiny-ang @sdminonne @pmorie Feel free to give this a try. It is still WIP but I wanted to get this out there.

---
This change is [Reviewable](https://reviewable.io/reviews/marun/federation-v2/36)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## Kubernetes Federation v2
+# Kubernetes Federation v2
 
 This repo contains an in-progress prototype of some of the
 foundational aspects of V2 of Kubernetes Federation.  The prototype
@@ -14,7 +14,7 @@ group](https://groups.google.com/forum/#!forum/kubernetes-sig-multicluster).
 
 <p align="center"><img src="docs/images/propagation.png" width="711"></p>
 
-## Concepts
+# Concepts
 
 The following abstractions support the propagation of a logical
 federated type:
@@ -30,16 +30,16 @@ well-suited to serve as the glue between any given propagation
 mechanism and higher-order behaviors like policy-based placement and
 dynamic scheduling.
 
-## Getting started
+# Getting started
 
-### Required: `apiserver-builder`
+## Required: `apiserver-builder`
 This repo depends on
 [apiserver-builder](https://github.com/kubernetes-incubator/apiserver-builder)
 to generate code and build binaries.  Download a [recent
 release](https://github.com/kubernetes-incubator/apiserver-builder/releases)
 and install it in your `PATH`.
 
-### Adding a new API type
+## Adding a new API type
 
 As per the
 [docs](https://github.com/kubernetes-incubator/apiserver-builder/blob/master/docs/tools_user_guide.md#create-an-api-resource)
@@ -66,7 +66,7 @@ The generated code will need to be updated whenever the code for a
 type is modified. Care should be taken to separate generated from
 non-generated code in the commit history.
 
-### Enabling federation of a type
+## Enabling federation of a type
 
 Implementing support for federation of a Kubernetes type requires
 the following steps:
@@ -92,9 +92,9 @@ the following steps:
      propagated by the reconciler to member clusters.
    - e.g. [FederatedSecretAdapter](https://github.com/marun/federation-v2/blob/master/pkg/federatedtypes/secret.go)
 
-### Testing
+## Testing
 
-#### Integration
+### Integration
 
 The integration tests will spin up a federation consisting of kube
 api + cluster registry api + federation api + 2 member clusters and
@@ -122,7 +122,7 @@ the components involved in the test:
 cd test/integration && dlv test -- -test.run ^TestCrud$
 ``
 
-#### E2E
+### E2E
 
 The Federation-v2 E2E tests can run in an *unmanaged*, *managed*, or *hybrid*
 modes. For both unmanaged and hybrid modes, you will need to bring your own
@@ -139,7 +139,7 @@ verify that:
 
 The read operation is implicit.
 
-##### Managed
+#### Managed
 
 The E2E managed tests will spin up the same environment as the
 [Integration](README.md#integration) tests described above and run [CRUD
@@ -173,7 +173,7 @@ cd test/e2e
 dlv test -- -v=4 -test.v --ginkgo.focus='"FederatedSecret"'
 ```
 
-##### Unmanaged and Hybrid Cluster Setup
+#### Unmanaged and Hybrid Cluster Setup
 
 The difference between unmanaged and hybrid is that with hybrid, you run
 the federation-v2 controllers in-process as part of executing the test. This
@@ -181,7 +181,7 @@ helps in running a debugger to debug anything in the controllers. On the other
 hand with unmanaged, the controllers are already running in the K8s cluster.
 Both methods require the clusters to have already been joined.
 
-###### Create Clusters
+##### Create Clusters
 
 The quickest way to set up clusters for use with unmanaged and hybrid modes is
 to use [minikube](https://kubernetes.io/docs/getting-started-guides/minikube/).
@@ -208,7 +208,7 @@ kubectl get all --all-namespaces
 
 Once all pods are running you can move on to deploy the cluster registry.
 
-###### Deploy the Cluster Registry
+##### Deploy the Cluster Registry
 
 Make sure the storage provisioner is ready before deploying the Cluster
 Registry.
@@ -227,7 +227,7 @@ crinit aggregated init mycr --host-cluster-context=clusterA
 
 You can also specify your own cluster registry image using the `--image` flag.
 
-###### Deploy Federation
+##### Deploy Federation
 
 First you'll need to create the namespace to deploy the federation into:
 
@@ -280,7 +280,7 @@ kubectl get federatedcluster
 
 It should successfully report that no resources are found.
 
-###### Join Clusters
+##### Join Clusters
 
 First, make sure the federation deployment succeeded by verifying it using the
 step above. Next, you'll want to use the `kubefnord` tool to join all your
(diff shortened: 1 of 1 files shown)


sdminonne:
@font, giving a try a let you know.

marun:
lgtm, thank you!  I'll wait for @sdminonne's feedback before merging.

onyiny-ang:
The Managed setup test skips "Namespaces" when I run it. I'm not sure if this is desired behaviour. It also fails to test Namespaces when it is specified with `gingko.focus='"FederatedNamespace"'.
That's the only set up I tried so far.
Otherwise, looks good and the write up is easy to understand.

font:
@onyiny-ang thanks for trying things out! The parameter to `ginkgo.focus` is `Namespace` as we do not have a `FederatedNamespace` type resource. In any case, the managed setup test skips `Namespace` intentionally, as it is not supported in that environment yet due to a lack of a k8s namespace controller.

marun:
@font Please remove reference to fnord from the proposed changes.

sdminonne:
@font @marun  sorry for the delay on this.
LGTM

README.md
```diff
@@ -120,3 +120,152 @@ the components involved in the test:
 ``
 cd test/integration && dlv test -- -test.run ^TestCrud$
 ``
+
+#### E2E
+
+The fnord E2E tests can run in an unmanaged, managed, or hybrid modes. For both
```
pmorie: We should probably change this to federation-v2

README.md
```diff
@@ -120,3 +120,152 @@ the components involved in the test:
...
+unmanaged and hybrid modes, you will need to bring your own clusters (BYOC).
+The managed mode runs similarly to integration as it uses the same test fixture
+setup.
+
+##### Managed
+
+The E2E managed tests will spin up the same environment as the
+[Integration](README.md#integration) tests described above and run [CRUD
```
pmorie: It would be helpful to expand a tiny bit on what CRUD means here - do the tests ensure that resources are put into a target cluster?

README.md
```diff
@@ -120,3 +120,152 @@ the components involved in the test:
...
+    ```
+    cd test/e2e
+    go test -c -v
+    ./e2e.test -v=4 -kubeconfig=/path/to/kubeconfig -test.v --ginkgo.focus='"FederatedSecret" resources'
+    ```
+
+You can repeat these steps to join any additional clusters.
+
+###### Hybrid
+
```
pmorie: Can you explain what is meant by "hybrid" here?

README.md
font: Fixed.

README.md
font: Yup! This was still WIP so did not elaborate and instead just provided the commands so others could get started.

README.md
font: Done.

font:
@marun This is ready to be re-reviewed/merged now that the other PRs are merged.

font:
The testing issue has been fixed by #55.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubernetes-sigs/federation-v2/36#-:-LA_J4kf1MdoiOOVYvu-:b-8nbshn)*

README.md
```diff
@@ -122,6 +122,239 @@ the components involved in the test:
...
+Both methods require the clusters to have already been joined.
+
+###### Create Clusters
+
+The quickest way to set up clusters for use with unmanaged and hybrid modes is
+to use [minikube](https://kubernetes.io/docs/getting-started-guides/minikube/).
+Once you have minikube installed run:
+
+```bash
+minikube start -p clusterA
```
marun: I ran into trouble specifying a profile, as per the following issue:

https://github.com/kubernetes/minikube/issues/2717

Consider removing use of a profile until this is fixed, and maybe documenting why profiles are not recommended at this time.

README.md
```diff
@@ -122,6 +122,239 @@ the components involved in the test:
...
+```
+
+###### Deploy the Cluster Registry
+
+Get the latest version of the [Cluster
+Registry](https://github.com/kubernetes/cluster-registry/releases) and run the
+following:
+
+```bash
+crinit aggregated init mycr --host-cluster-context=clusterA
```
marun: I ran into an issue with `crinit` when I tried to run it too soon after minikube initialization.  If the storage provisioner hadn't been started before `crinit` was invoked, the etcd pvc would never be bound.  It may be worth documenting this and/or providing a sample check similar to what my deployment script uses:

https://github.com/kubernetes-sigs/federation-v2/pull/57/files#diff-1ed12d3b5689c7a45b1c21e93e23e3d3R53

README.md
```diff
@@ -122,6 +122,239 @@ the components involved in the test:
...
+
+###### Deploy Federation
+
+First you'll need to create the namespace to deploy the federation into:
+
+```bash
+kubectl create ns federation
+```
+
+Then run `apiserver-boot` to deploy. This will build the federated API server
```
marun: Consider documenting the requirement to configure login to the docker hub:

https://github.com/kubernetes-sigs/federation-v2/pull/57/files#diff-1ed12d3b5689c7a45b1c21e93e23e3d3R40

Also, have you tested the deployed controller manager?  I needed to create a permissive role binding or the controller manager wouldn't be able to perform cluster list/watch and the apiserver reported errors related to webhook configuration:

https://github.com/kubernetes-sigs/federation-v2/pull/57/files#diff-1ed12d3b5689c7a45b1c21e93e23e3d3R64

README.md
```diff
@@ -122,6 +122,239 @@ the components involved in the test:
...
+    --image <containerregistry>/<username>/<imagename>:<tagname> \
+    --controller-args="-logtostderr,-v=4"
+```
+
+You will most likely need to increase the limit on the amount of memory
+required by the API server as there isn't a way to request it via
+`apiserver-boot` at the moment. Otherwise the federation pod may be terminated
+with `OOMKilled`. In order to do that you will need to modify the deployment:
+
+```bash
```
marun: Consider supplying a 'kubectl patch` invocation instead: 

https://github.com/kubernetes-sigs/federation-v2/pull/57/files#diff-1ed12d3b5689c7a45b1c21e93e23e3d3R69

README.md
```diff
@@ -122,6 +122,239 @@ the components involved in the test:
...
+            cpu: 100m
+            memory: 128Mi
+          requests:
+            cpu: 100m
+            memory: 64Mi
+```
+
+###### Join Clusters
+
+Next, join all your clusters that you want to test against using the `kubefnord` tool.
```
marun: I also ran into an issue trying to join too soon after deploying with `apiserver-builder` - join would fail if the apiserver had not yet come online.  Consider documenting how to verify that the apiserver is up and aggregating properly:

https://github.com/kubernetes-sigs/federation-v2/pull/57/files#diff-1ed12d3b5689c7a45b1c21e93e23e3d3R71

README.md
font: Ah yes, I did as well but the solution I went for was to downgrade my version of `minikube` because specifying a profile is necessary to work with multiple clusters. Have you tried using `v0.25.2`? The issue with profiles should probably be captured, but if we're going to remove the profile from the command, I think a comment also needs to be captured when using multiple clusters. I'm still not sure whether we should remove `-p` from the command as that's a bug in the latest version of minikube anyway.

README.md
font: Yes, that can happen. Personally, I think the minikube command should not return until all the components are up and ready, or timeout saying they have not properly come up. That may already be an option.

README.md
font: I don't necessarily want to be prescriptive to docker hub e.g. I use google's container registry. But I think a generic "login" comment perhaps with docker as an example is useful.

Yes, I have tested the deployed controller manager in `minikube` version `v0.25.2`, but did not run into any permission issues. Interestingly enough though, during my testing of this deployment in a public cloud environment on Friday, both the `apiserver` and `controller` fail due to a lack of various RBAC permissions that were not encountered with `minikube`. Basically, `apiserver-boot` does not generate a deployment config that includes all the necessary RBAC objects for an aggregated API server.

I think that once we have an officially released alpha version of federation-v2 with a container image, we should commit the `apiserver.yaml` config with all the necessary pieces: pointing to the released container image, the memory limit update, and necessary RBAC objects. I think there is even benefit to committing it now with the expectation that users update the container image they are using.

README.md
font: Yes, thanks! I knew there'd be a way to do that but I wasn't familiar with it and needed to just capture the issue and move on. Similarly, we should include a patch command to delete the controller in the hybrid setup too. Ideally, I think this should just be part of the deployment config committed to the tree (see comment above).

README.md
font: Yep, probably worth a comment about that, especially if one is scripting this.

README.md
marun: If recommending the use of profiles, then, please recommend the version of minikube that is compatible with its use.

README.md
marun: It is not an option afaict.

pmorie:
I think we should merge this so that the information is there and iterate on it in a follow up, unless this PR has info that's no longer accurate. Once this is merged we can all iterate on the doc to make it better.

README.md
font: Done.

README.md
font: Done.

README.md
font: Done.

README.md
font: Done.

README.md
font: Done.

marun:
/lgtm
Thank you!

PR #37, opened 2018-03-30
To add FederatedJob

sdminonne:
@marun @font @onyiny-ang @pmorie. 
To add FederatedJob with a pretty basic override (only `spec.parallelism` currently)

Thanks for having a look


marun:
@sdminonne Please make sure to commit generated code for placement and overrides separate from manual changes.

marun:
Please run `sed -e s+marun/fnord+kubernetes-sigs/federation-v2+` on your changes.

sdminonne:
@marun PTAL

marun:
@sdminonne lgtm except for the last commit.  It isn't necessary to update generated code after adding an adapter since an adapter doesn't affect generated code.  The only reason there were changes is that `apiserver-boot build generated` doesn't render the contents of generated files in a deterministic order.

sdminonne:
@marun thanks for the feedback. 
No problem: I removed the last commit and renamed. 
PTAL

marun:
lgtm, thank you!

PR #40, opened 2018-04-04
Refactor reconcile loop

font:
This removes a lot of the anonymous function closures and indirection in the reconcile loop, which I think improves readability and maintenance.

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -455,50 +455,17 @@ func (s *FederationSyncController) reconcile(qualifiedName federatedtypes.Qualif
 		// A status update is pending
 		return statusNeedsRecheck
 	}
-	propagatedVersiongFromCache, err := s.objFromCache(s.propagatedVersionStore, "PropagatedVersion", propagatedVersionKey)
+	propagatedVersionFromCache, err := s.objFromCache(s.propagatedVersionStore,
+		"PropagatedVersion", propagatedVersionKey)
 	if err != nil {
 		return statusError
 	}
 	var propagatedVersion *fedv1a1.PropagatedVersion
-	if propagatedVersiongFromCache != nil {
-		propagatedVersion = propagatedVersiongFromCache.(*fedv1a1.PropagatedVersion)
+	if propagatedVersionFromCache != nil {
+		propagatedVersion = propagatedVersionFromCache.(*fedv1a1.PropagatedVersion)
 	}
 
-	operationsAccessor := func(adapter federatedtypes.FederatedTypeAdapter, selectedClusters, unselectedClusters []string, template, placement, override pkgruntime.Object) ([]util.FederatedOperation, error) {
-		clusterVersions := make(map[string]string)
-		if propagatedVersion != nil {
-			templateVersion := adapter.Template().ObjectMeta(template).ResourceVersion
-			overrideVersion := ""
-			if override != nil {
-				overrideVersion = adapter.Override().ObjectMeta(override).ResourceVersion
-			}
-			if templateVersion == propagatedVersion.Status.TemplateVersion && overrideVersion == propagatedVersion.Status.OverrideVersion {
-				for _, versions := range propagatedVersion.Status.ClusterVersions {
-					clusterVersions[versions.ClusterName] = versions.ResourceVersion
-				}
-			}
-		}
-		operations, err := clusterOperations(adapter, selectedClusters, unselectedClusters, template, override, key, clusterVersions, func(clusterName string) (interface{}, bool, error) {
-			return s.informer.GetTargetStore().GetByKey(clusterName, key)
-		})
-		if err != nil {
-			s.eventRecorder.Eventf(template, corev1.EventTypeWarning, "FedClusterOperationsError", "Error obtaining sync operations for %s: %s error: %s", kind, key, err.Error())
-		}
-		return operations, err
-	}
-
-	return syncToClusters(
-		operationsAccessor,
-		clusterNames,
-		s.updater.Update,
-		s.adapter,
-		s.informer,
-		template,
-		placement,
-		override,
-		propagatedVersion,
-		s.pendingVersionUpdates,
-	)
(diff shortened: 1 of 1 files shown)


pkg/controller/sync/controller.go
```diff
@@ -566,54 +533,98 @@ func (s *FederationSyncController) versionName(resourceName string) string {
...
 
-	clusterVersions, operationErrors := execute(operations)
+	updatedClusterVersions, operationErrors := s.updater.Update(operations)
 
-	err = updatePropagatedVersion(adapter, clusterVersions, template, override, version, selectedClusters, pendingVersionUpdates)
+	err = updatePropagatedVersion(s.adapter, updatedClusterVersions, template, override,
+		propagatedVersion, selectedClusters, s.pendingVersionUpdates)
 	if err != nil {
-		runtime.HandleError(fmt.Errorf("Failed to record propagated version for %s %q: %v", kind, key, err))
+		runtime.HandleError(fmt.Errorf("Failed to record propagated version for %s %q: %v", kind,
```
marun: (No action required) I think linebreaking at a consistent width can hurt readability, as in this case where the change is for the sake of 10 chars.

pkg/controller/sync/controller.go
```diff
@@ -566,54 +533,98 @@ func (s *FederationSyncController) versionName(resourceName string) string {
...
+			key, err))
 		// Don't return an error - failure to record the propagated
 		// version does not imply that propagation for the resource
 		// needs to be attempted again.
 	}
 
 	if len(operationErrors) > 0 {
-		runtime.HandleError(fmt.Errorf("Failed to execute updates for %s %q: %v", kind, key, operationErrors))
+		runtime.HandleError(fmt.Errorf("Failed to execute updates for %s %q: %v", kind,
```
marun: ditto

pkg/controller/sync/controller.go
```diff
@@ -566,54 +533,98 @@ func (s *FederationSyncController) versionName(resourceName string) string {
...
+
+	templateVersion := adapter.Template().ObjectMeta(template).ResourceVersion
+	overrideVersion := ""
+
+	if override != nil {
+		overrideVersion = adapter.Override().ObjectMeta(override).ResourceVersion
+	}
+
+	if templateVersion == propagatedVersion.Status.TemplateVersion &&
+		overrideVersion == propagatedVersion.Status.OverrideVersion {
```
marun: (No action required) Ugh, I really don't like seeing the conditional line up with the contents.  I'd like to see a clear visual differentiation between the inside and outside of a block.

pkg/controller/sync/controller.go
```diff
@@ -566,54 +533,98 @@ func (s *FederationSyncController) versionName(resourceName string) string {
...
+		}
+	}
+
+	return clusterVersions
+}
+
+// updatePropagatedVersion handles creating or updating the propagated version
+// resource in the federation API.
+func updatePropagatedVersion(adapter federatedtypes.FederatedTypeAdapter,
+	updatedVersions map[string]string, template, override pkgruntime.Object,
```
marun: ditto

pkg/controller/sync/controller.go
```diff
@@ -648,8 +659,10 @@ func updatePropagatedVersion(adapter federatedtypes.FederatedTypeAdapter, versio
 	return err
 }
 
-// newVersion initializes a new version resource for the given cluster versions and template and override
-func newVersion(clusterVersions map[string]string, templateMeta *metav1.ObjectMeta, targetKind, overrideVersion string) *fedv1a1.PropagatedVersion {
+// newVersion initializes a new propagated version resource for the given
```
marun: (No action required) In future try to avoid mixing functional and non-functional (formattting in this case) changes in a single commit.

pkg/controller/sync/controller.go
```diff
@@ -700,15 +714,20 @@ func clusterVersions(newVersions map[string]string, oldVersions []fedv1a1.Cluste
...
+// clusterOperations returns the list of operations needed to synchronize the
+// state of the given object to the provided clusters.
+func (s *FederationSyncController) clusterOperations(selectedClusters, unselectedClusters []string,
+	template, override pkgruntime.Object, key string,
+	clusterVersions map[string]string) ([]util.FederatedOperation, error) {
 
-// clusterOperations returns the list of operations needed to synchronize the state of the given object to the provided clusters
-func clusterOperations(adapter federatedtypes.FederatedTypeAdapter, selectedClusters, unselectedClusters []string, template, override pkgruntime.Object, key string, clusterVersions map[string]string, accessor clusterObjectAccessorFunc) ([]util.FederatedOperation, error) {
 	operations := make([]util.FederatedOperation, 0)
+	accessor := func(clusterName string) (interface{}, bool, error) {
```
marun: I don't think accessor is necessary where it's not passed in - just use the function directly.

pkg/controller/sync/controller.go
font: Fair enough. I'm just trying to avoid these hugely long lines that I see all over the k8s code base. I realize there is no standard for the width in k8s, but that really hurts readability IMO.

pkg/controller/sync/controller.go
font: Agreed. Perhaps an extra space would help as I've done to functions to split between the function parameters and its contents since opening braces are required to be on the same line. But again, trying to avoid these really long lines.

Personally, I prefer the boolean operator to indicate the beginning of continuing conditional expressions but it seems `gofmt` via my plugin does not like it.

pkg/controller/sync/controller.go
font: It's definitely not necessary. I just left this one only to save some line width space since each invocation is quite lengthy.

pkg/controller/sync/controller.go
font: Sorry about that. I was in refactoring mode and cleaning things up so I made these changes along the way. Afterwards it's difficult to split these changes up within the same file.

marun:
lgtm, thank!

PR #41, opened 2018-04-04
Use ReplicaSetSpec as override instead of  just replicas

shashidharatd:
This PR modifies the `FederatedReplicaSetOverride` api object to include the complete ReplicaSet spec instead of just a single field.
Note: This pr uses the https://github.com/imdario/mergo package to merge two structs.

/assign @marun @font 
/cc @irfanurrehman


shashidharatd:
okie, near-term goal SGTM. The only reason i did this PR is to stabilise the API's asap and by accepting the k8s object itself as override, user can override any field(s) and just input that.

PR #42, opened 2018-04-04
Refactor placement computation into plugin

marun:
This is in anticipation of computing namespace placement differently.

cc: @font

### DIFF
--- a/pkg/controller/sync/placement/namespace.go
+++ b/pkg/controller/sync/placement/namespace.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package placement
+
+import (
+	"github.com/marun/federation-v2/pkg/controller/util"
+	"github.com/marun/federation-v2/pkg/federatedtypes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


pkg/controller/sync/placement/resource.go
```diff
@@ -0,0 +1,80 @@
...
+type ResourcePlacementPlugin struct {
+	adapter federatedtypes.PlacementAdapter
+	// Store for the placement directives of the federated type
+	store cache.Store
+	// Informer controller for placement directives of the federated type
+	controller cache.Controller
+}
+
+func NewResourcePlacementPlugin(adapter federatedtypes.PlacementAdapter, triggerFunc func(pkgruntime.Object)) PlacementPlugin {
+	store, controller := cache.NewInformer(
```
font: The sync controller has a function that instantiates a store and controller based on a FedApiAdapter. Seems like we could perhaps move that function to a common package and continue using it in the sync controller for templates and overrides, and here for placement.

pkg/controller/sync/placement/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (p *NamespacePlacementPlugin) Run(stopCh <-chan struct{}) {
+	p.controller.Run(stopCh)
+}
+
+func (p *NamespacePlacementPlugin) HasSynced() bool {
+	return p.controller.HasSynced()
+}
+
+func (p *NamespacePlacementPlugin) ComputePlacement(key string, clusterNames []string) (selectedClusters, unselectedClusters []string, err error) {
```
font: Since we don't have any naked returns, perhaps we can do without the named return values.

pkg/controller/sync/placement/resource.go
marun: This is temporary - it will be replaced with a generic informer in my next PR.

pkg/controller/sync/placement/namespace.go
marun: My thinking is that naming return values can provide for self-documenting return values.  Is there a side-effect of this that suggests not doing so?

font: Yeah, that's a good point. It does make the prototype self documenting.

No side-effect that I'm aware of - just that if we also had a naked return you have to look at the code a little more closely.

pkg/controller/sync/placement/resource.go
font: Gotcha.

font:
LGTM

PR #14, opened 2018-02-21
Add Kubefnord join and namespace PropagatedVersion changes

font:
This adds support for `kubefnord join` to join clusters to the federation.

You can build it by running the following command from the root of the repo:
```
go build -o bin/kubefnord cmd/kubefnord/kubefnord.go
```

PR #45, opened 2018-04-06
Add federated service

shashidharatd:
This pr defines the `FederatedService` API and adds the sync adaptor for it.
Added a new hook to sync adaptor which can be used to retain some fields in the cluster object during update operation.

/assign @marun @font 
/cc @pmorie @irfanurrehman


pkg/federatedtypes/service.go
```diff
@@ -107,6 +107,28 @@ func (a *FederatedServiceAdapter) ObjectForCluster(template, override pkgruntime
...
+func (a *FederatedServiceAdapter)ObjectForUpdateOp(desiredObj, clusterObj pkgruntime.Object) pkgruntime.Object {
+	desiredService := desiredObj.(*corev1.Service)
+	clusterService := clusterObj.(*corev1.Service)
+
+	// ClusterIP and NodePort are allocated to Service by cluster, so retain the same if any while updating
+	desiredService.Spec.ClusterIP = clusterService.Spec.ClusterIP
+	for _, cPort := range clusterService.Spec.Ports {
+		for i, fPort := range clusterService.Spec.Ports {
+			if fPort.Name == cPort.Name && fPort.Protocol == cPort.Protocol && fPort.Port == cPort.Port {
+				desiredService.Spec.Ports[i].NodePort = cPort.NodePort
```
marun: I don't think it can be assumed that the port layout of the desired and cluster services will match.  In the event of divergence between the 2, this line seems to introduce the possibility of an index error.

pkg/federatedtypes/service.go
```diff
@@ -107,6 +107,28 @@ func (a *FederatedServiceAdapter) ObjectForCluster(template, override pkgruntime
...
+			}
+		}
+	}
+	// ExternalIPs are not managed by Kubernetes, so retain the same if any while updating
+	desiredService.Spec.ExternalIPs = clusterService.Spec.ExternalIPs
+
+	// Pass the same ResourceVersion as in the cluster object for update operation, otherwise operation will fail.
+	desiredService.ResourceVersion = clusterService.ResourceVersion
```
marun: Why do you think this necessary?  

The problem of an implicit resource version in federation v1 (because a kube resource was the template vs a template being a field in v2) should not be present here.  And while a user could provide a resource version, I would expect that field to be ignored when constructing the desired form of a resource for a given cluster.

pkg/apis/federation/v1alpha1/federatedserviceoverride_types.go
```diff
@@ -0,0 +1,77 @@
...
+type FederatedServiceOverrideSpec struct {
+	Overrides []FederatedServiceClusterOverride
+}
+
+// FederatedServiceClusterOverride defines the overrides for a named cluster
+type FederatedServiceClusterOverride struct {
+	// TODO(marun) Need to ensure that a cluster name only appears
+	// once.  Why can't maps be used so this validation is automatic?
+	ClusterName string         `json:"clustername,omitempty"`
+	Override    corev1.Service `json:"template,omitempty"`
```
marun: As previously discussed, I don't think it's desirable to define an override as the entire type.  Are there specific fields you think a user would want to override?  If not, consider avoiding implementing an override type at this time (it is optional in the sync controller).

shashidharatd: oh, I was assuming it was mandatory for the sync controller. I will remove the override type for service for now. Thanks for updating me.

pkg/federatedtypes/service.go
shashidharatd: oops, there is a bug in line no.117. it should have been desiredService instead of clusterService. Will fix it. Basically i am iterating through all ports in clusterService to find a matching port in desiredService. if there is a match then only the NodePort in desiredService is retained to the same as that of clusterService port. Let me know if that's ok, after the fix.

pkg/federatedtypes/service.go
shashidharatd: For ExternalIPs, lets deal with that later. (Essentially i was trying to bring what existed in federation v1. this was added by this pr https://github.com/kubernetes/kubernetes/pull/51894)

For Resource version, it seems the k8s is rejecting the update operation if it is empty for service type. For other types it seems to be ok, if ResourceVersion is empty. Here is the error log, if i don't retain the Resourceversion for update operation.
```
E0409 12:30:17.613205   21138 controller.go:598] Failed to execute updates for FederatedService "8b2e62c0-16a8-40d2-a0d2-bfaa59fc62ba/test-service-b8j25": [Failed to update FederatedService "8b2e62c0-16a8-40d2-a0d2-bfaa59fc62ba/test-service-b8j25" in cluster test-cluster-7nqlv: Service "test-service-b8j25" is invalid: metadata.resourceVersion: Invalid value: "": must be specified for an update Failed to update FederatedService "8b2e62c0-16a8-40d2-a0d2-bfaa59fc62ba/test-service-b8j25" in cluster test-cluster-5z8kw: Service "test-service-b8j25" is invalid: metadata.resourceVersion: Invalid value: "": must be specified for an update]
```

marun: Ok, very curious.  Thanks for the info.

pkg/federatedtypes/service.go
```diff
@@ -92,6 +92,28 @@ func (a *FederatedServiceAdapter) ObjectForCluster(template, override pkgruntime
...
 
+// TODO(shashi) Avoid the need for this adapter method by handling the scenario generically
+// for cases where the field in spec is updated by controllers instead of status.
+func (a *FederatedServiceAdapter) ObjectForUpdateOp(desiredObj, clusterObj pkgruntime.Object) pkgruntime.Object {
+	desiredService := desiredObj.(*corev1.Service)
+	clusterService := clusterObj.(*corev1.Service)
+
+	// ClusterIP and NodePort are allocated to Service by cluster, so retain the same if any while updating
+	desiredService.Spec.ClusterIP = clusterService.Spec.ClusterIP
+	for i, fPort := range desiredService.Spec.Ports {
```
marun: (No action required) Maybe choose a more descriptive name than fPort?  I parse cPort as c[luster]Port, but I don't know what the `f` stands for.

pkg/federatedtypes/service.go
```diff
@@ -92,6 +92,28 @@ func (a *FederatedServiceAdapter) ObjectForCluster(template, override pkgruntime
 	return service
 }
 
+// TODO(shashi) Avoid the need for this adapter method by handling the scenario generically
```
marun: Thank you for documenting this.  

I'm looking forward to deprecating adapters in the not-too-distant future - in favor of generic handling of types - in which case this behavior could be handled instead via a callback provided to controller instantiation.

test/common/crudtester.go
```diff
@@ -149,11 +149,15 @@ func (c *FederatedTypeCrudTester) createFedResource(adapter federatedtypes.FedAp
 func (c *FederatedTypeCrudTester) CheckCreate(desiredTemplate, desiredPlacement, desiredOverride pkgruntime.Object) (pkgruntime.Object, pkgruntime.Object, pkgruntime.Object) {
 	template, placement, override := c.Create(desiredTemplate, desiredPlacement, desiredOverride)
 
+	overrideResourceVersion := ""
```
marun: Good catch! I think the override being optional hasn't been tested in integration and e2e testing is pending @font's join PR.

font: This was also fixed in #14 that isn't merged, except that the log was re-written to not include "override" if the resource does not contain one.

test/common/crudtester.go
```diff
@@ -149,11 +149,15 @@ func (c *FederatedTypeCrudTester) createFedResource(adapter federatedtypes.FedAp
...
+	if c.adapter.Override() != nil {
+		overrideResourceVersion = c.adapter.Override().ObjectMeta(override).ResourceVersion
+	}
 	c.tl.Logf("Resource versions for %s: template %q, placement %q, override %q",
```
font: To prevent any confusion, I suggest re-writing the log to not include "override" if the resource doesn't contain one. Otherwise you can wait for that part of #14 to merge, which will be in a new PR.

shashidharatd:
@font , addressed the nit. Can you please merge it now. Thanks!

pmorie:
Thanks for the patch @shashidharatd - i have some ideas around federated DNS that can dovetail off this - will try to post an issue later on today so you can weigh in.

font:
lgtm

shashidharatd:
>i have some ideas around federated DNS that can dovetail off this - will try to post an issue later on today so you can weigh in.

@pmorie, looking forward to it, meanwhile i am working on federated service status, which also include a simple healthcheck that endpoint exist for the service or not.

PR #46, opened 2018-04-09
Scheduling API and controller for federated deployments

irfanurrehman:
Pushing only one commit into this PR to open up discussion on definition/naming of groups for advanced (or other API) types in federation-v2.
I could have opened an issue to discuss the same, but the reference code is handy here.
My original understanding is that the base/core API group is named as `federation.k8s.io` and we should be able to bootstrap, say 
`scheduling.federation.k8s.io` (example for scheduling) and say `simple.scheduling.federation.k8s.io` as a subgroup within scheduling group, and thus possibility of different implementations within the same group.

The catch is that `apiserver-builder` restricts a group to be a simple string (`^[a-z]+$`). The group currently cannot also be empty (This ideally should be possible similar to k8s, where core api group is identified with ""). The domain used to initialise the repo can however be a FQDN.

We now have couple of options (for immediate solution):
1. Distingiush API types into groups with elucidative naming (eg `scheduling` `simplescheduling` `complexscheduling`, etc). 
I have used `federatedscheduling` in this PR not only as representation, but also to keep it different from k8s core api group `scheduling`. As we have initiated federation-v2 also with domain `k8s.io`, is there a possible conflict with k8s core API group `scheduling`, when agregated? (both will become `scheduling.k8s.io`, not sure how the agregation strategy works out to resolve such conflicts).
2. Change the domain used to init the project to `federation.k8s.io`, the building blocks api group to `base/core.federation.k8s.io`, and all other APIs into their named groups.
3. Initiate development of higher level APIs into other repos initialised with lower level subdomain viz `federation.k8s.io`. 
4. Discussions on this PR .....  = more suggestions.

5. CRDs do not have restrictions like we have here. The group name is treated as FQDN. But CRDs have much lesser flexibility in implementation. 

@marun @pmorie @shashidharatd
Do we also need a federation subgroup alias?

---
This change is [Reviewable](https://reviewable.io/reviews/marun/federation-v2/46)


marun:
@irfanurrehman I'm fine with `federatedscheduling` for now.  Assuming it's possible to supply a subdomain (e.g. `federation.k8s.io`) as the init domain instead of `k8s.io`, that seems like a relatively easy near-term path too.  Anyone should be able to work in other repos, but that introduces the challenges of vendoring and the associated requirement for api stability.  I'm not sure it's time for that quite yet.

I think it would be ideal to start using CRDs sooner than latter - that's certainly my intent.  As with where cluster registry is going, I think CRDs represent the future for defining federation types (no more separate api server!).  The catch is that switching to CRDs so may entail a move from `apiserver-builder` to [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) to avoid having to hand-curate clients.

irfanurrehman:
@marun moving to CRDs, sounds like a plan! And I share you opinion of continue building the functionality in the interim.

irfanurrehman:
**A note to reviewers:**
As mentioned in work group sync meetings, This is not complete and probably first step in a series of PRs. 
What this lacks as of now, which I intend to focus on into next set of PRs
- Tests
- Make the overall code generic (use a adapter like mechanism used in sync) to be able to take scheduling types. 
- ReplicaSchedulingPreference has a targetRef (kind+name), which is not used in this PR, update to ensure the same is used.
- There are couple of TODOs which needs to be worked on.

irfanurrehman:
To reviewers; this can be taken up for review now.
As mentioned in the previous comment; I intend to commit further items as separate PRs to make the reviews easier.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -15,6 +14,709 @@ See the License for the specific language governing permissions and
...
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return fedClient.FederatedschedulingV1alpha1().ReplicaSchedulingPreferences(metav1.NamespaceAll).Watch(options)
+			},
+		},
+		&fedschedulingv1a1.ReplicaSchedulingPreference{},
+		util.NoResyncPeriod,
+		util.NewTriggerOnAllChanges(deliverObj),
+	)
+
+	s.templateStore, s.templateController = cache.NewInformer(
```
marun: Is this actually ready for review?  This line and 161 would appear to conflict and cause me to question whether this controller is functional.  Are you intending to add tests to verify the proposed functionality?

cmd/controller-manager/main.go
```diff
@@ -54,6 +55,11 @@ func main() {
 		}
 	}
 
+	err = rspcontroller.StartReplicaSchedulingPreferenceController(config, config, config, stopChan, true)
```
marun: Up till now the only functionality of v2 has been propagation.  Adding scheduling suggests a requirement for a configuration mechanism that would allow scheduling to be optionally enabled.

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: I did an initial verification of api availability only earlier; have verified the complete flow of overrides, placement and actual deployment being created in fcp and cluster respectively after a replicaschedulingpreference and template is created in fcp, today.
Tests us what I am working on now and intend to submit that as another PR.

cmd/controller-manager/main.go
irfanurrehman: Configuration seems reasonable to me. I can take that up once I complete relevant tasks needed to make the scheduling API usable ([listed here](https://github.com/kubernetes-sigs/federation-v2/pull/46#issuecomment-384913451)).

pkg/controller/replicaschedulingpreference/controller.go
marun: I think minimal automated testing is a prereq for reviewing, let alone merging.  It doesn't have to be exhaustive by any means, but I would like to be reviewing code that had some automatic validation in place.  I'm not a very good runtime.

irfanurrehman: :-); I will update some tests as part of this PR.

irfanurrehman: @marun updated, took me some time to absorb and rebase on your recent changes. PTAL.
@shashidharatd @font if possible, you can also pitch in for review please.

irfanurrehman:
@marun, @font, @shashidharatd 
There seems to be an issue with the test fixture; if multiple specs are run in parallel in integration, this (the error listed in CI failure) seems to happen; 

`cd test/integration`
`go test` **fails**
`go test -run <single test name>` **passes**
Looking into it, any pointers to speed up resolution appreciated.

pmorie:
Looks like a legit test failure in CI:

```
=== RUN   TestReplicaSchedulingPreference
I0521 19:33:56.720697    6729 serving.go:283] Generated self-signed cert (/tmp/fed-test-certs814626197/apiserver.crt, /tmp/fed-test-certs814626197/apiserver.key)
I0521 19:33:56.897329    6729 serve.go:85] Serving securely on 127.0.0.1:39391
I0521 19:33:57.571366    6735 server.go:121] Version: v1.10.0-alpha.1
I0521 19:33:57.571787    6735 services.go:35] Network range for service cluster IPs is unspecified. Defaulting to 10.0.0.0/24.
I0521 19:33:57.622350    6642 serving.go:295] Generated self-signed cert (/tmp/fed-test-certs931353250/apiserver.crt, /tmp/fed-test-certs931353250/apiserver.key)
panic: the group "federatedscheduling.k8s.io" has already been announced.
goroutine 5479 [running]:
github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/apimachinery/pkg/apimachinery/announced.(*GroupMetaFactory).Announce(0xc423b81b60, 0xc420284f60, 0xc423b81b60)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/apimachinery/pkg/apimachinery/announced/group_factory.go:85 +0x6c
github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/builders.(*APIGroupBuilder).Install(0xc42054f500, 0xc420284f60, 0xc420284f90, 0xc420224300)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/builders/api_group_builder.go:134 +0x1cf
github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/builders.(*APIGroupBuilder).Announce(0xc42054f500)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/builders/api_group_builder.go:142 +0x4f
github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/apiserver.(*Config).Init(0xc4210f8048, 0xc4204b4ef0)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/apiserver/apiserver.go:39 +0x5d
github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/cmd/server.(*ServerOptions).RunServer(0xc4248e9bc0, 0xc422fb5800, 0x1a4a3f1, 0x3, 0x1a4a126, 0x2, 0xc421e80120, 0xc4202fc570)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/cmd/server/start.go:218 +0x1d7
github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/cmd/server.NewCommandStartServer.func1(0xc4217a0900, 0xc4200be420, 0x0, 0xb, 0x0, 0x0)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/kubernetes-incubator/apiserver-builder/pkg/cmd/server/start.go:115 +0x8b
github.com/kubernetes-sigs/federation-v2/vendor/github.com/spf13/cobra.(*Command).execute(0xc4217a0900, 0xc4202fc4d0, 0xb, 0xb, 0xc4217a0900, 0xc4202fc4d0)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/spf13/cobra/command.go:599 +0x3db
github.com/kubernetes-sigs/federation-v2/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc4217a0900, 0x1, 0x0, 0x0)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/spf13/cobra/command.go:689 +0x2d4
github.com/kubernetes-sigs/federation-v2/vendor/github.com/spf13/cobra.(*Command).Execute(0xc4217a0900, 0x1, 0x0)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/github.com/spf13/cobra/command.go:648 +0x2b
github.com/kubernetes-sigs/federation-v2/test/integration/framework.(*FederationApiFixture).setUp.func1(0xc4217a0900, 0x441cb00, 0xc420d15450)
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/test/integration/framework/fedapi.go:79 +0x2f
created by github.com/kubernetes-sigs/federation-v2/test/integration/framework.(*FederationApiFixture).setUp
	/home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/test/integration/framework/fedapi.go:78 +0x5df
FAIL	github.com/kubernetes-sigs/federation-v2/test/integration	53.856s
```

irfanurrehman:
> Looks like a legit test failure in CI:

Yes, I could pinpoint the root of the problem as some global variables retaining the values set in apiserver run as part of one test spec in integration test (via federation test fixture) into another specs federation test fixture (https://github.com/kubernetes-incubator/apiserver-builder/blob/master/pkg/builders/scheme.go). Please note that this will happen even if the test run is serial (setup, teardown, setup, .... ) . All in all, it seems, using the code from apiserver-builder, we cannot instantiate two apiservers from the same context (or probably its not meant to be done anyways).
Possible solutions: 
1. We do not use the test fixture setup and tear down multiple times in the same test package, rather use a suite setup to setup single integration test fixture and run multiple specs against it (working on it). 
2. Or some simpler workaround (working on this too.. :) )

irfanurrehman:
Implementes #88

irfanurrehman:
@marun you might want to have (atleast a superficial) look at this. Its valuable :-).
@shashidharatd will need a merge approval if everything seems allright.

test/integration/integration_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"testing"
+
+	"github.com/kubernetes-sigs/federation-v2/test/integration/framework"
+)
+
+var FedFixture *framework.FederationFixture
+
+// TestIntegration provides a common setup and teardown for all the integration tests.
```
marun: (No action required) If the goal is shared setup, I'm not sure why manually composing a test suite would be the strategy.  Seems like administrivia when we already have automated test discovery.   I don't care too much, though, since I expect integration to be deprecated before too long in preference to managed e2e.

test/integration/replicaschedulingpreference_test.go
```diff
@@ -0,0 +1,237 @@
...
+// TestReplicaSchedulingPreference validates basic replica scheduling preference calculations.
+var TestReplicaSchedulingPreference = func(t *testing.T) {
+	t.Parallel()
+	tl := framework.NewIntegrationLogger(t)
+
+	controllerFixture, fedClient := initRSPTest(tl, FedFixture)
+	defer controllerFixture.TearDown(tl)
+
+	clusters := getClusterNames(fedClient)
+	if len(clusters) != 2 {
```
marun: '< 2' ?

test/integration/replicaschedulingpreference_test.go
```diff
@@ -0,0 +1,237 @@
...
+		tl.Fatalf("Expected two clusters to be part of Federation Fixture setup")
```
marun: (No action required) Is failing preferable to skipping here?

It seems strange to me that these tests effectively require all other tests to have 2 member clusters.  Again, this is fine in the near term but I think any test that has specific requirements should probably be deploying its own fixture rather than constraining fixture for others.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,751 @@
...
+		s.podInformer.Stop()
+		s.workQueue.ShutDown()
+		s.deliverer.Stop()
+		s.clusterDeliverer.Stop()
+	}()
+}
+
+type reconciliationStatus int
+
+const (
```
marun: (No action required) Seems like there is common controller code that could be factored out for reuse at some point.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,751 @@
...
+		delay = delay + s.backoff.Get(key)
+	} else {
+		s.backoff.Reset(key)
+	}
+	s.deliverer.DeliverAfter(key, &qualifiedName, delay)
+}
+
+// Check whether all data stores are in sync. False is returned if any of the informer/stores is not yet
+// synced with the corresponding api server.
+func (s *ReplicaSchedulingPreferenceController) isSynced() bool {
```
marun: What about the sync status for pods, templates, overrides and placement?

cmd/controller-manager/main.go
```diff
@@ -60,6 +61,11 @@ func main() {
 	c := manager.NewFederatedTypeConfigController(config, si)
 	c.Run(stopChan)
 
+	err = rspcontroller.StartReplicaSchedulingPreferenceController(config, config, config, stopChan, true)
```
marun: (No action required) I would like to see execution of this controller configurable before alpha, as per #101.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,751 @@
...
+		return statusAllOK
+	}
+
+	typedRsp, ok := rsp.(*fedschedulingv1a1.ReplicaSchedulingPreference)
+	if !ok {
+		runtime.HandleError(fmt.Errorf("Incorrect runtime object for RSP: %v", rsp))
+		return statusNotSynced
+	}
+
+	if typedRsp.ObjectMeta.DeletionTimestamp != nil {
```
marun: Without a finalizer or the cleanup on deletion that would suggest one, consider removing this block and the `delete()` function.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,751 @@
...
+		}
+		return statusAllOK
+	}
+
+	clusterNames, err := s.clusterNames()
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("Failed to get cluster list: %v", err))
+		return statusNotSynced
+	}
+	if len(clusterNames) < 1 {
```
marun: `== 0` ?

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,751 @@
...
+}
+
+func updateOverrides(fedClient fedclientset.Interface, qualifiedName util.QualifiedName, result map[string]int64) error {
+	override, err := fedClient.FederationV1alpha1().FederatedDeploymentOverrides(qualifiedName.Namespace).Get(qualifiedName.Name, metav1.GetOptions{})
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+
+		// TODO: remove after review and
```
marun: I don't understand this comment in its current form.  Please clarify or remove.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,751 @@
...
+	checkLen := 0
+	for _, override := range overrideSpec.Overrides {
+		replicas, ok := result[override.ClusterName]
+		if !ok || (override.Replicas == nil) || (int32(replicas) != *override.Replicas) {
+			return true
+		}
+		checkLen += 1
+	}
+
+	if checkLen != resultLen {
```
marun: Code golf:

```golang
return checkLen != resultLen
```

test/integration/integration_test.go
irfanurrehman: This  provides the mechanism to let the subtests run in parallel.

test/integration/replicaschedulingpreference_test.go
irfanurrehman: Some of the tabular tests here expect exactly 2 clusters. This could be a difference number also, if the tests cases need that (may be in future).

test/integration/replicaschedulingpreference_test.go
irfanurrehman: Actually, I did intend to and deploy a different fixture as part of these tests, but did meet an issue because of how apiserver init code is setup (as per the autogenerated setup code in this repo), the root of which is explained [here](https://github.com/kubernetes-sigs/federation-v2/pull/46#issuecomment-391401574). Two fixtures do not work out fine within the same test context. Perhaps some mechanism to have different test binaries per test setup is the workaround to this problem, but did not seem like necessary at this point of time.

cmd/controller-manager/main.go
irfanurrehman: Agree, I will take this up as soon as I finish #92, which is the next priority item in my list.

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Filed https://github.com/kubernetes-sigs/federation-v2/issues/102

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Updated!

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Updated!

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Updated!

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Removed!

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Updated!

test/integration/replicaschedulingpreference_test.go
```diff
@@ -0,0 +1,237 @@
...
+			expected: map[string]int32{
+				clusters[0]: int32(3),
+				clusters[1]: int32(3),
+			},
+		},
+	}
+
+	for testName, testCase := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			name, err := createTestObjs(testCase.prefSpec, framework.DefaultTestNS, fedClient)
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

irfanurrehman:
@marun, updated as per your comments. PTAL!

test/integration/replicaschedulingpreference_test.go
marun: That sounds like a point for moving to kubebuilder and CRDs, then, since that would remove the need for an autogenerated apiserver.

test/integration/integration_test.go
marun: Why does shared fixture instantiation/teardown require the usage of a suite?  AFAICT a test calling  `t.Parallel` doesn't require that the test be called as part of a suite.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -0,0 +1,740 @@
...
+	clusters, err = s.podInformer.GetReadyClusters()
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("Failed to get ready clusters: %v", err))
+		return false
+	}
+	if !s.podInformer.GetTargetStore().ClustersSynced(clusters) {
+		return false
+	}
+
+	if !s.controller.HasSynced() ||
```
marun: (No action required) Code golf

```golang
return [boolean expression]

// this is a potentially more readable version
controllersSynced := [boolean expression]
return controllersSynced
```

In any case, I think it's preferable to avoid using an `if` conditional to decide between `true` or `false`.  The conditional is already evaluating to a boolean, just use it directly.

test/integration/replicaschedulingpreference_test.go
marun: (No action required) Requiring such specific setup is a smell to me.  I think integration testing would ideally be limited to retrieving details from an arbitrary number of member clusters and the federation api, and ensuring the resulting computation is applied correctly to the correct resources.  The specifics tested here that depend on an exact number of clusters would seem to be easier to evaluate in a unit test that provided specific inputs without the overhead of fixture.  I'm not suggesting this gets fixed here, but I anticipate this guideline being usefully applied in the future.

marun:
Thank you for cleanup!

/lgtm

I'll leave it to @shashidharatd to approve.

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: Thanks for the example; my past coding style however has always been use of `if` statements. I found them to be more readable. Meanwhile suggestion well taken, will try to adapt in future code.

test/integration/framework/controller.go
```diff
@@ -54,6 +55,18 @@ func NewClusterControllerFixture(fedConfig, kubeConfig, crConfig *restclient.Con
 	return f
 }
 
+// NewSyncControllerFixture initializes a new sync controller fixture.
```
shashidharatd: Need fixing comment

test/integration/integration_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestIntegration(t *testing.T) {
+	tl := framework.NewIntegrationLogger(t)
+	FedFixture = framework.SetUpFederationFixture(tl, 2)
+	defer FedFixture.TearDown(tl)
+
+	t.Run("Parallel-Integration-Test-Group", func(t *testing.T) {
```
shashidharatd: Is nesting all the tests under `Parallel-Integration-Test-Group` required? It would be unnecessarily long test case name to call out individually.

pkg/controller/replicaschedulingpreference/controller.go
```diff
@@ -15,6 +14,727 @@ See the License for the specific language governing permissions and
...
+
+	// Store for the override directives of the federated type
+	overrideStore cache.Store
+	// Informer controller for override directives of the federated type
+	overrideController cache.Controller
+
+	// Store for the placements of the federated type
+	placementStore cache.Store
+	// Informer controller for placements of the federated type
+	placementController cache.Controller
```
shashidharatd: I didn't find the need for override and placement controllers. This rspcontroller is only creating/updating the override, placement objects as needed. Please remove these controllers.

test/integration/integration_test.go
irfanurrehman: Yes, I followed the examples [here](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks). Without using a parent test group if you use t.Parallel() to indicate parallelism into individual tests your main test can still finish before the individual tests finish, at which point `FedFixture.TearDown` is called, causing problems. Writing it this way ensures that `Parallel-Integration-Test-Group` does not return until all the sub tests finish.

pkg/controller/replicaschedulingpreference/controller.go
irfanurrehman: I do need it to trigger reconcile, in case the target override or placements are changed by a different entity (user, automated stuff) :).

test/integration/framework/controller.go
irfanurrehman: Updated!

irfanurrehman:
@shashidharatd updated, PTAL!

irfanurrehman:
test continuous-integration/travis-ci/pr

irfanurrehman:
@shashidharatd It seems we need to be members of `kubernetes-sigs` org; but not sure whom to request for membership. @pmorie any suggestions? I see that you have `kubernetes-sigs` membership.

marun:
@irfanurrehman Yes, you'll need to be added to the org.  Neither @pmorie or myself can do this, but it looks like @quinton-hoole is an owner and presumably has the rights.

quinton-hoole:
I have added @irfanurrehman and @shashidharatd as members of the kubernetes-sigs org, and as owners and admins of the kubernetes-sigs/federation-v2 repo.

PR #47, opened 2018-04-10
Track resource propagation using ResourceVersion or Generation

font:


pkg/federatedtypes/adapter.go
```diff
@@ -30,7 +31,7 @@ import (
 // target resource in member clusters.
 type FederatedTypeAdapter interface {
 	FedClient() fedclientset.Interface
-	Template() FedApiAdapter
+	Template() TemplateAdapter
```
marun: Why is this change desirable if `TemplateAdapter` doesn't provide different behavior from `FedApiAdapter`?  `OverrideAdapter` is arguably the bad example here.  Also, this kind of non-functional change would ideally be broken out into a separate commit rather than mixed in with unrelated changes.

pkg/controller/sync/controller.go
```diff
@@ -759,15 +770,24 @@ func (s *FederationSyncController) clusterOperations(selectedClusters, unselecte
...
+				targetVersion := util.GetVersion(s.versionCompareType,
+					s.adapter.Target().ObjectMeta(clusterObj))
+
+				// Check if versions don't match. If they match and the resource uses
+				// Generation to compare versions, then check its ObjectMeta.
+				// This is because Generation is only updated when Spec
+				// changes.
+				if version != targetVersion {
 					operationType = util.OperationTypeUpdate
+				} else if s.versionCompareType == util.Generation {
```
marun: Consider skipping the check of the compare type since the equivalent method alone will produce the correct result.

pkg/apis/federation/v1alpha1/propagatedversion_types.go
```diff
@@ -48,14 +48,14 @@ type PropagatedVersionSpec struct {
...
-	ClusterVersions []ClusterResourceVersion `json:"clusterversions,omitempty"`
+	TemplateVersion string           `json:"templateversion,omitempty"`
+	OverrideVersion string           `json:"overrideversion,omitempty"`
+	ClusterVersions []ClusterVersion `json:"clusterversions,omitempty"`
 }
 
-type ClusterResourceVersion struct {
-	ClusterName     string `json:"clustername,omitempty"`
-	ResourceVersion string `json:"resourceversion,omitempty"`
+type ClusterVersion struct {
```
marun: To me `ClusterVersion` reads as 'version of the cluster'.  Consider using a more descriptive name (e.g. `ClusterObjVersion`).

pkg/controller/sync/controller.go
```diff
@@ -605,8 +613,9 @@ func (s *FederationSyncController) syncToClusters(selectedClusters, unselectedCl
 
 // getClusterVersions returns the cluster versions populated in the current
 // propagated version object if it exists.
-func getClusterVersions(adapter federatedtypes.FederatedTypeAdapter, template,
-	override pkgruntime.Object, propagatedVersion *fedv1a1.PropagatedVersion) map[string]string {
+func getClusterVersions(adapter federatedtypes.FederatedTypeAdapter,
```
marun: (No action required) In future please separate non-functional changes from functional changes.

pkg/controller/sync/controller.go
```diff
@@ -759,14 +777,22 @@ func (s *FederationSyncController) clusterOperations(selectedClusters, unselecte
...
+				desiredObjMeta := s.adapter.Target().ObjectMeta(desiredObj)
+				targetVersion := s.comparisonHelper.GetVersion(clusterObjMeta)
+
+				// Check if versions don't match. If they match then check its
+				// ObjectMeta which only applies to resources where Generation
+				// is used to track versions because Generation is only updated
+				// when Spec changes.
+				if version != targetVersion {
+					operationType = util.OperationTypeUpdate
+				} else if !s.comparisonHelper.Equivalent(desiredObjMeta, clusterObjMeta) { // non-federated
```
marun: Remove `// non-federated` comment.

pkg/controller/util/propagatedversion.go
```diff
@@ -0,0 +1,101 @@
...
+// comparison type passed in.
+func NewComparisonHelper(compareType VersionCompareType) (ComparisonHelper, error) {
+	switch compareType {
+	case ResourceVersion:
+		return &ResourceHelper{}, nil
+	case Generation:
+		return &GenerationHelper{}, nil
+	default:
+		err := fmt.Errorf("Unrecognized version comparison type %v", compareType)
+		runtime.HandleError(err)
```
marun: Use of `HandleError` should be restricted to functions that don't return an error.  If an error is returned, the caller can decide what to do with it.

pkg/controller/util/propagatedversion.go
```diff
@@ -0,0 +1,101 @@
...
+type ResourceHelper struct{}
+
+// GetVersion returns a string containing the version in the resource's
+// ObjectMeta using the resource comparison type to perform for that
+// resource.
+func (ResourceHelper) GetVersion(objectMeta *metav1.ObjectMeta) string {
+	return objectMeta.ResourceVersion
+}
+
+// Equivalent returns true (not implemented).
```
marun: (No action required) I'm not sure 'not implemented' is true.  It would be more accurate to say that equivalency for resourceVersion-based comparison doesn't require evaluation of object meta.

pkg/controller/util/propagatedversion.go
```diff
@@ -0,0 +1,101 @@
...
+	})
+}
+
+// PropagatedVersionStatusEquivalent returns true if both statuses are equal by
+// comparing Template and Override version, and their ClusterVersion slices;
+// false otherwise.
+func PropagatedVersionStatusEquivalent(pvs1, pvs2 *fedv1a1.PropagatedVersionStatus) bool {
+	return pvs1.TemplateVersion == pvs2.TemplateVersion &&
+		pvs1.OverrideVersion == pvs2.OverrideVersion &&
+		len(pvs1.ClusterVersions) == len(pvs2.ClusterVersions)
```
marun: There should be a comparison of the `ClusterVersions` slices with `reflect.DeepEqual`.  The fact that this omission wasn't detected in testing is reminding me that unit testing needs to start real soon.

pkg/federatedtypes/configmap.go
```diff
@@ -103,7 +103,7 @@ func (a *FederatedConfigMapAdapter) ObjectForCluster(template, override pkgrunti
 	return configMap
 }
 
-func (a *FederatedConfigMapAdapter)ObjectForUpdateOp(desiredObj, clusterObj pkgruntime.Object) pkgruntime.Object {
```
marun: (No action required) Again, in future please ensure that functional changes are separated from non-functional.

test/common/crudtester.go
```diff
@@ -364,9 +343,15 @@ func (c *FederatedTypeCrudTester) CheckPropagation(template, placement, override
 
 func (c *FederatedTypeCrudTester) waitForResource(client clientset.Interface, qualifiedName federatedtypes.QualifiedName, expectedVersion string) error {
 	targetAdapter := c.adapter.Target()
-	err := wait.PollImmediate(c.waitInterval, c.clusterWaitTimeout, func() (bool, error) {
+	compareHelper, err := util.NewComparisonHelper(targetAdapter.VersionCompareType())
```
marun: I think the comparison helper should be initialized on crudtester initialization, as with the sync controller.  The error that can occur won't change, so detecting as early as possible is desirable.

pkg/federatedtypes/configmap.go
font: This is the `gofmt` automatic formatting at work.

test/common/crudtester.go
font: I think so too but opted for this as an expedient to get to e2e testing asap.

pkg/controller/util/propagatedversion.go
font: Yeah that's better. I knew there had to be an easier way to compare.

marun:
Thank you!

PR #48, opened 2018-04-13
Add kubefnord to join clusters to the federation

font:
This adds support for kubefnord join to join clusters to the federation.

You can build it by running the following command from the root of the repo:

go build -o bin/kubefnord cmd/kubefnord/kubefnord.go

---
This change is [Reviewable](https://reviewable.io/reviews/marun/federation-v2/48)


test/e2e/framework/test_context.go
```diff
@@ -33,23 +33,26 @@ type TestContextType struct {
 var TestContext TestContextType
 
 func registerFlags(t *TestContextType) {
-	flag.BoolVar(&t.TestManagedFederation, "test-managed-federation", false, "Whether the test run should rely on a test-managed federation.")
-	flag.BoolVar(&t.InMemoryControllers, "in-memory-controllers", true, "Whether federation controllers should be started in memory.  Ignored if test-managed-cluster is false")
-	flag.StringVar(&t.KubeConfig, "kubeconfig", os.Getenv("KUBECONFIG"), "Path to kubeconfig containing embedded authinfo.  Ignored if test-managed-cluster is true")
-	flag.StringVar(&t.KubeContext, "context", "", "kubeconfig context to use/override. If unset, will use value from 'current-context'")
+	flag.BoolVar(&t.TestManagedFederation, "test-managed-federation",
+		false, "Whether the test run should rely on a test-managed federation.")
+	flag.BoolVar(&t.InMemoryControllers, "in-memory-controllers", false,
```
font: `in-memory-controllers` is now defaulted to false.

pkg/controller/util/cluster_util.go
```diff
@@ -92,9 +91,20 @@ func BuildClusterConfig(fedCluster *fedv1a1.FederatedCluster, kubeClient kubecli
 			return nil, err
 		}
 
-		// TODO(marun) Use a service account instead of a serialized kubeconfig
-		kubeconfigGetter := KubeconfigGetterForSecret(secret)
-		clusterConfig, err = clientcmd.BuildConfigFromKubeconfigGetter(serverAddress, kubeconfigGetter)
+		token, tokenFound := secret.Data["token"]
```
font: These changes support both integration and e2e tests.

pkg/kubefnord/join.go
```diff
@@ -0,0 +1,678 @@
...
+	return nil, fmt.Errorf("could not find the deployment for federation controller manager in host cluster")
+}
+
+// getFederationName gets the federation name from the federation
+// deployment.
+func getFederationName(hostClientSet client.Interface, fedNamespace string) (string, error) {
+	d, err := getCMDeployment(hostClientSet, fedNamespace)
+	if err != nil {
+		// TODO: Figure out how to handle this for integration tests.
+		return "", err
```
font: This will need to change when we support running `JoinCluster` from the integration tests because we do not have a deployment running the federation controller/apiserver. Once we get a k8s namespace controller in integration tests we can look at using `JoinCluster` and revisit this then.

pkg/kubefnord/util/util.go
```diff
@@ -0,0 +1,216 @@
...
+	client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	crclient "k8s.io/cluster-registry/pkg/client/clientset_generated/clientset"
+)
+
+// FedConfig provides a rest config based on the filesystem kubeconfig (via
+// `PathOptions()`) and context in order to talk to the host kubernetes cluster
+// and the joining kubernetes cluster
+type FedConfig interface {
```
font: There is potential for a common interface like this to be shared with tests and controllers.

pkg/kubefnord/util/util.go
```diff
@@ -0,0 +1,216 @@
...
+	overrides := &clientcmd.ConfigOverrides{
+		CurrentContext: context,
+	}
+
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&loadingRules, overrides)
+}
+
+// FedClientset provides the clients to talk to the kubernetes
+// API server, the cluster registry API server, and the federation API server.
+type FedClientset interface {
```
font: There is potential for a common interface like this to be shared across tests and controllers.

font:
Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---

*[pkg/kubefnord/util/util.go, line 32 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFRmWACB64r3MwK3Jb:-LAFRmWACB64r3MwK3Jc:b-oofr4r) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/util/util.go#L32)):*
> ```Go
> // `PathOptions()`) and context in order to talk to the host kubernetes cluster
> // and the joining kubernetes cluster
> type FedConfig interface {
> ```

There is potential for a common interface like this to be shared across tests and controllers.

---

*[pkg/kubefnord/util/util.go, line 131 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFS4hU4o9mggixoLLy:-LAFS4hU4o9mggixoLLz:b-ukzvwb) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/util/util.go#L131)):*
> ```Go
> // FedClientset provides the clients to talk to the kubernetes
> // API server, the cluster registry API server, and the federation API server.
> type FedClientset interface {
> ```

There is potential for a common interface like this to be shared across tests and controllers.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/48)*

marun:
Mostly minor nits/invitation for discussion.  I'm guessing much of this code is following conventions from kubefed, which was not what I would call a clean design.

---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[pkg/controller/util/cluster_util.go, line 24 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFUQDe55Y0-NhbUYSn:-LAFUQDe55Y0-NhbUYSo:b30nfcv) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/controller/util/cluster_util.go#L24)):*
> ```Go
> 	"time"
> 
> 	"github.com/golang/glog"
> ```

The general convention for imports are the following line-separated sections:

system 

non-kube 3rd party (e.g. glog)

kube ecosystem

My guess this is in relative order of dependency, since system won't depend on anything, 3rd party could depend on system, kube could depend on 3rd party and system.  By that logic, maybe this project should break out its imports to appear after the kube ecosystem?  For now, though, just separate out the glog import.

---

*[pkg/controller/util/cluster_util.go, line 103 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFVZ8JB0BmSaTi4IOX:-LAFVZ8JB0BmSaTi4IOY:b-hx334k) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/controller/util/cluster_util.go#L103)):*
> ```Go
> 			clusterConfig.CAData = ca
> 			clusterConfig.BearerToken = string(token)
> 		} else {
> ```

Is it desirable to continue supporting this legacy mode given that federation will only support kube >= 1.10 where service accounts are the norm?

---

*[pkg/kubefnord/join.go, line 54 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFbkFUAkBpranGtSfm:-LAFbkFUAkBpranGtSfn:bb5qayq) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L54)):*
> ```Go
> 
>         Current context is assumed to be a federation API
>         server. Please use the --context flag otherwise.`
> ```

Did you mean `--host-cluster-context`?

---

*[pkg/kubefnord/join.go, line 81 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFc8oI1Ax0UlK-zbs0:-LAFc8oI1Ax0UlK-zbs1:b-oo1pbv) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L81)):*
> ```Go
> 	flags.StringVar(&o.clusterContext, "cluster-context", "",
> 		"Name of the cluster's context in the local kubeconfig. Defaults to cluster name if unspecified.")
> 	flags.StringVar(&o.secretName, "secret-name", "",
> ```

Why is it desirable to allow a user to specify the secret name?  It might also be helpful to document that the secret will be created in the federation namespace.

---

*[pkg/kubefnord/join.go, line 84 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFcPklB0VIVxphqzo5:-LAFcPklB0VIVxphqzo6:b-fjovtv) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L84)):*
> ```Go
> 		"Name of the secret where the cluster's credentials will be stored in the host cluster. This name should be a valid RFC 1035 label. If unspecified, defaults to a generated name containing the cluster name.")
> 	flags.BoolVar(&o.addToRegistry, "add-to-registry", false,
> 		"Add the cluster to the cluster registry that is aggregated with the kubernetes API server running in the host cluster context.")
> ```

This doc confirms that only a single context and kubeconfig be required to configure join, since an aggregated cluster registry would share the same context as the federation api and host api.  Consider reflecting this in the config loader.

---

*[pkg/kubefnord/join.go, line 125 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFcpWCBymij-oUG0U7:-LAFcpWCBymij-oUG0U8:bx52vx1) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L125)):*
> ```Go
> 
> 	if j.clusterContext == "" {
> 		j.clusterContext = j.ClusterName
> ```

Consider logging that the name is being defaulted.

---

*[pkg/kubefnord/join.go, line 160 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFd5EO4Bcf6I8TDD6a:-LAFd5EO4Bcf6I8TDD6b:by6kcjc) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L160)):*
> ```Go
> 	_, err := config.NewHostConfig(j.Host, j.Kubeconfig)
> 	if err != nil {
> 		glog.V(2).Infof("Failed to get host cluster config: %v", err)
> ```

Consider returning a new error that includes the full text of this info message so that it can be output by the caller.  As written, only the bare error will be shown to the user.  This can wait for a subsequent PR (if ever) given the amount of work involved.  Maybe add a TODO though.

---

*[pkg/kubefnord/join.go, line 412 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFha34DdGISg8cB5La:-LAFha34DdGISg8cB5Lb:b341o9x) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L412)):*
> ```Go
> 	}
> 
> 	fedCluster, err := fedClientset.FederationV1alpha1().FederatedClusters().Create(fedCluster)
> ```

(No action required) Might need to make this namespaced if the CR discussion on Friday ends up with a namespaced cluster resource.

---

*[pkg/kubefnord/join.go, line 441 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFgzsQ5CN5wEL4asZa:-LAFgzsQ5CN5wEL4asZb:b-l8bnmh) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L441)):*
> ```Go
> // getFederationName gets the federation name from the federation
> // deployment.
> func getFederationName(hostClientSet client.Interface, fedNamespace string) (string, error) {
> ```

What exactly is the 'federation name', and what is it used for?

---

*[pkg/kubefnord/join.go, line 461 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFg10iD7TDVNwkKJXg:-LAFg10iD7TDVNwkKJXh:bgixnt0) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L461)):*
> ```Go
> 	}
> 
> 	if dryRun {
> ```

(No action required) I don't really understand the point of dryRun, it seems to add complexity to little benefit.  I'd suggest removing, but it can be done separately given the effort implied.

---

*[pkg/kubefnord/join.go, line 467 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFhouZ0ZveRYepsOeG:-LAFhouZ0ZveRYepsOeH:b-p5rupa) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L467)):*
> ```Go
> 	_, err := clusterClientset.CoreV1().Namespaces().Create(federationNS)
> 	if err != nil && !errors.IsAlreadyExists(err) {
> 		glog.V(2).Infof("Could not create %s namespace in client: %v", federationNamespace, err)
> ```

What does 'in client' mean?

---

*[pkg/kubefnord/join.go, line 514 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFiBMi5T6serBggtW2:-LAFiBMi5T6serBggtW3:b-gr3vfb) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L514)):*
> ```Go
> 	glog.V(2).Info("Created secret in host cluster")
> 
> 	return secret, saName, crb.Name, nil
> ```

Why are `saName` and `crb.Name` returned when the sole caller ignores them?

---

*[pkg/kubefnord/join.go, line 572 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFgZwJBRbEHmGHeyp4:-LAFgZwJBRbEHmGHeyp5:b-x2j5p0) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L572)):*
> ```Go
> 	}
> 
> 	// TODO: This should limit its access to only necessary resources.
> ```

(No action required) What resources shouldn't be available?  Given that any resource could be federated, I'm not sure there's value in being selective.

---

*[pkg/kubefnord/join.go, line 617 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFj0lXBbe2VURHJ_F8:-LAFj0lXBbe2VURHJ_F9:betl4zn) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L617)):*
> ```Go
> // hostClientset, putting it in a secret named secretName in the provided
> // namespace.
> func populateSecretInHostCluster(clusterClientset, hostClientset client.Interface,
> ```

(No action required) Retrieving the secret from one cluster and creating it in another seems ripe for separation into independent methods.

---

*[pkg/kubefnord/kubefnord.go, line 36 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFVzLb0lb9jEOoxcLA:-LAFVzLb0lb9jEOoxcLB:bwtlabu) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/kubefnord.go#L36)):*
> ```Go
> 		Use:   "kubefnord",
> 		Short: "kubefnord controls a Kubernetes Cluster Federation",
> 		Long:  "kubefnord controls a Kubernetes Cluster Federation",
> ```

(No action required) Consider adding a TODO to differentiate the `Long` description from the short.

---

*[pkg/kubefnord/options/options.go, line 38 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAF_kAt5GUoK2HecCSC:-LAF_kAt5GUoK2HecCSD:b-iuaweq) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/options/options.go#L38)):*
> ```Go
> // CommonBind adds the common flags to the flagset passed in.
> func (o *SubcommandOptions) CommonBind(flags *pflag.FlagSet) {
> 	flags.StringVar(&o.Kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
> ```

If only a single context and kubeconfig are provided, why does it make sense to load 3 separate configurations?

---

*[pkg/kubefnord/options/options.go, line 40 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFWrze3TicEtIg8g_S:-LAFWrzf-f918nHGv737:b-fdahux) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/options/options.go#L40)):*
> ```Go
> 	flags.StringVar(&o.Kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
> 	flags.StringVar(&o.Host, "host-cluster-context", "", "Host cluster context")
> 	flags.StringVar(&o.FederationNamespace, "federation-namespace", util.FederationSystemNamespace, "Namespace in the host cluster where the federation system components are installed")
> ```

If this is configurable, how does the system discover it when non-default value is configured?

---

*[pkg/kubefnord/util/util.go, line 32 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFRmWACB64r3MwK3Jb:-LAFXILgAujzKNMW3hsp:b-60t2xa) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/util/util.go#L32)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

There is potential for a common interface like this to be shared across tests and controllers.
</blockquote></details>

Why is it necessary to have separate configurations for the host and federation config?  Won't they always be the same given that federation needs to be aggregated to the host cluster in a deployment (i.e. non-integration) scenario?

Cluster registry is likely to be aggregated in the same manner, at least in the near term, which suggests a single configuration may be desirable.

---

*[pkg/kubefnord/util/util.go, line 64 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFa01k-3yOW1Fny4L4:-LAFa01k-3yOW1Fny4L5:b1yaa0c) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/util/util.go#L64)):*
> ```Go
> // NewHostConfig provides a rest config to talk to the host kubernetes cluster
> // based on the context and kubeconfig passed in.
> func (a *fedConfig) NewHostConfig(context, kubeconfigPath string) (*rest.Config, error) {
> ```

The name of this method suggests a constructor, yet it is setting an instance variable that the `HostConfig` method depends on.  Is there a reason to have this type mix the loading and provision of configuration?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/48#-:-LAFkCoZ7LYuzuItZykg:b59kuyi)*

font:
Yes, lots of it was following existing `kubefed` conventions. This code should be further optimized in the future. I've addressed your comments and will now push another update. PTAL.

---

Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[pkg/controller/util/cluster_util.go, line 24 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFUQDe55Y0-NhbUYSn:-LAFtr5ADVLZN9nqkFLl:b-6c79i7) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/controller/util/cluster_util.go#L24)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

The general convention for imports are the following line-separated sections:

system 

non-kube 3rd party (e.g. glog)

kube ecosystem

My guess this is in relative order of dependency, since system won't depend on anything, 3rd party could depend on system, kube could depend on 3rd party and system.  By that logic, maybe this project should break out its imports to appear after the kube ecosystem?  For now, though, just separate out the glog import.
</blockquote></details>

Makes sense. This was actually `goimports` auto-arranging the imports, but it did not add a line between the 3rd party packages and k8s ecosystem packages.

---

*[pkg/controller/util/cluster_util.go, line 103 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFVZ8JB0BmSaTi4IOX:-LAFuBRl75rgjjnbmg1E:b-7fe5cj) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/controller/util/cluster_util.go#L103)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Is it desirable to continue supporting this legacy mode given that federation will only support kube >= 1.10 where service accounts are the norm?
</blockquote></details>

Unfortunately yes. I had added a comment to this PR on github about this, but it did not carry over to reviewable. Basically, these changes support both integration (legacy) and e2e tests (SAs). We cannot use `JoinCluster` in integration until we have the necessary RBAC controllers, specifically the token controller. Until then, we have to rely on the legacy method to allow integration tests to pass. I have added a comment about this.

---

*[pkg/kubefnord/join.go, line 54 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFbkFUAkBpranGtSfm:-LAFy-ni-OKNVQMlBW-2:b-pkzycw) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L54)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Did you mean `--host-cluster-context`?
</blockquote></details>

Yes, fixed.

---

*[pkg/kubefnord/join.go, line 81 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFc8oI1Ax0UlK-zbs0:-LAFyjGb8U1Y3jKiFOeB:b-g0hac0) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L81)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Why is it desirable to allow a user to specify the secret name?  It might also be helpful to document that the secret will be created in the federation namespace.
</blockquote></details>

This was legacy code from `kubefed`. I can remove it in the future but it's possible somebody would ask for this eventually if that's the reason it was added to `kubefed`.

---

*[pkg/kubefnord/join.go, line 84 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFcPklB0VIVxphqzo5:-LAGuV0pDJVMe4bjuIsA:b-bht4jl) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L84)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

This doc confirms that only a single context and kubeconfig be required to configure join, since an aggregated cluster registry would share the same context as the federation api and host api.  Consider reflecting this in the config loader.
</blockquote></details>

I've updated the config loader to reflect this for now. But we will potentially have to revisit it when we want to call `JoinCluster` from the integration tests. See my comment below about this for more details.

---

*[pkg/kubefnord/join.go, line 125 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFcpWCBymij-oUG0U7:-LAGzjrV27Og1KKlNS0i:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L125)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Consider logging that the name is being defaulted.
</blockquote></details>

Done.

---

*[pkg/kubefnord/join.go, line 160 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFd5EO4Bcf6I8TDD6a:-LAGvCFx8-Nb1ocqCjfu:b9bsjot) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L160)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Consider returning a new error that includes the full text of this info message so that it can be output by the caller.  As written, only the bare error will be shown to the user.  This can wait for a subsequent PR (if ever) given the amount of work involved.  Maybe add a TODO though.
</blockquote></details>

I removed this method and moved the simplified code to `Run`. I've added a TODO for this. Alternatively, we could just output the error directly to `stdout` or lower the default log level for these messages.

---

*[pkg/kubefnord/join.go, line 412 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFha34DdGISg8cB5La:-LAFzyyS-APcXGQWxcDq:brozs7p) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L412)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

(No action required) Might need to make this namespaced if the CR discussion on Friday ends up with a namespaced cluster resource.
</blockquote></details>

Acknowledged, but do you mean for the CR cluster resource? This is for the federatedcluster resource although ideally we need to figure out the UX for the use of both.

---

*[pkg/kubefnord/join.go, line 441 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFgzsQ5CN5wEL4asZa:-LAG0S-p17dZvXqxLK6p:b-vbcfao) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L441)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

What exactly is the 'federation name', and what is it used for?
</blockquote></details>

It's the name of the controller manager that's part of the federation deployment. It is subsequently used to generate the cluster role name for the creation of RBAC objects. This is legacy code as well and is a good candidate for removal because it actually poses problems for calling `JoinCluster` from integration tests too. I've now removed it.

---

*[pkg/kubefnord/join.go, line 461 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFg10iD7TDVNwkKJXg:-LAG1Vkc9_Jxr77mgkIF:b8poqse) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L461)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

(No action required) I don't really understand the point of dryRun, it seems to add complexity to little benefit.  I'd suggest removing, but it can be done separately given the effort implied.
</blockquote></details>

Yeah, legacy code, although lots of k8s tools provide a `--dry-run` option so while it does add complexity, it seems to be part of the UX in the k8s ecosystem. I wasn't enjoying having to test the different code paths during development though. 

---

*[pkg/kubefnord/join.go, line 467 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFhouZ0ZveRYepsOeG:-LAGzh4SEInvXVVWC-6u:bt2vr1b) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L467)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

What does 'in client' mean?
</blockquote></details>

Fixed.

---

*[pkg/kubefnord/join.go, line 514 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFiBMi5T6serBggtW2:-LAG2_x96uzw2SL8mR-H:bzdpn9d) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L514)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Why are `saName` and `crb.Name` returned when the sole caller ignores them?
</blockquote></details>

Artifact left over from refactoring `kubefed`. I've removed for now, but we may want access to them again in the future for testing.

---

*[pkg/kubefnord/join.go, line 572 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFgZwJBRbEHmGHeyp4:-LAG2oHS8OZQOhRbfXnS:bdmajew) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L572)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

(No action required) What resources shouldn't be available?  Given that any resource could be federated, I'm not sure there's value in being selective.
</blockquote></details>

I left this TODO in from `kubefed` as it seemed applicable because this provides full privileges to the joining cluster and seems like it may not be conducive to a multi-tenancy environment.

---

*[pkg/kubefnord/kubefnord.go, line 36 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFVzLb0lb9jEOoxcLA:-LAG6zho7Qr4WKCLk_4H:biwccwb) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/kubefnord.go#L36)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

(No action required) Consider adding a TODO to differentiate the `Long` description from the short.
</blockquote></details>

Added the URL to the repo for more info.

---

*[pkg/kubefnord/options/options.go, line 38 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAF_kAt5GUoK2HecCSC:-LAGxrP83otNfGnplCRO:bxk4hr4) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/options/options.go#L38)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

If only a single context and kubeconfig are provided, why does it make sense to load 3 separate configurations?
</blockquote></details>

It doesn't for just the CLI. But when I was working through this I was also trying to integrate it with integration tests because that was part of the original goal. With integration tests, each fixture has its own config to the loopback interface on different ports for example. Since we cannot integrate with integration tests for now, I've simplified the code to only use host and cluster config and we can add the others back when/if needed.

---

*[pkg/kubefnord/options/options.go, line 40 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFWrze3TicEtIg8g_S:-LAG3tSi0WZO7UOmcv41:b3whpta) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/options/options.go#L40)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

If this is configurable, how does the system discover it when non-default value is configured?
</blockquote></details>

Yep, we need a way to discover it. It's actually an existing TODO that you created in `pkg/controller/util/cluster_util.go`. I think it needs to be allowed to be configurable to avoid any namespace name clashes, particularly in a multi-tenancy environment. It seems like having support for nested-namespaces would make this simpler.

---

*[pkg/kubefnord/util/util.go, line 32 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFRmWACB64r3MwK3Jb:-LAG4Vw83liu8q8TPTnb:b-hna48l) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/util/util.go#L32)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Why is it necessary to have separate configurations for the host and federation config?  Won't they always be the same given that federation needs to be aggregated to the host cluster in a deployment (i.e. non-integration) scenario?

Cluster registry is likely to be aggregated in the same manner, at least in the near term, which suggests a single configuration may be desirable.
</blockquote></details>

See my comment above about integration tests being the reason for it. This code has now been simplified too.

---

*[pkg/kubefnord/util/util.go, line 64 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFa01k-3yOW1Fny4L4:-LAGy6i-AbvmEfvwb859:b-w5qq0w) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/util/util.go#L64)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

The name of this method suggests a constructor, yet it is setting an instance variable that the `HostConfig` method depends on.  Is there a reason to have this type mix the loading and provision of configuration?
</blockquote></details>

Originally it was to avoid having multiple calls to construct a new instance of a config even though the garbage collector should take care of that. I've now simplified this code and made it more straight-forward.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/48#-:-LAH25uI4qFQMOQ_V2GC:b-2dsao7)*

marun:
Thank you for cleanup! There are a few tiny things to fix and then I think this is ready.

---

Reviewed 2 of 3 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[pkg/controller/util/cluster_util.go, line 24 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFUQDe55Y0-NhbUYSn:-LAJ_1Zc6NH-q60njMOJ:b4e2eej) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/controller/util/cluster_util.go#L24)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Makes sense. This was actually `goimports` auto-arranging the imports, but it did not add a line between the 3rd party packages and k8s ecosystem packages.
</blockquote></details>

Please add a line break to separate the glog import.

---

*[pkg/kubefnord/join.go, line 54 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFbkFUAkBpranGtSfm:-LAJ_7XCCTBWZQAGYJ1x:b-bsumq7) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L54)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Yes, fixed.
</blockquote></details>

`--host-cluster-context`?

---

*[pkg/kubefnord/join.go, line 81 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFc8oI1Ax0UlK-zbs0:-LAJ_CZW9Q10rgvI7vdO:bi1aj88) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L81)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

This was legacy code from `kubefed`. I can remove it in the future but it's possible somebody would ask for this eventually if that's the reason it was added to `kubefed`.
</blockquote></details>

Fair enough for now, but let's remove it in the future.  I haven't heard of any use case that requires that the user to control the name of a system-only secret.

---

*[pkg/kubefnord/join.go, line 412 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFha34DdGISg8cB5La:-LAJaLZQ3dlS--IzF-YQ:bjjgo0l) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L412)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Acknowledged, but do you mean for the CR cluster resource? This is for the federatedcluster resource although ideally we need to figure out the UX for the use of both.
</blockquote></details>

If the CR cluster gets namespaced, I think it will make sense to make the FederatedCluster similarly namespaced for uniformity with the 2 resources it references (cluster and secret).

---

*[pkg/kubefnord/join.go, line 461 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFg10iD7TDVNwkKJXg:-LAJah8d6f1CXZutoumK:b5089sc) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L461)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Yeah, legacy code, although lots of k8s tools provide a `--dry-run` option so while it does add complexity, it seems to be part of the UX in the k8s ecosystem. I wasn't enjoying having to test the different code paths during development though. 
</blockquote></details>

I don't think we need to cargo cult kube convention on stuff like this, and follow YAGNI wherever possible.

---

*[pkg/kubefnord/join.go, line 514 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFiBMi5T6serBggtW2:-LAJayPp9FO5tsJ-Bqm3:b6vlcet) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L514)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Artifact left over from refactoring `kubefed`. I've removed for now, but we may want access to them again in the future for testing.
</blockquote></details>

I think its likely that refactoring for testing will imply setting up the service account separately from creating the secret in the host cluster.

---

*[pkg/kubefnord/join.go, line 572 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFgZwJBRbEHmGHeyp4:-LAJbFEI9VdVk4-Zobpu:b-fbd28q) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L572)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

I left this TODO in from `kubefed` as it seemed applicable because this provides full privileges to the joining cluster and seems like it may not be conducive to a multi-tenancy environment.
</blockquote></details>

I don't think there is another option for a push reconciler.  The cost of doing per-namespace reconciliation could be considerable and where more granular access control is required a pull reconciler running in a given cluster may be preferable to to the reduced scope involved.

---

*[pkg/kubefnord/options/options.go, line 40 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFWrze3TicEtIg8g_S:-LAJbn8nAHujj39F3q70:bdxnnxd) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/options/options.go#L40)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Yep, we need a way to discover it. It's actually an existing TODO that you created in `pkg/controller/util/cluster_util.go`. I think it needs to be allowed to be configurable to avoid any namespace name clashes, particularly in a multi-tenancy environment. It seems like having support for nested-namespaces would make this simpler.
</blockquote></details>

I don't think hierarchical namespacing is going to be a thing anytime soon.  :/

---

*[pkg/kubefnord/util/util.go, line 40 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAJctTK3Cu929fdJU05:-LAJctTK3Cu929fdJU06:b-mdc8ak) ([raw file](https://github.com/marun/federation-v2/blob/2d8357ed982a0f2d13726874b03a75b7849525dc/pkg/kubefnord/util/util.go#L40)):*
> ```Go
> type fedConfig struct {
> 	pathOptions   *clientcmd.PathOptions
> 	hostConfig    *rest.Config
> ```

It appears that these instance variables are unused and can be removed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/48#-:-LAJdMtr0eaFHA1pBuSL:b7r8gjs)*

font:
No problem! Addressed the last remaining things and I've further simplified the `FedConfig` interface and removed the `FedClientset` interface. Will now push an update. PTAL.

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[pkg/controller/util/cluster_util.go, line 24 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFUQDe55Y0-NhbUYSn:-LAJoK48EKSmMVqhBXdR:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/controller/util/cluster_util.go#L24)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Please add a line break to separate the glog import.
</blockquote></details>

Done.

---

*[pkg/kubefnord/join.go, line 54 at r1](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAFbkFUAkBpranGtSfm:-LAJhbDe1YjcRZDfbWrx:bg5gwno) ([raw file](https://github.com/marun/federation-v2/blob/588f26a79fee36a054b4f2dd738ed97c4a5e5298/pkg/kubefnord/join.go#L54)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

`--host-cluster-context`?
</blockquote></details>

Really fixed now.

---

*[pkg/kubefnord/util/util.go, line 40 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAJctTK3Cu929fdJU05:-LAJlMObEn1vrVADZrSH:b36bo37) ([raw file](https://github.com/marun/federation-v2/blob/2d8357ed982a0f2d13726874b03a75b7849525dc/pkg/kubefnord/util/util.go#L40)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

It appears that these instance variables are unused and can be removed.
</blockquote></details>

`pathOptions` is used and after simplifying this the rest can be removed. I've also removed the `FedClientset` interface.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/48#-:-LAJvXOP93fTo-vGcKSX:bw6s2k6)*

marun:
:lgtm:

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[pkg/kubefnord/util/util.go, line 40 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/48#-LAJctTK3Cu929fdJU05:-LAJwfu30lyF3URKdXh9:b-q74txj) ([raw file](https://github.com/marun/federation-v2/blob/2d8357ed982a0f2d13726874b03a75b7849525dc/pkg/kubefnord/util/util.go#L40)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

`pathOptions` is used and after simplifying this the rest can be removed. I've also removed the `FedClientset` interface.
</blockquote></details>

great!

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/48#-:-LAJwo9yCviap6PdJVXX:bnfp4nl)*

PR #49, opened 2018-04-13
PropagatedVersion changes to support namespaces

font:
- Changes to support`PropagatedVersion` for namespaces.
- Adds crudtester check for primary cluster so that we can run E2E tests with or without the host cluster being part of the federation.

---
This change is [Reviewable](https://reviewable.io/reviews/marun/federation-v2/49)

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -426,6 +426,17 @@ func (s *FederationSyncController) reconcile(qualifiedName federatedtypes.Qualif
 	templateAdapter := s.adapter.Template()
 	kind := templateAdapter.Kind()
 	key := qualifiedName.String()
+	namespace := qualifiedName.Namespace
+
+	if federatedtypes.IsNamespaceKind(kind) {
+		namespace = qualifiedName.Name
+		// TODO(font): Need a configurable or discoverable list of namespaces
+		// to not propagate beyond just the default system namespaces e.g.
+		// clusterregistry.
+		if isSystemNamespace(namespace) {
+			return statusAllOK
+		}
+	}
 
 	glog.V(4).Infof("Starting to reconcile %v %v", kind, key)
 	startTime := time.Now()
@@ -480,7 +491,7 @@ func (s *FederationSyncController) reconcile(qualifiedName federatedtypes.Qualif
 	}
 
 	propagatedVersionKey := federatedtypes.QualifiedName{
-		Namespace: qualifiedName.Namespace,
+		Namespace: namespace,
 		Name:      s.versionName(qualifiedName.Name),
 	}.String()
 	// TODO(marun) LOCK!
@@ -537,7 +548,7 @@ func (s *FederationSyncController) delete(template pkgruntime.Object, meta *meta
 		return err
 	}
 
-	if kind == federatedtypes.NamespaceKind {
+	if federatedtypes.IsNamespaceKind(kind) {
 		// Return immediately if we are a namespace as it will be deleted
 		// simply by removing its finalizers.
 		return nil
@@ -658,9 +669,11 @@ func updatePropagatedVersion(adapter federatedtypes.FederatedTypeAdapter,
 		if err != nil {
 			return err
 		}
+
 		key := federatedtypes.NewQualifiedName(version).String()
 		// TODO(marun) add timeout to ensure against lost updates blocking propagation of a given resource
 		pendingVersionUpdates.Insert(key)
+
 		_, err = adapter.FedClient().FederationV1alpha1().PropagatedVersions(version.Namespace).UpdateStatus(version)
 		if err != nil {
 			pendingVersionUpdates.Delete(key)
@@ -688,6 +701,7 @@ func updatePropagatedVersion(adapter federatedtypes.FederatedTypeAdapter,
 
 	key := federatedtypes.NewQualifiedName(version).String()
 	pendingVersionUpdates.Insert(key)
+
 	_, err := adapter.FedClient().FederationV1alpha1().PropagatedVersions(version.Namespace).UpdateStatus(version)
 	if err != nil {
 		pendingVersionUpdates.Delete(key)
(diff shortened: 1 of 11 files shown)


pkg/controller/sync/controller.go
```diff
@@ -658,9 +669,11 @@ func updatePropagatedVersion(adapter federatedtypes.FederatedTypeAdapter,
 		if err != nil {
 			return err
 		}
+
```
marun: (No action required) functional vs non-functional...

test/common/crudtester.go
```diff
@@ -158,15 +163,14 @@ func (c *FederatedTypeCrudTester) CheckUpdate(template, placement, override pkgr
...
-	kind := adapter.Kind()
 
 	var initialAnnotation string
 	meta := adapter.ObjectMeta(template)
 	if meta.Annotations != nil {
 		initialAnnotation = meta.Annotations[AnnotationTestFederationCrudUpdate]
 	}
 
-	c.tl.Logf("Updating %s %q", kind, qualifiedName)
+	c.tl.Logf("Updating %s %q", c.kind, qualifiedName)
```
marun: (No action required) Again, from a review perspective, this kind of change would ideally be separated out.

test/common/crudtester.go
```diff
@@ -412,11 +422,20 @@ func (c *FederatedTypeCrudTester) waitForPropagatedVersion(template, placement,
 
 	clusterNames := c.adapter.Placement().ClusterNames(placement)
 	selectedClusters := sets.NewString(clusterNames...)
+	if c.kind == federatedtypes.NamespaceKind {
+		// Delete the primary cluster as it will never be included in
+		// PropagatedVersion's list of cluster verions.
```
marun: s/verions/versions/

test/common/crudtester.go
```diff
@@ -412,11 +422,20 @@ func (c *FederatedTypeCrudTester) waitForPropagatedVersion(template, placement,
...
+		selectedClusters.Delete(c.getPrimaryClusterName())
+	}
 
 	var version *fedv1a1.PropagatedVersion
 	err := wait.PollImmediate(c.waitInterval, c.clusterWaitTimeout, func() (bool, error) {
 		var err error
-		version, err = client.FederationV1alpha1().PropagatedVersions(namespace).Get(name, metav1.GetOptions{})
+		if c.kind == federatedtypes.NamespaceKind {
```
marun: (No action required) Might make sense to add a helper for this pattern in a future PR.

test/common/crudtester.go
```diff
@@ -451,7 +470,35 @@ func (c *FederatedTypeCrudTester) versionName(resourceName string) string {
...
+			continue
+		} else {
+			*clusterNames = append((*clusterNames)[:i], (*clusterNames)[i+1:]...)
+			break
+		}
+	}
+}
+
+func (c *FederatedTypeCrudTester) propagatedVersion(version *fedv1a1.PropagatedVersion, clusterName string) string {
+	// The primary cluster's namespace version is stored as the
```
marun: Can you rephrase this?  It doesn't make sense to me as-is.

test/e2e/framework/unmanaged.go
```diff
@@ -95,28 +97,33 @@ func (f *UnmanagedFramework) AfterEach() {
...
-	// Print events if the test failed and ran in a namep.
+	// Print events if the test failed and ran in a namespace.
 	if CurrentGinkgoTestDescription().Failed && f.testNamespaceName != "" {
 		kubeClient := f.KubeClient(userAgent)
 		DumpEventsInNamespace(func(opts metav1.ListOptions, ns string) (*corev1.EventList, error) {
 			return kubeClient.Core().Events(ns).List(opts)
 		}, f.testNamespaceName)
 	}
+
+	// DeleteNamespace at the very end but before tearing down the namespace
```
marun: Cleanup needs to remain as a defer function because `DumpEventsInNamespace` contains assertions that could exit the function.

test/e2e/framework/unmanaged.go
```diff
@@ -152,15 +159,29 @@ func (f *UnmanagedFramework) ClusterClients(userAgent string) map[string]kubecli
...
-	clusterClients := make(map[string]kubeclientset.Interface)
+	testClusters := make(map[string]common.TestCluster)
 	for _, cluster := range clusterList.Items {
 		ClusterIsReadyOrFail(fedClient, cluster)
 		config, err := util.BuildClusterConfig(&cluster, kubeClient, crClient)
 		Expect(err).NotTo(HaveOccurred())
 		restclient.AddUserAgent(config, userAgent)
-		clusterClients[cluster.Name] = kubeclientset.NewForConfigOrDie(config)
+
+		// Check if this cluster is the same name as the current-context to
```
marun: Is this the best way to do this in the absence of a first-class cluster id?  I'm wondering if it might make sense to require a configuration parameter naming the host cluster instead.

test/e2e/framework/unmanaged.go
```diff
@@ -152,15 +159,29 @@ func (f *UnmanagedFramework) ClusterClients(userAgent string) map[string]kubecli
...
+		// make it the primary cluster.
+		if cluster.Name == f.Kubeconfig.CurrentContext {
+			testClusters[cluster.Name] = common.TestCluster{
```
marun: Also, consider simplifying: 

```golang
testClusters[cluster.Name] = common.TestCluster{
  kubeclientset.NewForConfigOrDie(config),
  (cluster.Name == hostClusterName),
}

File: test/common/crudtester.go (hunk 2 above)
Comment (font) [reply test/common/crudtester.go]: I definitely understand and mostly agree on your desire to have these things separated out from a review perspective. But it's too tedious to go back and undo each non-functional change from each individual commit, especially when a lot of this code was written during a mostly prototype phase. Moving forward, you can expect these things to be separated out.

File: test/common/crudtester.go (hunk 3 above)
Comment (font) [reply test/common/crudtester.go]: Fixed.

File: test/common/crudtester.go (hunk 5 above)
Comment (font) [reply test/common/crudtester.go]: Updated with the below phrasing. Let me know if that is more clear.
```
	// For namespaces, since we do not store the primary cluster's namespace
	// version in PropagatedVersion's ClusterVersions slice, grab it from the
	// TemplateVersion field instead.
```

test/e2e/framework/unmanaged.go
font: Ah, interesting. I added a comment about that and moved the original code, plus the fixture cleanup which depends on #48, to a defer function.

PR #50, opened 2018-04-13
Add hybrid mode to E2E testing

font:
**NOTE: This should ideally be merged immediately after #48 but before #49.**

This adds support for running E2E tests with in-memory-controllers in an unmanaged federation (hybrid).

### DIFF
--- a/test/e2e/framework/test_context.go
+++ b/test/e2e/framework/test_context.go
@@ -33,10 +33,14 @@ type TestContextType struct {
 var TestContext TestContextType
 
 func registerFlags(t *TestContextType) {
-	flag.BoolVar(&t.TestManagedFederation, "test-managed-federation", false, "Whether the test run should rely on a test-managed federation.")
-	flag.BoolVar(&t.InMemoryControllers, "in-memory-controllers", true, "Whether federation controllers should be started in memory.  Ignored if test-managed-cluster is false")
-	flag.StringVar(&t.KubeConfig, "kubeconfig", os.Getenv("KUBECONFIG"), "Path to kubeconfig containing embedded authinfo.  Ignored if test-managed-cluster is true")
-	flag.StringVar(&t.KubeContext, "context", "", "kubeconfig context to use/override. If unset, will use value from 'current-context'")
+	flag.BoolVar(&t.TestManagedFederation, "test-managed-federation",
+		false, "Whether the test run should rely on a test-managed federation.")
+	flag.BoolVar(&t.InMemoryControllers, "in-memory-controllers", false,
+		"Whether federation controllers should be started in memory. This works like a hybrid setup if test-managed-federation is false by launching the federation controllers outside the unmanaged cluster.")
+	flag.StringVar(&t.KubeConfig, "kubeconfig", os.Getenv("KUBECONFIG"),
+		"Path to kubeconfig containing embedded authinfo.  Ignored if test-managed-federation is true.")
+	flag.StringVar(&t.KubeContext, "context", "",
+		"kubeconfig context to use/override. If unset, will use value from 'current-context'.")
 }
 
 func validateFlags(t *TestContextType) {
@@ -48,8 +52,7 @@ func validateFlags(t *TestContextType) {
 	}
 
 	if !t.TestManagedFederation && t.InMemoryControllers {
-		glog.Info("in-memory-controllers require test-managed-federation=true - disabling.")
-		t.InMemoryControllers = false
+		glog.Info("in-memory-controllers=true while test-managed-federation=false - this will launch the federation controllers outside the unmanaged cluster.")
 	}
 }
 
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -52,8 +52,14 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 		framework.SetUpManagedFederation()
 		// TODO(marun) support parallel execution by sending
 		// configuration for test-managed federation to other nodes
+	} else if framework.TestContext.InMemoryControllers {
+		// This supports a hybrid setup where the user asks for an unmanaged
+		// federation with in memory controllers. This means the k8s clusters
+		// are already running along with the federation cluster registry API
+		// servers. This will then join the clusters and launch the required
+		// federation controllers e.g. cluster and sync controllers.
+		framework.SetUpUnManagedFederation()
 	}
-	// TODO(marun) support deployed federation
 
 	return nil
 
@@ -73,5 +79,7 @@ var _ = ginkgo.SynchronizedAfterSuite(func() {
 	framework.Logf("Running AfterSuite actions on node 1")
 	if framework.TestContext.TestManagedFederation {
 		framework.TearDownManagedFederation()
+	} else if framework.TestContext.InMemoryControllers {
+		framework.TearDownUnManagedFederation()
 	}
 })
--- a/test/e2e/framework/managed.go
+++ b/test/e2e/framework/managed.go
@@ -29,7 +29,6 @@ import (
 
 var (
 	fedFixture *framework.FederationFixture
-	logger     common.TestLogger
 )
 
 func SetUpManagedFederation() {
@@ -106,6 +105,6 @@ func (f *ManagedFramework) SetUpControllerFixture(kind string, adapterFactory fe
 	fedConfig := fedFixture.FedApi.NewConfig(f.logger)
 	kubeConfig := fedFixture.KubeApi.NewConfig(f.logger)
 	crConfig := fedFixture.CrApi.NewConfig(f.logger)
-	fixture := framework.NewControllerFixture(f.logger, kind, adapterFactory, fedConfig, kubeConfig, crConfig)
+	fixture := framework.NewSyncControllerFixture(f.logger, kind, adapterFactory, fedConfig, kubeConfig, crConfig)
 	f.fixtures = append(f.fixtures, fixture)
 }
--- a/test/e2e/framework/unmanaged.go
+++ b/test/e2e/framework/unmanaged.go
@@ -4,7 +4,6 @@ Copyright 2018 The Kubernetes Authors.
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
     http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
@@ -26,6 +25,9 @@ import (
 	fedclientset "github.com/marun/federation-v2/pkg/client/clientset_generated/clientset"
 	"github.com/marun/federation-v2/pkg/controller/util"
 	"github.com/marun/federation-v2/pkg/federatedtypes"
+	"github.com/marun/federation-v2/pkg/kubefnord"
+	"github.com/marun/federation-v2/test/common"
+	"github.com/marun/federation-v2/test/integration/framework"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,6 +35,7 @@ import (
 	kubeclientset "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	crclientset "k8s.io/cluster-registry/pkg/client/clientset_generated/clientset"
 
 	. "github.com/onsi/ginkgo"
@@ -40,7 +43,41 @@ import (
 )
 
 // Only check that the api is available once
-var checkedApi bool
+var (
+	clusterControllerFixture *framework.ControllerFixture
+	checkedApi               bool
+)
+
+func SetUpUnManagedFederation() {
+	if clusterControllerFixture != nil {
+		return
+	}
+
+	if TestContext.KubeConfig == "" {
+		Failf("--kubeconfig or KUBECONFIG must be specified to load client config")
+	}
+
+	config, kubeConfig, err := loadConfig(TestContext.KubeConfig, TestContext.KubeContext)
+	Expect(err).NotTo(HaveOccurred())
+
+	clusterControllerFixture = framework.NewClusterControllerFixture(config, config, config)
+
+	// TODO(font): Support joining multiple clusters.
+	hostClusterName := kubeConfig.CurrentContext
+	err = kubefnord.JoinCluster(config, config, config,
+		util.FederationSystemNamespace, hostClusterName, hostClusterName, "", true, false)
+	if err != nil {
+		Failf("Failed to join cluster %s: %v", hostClusterName, err)
+	}
+}
+
+func TearDownUnManagedFederation() {
+	if clusterControllerFixture != nil {
+		clusterControllerFixture.TearDown(NewE2ELogger())
+		clusterControllerFixture = nil
+	}
+	// TODO(font): Unjoin clusters.
+}
 
 type UnmanagedFramework struct {
 	// To make sure that this framework cleans up after itself, no matter what,
@@ -53,11 +90,18 @@ type UnmanagedFramework struct {
 	Config *restclient.Config
 
 	BaseName string
+
+	logger common.TestLogger
+
+	// Fixtures to cleanup after each test
+	fixtures []framework.TestFixture
 }
 
 func NewUnmanagedFramework(baseName string) FederationFramework {
 	f := &UnmanagedFramework{
 		BaseName: baseName,
+		logger:   NewE2ELogger(),
+		fixtures: []framework.TestFixture{},
 	}
 	return f
 }
@@ -74,7 +118,7 @@ func (f *UnmanagedFramework) BeforeEach() {
 			Failf("--kubeconfig or KUBECONFIG must be specified to load client config")
 		}
 		var err error
-		f.Config, err = loadConfig(TestContext.KubeConfig, TestContext.KubeContext)
+		f.Config, _, err = loadConfig(TestContext.KubeConfig, TestContext.KubeContext)
 		Expect(err).NotTo(HaveOccurred())
 	}
 
@@ -110,7 +154,13 @@ func (f *UnmanagedFramework) AfterEach() {
 		deleteNamespace(client, namespaceName)
 	}()
 
-	// Print events if the test failed and ran in a namep.
+	for len(f.fixtures) > 0 {
+		fixture := f.fixtures[0]
+		fixture.TearDown(f.logger)
+		f.fixtures = append(f.fixtures[:0], f.fixtures[1:]...)
+	}
+
+	// Print events if the test failed and ran in a namespace.
 	if CurrentGinkgoTestDescription().Failed && f.testNamespaceName != "" {
 		kubeClient := f.KubeClient(userAgent)
 		DumpEventsInNamespace(func(opts metav1.ListOptions, ns string) (*corev1.EventList, error) {
@@ -176,7 +226,14 @@ func (f *UnmanagedFramework) TestNamespaceName() string {
 }
 
 func (f *UnmanagedFramework) SetUpControllerFixture(kind string, adapterFactory federatedtypes.AdapterFactory) {
-	// Not supported
+	// Hybrid setup where just the sync controller is run and we do not rely on
+	// the already deployed (unmanaged) controller manager. Only do this if
+	// in-memory-controllers is true.
+	if TestContext.InMemoryControllers {
+		fixture := framework.NewSyncControllerFixture(f.logger, kind, adapterFactory, f.Config,
+			f.Config, f.Config)
+		f.fixtures = append(f.fixtures, fixture)
+	}
 }
 
 func createNamespace(client kubeclientset.Interface, baseName string) (string, error) {
@@ -234,11 +291,11 @@ func deleteNamespace(client kubeclientset.Interface, namespaceName string) {
 	}
 }
 
-func loadConfig(configPath, context string) (*restclient.Config, error) {
+func loadConfig(configPath, context string) (*restclient.Config, *clientcmdapi.Config, error) {
 	Logf(">>> kubeConfig: %s", configPath)
 	c, err := clientcmd.LoadFromFile(configPath)
 	if err != nil {
-		return nil, fmt.Errorf("error loading kubeConfig %s: %v", configPath, err.Error())
+		return nil, nil, fmt.Errorf("error loading kubeConfig %s: %v", configPath, err.Error())
 	}
 	if context != "" {
 		Logf(">>> kubeContext: %s", context)
@@ -246,9 +303,9 @@ func loadConfig(configPath, context string) (*restclient.Config, error) {
 	}
 	cfg, err := clientcmd.NewDefaultClientConfig(*c, &clientcmd.ConfigOverrides{}).ClientConfig()
 	if err != nil {
-		return nil, fmt.Errorf("error creating default client config: %v", err.Error())
+		return nil, nil, fmt.Errorf("error creating default client config: %v", err.Error())
 	}
-	return cfg, nil
+	return cfg, c, nil
 }
 
 // waitForApiserver waits for the apiserver to be ready.  It tests the
@@ -306,14 +363,15 @@ func ClusterIsReadyOrFail(client fedclientset.Interface, cluster fedv1a1.Federat
 	clusterName := cluster.Name
 	By(fmt.Sprintf("Checking readiness of cluster %q", clusterName))
 	err := wait.PollImmediate(PollInterval, SingleCallTimeout, func() (bool, error) {
+		cluster, err := client.FederationV1alpha1().FederatedClusters().Get(clusterName,
+			metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
 		for _, condition := range cluster.Status.Conditions {
 			if condition.Type == fedcommon.ClusterReady && condition.Status == corev1.ConditionTrue {
 				return true, nil
 			}
-		}
-		_, err := client.FederationV1alpha1().FederatedClusters().Get(clusterName, metav1.GetOptions{})
-		if err != nil {
-			return false, err
 		}
 		return false, nil
 	})
--- a/test/e2e/framework/util.go
+++ b/test/e2e/framework/util.go
@@ -32,7 +32,7 @@ const (
 	// transient failures from failing tests.
 	// TODO: client should not apply this timeout to Watch calls. Increased from 30s until that is fixed.
 	PollInterval      = 2 * time.Second
-	SingleCallTimeout = 5 * time.Minute
+	SingleCallTimeout = 30 * time.Second
 )
 
 // unique identifier of the e2e run
--- a/test/integration/crud_test.go
+++ b/test/integration/crud_test.go
@@ -60,7 +60,7 @@ func initCrudTest(tl common.TestLogger, fedFixture *framework.FederationFixture,
 	fedConfig := fedFixture.FedApi.NewConfig(tl)
 	kubeConfig := fedFixture.KubeApi.NewConfig(tl)
 	crConfig := fedFixture.CrApi.NewConfig(tl)
-	fixture := framework.NewControllerFixture(tl, kind, adapterFactory, fedConfig, kubeConfig, crConfig)
+	fixture := framework.NewSyncControllerFixture(tl, kind, adapterFactory, fedConfig, kubeConfig, crConfig)
 
 	userAgent := fmt.Sprintf("crud-test-%s", kind)
 
--- a/test/integration/framework/controller.go
+++ b/test/integration/framework/controller.go
@@ -17,6 +17,9 @@ limitations under the License.
 package framework
 
 import (
+	"time"
+
+	"github.com/marun/federation-v2/pkg/controller/federatedcluster"
 	"github.com/marun/federation-v2/pkg/controller/sync"
 	"github.com/marun/federation-v2/pkg/federatedtypes"
 	"github.com/marun/federation-v2/test/common"
@@ -28,8 +31,8 @@ type ControllerFixture struct {
 	stopChan chan struct{}
 }
 
-// NewControllerFixture initializes a new controller fixture
-func NewControllerFixture(tl common.TestLogger, kind string, adapterFactory federatedtypes.AdapterFactory, fedConfig, kubeConfig, crConfig *restclient.Config) *ControllerFixture {
+// NewSyncControllerFixture initializes a new sync controller fixture.
+func NewSyncControllerFixture(tl common.TestLogger, kind string, adapterFactory federatedtypes.AdapterFactory, fedConfig, kubeConfig, crConfig *restclient.Config) *ControllerFixture {
 	f := &ControllerFixture{
 		stopChan: make(chan struct{}),
 	}
@@ -37,6 +40,17 @@ func NewControllerFixture(tl common.TestLogger, kind string, adapterFactory fede
 	if err != nil {
 		tl.Fatalf("Error starting sync controller: %v", err)
 	}
+	return f
+}
+
+// NewClusterControllerFixture initializes a new cluster controller fixture.
+func NewClusterControllerFixture(fedConfig, kubeConfig, crConfig *restclient.Config) *ControllerFixture {
+	f := &ControllerFixture{
+		stopChan: make(chan struct{}),
+	}
+	monitorPeriod := 1 * time.Second
+	federatedcluster.StartClusterController(fedConfig, kubeConfig, crConfig,
+		f.stopChan, monitorPeriod)
 	return f
 }
 
--- a/test/integration/framework/federation.go
+++ b/test/integration/framework/federation.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	fedv1a1 "github.com/marun/federation-v2/pkg/apis/federation/v1alpha1"
-	"github.com/marun/federation-v2/pkg/controller/federatedcluster"
 	"github.com/marun/federation-v2/pkg/controller/util"
 	"github.com/marun/federation-v2/test/common"
 	apiv1 "k8s.io/api/core/v1"
@@ -41,13 +40,11 @@ const userAgent = "federation-framework"
 // FederationFixture manages servers for kube, cluster registry and
 // federation along with a set of member clusters.
 type FederationFixture struct {
-	stopChan chan struct{}
-
-	KubeApi *KubernetesApiFixture
-	CrApi   *ClusterRegistryApiFixture
-	FedApi  *FederationApiFixture
-
-	Clusters map[string]*KubernetesApiFixture
+	KubeApi           *KubernetesApiFixture
+	CrApi             *ClusterRegistryApiFixture
+	FedApi            *FederationApiFixture
+	Clusters          map[string]*KubernetesApiFixture
+	ClusterController *ControllerFixture
 }
 
 func SetUpFederationFixture(tl common.TestLogger, clusterCount int) *FederationFixture {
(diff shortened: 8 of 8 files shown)


test/e2e/framework/test_context.go
```diff
@@ -33,10 +33,14 @@ type TestContextType struct {
 var TestContext TestContextType
 
 func registerFlags(t *TestContextType) {
-	flag.BoolVar(&t.TestManagedFederation, "test-managed-federation", false, "Whether the test run should rely on a test-managed federation.")
-	flag.BoolVar(&t.InMemoryControllers, "in-memory-controllers", true, "Whether federation controllers should be started in memory.  Ignored if test-managed-cluster is false")
-	flag.StringVar(&t.KubeConfig, "kubeconfig", os.Getenv("KUBECONFIG"), "Path to kubeconfig containing embedded authinfo.  Ignored if test-managed-cluster is true")
-	flag.StringVar(&t.KubeContext, "context", "", "kubeconfig context to use/override. If unset, will use value from 'current-context'")
+	flag.BoolVar(&t.TestManagedFederation, "test-managed-federation",
+		false, "Whether the test run should rely on a test-managed federation.")
+	flag.BoolVar(&t.InMemoryControllers, "in-memory-controllers", false,
```
font: The `in-memory-controllers flag` is now false by default.

marun:
Yay easier debugging!  Aside from some minor fixes, I have a concern (inline) about not having automatically joined clusters cleaned up on teardown.

---

Reviewed 7 of 8 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[test/e2e/e2e.go, line 61 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOqaCj0Y9xj_yLJIn0:-LAOqaCj0Y9xj_yLJIn1:bq4ecur) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/e2e.go#L61)):*
> ```Go
> 		// API servers. This will then join the clusters and launch the
> 		// required federation controllers e.g. cluster and sync controllers.
> 		framework.SetUpUnManagedFederation()
> ```

s/UnManaged/Unmanaged/

---

*[test/e2e/framework/unmanaged.go, line 57 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOzyra0QxBndwjxxeS:-LAOzyra0QxBndwjxxeT:b-4l6fu) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L57)):*
> ```Go
> 	}
> 
> 	if TestContext.KubeConfig == "" {
> ```

This (and the related check on 119) are not necessary.  The logic in `test_context.go:validateFlags` ensures that `TestManagedFederation = true` if `KubeConfig == ""`, so the unmanaged framework will never be used if kubeconfig is not provided.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAP-Zs69V8sukkrKmfm:b-3kqc49) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
> ```Go
> 	clusterControllerFixture = framework.NewClusterControllerFixture(config, config, config)
> 
> 	// TODO(font): Support joining multiple clusters.
> ```

I think configuration of joining/unjoining needs to be orthogonal from deployment of in-memory controllers.  Do you think you'll have time to rectify this in this PR or do your other responsibilities suggest that I implement it in a subsequent PR? 

---

*[test/e2e/framework/unmanaged.go, line 80 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP06wP3f3kkAJy01hF:-LAP06wP3f3kkAJy01hG:bvn4hhw) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L80)):*
> ```Go
> 		clusterControllerFixture = nil
> 	}
> 	// TODO(font): Unjoin clusters.
> ```

Without unjoin it's not possible to repeatedly run the e2e tests against a given cluster without manual cleanup.  I think either unjoin needs to be implemented (which implies explicit configuration to avoid unjoining manually joined clusters) or automatic join needs to be removed pending implementation of unjoin. 

---

*[test/e2e/framework/unmanaged.go, line 161 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOxp_GBjlK_8MGHQjJ:-LAOxp_GBjlK_8MGHQjK:bblxs2r) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L161)):*
> ```Go
> 		deleteNamespace(client, namespaceName)
> 
> 		for _, fixture := range f.fixtures {
> ```

This deferred function is only required to ensure best-effort namespace removal in the event of failure during dumping of events.  Since fixture teardown is orthogonal (and as written the fixtures won't be torn down if any error occurs on namespace removal), consider performing fixture teardown outside of this function and in advance of event dumping.

Also,`f.fixtures` needs to be cleaned up (remove fixtures from the slice as they are torn down) since the framework and its slice are reused across tests.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAOsw3U-z-fv_HPqLjt:b-2n39sr) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
> ```Go
> 	By(fmt.Sprintf("Checking readiness of cluster %q", clusterName))
> 	err := wait.PollImmediate(PollInterval, SingleCallTimeout, func() (bool, error) {
> 		cluster, err := client.FederationV1alpha1().FederatedClusters().Get(clusterName, metav1.GetOptions{})
> ```

Why is this desirable?  The caller already retrieved the list of clusters that `cluster` came from on 199 in this file, so retrieving only on failure avoids unnecessary rest calls.

---

*[test/e2e/framework/util.go, line 33 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOlae7CDJij2NB9bHq:-LAOlae8CvtfDoAg_ybJ:bj2j7mt) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/util.go#L33)):*
> ```Go
> 	// How long to try single API calls (like 'get' or 'list'). Used to prevent
> 	// transient failures from failing tests.
> 	// TODO: client should not apply this timeout to Watch calls. Increased from 30s until that is fixed.
> ```

This comment looks like it can be removed.

---

*[test/integration/framework/federation.go, line 46 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOpJfy-uFth-bSjGiM:-LAOpJfy-uFth-bSjGiN:bl78b41) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/integration/framework/federation.go#L46)):*
> ```Go
> 	FedApi            *FederationApiFixture
> 	Clusters          map[string]*KubernetesApiFixture
> 	ClusterController *ControllerFixture
> ```

This could be of type TestFixture, since the only use of this is for teardown.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50#-:-LAP1iGj8e0h9P5NYhGD:b-ddjp64)*

font:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAPGF8J08lYecqFuRLi:b-nsxa71) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

I think configuration of joining/unjoining needs to be orthogonal from deployment of in-memory controllers.  Do you think you'll have time to rectify this in this PR or do your other responsibilities suggest that I implement it in a subsequent PR? 
</blockquote></details>

I think the ability to join multiple clusters could possibly introduce a `--clusters` flag that allows users to specify the list of clusters they want joined (along with the `--automatic-join` option), and perhaps they should have to specify even the `--host-cluster` as well so we don't rely on current-context. This could potentially also serve as the option to signify which clusters you are wanting to test against regardless of whether automatic or manual join is performed to ease the concern about unjoining manually joined clusters.

For right now, I could possibly add an `--automatic-join` or `--join-clusters` option (defaulted to `false`) that will automatically join clusters via `kubefnord.JoinCluster` if specified with unmanaged-federation and in-memory-controllers. Then later we can add the ability to support automatic unjoin and multiple clusters. Thoughts?

---

*[test/e2e/framework/unmanaged.go, line 80 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP06wP3f3kkAJy01hF:-LAPCgHd-ADErC-eTXW3:b-knvvi0) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L80)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Without unjoin it's not possible to repeatedly run the e2e tests against a given cluster without manual cleanup.  I think either unjoin needs to be implemented (which implies explicit configuration to avoid unjoining manually joined clusters) or automatic join needs to be removed pending implementation of unjoin. 
</blockquote></details>

I don't think join needs to be removed until unjoin is implemented, because I'd rather have manual unjoin than nothing at all. It's handy to have when debugging the cluster controller e.g. it was helpful in fixing the issue on line 378. I do think the automatic join should be an option though e.g. `--automatic-join`.

---

*[test/e2e/framework/unmanaged.go, line 161 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOxp_GBjlK_8MGHQjJ:-LAP7uYCA1hbgdvGQwD5:bfle5y1) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L161)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

This deferred function is only required to ensure best-effort namespace removal in the event of failure during dumping of events.  Since fixture teardown is orthogonal (and as written the fixtures won't be torn down if any error occurs on namespace removal), consider performing fixture teardown outside of this function and in advance of event dumping.

Also,`f.fixtures` needs to be cleaned up (remove fixtures from the slice as they are torn down) since the framework and its slice are reused across tests.
</blockquote></details>

For namespaces, in the event of success, we cannot tear down the namespace sync controller fixture prior to deleting the test namespace in this defer. This is because the test namespace will never be deleted properly without the federated namespace sync controller removing the necessary finalizer.

I would need to understand what causes the dumping of events to exit the function. But I understand that in the event of failure the defer cleanup is just best-effort. If it's just best-effort, perhaps it's not too significant if the fixtures were not properly cleaned up as the test namespace would also be left as an artifact. I'm not sure if it could introduce weird test issues for the follow-on test fixtures. Perhaps we could also check if we're testing a namespace to call teardown differently - that could help minimize the impact.

Good catch on the slice needing to be updated. I actually had it correct the first time but your comment in another PR made me think it was incorrect so I updated it without digging into the ramifications.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAP52Ny9Ys4wb9u6QxR:b-n7j24l) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Why is this desirable?  The caller already retrieved the list of clusters that `cluster` came from on 199 in this file, so retrieving only on failure avoids unnecessary rest calls.
</blockquote></details>

What if the cluster was not ready when it was first retrieved on line 199? This is fixing an issue where we never re-check the latest status of the retrieved cluster if it wasn't ready the first time we retrieved it on line 199, because we ignore the cluster returned.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

marun:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAPSZMh11erQ8Ttj1Qs:b36vlsr) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

I think the ability to join multiple clusters could possibly introduce a `--clusters` flag that allows users to specify the list of clusters they want joined (along with the `--automatic-join` option), and perhaps they should have to specify even the `--host-cluster` as well so we don't rely on current-context. This could potentially also serve as the option to signify which clusters you are wanting to test against regardless of whether automatic or manual join is performed to ease the concern about unjoining manually joined clusters.

For right now, I could possibly add an `--automatic-join` or `--join-clusters` option (defaulted to `false`) that will automatically join clusters via `kubefnord.JoinCluster` if specified with unmanaged-federation and in-memory-controllers. Then later we can add the ability to support automatic unjoin and multiple clusters. Thoughts?
</blockquote></details>

Why is it desirable to support joining through the test framework for unmanaged federations vs just using kubefnord directly to do so?  Given that it will often be desirable to persist join state across test runs, I can't see the point of having to support joining via both kubefnord and test setup.  Maybe this made sense before kubefnord was ready, but now that it's in tree I need more convincing.

---

*[test/e2e/framework/unmanaged.go, line 80 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP06wP3f3kkAJy01hF:-LAPRHpF2NN7sRx7kWPX:bvim3k2) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L80)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

I don't think join needs to be removed until unjoin is implemented, because I'd rather have manual unjoin than nothing at all. It's handy to have when debugging the cluster controller e.g. it was helpful in fixing the issue on line 378. I do think the automatic join should be an option though e.g. `--automatic-join`.
</blockquote></details>

If automatic join were an option, it would need to default to false, and its doc would need to warn that cleanup would not be performed.  A key tenant of testing is that tests (and test runs) should not leave state that can trip up subsequent execution.

---

*[test/e2e/framework/unmanaged.go, line 161 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOxp_GBjlK_8MGHQjJ:-LAPUCXG0hWnD7dcPHqf:bpxqcyl) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L161)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

For namespaces, in the event of success, we cannot tear down the namespace sync controller fixture prior to deleting the test namespace in this defer. This is because the test namespace will never be deleted properly without the federated namespace sync controller removing the necessary finalizer.

I would need to understand what causes the dumping of events to exit the function. But I understand that in the event of failure the defer cleanup is just best-effort. If it's just best-effort, perhaps it's not too significant if the fixtures were not properly cleaned up as the test namespace would also be left as an artifact. I'm not sure if it could introduce weird test issues for the follow-on test fixtures. Perhaps we could also check if we're testing a namespace to call teardown differently - that could help minimize the impact.

Good catch on the slice needing to be updated. I actually had it correct the first time but your comment in another PR made me think it was incorrect so I updated it without digging into the ramifications.
</blockquote></details>

Ah, I wasn't thinking of the finalizer 2-step.  I think for the purposes of testing it would be desirable to separate that dependency and delete the finalizer manually rather than relying on the controller.  This would remove the dependency of namespace removal on fixture teardown. 

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAPRsamCDjAy4wXcElY:ba8xiql) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

What if the cluster was not ready when it was first retrieved on line 199? This is fixing an issue where we never re-check the latest status of the retrieved cluster if it wasn't ready the first time we retrieved it on line 199, because we ignore the cluster returned.
</blockquote></details>

Good point that the cluster should not be ignored on retrieval - please fix that rather than reordering retrieval.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

font:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAPWinvAdoIk6MdsJL4:b-vo2f8w) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Why is it desirable to support joining through the test framework for unmanaged federations vs just using kubefnord directly to do so?  Given that it will often be desirable to persist join state across test runs, I can't see the point of having to support joining via both kubefnord and test setup.  Maybe this made sense before kubefnord was ready, but now that it's in tree I need more convincing.
</blockquote></details>

It's desirable for debugging the cluster controller, which I already had to do once, although admittedly I have not had to do since then. I agree with you that `kubefnord` works well-enough, except this is for a debug path, not the normal use-case of E2E tests. Are you suggesting to do-away with the automatic join feature entirely? Do you foresee yourself never having to use it?

---

*[test/e2e/framework/unmanaged.go, line 80 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP06wP3f3kkAJy01hF:-LAPeApkCGIEHgScux_9:b-wpzrjc) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L80)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

If automatic join were an option, it would need to default to false, and its doc would need to warn that cleanup would not be performed.  A key tenant of testing is that tests (and test runs) should not leave state that can trip up subsequent execution.
</blockquote></details>

Yes, defaulting to `false` is what I proposed above. The lack of an `unjoin` is also temporary; it should eventually be enabled. Also, a gentle reminder that this is a debug path for tests. We could optionally hide the flag if you are concerned about exposing it.

---

*[test/e2e/framework/unmanaged.go, line 161 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOxp_GBjlK_8MGHQjJ:-LAPbt3i6wJSwMKklA1T:b-d4pfij) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L161)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Ah, I wasn't thinking of the finalizer 2-step.  I think for the purposes of testing it would be desirable to separate that dependency and delete the finalizer manually rather than relying on the controller.  This would remove the dependency of namespace removal on fixture teardown. 
</blockquote></details>

Since this is only used for `in-memory-controllers` as a debug test path, I'd like to propose a `TODO` to come back to this while leaving it here with the fix to remove the fixture from the slice. Otherwise, I can try to get to this soon, but make no promises within the next 3 weeks.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAPZ1gH4iHkfiHLehbw:bjvli6) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Good point that the cluster should not be ignored on retrieval - please fix that rather than reordering retrieval.
</blockquote></details>

The reason I reordered it is because I wasn't sure on the golang scoping rules that would take hold here. If I just replace `_` with `cluster` will this overwrite the cluster variable passed in or create a new one within this scope? Will the for loop always check the last overwritten value of cluster, or the cluster passed into `ClusterIsReadyOrFail`? I would need to understand how both `:=` and closure functions work within this scope.

Then I thought about just checking status prior to the poll but would have to rearrange the error check afterwards. Anyway, that was my dilemma so I opted for this.

Ideally I'd like to understand the golang implications here (a pointer to a doc would be greatly appreciated). Otherwise probably checking status prior to the poll and setting an `err` variable seems appropriate.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

marun:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAPoh8u6-W2M3AfrQEE:bkf2rsp) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

It's desirable for debugging the cluster controller, which I already had to do once, although admittedly I have not had to do since then. I agree with you that `kubefnord` works well-enough, except this is for a debug path, not the normal use-case of E2E tests. Are you suggesting to do-away with the automatic join feature entirely? Do you foresee yourself never having to use it?
</blockquote></details>

Can you elaborate on 'debugging the cluster controller'?  

Having a cluster joined is useful, sure.  I'm not sure I understand the value in joining on test setup vs before test execution.  What are the pros and cons you are thinking of?

---

*[test/e2e/framework/unmanaged.go, line 161 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOxp_GBjlK_8MGHQjJ:-LAPp8QN0hC-Q5m4qVhW:b-mxwx7v) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L161)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Since this is only used for `in-memory-controllers` as a debug test path, I'd like to propose a `TODO` to come back to this while leaving it here with the fix to remove the fixture from the slice. Otherwise, I can try to get to this soon, but make no promises within the next 3 weeks.
</blockquote></details>

Ok, a todo it is.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAPpCkY7_Sih3YjwCHD:b-1ofpy8) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

The reason I reordered it is because I wasn't sure on the golang scoping rules that would take hold here. If I just replace `_` with `cluster` will this overwrite the cluster variable passed in or create a new one within this scope? Will the for loop always check the last overwritten value of cluster, or the cluster passed into `ClusterIsReadyOrFail`? I would need to understand how both `:=` and closure functions work within this scope.

Then I thought about just checking status prior to the poll but would have to rearrange the error check afterwards. Anyway, that was my dilemma so I opted for this.

Ideally I'd like to understand the golang implications here (a pointer to a doc would be greatly appreciated). Otherwise probably checking status prior to the poll and setting an `err` variable seems appropriate.
</blockquote></details>

Yes, `cluster` will be overwritten the context of this function if using `=` (assign to existing value) instead of `:=` (assign to new variable).  The change will not propagate to the calling function (i.e. the for loop) since the cluster is passed in by value (the only way things are passed in golang).  The only way modifications to parameters can be reflected in the calling scope is if they are reference types (e.g. slices/maps/etc) or pointers (and explicitly dereferenced) - neither of which is true for cluster.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

font:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAQ7bzm9AjLVKVnyuij:b948sic) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Can you elaborate on 'debugging the cluster controller'?  

Having a cluster joined is useful, sure.  I'm not sure I understand the value in joining on test setup vs before test execution.  What are the pros and cons you are thinking of?
</blockquote></details>

I think the nature of doing an automatic join in the test setup, particularly around dealing with cluster readiness, is mostly around timing. Manually joining introduced human delay that masqueraded the issue fixed in this PR. Automatic join helped surface it.

Like I mentioned, I have not used it since, so I'm sort of indifferent at this point. If you are adamant about removing it, it's easier for me to remove it at this point than introduce new code to enable its option and all its permutations.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAPzNTZ28iwWdc4V21C:b-xw499z) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Yes, `cluster` will be overwritten the context of this function if using `=` (assign to existing value) instead of `:=` (assign to new variable).  The change will not propagate to the calling function (i.e. the for loop) since the cluster is passed in by value (the only way things are passed in golang).  The only way modifications to parameters can be reflected in the calling scope is if they are reference types (e.g. slices/maps/etc) or pointers (and explicitly dereferenced) - neither of which is true for cluster.
</blockquote></details>

I don't understand how passing in by value vs reference has to do with this particular problem as we're dealing with cluster as a value in all of these cases. Let me clarify my confusion:

If leaving the original ordering i.e. for loop to check for status followed by a Get(), then in the first iteration the for loop will definitely check the value of cluster status using the cluster variable passed into `ClusterIsReadyOrFail`. However, what happens in the next iteration if also leaving `cluster := Get()` , does the for loop see the latest updated cluster? I assume not since the cluster (in the previous iteration) was assigned to a new variable using `:=` and that variable loses scope after this closure function returns. Rinse and repeat and you'd still have the same bug.

I would expect it should instead be `cluster = Get()`, so that we just overwrite the value of cluster used in the entire scope of `ClusterIsReadyOrFail`. Then all subsequent calls to this closure function would allow the for loop to be checking the last known value of cluster i.e. from the previous iteration's call to Get(). That would be my expectation and my understanding of golang thus far (do you agree?). But due to my lack of certainty, I opted for a different approach albeit with a little inefficiency i.e. 1 extra *potentially* unnecessary call to Get().

Let's chat if that isn't clear.

On a separate note, I have also read that in some blog post i.e. "pass by value is the only way things are passed in golang". But I would disagree based on the definition of pass by value vs pass by reference. Unless the industry standard definition for this has changed, pass by reference just means passing a variable's address instead of its value, of which golang is very capable of doing. If a function is invoked by passing a variable's address as in pass by reference, the function will always receive that as a **copy** into a new variable, typically on the stack such as in C, but that doesn't make it pass by value. Just my $.02.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

marun:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LAQIdzj5LSKG1OCXDp8:b75jp6s) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

I think the nature of doing an automatic join in the test setup, particularly around dealing with cluster readiness, is mostly around timing. Manually joining introduced human delay that masqueraded the issue fixed in this PR. Automatic join helped surface it.

Like I mentioned, I have not used it since, so I'm sort of indifferent at this point. If you are adamant about removing it, it's easier for me to remove it at this point than introduce new code to enable its option and all its permutations.
</blockquote></details>

I think YAGNI applies then.  There will need to be testing of join in the future that should cover the case you found automatic join useful for.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAQJ1bO6lFAAMYIwAAH:bpkudyv) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

I don't understand how passing in by value vs reference has to do with this particular problem as we're dealing with cluster as a value in all of these cases. Let me clarify my confusion:

If leaving the original ordering i.e. for loop to check for status followed by a Get(), then in the first iteration the for loop will definitely check the value of cluster status using the cluster variable passed into `ClusterIsReadyOrFail`. However, what happens in the next iteration if also leaving `cluster := Get()` , does the for loop see the latest updated cluster? I assume not since the cluster (in the previous iteration) was assigned to a new variable using `:=` and that variable loses scope after this closure function returns. Rinse and repeat and you'd still have the same bug.

I would expect it should instead be `cluster = Get()`, so that we just overwrite the value of cluster used in the entire scope of `ClusterIsReadyOrFail`. Then all subsequent calls to this closure function would allow the for loop to be checking the last known value of cluster i.e. from the previous iteration's call to Get(). That would be my expectation and my understanding of golang thus far (do you agree?). But due to my lack of certainty, I opted for a different approach albeit with a little inefficiency i.e. 1 extra *potentially* unnecessary call to Get().

Let's chat if that isn't clear.

On a separate note, I have also read that in some blog post i.e. "pass by value is the only way things are passed in golang". But I would disagree based on the definition of pass by value vs pass by reference. Unless the industry standard definition for this has changed, pass by reference just means passing a variable's address instead of its value, of which golang is very capable of doing. If a function is invoked by passing a variable's address as in pass by reference, the function will always receive that as a **copy** into a new variable, typically on the stack such as in C, but that doesn't make it pass by value. Just my $.02.
</blockquote></details>

I'm telling you how I've seen the golang community describes things - 'everything is passed by value'.  The implication is that a copy is provided to the callee is implicit in that statement.

And yes, line 326 in the master version of this should be switched to `=`.   It was `:=` previously because err was the only variable being assigned to and it didn't exist yet.  As soon as an assignment to the existing value `cluster` is made, `=` is required and err needs to be defined before use:

```golang
var err error
cluster, err = ...
```

I feel frustration at this exchange, it seems like pretty basic stuff and we've been spinning on it for a while.  I hope you're learning something from this because otherwise we just wasted alot of time writing back and forth over making the above change.  Maybe we need to switch to higher bandwidth next time?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

font:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LATWn158TqvppeVBsLi:b5a317e) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

I'm telling you how I've seen the golang community describes things - 'everything is passed by value'.  The implication is that a copy is provided to the callee is implicit in that statement.

And yes, line 326 in the master version of this should be switched to `=`.   It was `:=` previously because err was the only variable being assigned to and it didn't exist yet.  As soon as an assignment to the existing value `cluster` is made, `=` is required and err needs to be defined before use:

```golang
var err error
cluster, err = ...
```

I feel frustration at this exchange, it seems like pretty basic stuff and we've been spinning on it for a while.  I hope you're learning something from this because otherwise we just wasted alot of time writing back and forth over making the above change.  Maybe we need to switch to higher bandwidth next time?
</blockquote></details>

Ditto. I suggested we chat and was just looking for an answer to my question to make sure you agree and not expecting any further details. Higher bandwidth sgtm.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50)*

font:
Yes, I have used this quite a bit! Thanks for your thorough review! I have addressed all the feedback and will now push another update. PTAL.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[test/e2e/e2e.go, line 61 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOqaCj0Y9xj_yLJIn0:-LATuLD35fYExXx3fcPC:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/e2e.go#L61)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

s/UnManaged/Unmanaged/
</blockquote></details>

Done.

---

*[test/e2e/framework/unmanaged.go, line 57 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOzyra0QxBndwjxxeS:-LATvI4-BHm2UqZ0HHat:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L57)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

This (and the related check on 119) are not necessary.  The logic in `test_context.go:validateFlags` ensures that `TestManagedFederation = true` if `KubeConfig == ""`, so the unmanaged framework will never be used if kubeconfig is not provided.
</blockquote></details>

Done.

---

*[test/e2e/framework/unmanaged.go, line 66 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAP-dzX7h7YbV9vqqi0:-LATvRxdAgt2RPZH2JRg:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L66)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

I think YAGNI applies then.  There will need to be testing of join in the future that should cover the case you found automatic join useful for.
</blockquote></details>

Done.

---

*[test/e2e/framework/unmanaged.go, line 161 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOxp_GBjlK_8MGHQjJ:-LATw_uc11ZWZEhr4i0R:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L161)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

Ok, a todo it is.
</blockquote></details>

Done.

---

*[test/e2e/framework/unmanaged.go, line 378 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOsw3U-z-fv_HPqLjs:-LAU4Jb9BK6Z3wU0SRX7:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/unmanaged.go#L378)):*
<details><summary><i>Previously, font (Ivan Font) wrote…</i></summary><blockquote>

Ditto. I suggested we chat and was just looking for an answer to my question to make sure you agree and not expecting any further details. Higher bandwidth sgtm.
</blockquote></details>

Done.

---

*[test/e2e/framework/util.go, line 33 at r3](https://beta.reviewable.io/reviews/marun/federation-v2/50#-LAOlae7CDJij2NB9bHq:-LAU4RO36sV2C3ET9l0G:b-896fix) ([raw file](https://github.com/marun/federation-v2/blob/e949c1b82c12d23c6a12eca51637e5b61a65ba4e/test/e2e/framework/util.go#L33)):*
<details><summary><i>Previously, marun (Maru Newby) wrote…</i></summary><blockquote>

This comment looks like it can be removed.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/marun/federation-v2/50#-:-LAURhY23zTVTf51FDsZ:blzryj4)*

PR #55, opened 2018-04-20
Rename after repo donation

font:


pmorie:
/lgtm

Looks like travis is switched over

PR #57, opened 2018-04-22
Add deployment script

marun:
The included scripts automate deployment and deletion of a federation (cr api + fedv2 api + fedv2 controllers) that was documented in #36.  The active kubectl context is targeted, so theoretically this will work with more than just minikube, but it's only been tested with minikube.

### DIFF
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Federation v2 Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script automates deployment of a federation - as documented in
+# the README - to the current kubectl context.  It also joins the
+# hosting cluster as a member of the federation.
+#
+# WARNING: The service account for the federation namespace will be
+# granted the cluster-admin role.  Until more restrictive permissions
+# are used, access to the federation namespace should be restricted to
+# trusted users.
+#
+# If using minikube, a cluster must be started prior to invoking this
+# script:
+#
+#   $ minikube start
+#
+# This script depends on crinit, kubectl, and apiserver-boot being
+# installed in the path.  These binaries can be installed via the
+# download-test-binaries.sh script, which downloads to ./bin:
+#
+#   $ ./scripts/download-test-binaries.sh
+#   $ export PATH=$(pwd)/bin:${PATH}
+#
+# To redeploy federation from scratch, prefix the deploy invocation with the deletion script:
+#
+#   # WARNING: The deletion script will remove federation and cluster registry data
+#   $ ./scripts/delete-federation.sh && ./scripts/deploy-federation.sh <image>
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+
+NS=federation
+IMAGE_NAME="${1:-}"
+
+if [[ -z "${IMAGE_NAME}" ]]; then
+  >&2 echo "Usage: $0 <image>
+
+<image> should be in the form <containerregistry>/<username>/<imagename>:<tagname>
+
+Example: docker.io/<username>/federation-v2:test
+
+If intending to use the docker hub as the container registry to push
+the federation image to, make sure to login to the local docker daemon
+to ensure credentials are available for push:
+
+  $ docker login --username <username>
+"
+  exit 2
+fi
+
+# Wait for the storage provisioner to be ready.  If crinit is called
+# before dynamic provisioning is enabled, the pvc for etcd will never
+# be bound.
+function storage-provisioner-ready() {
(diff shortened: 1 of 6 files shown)


scripts/deploy-federation.sh
```diff
@@ -0,0 +1,81 @@
...
+
+crinit aggregated init mycr
+
+kubectl create ns "${NS}"
+
+apiserver-boot run in-cluster --name "${NS}" --namespace "${NS}" --image "${IMAGE_NAME}" --controller-args="-logtostderr,-v=4"
+
+# Create a permissive rolebinding to allow federation controllers to run.
+# TODO(marun) Make this more restrictive.
+kubectl create clusterrolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
```
font: For more restrictive access, we need just the following for the API server:

```
kubectl create rolebinding -n kube-system \
    federation.k8s.io:extension-apiserver-authentication-reader \
    --role=extension-apiserver-authentication-reader \
    --serviceaccount=federation:default

kubectl create clusterrolebinding federation.k8s.io:apiserver-auth-delegator \
    --clusterrole=system:auth-delegator \
    --serviceaccount=federation:default
```

For the controller manager, we need a new cluster role that includes all the verbs for the operations we want the controller manager to do on the `federation.k8s.io` resources. Then, a new cluster role binding from the service account `federation:default` to this new cluster role we just created. I don't have the exact commands for this yet.

Anyway, this one permissive cluster role binding works easily enough for now.

scripts/deploy-federation.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+
+# This script automates deployment of a federation - as documented
+# in the README - to the current kubectl context.
```
pmorie: Might be helpful to mention that the cluster is also joined into the federation here.

scripts/deploy-federation.sh
pmorie: I would avoid granting these cluster-admin to the default service account - recommend making a federation-controller service account and granting these perms to that. Otherwise any pod in this NS will run with a highly privileged service account token.

scripts/util.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
pmorie: I notice the boilerplate checker isn't in place yet :)

scripts/deploy-federation.sh
```diff
@@ -0,0 +1,81 @@
...
+  $ docker login --username <username>
+"
+  exit 2
+fi
+
+# Wait for the storage provisioner to be ready.  If crinit is called
+# before dynamic provisioning is enabled, the pvc for etcd will never
+# be bound.
+function storage-provisioner-ready() {
+  result="$(kubectl -n kube-system get pod storage-provisioner -o jsonpath='{.status.conditions[?(@.type == "Ready")].status}' 2> /dev/null)"
```
font: `local result`

scripts/deploy-federation.sh
```diff
@@ -0,0 +1,81 @@
...
+# installed in the path.  These binaries can be installed via the
+# download-test-binaries.sh script, which downloads to ./bin:
+#
+#   $ ./scripts/download-test-binaries.sh
+#   $ export PATH=$(pwd)/bin:${PATH}
+#
+# To redeploy federation from scratch, prefix the deploy invocation with the deletion script:
+#
+#   # WARNING: The deletion script will remove federation and cluster registry data
+#   $ ./scripts/delete-federation.sh && ./scripts/deploy-test-federation.sh <image>
```
onyiny-ang: ./scripts/deploy-test-federation.sh should be ./scripts/deploy-federation.sh

scripts/deploy-federation.sh
marun: I'm going to leave this as a TODO for now, with an explicit warning in the header.  Nobody should be deploying this to a production cluster in its current state.

scripts/deploy-federation.sh
marun: Done.

scripts/util.sh
```diff
@@ -0,0 +1,47 @@
...
+    if [[ "${counter}" = "0" ]]; then
+      echo -n "${start_msg}"
+    fi
+
+    if [[ -z "${timeout}" || "${counter}" -lt "${timeout}" ]]; then
+      counter=$((counter + 1))
+      if [[ -n "${timeout}" ]]; then
+        echo -n '.'
+      fi
+      sleep 1
```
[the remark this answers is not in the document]
marun: YAGNI - an optional argument can be added if ever someone requires the behavior.

scripts/download-binaries.sh
```diff
@@ -39,12 +45,19 @@ kubectl_version="$(curl "$curl_args" https://storage.googleapis.com/kubernetes-r
 kubectl_url="https://storage.googleapis.com/kubernetes-release/release/${kubectl_version}/bin/${os_lowercase}/amd64/kubectl"
 curl "${curl_args}" "$kubectl_url" --output "$kubectl_dest"
 
-cr_dest="${dest_dir}/clusterregistry"
-cr_tgz="clusterregistry-server.tar.gz"
-cr_url="https://storage.googleapis.com/crreleases/v0.0.3/${cr_tgz}"
-curl "${curl_args}O" "${cr_url}" \
-  && tar -xzf "${cr_tgz}" -C "${dest_dir}" ./clusterregistry \
-  && rm "${cr_tgz}"
+crc_dest="${dest_dir}/crinit"
```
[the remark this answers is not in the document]
marun: Again, YAGNI.  I don't think maintaining pristine bash is suggested for a limited-scope effort.

scripts/deploy-federation.sh
marun: Done.

scripts/deploy-federation.sh
marun: Done.

marun:
I've responded to comments.  I don't think some of the issues raised suggested blocking merge so long as the concerns were noted for future resolution.

PR #58, opened 2018-04-24
Enable auth for cloud providers

font:
This enables authenticating with various cloud providers e.g. GCP, Azure, etc.


cmd/kubefnord/kubefnord.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	"github.com/kubernetes-sigs/federation-v2/pkg/kubefnord"
 	"k8s.io/apiserver/pkg/util/logs"
+	_ "k8s.io/client-go/plugin/pkg/client/auth" // Enable cloud provider auth
```
pmorie: It's worth a note here advertising that the side effect is desired, lgtm otherwise - i've made similar PRs in the past :)

PR #62, opened 2018-04-26
Refine usage of the dynamic client

marun:
This change simplifies creation of the dynamic client by using the client pool facility of `client-go`.  A new wrapper for the dynamic client is added - `ResourceClient` - which approximates the behavior of standard clientsets.

```golang
  // common
  config := ... // instance of rest.Config
  resource := schema.GroupVersionResource{
    Version: "v1",
    Resource: "secrets",
  }
  namespace := "foo"

  ////////////////////////
  // Before
  ////////////////////////
  
  // This logic was copied from dynamic.ClientPool
  config.APIPath = "/apis"
  groupVersion := resource.GroupVersion()
  config.GroupVersion = &groupVersion
  client, _ := dynamic.NewClient(config)

  // Resource interface creation was complicated and embedded the namespace 
  apiResource := metav1.APIResource{Name: resource.Resource}
  resourceInterface := client.ParameterCodec(dynamic.VersionedParameterEncoderWithV1Fallback).
    Resource(&apiResource, namespace)
  resourceInterface.Get(...)
 
  ////////////////////////
  // After
  ////////////////////////

  // Use the pool to create the client to avoid duplicating its logic 
  pool := dynamic.NewDynamicClientPool(config)
  // Use a ResourceClient wrapper to simplify varying the namespace
  client, _ := util.NewResourceClient(pool, resource)
  client.Resources(namespace).Get(...)
```

This change also replaces `NewUnstructuredInformer` with `NewResourceInformer`, switching to use the `ResourceClient` wrapper and casting to `pkgruntime.Object` instead of `Unstructured`.  Using `pkgruntime.Object` will simplify using dynamic informers in the sync controller.


marun:
@pmorie Looks like the bot needs lgtm needs to be on its own line.

Issue #64, opened 2018-04-26
Federated namespace placement doesn't recreate namespace after adding cluster back in


font:
Going through instructions at https://github.com/font/k8s-example-apps/tree/master/sample-fed-v2-demo, particularly step https://github.com/font/k8s-example-apps/tree/master/sample-fed-v2-demo#update-federatednamespaceplacement, the `test-namespace` is successfully removed from `us-east` after updating the namespace placement to remove `us-east` from the list of clusters.

But when the cluster `us-east` is added back in, the `test-namespace` does not get recreated in the `us-east` cluster causing all the other resources to no get added back due to for example:

```
E0426 19:44:46.756125       1 controller.go:617] Failed to execute updates for FederatedReplicaSet "test-namespace/test-replicaset": [Failed to create FederatedReplicaSet "test-namespace/test-replicaset" in cluster us-east: namespaces "test-namespace" not found]
```
Waiting long enough will eventually see the `test-namespace` created, possibly because the watch operation expires and the controller retries a list operation.

If you restart the federation pod, the `test-namespace` will quickly reappear in the `us-east` cluster after the controller receives the list of namespaces and namespace placements. The rest of the resources (configmap, secret, replicaset) then reappear as well once the `test-namespace` exists.

marun:
Fixed by #65.

PR #65, opened 2018-04-28
Workaround ns version handling issue

marun:
I traced #64 to the version cache.  I'll pursue a proper fix on Monday, but disabling the check for namespaces fixes the problem (at the cost of potentially performing some unnecessary updates).

cc: @font @onyiny-ang

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -496,8 +496,13 @@ func (s *FederationSyncController) reconcile(qualifiedName federatedtypes.Qualif
 	}.String()
 	// TODO(marun) LOCK!
 	if s.pendingVersionUpdates.Has(propagatedVersionKey) {
-		// A status update is pending
-		return statusNeedsRecheck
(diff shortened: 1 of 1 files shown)


pkg/controller/sync/controller.go
```diff
@@ -496,8 +496,13 @@ func (s *FederationSyncController) reconcile(qualifiedName federatedtypes.Qualif
 	}.String()
 	// TODO(marun) LOCK!
 	if s.pendingVersionUpdates.Has(propagatedVersionKey) {
-		// A status update is pending
-		return statusNeedsRecheck
+		// TODO(marun) Need to revisit how namespace deletion affects
+		// the version cache.  Ignoring may cause some unnecessary
+		// updates, but that's better than looping endlessly.
+		if kind != federatedtypes.NamespaceKind {
```
font: There is `federatedtypes.IsNameSpaceKind()` if this ends up being used beyond just as a workaround.

PR #67, opened 2018-05-02
WIP: Switch to metav1.APIResource form schema.GroupVersionResource

marun:
APIResource indicates whether or not a a resource is namespaced, which is important for Create and Update calls made with the dynamic client.  It also has a field for Kind.

### DIFF
--- a/pkg/federatedtypes/apiresource.go
+++ b/pkg/federatedtypes/apiresource.go
@@ -17,13 +17,14 @@ limitations under the License.
 package federatedtypes
 
 import (
-	"k8s.io/apimachinery/pkg/runtime/schema"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/federatedtypes/apiresource.go
```diff
@@ -17,13 +17,14 @@ limitations under the License.
...
-	return schema.GroupVersionResource{
-		Group:    "federation.k8s.io",
-		Version:  "v1alpha1",
-		Resource: resourceName,
+func apiResource(name string) *metav1.APIResource {
+	return &metav1.APIResource{
+		Group:      "federation.k8s.io",
+		Version:    "v1alpha1",
+		Name:       name,
+		Namespaced: true,
```
marun: This should be false for namespaces.

marun:
Not ready for review - more changes on the way.

marun:
Replaced by #68

PR #71, opened 2018-05-12
Enable federation of CRDs

marun:
This PR moves federated type configuration into the API, and spins up sync controllers in response to API state.  This enables dynamic registration of federated types (including for CRDs).

TODO:
- [x] Update README to reflect replacement of type adapters with FederatedTypeConfig API resources
- [x] Add e2e test for propagation of a CRD


marun:
Still WIP, need to cleanup the controller manager.

PR #77, opened 2018-05-18
adds test for boilerplate, verify script passes

onyiny-ang:


font:
Prior to donating the repo we had to modify the copyright from `The Kubernetes Authors` to `The Federation v2 Authors`. With this change, we are moving back to `The Kubernetes Authors`. Was the former change just for donation purposes?

cc @pmorie

pmorie:
@font Yep, the copyright should now be 'The Kubernetes Authors'

Issue #76, opened 2018-05-18
Bump dependency on cluster-registry to 0.0.5


pmorie:
The 0.0.5 version of the cluster-registry contains the CRD refactor and a new generated client (which should be very close to the old one, if not identical). We should revendor this version so that we're working with the current version of the cluster registry.

gyliu513:
Being handled by #120

pmorie:
This will be fixed in #120

PR #94, opened 2018-05-28
Bump dependency on cluster-registry to 0.0.5

onyiny-ang:
/multi-cluster
Addresses issue #76


marun:
As per #100, it should be possible to transition to CRD-based cluster registry with the existing fixture.  Aggregation is not required, as I had previously surmised - kube-apiserver embeds extensions-apiserver.  It will be necessary to [check that the API for the CRD has become established](https://github.com/kubernetes-sigs/federation-v2/pull/100/files#diff-65b5f94418f79144123722866a91ff41R90) before testing commences, as there may be a lag between creation of the CRD resource and when the CRD registration controller enables the endpoint.

onyiny-ang:
@marun well that's a relief (hopefully)! Thanks for your work on the POC. Hopefully this will be fairly straightforward to implement :smile:

pmorie:
@onyiny-ang looks like the build is still failing.

marun:
Superceded by #120

gyliu513:
#120 is trying to bump cluster registry version to v0.0.6 but not v0.0.5, checkout here https://github.com/kubernetes-sigs/federation-v2/pull/120/commits/9f925689062b39adcb97862c79739f9557dc8d55 for detail.

Issue #101, opened 2018-05-31
Add controller manager configuration (for alpha)


marun:
The controller manager currently runs all controllers.  Ideally it would be possible to control whether each controller is started.  Flags is a quick-n-dirty solution to this, and probably preferable if controllers are likely to live in separate repos post-alpha.  A more maintainable solution might be storing configuration as ConfigMap (or whatever the current best practice is for Kube components).

pmorie:
We should be using feature gates for this type of configuration.

marun:
I think this is satisfied by #105.

PR #105, opened 2018-06-06
Add feature gates

shashidharatd:
Ref: #101 
here is an first cut attempt to implement feature-gates to control which features(controllers as of now) should be enabled. Is this what is desired?

/cc @marun @pmorie @irfanurrehman

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"log"
+	"strings"
 	"time"
 
 	controllerlib "github.com/kubernetes-incubator/apiserver-builder/pkg/controller"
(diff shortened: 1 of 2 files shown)


marun:
I'm not familiar with this approach, does kube use it?  I was expecting something more like `--schedule-deployments=true|false` and `--push-reconciler=true|false`.  That said, if @pmorie wants this and you're happy with it, I think it's fine.

shashidharatd:
@marun, it is more or less similar to what you are expecting. instead of multiple individual flags for each feature it would be something like `--feature-gates=SchedulerPreferences=true,PushReconciler=true`

marun:
@shashidharatd I think that looks more like a way to enable/disable a feature in CI than something I would want to expose to a user.  For alpha, though, it's probably fine.

pmorie:
@marun the feature gate mechanism is used by k8s. We can configure parameters of features with CLI options, enable/disable with feature gates.

pmorie:
Thanks a lot for this @shashidharatd !

/lgtm

Issue #110, opened 2018-06-12
Align FederatedCluster with cluster-registry


pmorie:
In 0.0.5, the cluster-registry becomes namespaced; currently, our `FederatedCluster` resource is cluster-scoped. We should align `FederatedCluster` with `Cluster` by making it namespaced. As part of this, we should decide how federation should use the 'globally readable' namespace in the cluster-registry.

gyliu513:
This is related to https://github.com/kubernetes-sigs/federation-v2/issues/76

PR #134, opened 2018-07-03
Change scope of FederatedCluster to be Namespaced

### LINKED ISSUE #107: Namespace placement should be namespaced
Currently namespace placement is cluster-scoped, but this complicates the application of fine-grained authz.  If only the 'tenant' for a given namespace should be allowed to affect its placement, then an explicit rbac rule for just that placement resource would be required.  Changing the namespace placement type to be namespaced would allow for namespace-based authz instead.

marun:
This is to align with the switch of cluster-registry's Cluster resource's scope.  FederatedCluster resources will be stored in the `federation-system` namespace (where the secrets holding the auth credentials for the clusters are already stored).

Fixes ~#107~ #110.


onyiny-ang:
lgtm

PR #135, opened 2018-07-03
Hardcode kubefnord namespace

marun:
As per the inline comment, allowing the federation namespace to be configured by kubefnord is likely to trip users up given that the namespace name is currently hard-coded in the controllers.


pkg/kubefnord/options/options.go
```diff
@@ -49,6 +48,12 @@ func (o *SubcommandOptions) SetName(args []string) error {
 		return fmt.Errorf("NAME is required")
 	}
 
+	// Hard-code the federation namespace until a canonical way of
+	// configuring federation (e.g. via configmap) exists.  In the
+	// absence of canonical configuration, allowing override of the
+	// system namespace for a kubefnord command is a likely source of
+	// problems since only the hard-coded namespace is valid.
+	o.FederationNamespace = util.FederationSystemNamespace
```
onyiny-ang: mstm

PR #144, opened 2018-07-05
Clarify the `cluster-context` for joning cluster.

gyliu513:
/cc @marun @shashidharatd @onyiny-ang

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -184,8 +184,8 @@ clusters that you want to test against.
     ```
 1. Join Cluster(s)
     ```bash
-    ./bin/kubefnord join cluster1 --host-cluster-context cluster1 --add-to-registry --v=2
-    ./bin/kubefnord join cluster2 --host-cluster-context cluster1 --add-to-registry --v=2
+    ./bin/kubefnord join cluster1 --cluster-context cluster1 --host-cluster-context cluster1 --add-to-registry --v=2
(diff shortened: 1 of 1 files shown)


docs/userguide.md
```diff
@@ -184,8 +184,8 @@ clusters that you want to test against.
     ```
 1. Join Cluster(s)
     ```bash
-    ./bin/kubefnord join cluster1 --host-cluster-context cluster1 --add-to-registry --v=2
-    ./bin/kubefnord join cluster2 --host-cluster-context cluster1 --add-to-registry --v=2
+    ./bin/kubefnord join cluster1 --cluster-context cluster1 --host-cluster-context cluster1 --add-to-registry --v=2
```
font: The cluster context, if not specified, will default to the name of the cluster provided. This is described in the `--cluster-context` option's usage statement. So I think specifying this is a bit verbose and would prefer to just call it out as an aside in the guide instead of adding it into the commands.

gyliu513: Thanks @font , but I think it is better to use a full command here and we can add some NOTES for this command to tell end user that `--cluster-context` will default to cluster name if it is not specified. Otherwise, I think end user will have some concern for how does the host cluster know my remote cluster?

gyliu513:
This is being fix at https://github.com/kubernetes-sigs/federation-v2/pull/157 by @font

PR #139, opened 2018-07-04
Do not need namespace for clusterrole and clusterrolebindings.

gyliu513:
/cc @marun @onyiny-ang

PR #146, opened 2018-07-05
Need to specify sa namespace when create clusterrolebindings.

gyliu513:
Fixed https://github.com/kubernetes-sigs/federation-v2/issues/145

/cc @marun @shashidharatd @onyiny-ang


font:
This is undoing some of #139, which makes me think we need to have a CI test for join to validate PRs.

PR #149, opened 2018-07-05
Rename kubefnord to kubefed and implement simple kubefed version

irfanurrehman:
The version naming is suggestive and intended to trigger comments.

### DIFF
--- a/cmd/kubefed2/kubefed2.go
+++ b/cmd/kubefed2/kubefed2.go
@@ -14,14 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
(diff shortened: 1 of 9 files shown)


pkg/kubefed/version.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	// TODO (irfanurrehman): Fix a versioning strategy consistent across federation binaries
+	kubefedVersion = "v2.0.0-alpha.0"
```
gyliu513: One question: Why the init version is 2.0.0?

Issue #104, opened 2018-06-05
Basic walkthrough and associated test script


pmorie:
There should be a relatively simple walkthrough that guides users through the experience of kicking the tires on federation-v2 by deploying a very simple application. Other attributes of this:

- Should be possible to do using just minikube (ie, no cloud provider required)
- Should cover installing federation and adding new clusters into a federation
- There should be a test script that exercises the walkthrough and ensures that the walkthrough 'works'.

font:
> - Should be possible to do using just minikube (ie, no cloud provider required)
> - Should cover installing federation and adding new clusters into a federation
> - There should be a test script that exercises the walkthrough and ensures that the walkthrough 'works'.

I think the first 2 bullet points are more-or-less already covered by #84 depending on what you mean by a *simple* application. I think the 3rd bullet point is what still needs to be done.

xunpan:
@pmorie 
What does your walkthrough means for a script?
From my understanding, below `kubectl` command line is enough to deploy a simple federation resource. 
```
kubectl apply -f example/sample1/federatednamespace.yaml
kubectl apply -f example/sample1/federateddeployment.yaml
```

marun:
I think a script is suggested because the existing walkthrough is not sufficient.  A revised walkthrough would ideally include something like the following:

 - apply yaml to create a namespace and resources in that namespace that comprise a web service (i.e. create a non-federated microservice)
 - federate the application with `kubefed federate ns --contents ` (e.g. #741)
 - check the propagation status of the deployment with `kubectl get federateddeployment -n <ns> myns` (e.g. #843)
 - check that the web service is healthy (e.g. via curl) on the clusters
 - adjust the propagation (e.g. remove one cluster) and check the propagation status and check with curl
 - change an override (e.g. have the message the web service returns be configured via configmap, adjust the value in the configmap with an override, and check that the message the service returns on the cluster has changed)

PR #157, opened 2018-07-07
Update dev and user guides to use kubebuilder and CRDs

### LINKED ISSUE #155: Update userguide.

font:
Resolves #155.
Resolves bullet items 1 and 2 for #104.


docs/development.md
```diff
@@ -5,33 +5,70 @@ help you get started.
...
+to generate code and build binaries. Download the [v0.1.12
+release](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v0.1.12)
 and install it in your `PATH`.
 
+#### kubectl
+
+Download a version of
+[`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/) >= v1.11 and
+install it in your `PATH`.
+
```
gyliu513: How about add a link of `kubectl` v1.11 like line 51?

docs/userguide.md
```diff
@@ -5,11 +5,20 @@ is a walkthrough tutorial for how to deploy the federation v2 control plane.
 
 ## Prerequisites
 
+The federation v2 deployment requires kubernetes version >= 1.11. The following
+is a detailed list of binaries required.
+
 ### Binaries
```
gyliu513: This is same as development guide, how about keep this and let development guide link to this?

docs/userguide.md
```diff
@@ -18,42 +27,48 @@ export PATH=$(pwd)/bin:${PATH}
...
+   [instructions](https://github.com/kubernetes/minikube/blob/master/docs/contributors/build_guide.md).
+2. Or use a recent CI build such as [this one from PR
+   2943](http://storage.googleapis.com/minikube-builds/2943/minikube-linux-amd64).
 
 Once you have minikube installed run:
 
 ```bash
-minikube start -p cluster1
-minikube start -p cluster2
+minikube start -p cluster1 --kubernetes-version v1.11.0
```
gyliu513: Does this mean that federation v2 do not support managing cluster pre v1.11.0 release?

docs/userguide.md
```diff
@@ -184,8 +173,10 @@ clusters that you want to test against.
...
 1. Join Cluster(s)
     ```bash
-    ./bin/kubefnord join cluster1 --host-cluster-context cluster1 --add-to-registry --v=2
-    ./bin/kubefnord join cluster2 --host-cluster-context cluster1 --add-to-registry --v=2
+    ./bin/kubefnord join cluster1 --cluster-context cluster1 \
+        --host-cluster-context cluster1 --add-to-registry --v=2
+    ./bin/kubefnord join cluster2 --cluster-context cluster2 \
+        --host-cluster-context cluster1 --add-to-registry --v=2
     ```
 You can repeat these steps to join any additional clusters.
```
gyliu513: How about add following highlight:

```
*NOTE*: `cluster-context` will default to cluster name if not specified.
```

docs/userguide.md
```diff
@@ -282,44 +282,17 @@ successfully verified a working federation-v2 deployment.
...
 
 ```bash
-kubectl delete federatednamespaceplacement test-namespace
 kubectl delete ns test-namespace
+kubectl delete federatednamespaceplacement test-namespace
 ```
 
 ## Cleanup
 
-### Unjoin Clusters
```
gyliu513: I prefer that we add `unjoin` back when https://github.com/kubernetes-sigs/federation-v2/pull/144 got merged, this can map to the above `join` command.

scripts/delete-federation.sh
```diff
@@ -28,41 +28,45 @@ NS=federation-system
...
-${KCD} clusterrole "${CONTROLLER_ROLE}"
+# Remove cluster registry CRD
+${KCD} -f vendor/k8s.io/cluster-registry/cluster-registry-crd.yaml
 
-# Delete federated cluster
-# TODO(marun) Remove when federated clusters are stored in federation namespace
-if kubectl get federatedcluster &>/dev/null; then
-  ${KCD} federatedcluster "${CONTEXT}"
-fi
+# Unjoin clusters by removing objects added by kubefed.
```
gyliu513: We can remove this once `unjoin` finished.

docs/development.md
font: I've added it because we have a special dependency on 1.11. Otherwise I don't normally like doing that because it quickly grows stale.

docs/userguide.md
font: Ideally, I'd like to have a separate `prerequisites.md` markdown file that can be imported and rendered in any number of files that you want e.g. `development.md` and `userguide.md`. But that feature does not exist AFAIK.

docs/userguide.md
font: In theory, it can manage pre v1.11.0 clusters for the other clusters that are joined to the federation, but the host cluster must be v1.11.0. We do not have a version compatibility matrix and we need to discuss our overall version support strategy.

docs/userguide.md
font: Added.

docs/userguide.md
font: I opted to move the unjoin operation to the `delete-federation.sh` script because it was a bunch of bash commands and makes it cleaner. When we support `kubefed unjoin` we can add it back here and simplify the script.

scripts/delete-federation.sh
font: I don't know that we'll necessarily want to remove this as it's part of cleanup. But the script can be simplified.

pmorie:
@font needs rebase

docs/development.md
```diff
@@ -239,3 +253,43 @@ these steps:
...
+docker login --username <username>
+```
+
+### Create Deployment Config
+
+Run the following command to build the deployment config `hack/install.yaml`
+that includes all the necessary kubernetes resources:
+
+```bash
+kubebuilder create config \
```
marun: The generated yaml will cause errors with kube < 1.12 unless the type fields of openapischema structs are removed for resources that define subresources.  The fix is as follows:

https://github.com/kubernetes-sigs/federation-v2/pull/160/files#diff-1ed12d3b5689c7a45b1c21e93e23e3d3R107

scripts/delete-federation.sh
```diff
@@ -28,41 +28,45 @@ NS=federation-system
...
 
-# Remove crds
-# Enable available types
+# Disable available types
 for filename in ./config/federatedtypes/*.yaml; do
   ${KCD} -f "${filename}"
 done
 
+# Remove federation CRDs, namespace, RBAC and deployment resources.
+${KCD} -f hack/install-latest.yaml
```
marun: Should this be configurable to allow targeting `hack/install.yaml` (for developers)?

docs/development.md
font: Added a note about this.

scripts/delete-federation.sh
font: Perhaps. Although I think we should use the same committed config for developers and users.

marun: Regarding developers, `hack/install.yaml` is the path that kubebuilder generates to, and so long as it is not idempotent (the order of groups changes with every invocation) I don't think that file should ever be committed.  For users, I don't think they should be using the tree at all.  Release artifacts should be available that allow deployment and use of federation without ever checking out the source tree.

font: Yes, I noticed the RBAC `federation-role` `ClusterRole` config changes the ordering of the `apiGroups` field. That's a bug that should be fixed in `kubebuilder`. I think that would allow us to commit `install.yaml`. Otherwise, we should make sure our docs are consistent with our scripts to only ever use 1 file for our releases and developer workflows, wherever we decide to commit it.

While I don't think users should have to rely on checking out the tree, they should use the tree to apply whatever release artifacts they want. That is, every released version of the config should be available via github at that specific tagged point in the tree e.g. `https://raw.githubusercontent.com/kubernetes-sigs/federation-v2/<release-version-tag>/path/to/install.yaml` and the image should point to that tagged/released version. I think this would allow us to use 1 config file for both workflows.

docs/userguide.md
gyliu513: Yes, so for now, how about highlight this in the document? Such as we need make sure the host cluster must be v1.11.0 but the joining clusters can be other versions?

docs/userguide.md
gyliu513: Yes, same thinking as you.

shashidharatd:
@font, Great work. User guide worked like a charm. Although in developer guide for running integration and e2e tests in my environment, i faced below issue, and you can note that `etcd` is already in the path.
```
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e# which etcd
/root/gopath/src/github.com/kubernetes-sigs/federation-v2/bin/etcd
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2# cd test/integration && go test -v
=== RUN   TestIntegration
--- FAIL: TestIntegration (0.00s)
	logger.go:52: Starting a federation of 2 clusters...
	logger.go:44: Error starting etcd: fork/exec /root/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/sigs.k8s.io/testing_frameworks/integration/assets/bin/etcd: no such file or directory
FAIL
exit status 1
FAIL	github.com/kubernetes-sigs/federation-v2/test/integration	0.011s
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2/test/integration# cd ../../
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2/bin# which etcd 
/root/gopath/src/github.com/kubernetes-sigs/federation-v2/bin/etcd
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2# cd test/e2e
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e# go test -v
=== RUN   TestE2E
Running Suite: Federation e2e suite
===================================
Random Seed: 1531234422
Will run 10 of 10 specs

Jul 10 20:23:42.957: INFO: Starting a federation of 2 clusters...
Jul 10 20:23:42.957: INFO: Error starting etcd: fork/exec /root/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/sigs.k8s.io/testing_frameworks/integration/assets/bin/etcd: no such file or directory
Failure [0.002 seconds]
[BeforeSuite] BeforeSuite 
/root/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/e2e.go:48

  Jul 10 20:23:42.957: Error starting etcd: fork/exec /root/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/sigs.k8s.io/testing_frameworks/integration/assets/bin/etcd: no such file or directory

  /root/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/framework/logger.go:39
------------------------------
Ran 10 of 0 Specs in 0.002 seconds
FAIL! -- 0 Passed | 10 Failed | 0 Pending | 0 Skipped
--- FAIL: TestE2E (0.00s)
FAIL
exit status 1
FAIL	github.com/kubernetes-sigs/federation-v2/test/e2e	0.020s
```
Except the above issue this lgtm to merge.
Agree with @marun about the comment on release artifacts. It is some work for the release process. @pmorie, what is your opinion on this? Definitely it should not be a blocker for alpha release.
Thanks @font for this great work.

font:
@shashidharatd That error appears to be because the necessary environment variables have not been set as described in the section https://github.com/font/federation-v2/blob/078550ee1d866f6d6641b3703ba4c6b1f972531f/docs/development.md#environment-setup. Try it again after you've set those variables and let me know how it goes.

shashidharatd:
oops, sorry i missed that part. it works great. lgtm from me. Thanks @font

docs/userguide.md
font: The doc explicitly describes the need for k8s version >= 1.11. I do not want to add any comments about using older versions of k8s in the joining clusters because we do not necessarily support that and have not tested it. In theory it may work, but we do not have a strategy for version compatibility yet. I think we should expect all clusters to be >= 1.11 at least for the first alpha release.

gyliu513: @font then this will be a big limitation for federation v2. Kubernetes 1.11 is the latest version and many customers are still using kubenretes 1.10, 1.9, 1,8 etc and most of them actually want manage those versions.

I have did some test with 1.11 host cluster, and 1.10 as join cluster, it works fine, so can we update the note a bit to clarify we only need host cluster to be 1.11, but the joining cluster does not request 1.11 and it can be a lower version?

gyliu513: Also append my test result here.

```
root@gyliu-dev1:~/go/src/github.com/kubernetes-sigs/federation-v2# kubectl get nodes
NAME             STATUS    ROLES     AGE       VERSION
172.16.250.138   Ready     <none>    5d        v1.11.0
172.16.250.140   Ready     <none>    5d        v1.11.0
```
```
root@gyliu-dev1:~/go/src/github.com/kubernetes-sigs/federation-v2# kubectl get nodes --context cluster141.k8s.local
NAME             STATUS    ROLES     AGE       VERSION
172.16.250.141   Ready     <none>    22m       v1.10.0+icp-ee
172.16.250.142   Ready     <none>    21m       v1.10.0+icp-ee
```
```
root@gyliu-dev1:~/go/src/github.com/kubernetes-sigs/federation-v2# for r in configmaps secrets deploy; do     for c in cluster138.k8s.local cluster141.k8s.local; do         echo; echo ------------ ${c} ------------; echo;         kubectl --context=${c} -n test-namespace get ${r};         echo; echo;     done; done

------------ cluster138.k8s.local ------------

NAME             DATA      AGE
test-configmap   1         4m



------------ cluster141.k8s.local ------------

NAME             DATA      AGE
test-configmap   1         4m



------------ cluster138.k8s.local ------------

NAME                  TYPE                                  DATA      AGE
default-token-bb5kr   kubernetes.io/service-account-token   3         5m
test-secret           Opaque                                1         4m



------------ cluster141.k8s.local ------------

NAME                  TYPE                                  DATA      AGE
default-token-tg5cp   kubernetes.io/service-account-token   3         4m
test-secret           Opaque                                1         4m



------------ cluster138.k8s.local ------------

NAME              DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
test-deployment   3         3         3            3           4m



------------ cluster141.k8s.local ------------

NAME              DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
test-deployment   5         5         5            5           4m
```

font: @gyliu513 We need to discuss version compatibility support more broadly with the entire group before we decide to put any notes around what versions we support.

Issue #152, opened 2018-07-05
Add support for building tagged images


pmorie:
Once we begin releasing tagged versions, we will need to release tagged images. This issue is to update the existing image build to work with tagged versions such that:

- When building a tag, the image pushed will be tagged with the version being built; example: travis build of `0.0.1` tag results in `federation-v2:0.0.1` image being pushed
- The binaries in the image report a version consistent with the tag; example: `controller-manager --version` reports `0.0.1` from the `federation-v2:0.0.1` image

shashidharatd:
@pmorie, this has been addressed. Please reopen if you feel otherwise.

PR #158, opened 2018-07-09
Update travis.yml to include support for tagging

onyiny-ang:
Addresses #152

### DIFF
--- a/scripts/imagebuild.sh
+++ b/scripts/imagebuild.sh
@@ -23,9 +23,15 @@ dockerfile_dir="${base_dir}/images/federation-v2"
 
 [ -f "$base_dir/bin/controller-manager" ] || { echo "$base_dir/bin/controller-manager not found" ; exit 1 ;}
 
-if [[ "${TRAVIS_BRANCH}" != "master" ]]; then
-  echo "Not on master branch. Image build skipped." >&2
-  exit 0
+if [[ "${TRAVIS_TAG}" =~ ^[0-9]+\.[0-9]+\.[0-9]?\d$ ]]; then
+    echo "Pushing images with tag '${TRAVIS_TAG}'."
+    TAG="${TRAVIS_TAG}"
+elif [[ "${TRAVIS_BRANCH}" == "master" ]]; then
+    echo "Pushing images with default tag 'canary'."
+    TAG="canary"
+else
+    echo "Nothing to deploy. Image build skipped." >&2
+    exit 0
 fi
 
 echo "Starting image build"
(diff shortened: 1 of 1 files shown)


onyiny-ang:
Test failing #159

scripts/imagebuild.sh
```diff
@@ -23,9 +23,15 @@ dockerfile_dir="${base_dir}/images/federation-v2"
 
 [ -f "$base_dir/bin/controller-manager" ] || { echo "$base_dir/bin/controller-manager not found" ; exit 1 ;}
 
-if [[ "${TRAVIS_BRANCH}" != "master" ]]; then
-  echo "Not on master branch. Image build skipped." >&2
-  exit 0
+if [[ "${TRAVIS_TAG}" =~ ^[0-9]+\.[0-9]+\.[0-9]?\d$ ]]; then
+    echo "Pushing images with tags '${TRAVIS_TAG}' and 'latest'."
```
gyliu513: What do you mean of `tags '${TRAVIS_TAG}' and 'latest'.`? Why including `latest` here?

shashidharatd: we are not pushing the image with latest tag. either message needs updating or we should push image with latest tag.

scripts/imagebuild.sh
```diff
@@ -23,9 +23,15 @@ dockerfile_dir="${base_dir}/images/federation-v2"
...
+    export TAG="${TRAVIS_TAG}"
+elif [[ "${TRAVIS_BRANCH}" == "master" ]]; then
+    echo "Pushing images with default tags (git sha and 'canary')."
```
shashidharatd: we are not pushing image with `git sha` as a tag, so the message is misleading.

onyiny-ang:
@shashidharatd @gyliu513 removed misleading tag messages from imagebuild.sh

shashidharatd:
@onyiny-ang, Thanks for the update. couple of things.
1. Why do we use the tag canary instead of latest? for an end user who is familiar with docker using the image without tag is normal if he does not care about version.
2. For every merged pr, it makes sense to have 2 docker pushes, one for `git sha` and another one with `latest` tag (considering the first comment).

onyiny-ang:
@shashidharatd 
re: canary, it's a convention: http://www.yourdictionary.com/canary-build
Though I haven't fully determined what system we want to go with, I was looking at the tagging in place for svc-catalog: https://github.com/kubernetes-incubator/service-catalog/blob/master/contrib/travis/deploy.sh
In the case of service catalog, version tag and 'latest' are pushed together as more stable and well tested versions whereas git sha and 'canary' are used on master to alert users to the fact that they don't represent a stable release. This is not what is implemented in my PR atm because I'm not sure what system we want to follow for federation.
I welcome feedback on this.
@pmorie @marun @font @irfanurrehman

scripts/imagebuild.sh
```diff
@@ -23,9 +23,15 @@ dockerfile_dir="${base_dir}/images/federation-v2"
 
 [ -f "$base_dir/bin/controller-manager" ] || { echo "$base_dir/bin/controller-manager not found" ; exit 1 ;}
 
-if [[ "${TRAVIS_BRANCH}" != "master" ]]; then
-  echo "Not on master branch. Image build skipped." >&2
-  exit 0
+if [[ "${TRAVIS_TAG}" =~ ^[0-9]+\.[0-9]+\.[0-9]?\d$ ]]; then
+    echo "Pushing images with tag '${TRAVIS_TAG}'."
+    export TAG="${TRAVIS_TAG}"
```
pmorie: You don't need to export this; you can just say `tag=` and then reference elsewhere as `${tag}`

onyiny-ang:
Addresses #152

PR #164, opened 2018-07-10
Fix join to use passed-in federation namespace instead of a constant

gyliu513:
/cc @marun


marun:
This is a non-functional change, so no, it can wait.

gyliu513:
@marun I think we already released alpha, so can you help review this to check if this can be merged?

marun:
The name of this PR and the associated commit suggests your intent is at odds with your code.  Consider rephrasing it to something like 'Fix join to use passed-in federation namespace instead of a constant'.

gyliu513:
@marun done

PR #166, opened 2018-07-10
Flush log immediately for controller manager.

gyliu513:
/cc @marun

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	// Import auth/gcp to connect to GKE clusters remotely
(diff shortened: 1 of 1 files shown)


gyliu513:
There is no logs for controller manager now if we are using `glog`, how about put this in alpha as well?

/cc @marun @font @shashidharatd

marun:
@gyliu513 This change doesn't enable logs - it just affects the flush interval.  I think it can wait until post-alpha.

gyliu513:
@marun Yes, this is also the official way how Kubernetes handle its logs, hope we can have this post-alpha.

gyliu513:
@marun I think we already released alpha, so can you help review this to check if this can be merged?

PR #167, opened 2018-07-10
Implement versioning with git tag and sha info

irfanurrehman:
Raising this pull to get feedback.
Some comments explaining, what I intend to do.
Alternative suggestions welcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,104 @@
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+TARGET = federation-v2
+GOTARGET = github.com/kubernetes-sigs/$(TARGET)
+REGISTRY ?= quay.io/kubernetes-federation
+IMAGE = $(REGISTRY)/$(TARGET)
+DIR := ${CURDIR}
+DOCKER ?= docker
+
+GIT_VERSION ?= $(shell git describe --always --dirty)
+GIT_TAG ?= $(shell git describe --tags --exact-match 2>/dev/null)
+GIT_HASH ?= $(shell git rev-parse HEAD)
+BUIDLDATE = $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+GIT_TREESTATE = "clean"
+DIFF = $(shell git diff --quiet >/dev/null 2>&1; if [ $$? -eq 1 ]; then echo "1"; fi)
+ifeq ($(DIFF), 1)
+    GIT_TREESTATE = "dirty"
+endif
+
+ifneq ($(VERBOSE),)
+VERBOSE_FLAG = -v
+endif
+BUILDMNT = /go/src/$(GOTARGET)
+BUILD_IMAGE ?= golang:1.9.3
+
+CONTROLLER_TARGET = bin/controller-manager
+KUBEFED2_TARGET = bin/kubefed2
+
+LDFLAG_OPTIONS = -ldflags "-X github.com/kubernetes-sigs/federation-v2/pkg/version.version=$(GIT_VERSION) \
+                      -X github.com/kubernetes-sigs/federation-v2/pkg/version.gitCommit=$(GIT_HASH) \
+                      -X github.com/kubernetes-sigs/federation-v2/pkg/version.gitTreeState=$(GIT_TREESTATE) \
+                      -X github.com/kubernetes-sigs/federation-v2/pkg/version.buildDate=$(BUIDLDATE)"
+BUILDCMD_CONTROLLER = go build -o $(CONTROLLER_TARGET) $(VERBOSE_FLAG) $(LDFLAG_OPTIONS)
+BUILDCMD_KUBEFED2 = go build -o $(KUBEFED2_TARGET) $(VERBOSE_FLAG) $(LDFLAG_OPTIONS)
+
+BUILD_CONTROLLER = $(BUILDCMD_CONTROLLER) cmd/controller-manager/main.go
+BUILD_KUBEFED2 = $(BUILDCMD_KUBEFED2) cmd/kubefed2/kubefed2.go
+
+TESTARGS ?= $(VERBOSE_FLAG) -timeout 60s
+TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
+TEST_CMD = go test $(TESTARGS)
+TEST = $(TEST_CMD) $(TEST_PKGS) 
+
+VET = go vet $(TEST_PKGS)
+FMT = go fmt $(TEST_PKGS)
+
+DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
+
+# TODO (irfanurrehman): can add local compile, and auto-generate targets also if needed
+.PHONY: all container push clean controller kubefed2 test local-test vet fmt
+
+all: container controller kubefed2
+
+local-test:
+	$(TEST)
+
+# Unit tests 
+test: controller kubefed2 vet
+	$(DOCKER_BUILD) '$(TEST)'
+
+vet:
+	$(DOCKER_BUILD) '$(VET)'
+
+fmt:
+	$(FMT)
+
+container:
+	$(DOCKER) build -f Dockerfile.controller \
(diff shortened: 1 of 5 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,53 @@
...
+*/
+
+package version
+
+import (
+	"fmt"
+	"runtime"
+)
+
+type Info struct {
```
shashidharatd: why don't we use the https://github.com/kubernetes/apimachinery/blob/master/pkg/version/types.go#L22, which can be vendored instead of defining it in our package.

marun: This repo already vendors apimachinery:

https://github.com/kubernetes-sigs/federation-v2/tree/master/vendor/k8s.io/apimachinery

pkg/version/base.go
```diff
@@ -0,0 +1,37 @@
...
+// As of now a static version value supplied in base.go is used instead.
+var (
+	// TODO: The current plan as per the todo above is to populate these
+	// fields using ldflags as per the comments against each.
+	// Worst case (for example, as is now), this file can have static info.
+	version = "v0.0.1-alpha.0" // output of "git describe"
+	// the prerequisite is that the branch should be
+	// tagged using the correct versioning strategy.
+
+	gitCommit    = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
```
font: Where is this intended to be interpolated? Perhaps supplying an `unknown` value is a sane default.

pkg/version/base.go
```diff
@@ -0,0 +1,37 @@
...
+	gitTreeState = ""            // state of git tree, either "clean" or "dirty"
```
font: Should we supply a default e.g. `dirty`?

pkg/version/base.go
```diff
@@ -0,0 +1,37 @@
...
+
+	buildDate = "1970-01-01T00:00:00Z" // build date in ISO8601 format, output of $(date -u +'%Y-%m-%dT%H:%M:%SZ')
+)
```
font: We should consider adding at least `GoVersion`, `Compiler` and possibly `Platform`.

pkg/version/version.go
irfanurrehman: The reason I did put in federation specific structure is that, I did not find some of the fields meaningful for us, viz `Major` and `Minor`. I find the details repetitive, the same info is part of `Version` and as mentioned in the comment where the global var is declared [here](https://github.com/irfanurrehman/federation-v2/blob/0a83ed76eeb93b5469620205681e0a6523144dd8/pkg/version/base.go#L29), the version info should be a tag in the repo (ideally on master). This will ensure we do not have multiple places tracking versions.

pkg/version/base.go
irfanurrehman: Updated the default to `unknown`
Updated a Makefile, where this is calculated and set into the bin!

pkg/version/base.go
irfanurrehman: Updated the default to `unknown`

pkg/version/base.go
irfanurrehman: Its directly set in version.go

irfanurrehman:
Updated a Makefile!

irfanurrehman:
A git tag would additionally need to be added to the repo. Adding a tag also creates a release at github. This step should ideally be done, when all else is in place for the release.

Makefile
```diff
@@ -0,0 +1,104 @@
...
+BUILDMNT = /go/src/$(GOTARGET)
+BUILD_IMAGE ?= golang:1.9.3
+
+CONTROLLER_TARGET = bin/controller-manager
+KUBEFED2_TARGET = bin/kubefed2
+
+LDFLAG_OPTIONS = -ldflags "-X github.com/kubernetes-sigs/federation-v2/pkg/version.version=$(GIT_VERSION) \
+                      -X github.com/kubernetes-sigs/federation-v2/pkg/version.gitCommit=$(GIT_HASH) \
+                      -X github.com/kubernetes-sigs/federation-v2/pkg/version.gitTreeState=$(GIT_TREESTATE) \
+                      -X github.com/kubernetes-sigs/federation-v2/pkg/version.buildDate=$(BUIDLDATE)"
```
pmorie: typo: `BUIDLDATE`

Makefile
```diff
@@ -0,0 +1,104 @@
...
+GOTARGET = github.com/kubernetes-sigs/$(TARGET)
+REGISTRY ?= quay.io/kubernetes-federation
+IMAGE = $(REGISTRY)/$(TARGET)
+DIR := ${CURDIR}
+DOCKER ?= docker
+
+GIT_VERSION ?= $(shell git describe --always --dirty)
+GIT_TAG ?= $(shell git describe --tags --exact-match 2>/dev/null)
+GIT_HASH ?= $(shell git rev-parse HEAD)
+BUIDLDATE = $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
```
pmorie: Typo: `BUIDLDATE`

Makefile
```diff
@@ -0,0 +1,104 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+TARGET = federation-v2
+GOTARGET = github.com/kubernetes-sigs/$(TARGET)
+REGISTRY ?= quay.io/kubernetes-federation
```
pmorie: It's actually `quay.io/kubernetes-multicluster`

irfanurrehman:
@font @pmorie, updated, PTAL!

Makefile
```diff
@@ -0,0 +1,104 @@
...
+	$(DOCKER_BUILD) '$(TEST)'
+
+vet:
+	$(DOCKER_BUILD) '$(VET)'
+
+fmt:
+	$(FMT)
+
+container:
+	$(DOCKER) build -f Dockerfile.controller \
```
pmorie: The image build in travis is using `images/federation-v2/Dockerfile` to do the build. We should pick one - the dockerfile referenced here, or the dockerfile the image build script is already using.

onyiny-ang: For reference, the `images/federation-v2/Dockerfile` doesn't do any of the go tests or builds done in Dockerfile.controller because they're already done earlier in the travis build with `precommit.sh`. Since it's possible to copy the binaries created with `download-binaries.sh` into the path for the docker build, there's no need to run through Dockerfile.controller again.

irfanurrehman: I found both Dockerfiles, while updating the Makefile; Both have flaws. Tests are not needed as part of `Dockerfile.controller` and `images/federation-v2/Dockerfile` can only be invoked from `scripts/imagebuild.sh`, because the script basically copies/provides the `controller-manager` binary to be copied into the image. I think `Dockerfile.controller` is better as it builds the binary and can be invoked without dependency. I also think that this change could be done subsequently, as followup, where we can rather use `make container` and `make push` to build and push the image from CI and should not be a blocker for making this release. When both merge, tests could also be removed from `Dockerfile.controller`.
Meanwhile, the current use of `Dockerfile.controller` is to provide the user a mechanism to build their own image.

pmorie: I would really prefer to avoid having two Dockerfiles - I don't have a strong opinion on which one we use, but I don't want to have two different Dockerfiles and ways of building the image. Let's pick one now and change whatever needs to in this PR.

How does that sound?

pmorie: Also, I don't expect users to normally want to build their own image. If you're building your own image locally, you're a federation developer, IMO

Makefile
```diff
@@ -78,10 +78,11 @@ vet:
 fmt:
 	$(FMT)
 
-container:
-	$(DOCKER) build -f Dockerfile.controller \
-		-t $(REGISTRY)/$(TARGET):$(GIT_VERSION) \
-		.
+container: controller
+	cp -f bin/controller-manager images/federation-v2/
+	$(DOCKER) build /root/work/src/github.com/kubernetes-sigs/federation-v2/images/federation-v2 \
```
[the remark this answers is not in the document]
irfanurrehman: Thanks for pointing that out, updated!

Makefile
```diff
@@ -0,0 +1,105 @@
...
+
+controller:
+	$(DOCKER_BUILD) '$(BUILD_CONTROLLER)'
+
+kubefed2:
+	$(DOCKER_BUILD) '$(BUILD_KUBEFED2)'
+
+push:
+	$(DOCKER) push $(REGISTRY)/$(TARGET):$(GIT_VERSION)
+	if git describe --tags --exact-match >/dev/null 2>&1; \
```
font: Consider checking the `GIT_TAG` variable for non-null instead of re-running the command here.

Makefile
```diff
@@ -0,0 +1,105 @@
...
+#
+
+TARGET = federation-v2
+GOTARGET = github.com/kubernetes-sigs/$(TARGET)
+REGISTRY ?= quay.io/kubernetes-multicluster
+IMAGE = $(REGISTRY)/$(TARGET)
+DIR := ${CURDIR}
+DOCKER ?= docker
+
+GIT_VERSION ?= $(shell git describe --always --dirty)
```
font: We should consider adding an `IMAGE_TAG` variable that can be assigned the `GIT_VERSION` by default. This would allow creating an image with any custom tag name the developer chooses.

PR #168, opened 2018-07-10
Fix tagging format for image build

onyiny-ang:
Slight change in regex for imagebuild to follow convention
/cc @pmorie @shashidharatd


onyiny-ang:
Should close out #152

PR #176, opened 2018-07-12
Remove Dockerfile.controller to prevent confusion

onyiny-ang:
Having two Dockerfiles and different ways to build images within the repo is confusing to users. It also may be preferable to remove the option to build your own image from the `deploy-federation.sh` script eventually, but for now the script is just modified to keep the image build consistent with the Makefile and `imagebuild.sh` script.


scripts/deploy-federation.sh
```diff
@@ -84,7 +84,16 @@ shift
 JOIN_CLUSTERS="${*}"
 
 if [[ ! "${USE_LATEST}" ]]; then
-  docker build . -f Dockerfile.controller -t "${IMAGE_NAME}"
+  base_dir="$(cd "$(dirname "$0")/.." ; pwd)"
+  dockerfile_dir="${base_dir}/images/federation-v2"
+  if [[ ! -f "$base_dir/bin/controller-manager" && ! -f "${dockerfile_dir}/controller-manager" ]] ; then
```
font: Use of `${base_dir}` instead of `$base_dir` here and below.

scripts/deploy-federation.sh
```diff
@@ -84,7 +84,16 @@ shift
...
+    echo "$base_dir/bin/controller-manager not found, building"
+    go build -o "${dockerfile_dir}"/controller-manager "${base_dir}"/cmd/controller-manager/main.go
```
font: I think it's preferable to use the Makefile now that it's available instead of running these commands explicitly.

scripts/deploy-federation.sh
```diff
@@ -84,7 +84,16 @@ shift
...
+  elif [[ -f "$base_dir/bin/controller-manager" && ! -f "${dockerfile_dir}/controller-manager" ]]; then
+    cp ${base_dir}/bin/controller-manager ${dockerfile_dir}/controller-manager
+  fi
+  docker build ${dockerfile_dir} -t "${IMAGE_NAME}"
```
font: Same comment as above.

docs/development.md
```diff
@@ -276,11 +276,12 @@ by following the below instructions.
 
 #### Build Federation Container Image
 
-Run the following commands using the committed `Dockerfile.controller` to build
+Run the following commands using the committed `images/scrips/Dockerfile` to build
 and push a container image to use for deployment:
 
 ```bash
-docker build . -f Dockerfile.controller -t <containerregistry>/<username>/federation-v2:test
+go build -o images/federation-v2/controller-manager cmd/controller-manager/main.go
```
font: Let's use the Makefile.

font:
Using the Makefile can be done in a follow-up PR.

scripts/deploy-federation.sh
```diff
@@ -51,7 +51,7 @@ NS=federation-system
 PUBLIC_NS=kube-multicluster-public
 IMAGE_NAME="${1:-}"
 
-LATEST_IMAGE_NAME=quay.io/kubernetes-multicluster/federation-v2:canary
+LATEST_IMAGE_NAME=quay.io/kubernetes-multicluster/federation-v2:latest
```
font: We also need to update `scripts/deploy-federation-latest.sh`.

scripts/deploy-federation.sh
```diff
@@ -84,7 +84,16 @@ shift
 JOIN_CLUSTERS="${*}"
 
 if [[ ! "${USE_LATEST}" ]]; then
-  docker build . -f Dockerfile.controller -t "${IMAGE_NAME}"
+  base_dir="$(cd "$(dirname "$0")/.." ; pwd)"
+  dockerfile_dir="${base_dir}/images/federation-v2"
+  if [[ ! -f "${base_dir}/bin/controller-manager" && ! -f "${dockerfile_dir}/controller-manager" ]] ; then
+    echo "$base_dir/bin/controller-manager not found, building"
```
font: One more `base_dir` here.

scripts/deploy-federation.sh
onyiny-ang: ok I will add that on.

scripts/deploy-federation.sh
```diff
@@ -84,7 +84,16 @@ shift
...
+  base_dir="$(cd "$(dirname "$0")/.." ; pwd)"
+  dockerfile_dir="${base_dir}/images/federation-v2"
+  if [[ ! -f "${base_dir}/bin/controller-manager" && ! -f "${dockerfile_dir}/controller-manager" ]] ; then
+    echo "${base_dir}/bin/controller-manager not found, building"
+    go build -o "${dockerfile_dir}"/controller-manager "${base_dir}"/cmd/controller-manager/main.go
+  elif [[ -f "${base_dir}/bin/controller-manager" && ! -f "${dockerfile_dir}/controller-manager" ]]; then
+    cp ${base_dir}/bin/controller-manager ${dockerfile_dir}/controller-manager
+  fi
+  docker build ${dockerfile_dir} -t "${IMAGE_NAME}"
+  rm ${dockerfile_dir}/controller-manager
```
font: I would add a `-f` to the command to prevent any interaction.

PR #177, opened 2018-07-12
Update deploy script and doc for developer workflow

font:


docs/development.md
```diff
@@ -261,12 +261,11 @@ to build an image and a deployment config.
...
 
-If you just want to have this automated, then jump to the [automated deployment
-steps in the userguide](userguide.md#automated-deployment) but be sure to
-specify your own image such as:
+If you just want to have this automated, then run the following command
+specifying your own image:
 
 ```bash
-./scripts/deploy-federation-latest.sh <containerregistry>/<username>/federation-v2:test cluster2
+./scripts/deploy-federation.sh <containerregistry>/<username>/federation-v2:test cluster2
```
gyliu513: @font I think using `cluster2` here is not good, also we may need to clarify that here `cluster2` need to be same as the joining cluster context.

How about update here as follows:

```bash
./scripts/deploy-federation.sh <containerregistry>/<username>/federation-v2:test <join-cluster1> <join-cluster2>
```

**NOTE**: You can list multiple joining cluster names in above command. Please make sure the joining cluster name is same as joining cluster context.

font: I think adding a note about the joining cluster context matching the cluster name is good. I will leave it as `cluster2` because it follows with the rest of the guide https://github.com/font/federation-v2/blob/840eb7709102c6f8bdede7eb0f0f35b0fc55c1cd/docs/development.md#setup-clusters-deploy-the-cluster-registry-and-federation-v2-control-plane but will make a note about specifying any number of clusters.

docs/userguide.md
```diff
@@ -181,7 +186,7 @@ clusters that you want to test against.
     ```
 You can repeat these steps to join any additional clusters.
 
-**NOTE: `cluster-context` will default to use cluster name if not specified.**
+**NOTE:** `cluster-context` will default to use cluster name if not specified.
```
gyliu513: How about 
s/cluster/joining cluster/

docs/userguide.md
```diff
@@ -93,6 +93,11 @@ automated, then invoke the deployment script by running:
 ./scripts/deploy-federation-latest.sh cluster2
 ```
 
+**NOTE:** You can list multiple joining cluster names in above command. Also,
```
marun: s/above command/the command above/

docs/development.md
```diff
@@ -261,14 +261,22 @@ to build an image and a deployment config.
...
+specifying your own image. This assumes you've used the steps [documented
+above](#setup-clusters-deploy-the-cluster-registry-and-federation-v2-control-plane) to
+set up two minikube clusters (`cluster1` and `cluster2`):
 
 ```bash
-./scripts/deploy-federation-latest.sh <containerregistry>/<username>/federation-v2:test cluster2
+./scripts/deploy-federation.sh <containerregistry>/<username>/federation-v2:test cluster2
 ```
 
+**NOTE:** You can list multiple joining cluster names in above command. Also,
```
marun: s/above command/the command above/

(No action required) This seems to be entirely duplicated in the userguide.  Is there an intent to reuse text snippets or use references in the future?

scripts/deploy-federation.sh
```diff
@@ -93,6 +93,7 @@ if [[ ! "${USE_LATEST}" ]]; then
     cp ${base_dir}/bin/controller-manager ${dockerfile_dir}/controller-manager
   fi
   docker build ${dockerfile_dir} -t "${IMAGE_NAME}"
+  docker push "${IMAGE_NAME}"
```
marun: (No action required) Hmm, I didn't see this additional complexity pre-merge.  @irfanurrehman is there a reason this script doesn't reuse the makefile?

irfanurrehman: The script predates Makefile; I guess it makes sense to update all scripts to use Makefile now. Created https://github.com/kubernetes-sigs/federation-v2/issues/182 for the same.

docs/userguide.md
font: Updated.

docs/userguide.md
font: Updated.

scripts/deploy-federation.sh
font: Yes, we need to migrate to the use of the Makefile throughout the repo e.g. docs and scripts.

docs/development.md
font: This section, *Automated Deployment*, contains a slightly different command than what's captured in the user guide but the **NOTE** still applies. Because of that, I opted to repeat the text instead of adding a reference to go look at the **NOTE** in the other user guide's equivalent section in order to avoid any confusion.

It'd be nice to have a feature to maintain content in separate markdown files and use some sort of import/include statement to source and render imported markdowns whenever there is any re-use of text.

PR #179, opened 2018-07-13
Fix docker tag for latest

onyiny-ang:


scripts/imagebuild.sh
```diff
@@ -54,6 +54,8 @@ echo "Pushing image with tag '${TAG}'."
 docker push ${REGISTRY}${REPO}/federation-v2:${TAG}
 
 if ! [ -z "${LATEST}" ]; then
```
pmorie: since `nounsent` is set, I think you're going to want to say `${LATEST:""}` here

scripts/imagebuild.sh
```diff
@@ -54,6 +54,8 @@ echo "Pushing image with tag '${TAG}'."
...
+
+   docker tag ${REGISTRY}${REPO}/federation-v2:${LATEST}
```
marun: This doesn't look correct - a source is needed as well:

```
docker tag ${REGISTRY}${REPO}/federation-v2:${LATEST} ${REGISTRY}${REPO}/federation-v2:${TAG}
```

Reference: https://docs.docker.com/engine/reference/commandline/tag/

irfanurrehman:
@pmorie @onyiny-ang merging this PR is not important right now, marking https://quay.io/kubernetes-multicluster/federation-v2:v0.0.1 as latest is. I guess that can be done manually this time.

scripts/imagebuild.sh
font: Yep. You'll need to put the source image first:

```
docker tag ${REGISTRY}${REPO}/federation-v2:${TAG} ${REGISTRY}${REPO}/federation-v2:${LATEST}
```

scripts/imagebuild.sh
```diff
@@ -53,7 +54,9 @@ docker build ${dockerfile_dir} -t ${REGISTRY}${REPO}/federation-v2:${TAG}
 echo "Pushing image with tag '${TAG}'."
 docker push ${REGISTRY}${REPO}/federation-v2:${TAG}
 
-if ! [ -z "${LATEST}" ]; then
+if [ "$LATEST" = "latest" ]; then
```
pmorie: The `=` operator isn't how the rest of the equality comparison tests in this file are written - would you use `==` so they're consistent?

Issue #165, opened 2018-07-10
Rename `Host` to `HostClusterContext` for `join`/`unjoin` options


gyliu513:
We are now using `Host` to identify the `HostClusterContext` for `join`/`unjoin` options, it is a bit difficult to understand, and changing this option to `HostClusterContext` is more simple to understand.

```
flags.StringVar(&o.Host, "host-cluster-context", "", "Host cluster context")
```

/cc @marun comments for this?

PR #180, opened 2018-07-17
Rename `Host` to `HostClusterContext` for `join`.

gyliu513:
Fixed https://github.com/kubernetes-sigs/federation-v2/issues/165

/cc @font @marun

### DIFF
--- a/pkg/kubefed2/join.go
+++ b/pkg/kubefed2/join.go
@@ -127,7 +127,7 @@ func (j *joinFederation) Complete(args []string) error {
 	}
 
 	glog.V(2).Infof("Args and flags: name %s, host: %s, host-system-namespace: %s, kubeconfig: %s, cluster-context: %s, secret-name: %s, dry-run: %v",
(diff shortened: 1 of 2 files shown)


marun:
Ugh, seeing transient travis build failures due to running both e2e and integration.

@gyliu513 I'll allow this.  However, the code is alpha quality, intended to lay down enough functionality for users to provide substantive feedback.  I would like to avoid filing issues for every single corner we've cut in the process.  I would rather see trivial refactors and renames such as this one incorporated into more substantive work to minimize the overhead involved.

gyliu513:
> I would rather see trivial refactors and renames such as this one incorporated into more substantive work to minimize the overhead involved.

@marun so you mean that we want to combine such fix when doing some other function level change?

marun:
@gyliu513 Correct.  Much of the code quality has not been exhaustively reviewed and lacks comprehensive test coverage, in keeping with work whose primary purpose is to validate assumptions about user requirements.  As such, I think minor refactoring and stabilization efforts are of a lower priority to work driven by user requirements.

PR #181, opened 2018-07-17
Correct the path for docker file.

gyliu513:
/cc @font @onyiny-ang


onyiny-ang:
Good catch!
/lgtm
/approve

PR #184, opened 2018-07-20
Enable CI e2e job against single minikube cluster

font:
This updates the pre-commit CI job. It continues to run integration tests but now runs [unmanaged](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md#unmanaged-and-hybrid-cluster-setup) instead of [managed](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md#managed) e2e tests i.e. a single minikube k8s cluster.

This works because we are requesting an Ubuntu 16.04 distro (Xenial) from Travis in order to use systemd. This distro is not actually documented in Travis, but appears to be available as an experimental feature.

### DIFF
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -85,9 +85,10 @@ JOIN_CLUSTERS="${*}"
 
 # Use DOCKER_PUSH=false ./scripts/deploy-federation.sh <image> to skip docker
 # push on container image when not using latest image.
-DOCKER_PUSH=${DOCKER_PUSH:-docker push ${IMAGE_NAME}}
+DOCKER_PUSH=${DOCKER_PUSH:-true}
+DOCKER_PUSH_CMD="docker push ${IMAGE_NAME}"
 if [[ "${DOCKER_PUSH}" == false ]]; then
-    DOCKER_PUSH=
+    DOCKER_PUSH_CMD=
 fi
 
 if [[ ! "${USE_LATEST}" ]]; then
@@ -100,7 +101,7 @@ if [[ ! "${USE_LATEST}" ]]; then
     cp ${base_dir}/bin/controller-manager ${dockerfile_dir}/controller-manager
   fi
   docker build ${dockerfile_dir} -t "${IMAGE_NAME}"
-  ${DOCKER_PUSH}
+  ${DOCKER_PUSH_CMD}
   rm -f ${dockerfile_dir}/controller-manager
 fi
 
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -83,6 +83,13 @@ fi
 shift
 JOIN_CLUSTERS="${*}"
 
+# Use DOCKER_PUSH=false ./scripts/deploy-federation.sh <image> to skip docker
+# push on container image when not using latest image.
+DOCKER_PUSH=${DOCKER_PUSH:-docker push ${IMAGE_NAME}}
+if [[ "${DOCKER_PUSH}" == false ]]; then
+    DOCKER_PUSH=
+fi
+
 if [[ ! "${USE_LATEST}" ]]; then
   base_dir="$(cd "$(dirname "$0")/.." ; pwd)"
   dockerfile_dir="${base_dir}/images/federation-v2"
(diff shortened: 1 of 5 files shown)


scripts/deploy-federation.sh
```diff
@@ -83,6 +83,13 @@ fi
 shift
 JOIN_CLUSTERS="${*}"
 
+# Use DOCKER_PUSH=false ./scripts/deploy-federation.sh <image> to skip docker
+# push on container image when not using latest image.
+DOCKER_PUSH=${DOCKER_PUSH:-docker push ${IMAGE_NAME}}
+if [[ "${DOCKER_PUSH}" == false ]]; then
```
irfanurrehman: (Action not necessary): Consider using a different flag for the user intent.
Say a user thinks this to be a flag and provides `DOCKER_PUSH=true ./script/deploy-federation.sh`

scripts/pre-commit.sh
```diff
@@ -20,31 +20,83 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+
+function build-binaries() {
+  go build -o bin/controller-manager ./cmd/controller-manager
```
irfanurrehman: (Action not necessary), we can use make commands now.
You can as well differ this to be done separately.

scripts/pre-commit.sh
```diff
@@ -20,31 +20,83 @@ set -o errexit
...
+  sudo chown -R $USER $HOME/.kube
+  sudo chgrp -R $USER $HOME/.kube
+
+  sudo chown -R $USER $HOME/.minikube
+  sudo chgrp -R $USER $HOME/.minikube
+}
+
+# Wait for the kubernetes API server to be ready.
+function kube-apiserver-ready() {
+  local result="$(kubectl -n kube-system get pod kube-apiserver-minikube -o jsonpath='{.status.conditions[?(@.type == "Ready")].status}' 2> /dev/null)"
```
marun: Calling the apiserver to get the pod status of the apiserver seems strange, am I missing something?  Regardless, consider calling `kubectl get --raw=/healthz` instead.

scripts/deploy-federation.sh
marun: +1 to using a flag separate from the command.  Also, consider using the pattern of `FLAG=y|[empty]` and `[[ -z "${FLAG}" ]]` instead of requiring a specific value.

marun:
Nice work, and it's passing!  Minor nits as per @irfanurrehman.  Also, would it be possible to redirect output of minikube setup to a file instead of dumping to stdout to have less noise to filter through when tests fail?

font:
> Also, would it be possible to redirect output of minikube setup to a file instead of dumping to stdout to have less noise to filter through when tests fail?

The minikube setup actually doesn't really output that much to stdout. I think the problem you're referring to is the `-x` debug output from the entire pre-commit.sh script that is passed in the [travis config](https://github.com/kubernetes-sigs/federation-v2/blob/d574abd121b4c596d2fa26e5bbfc973cb57b89c6/.travis.yml#L18). I also felt like it was excessive. However, with the recent update to check `/healthz` for cluster readiness, it now comes back as `ok` immediately (somewhat concerning) so the log becomes much shorter and manageable.

Alternatively, if we feel the debug output is not useful, I can add a `DEBUG` flag that is not set by default. If an error occurs we'd just have to update the travis config to investigate.

scripts/pre-commit.sh
font: After updating to use this, the cluster readiness comes back as `ok` almost immediately. This is somewhat concerning because several `kube-system` components are still not ready. I'd have to investigate it further, but for now it seems to be working.

irfanurrehman: Would using this additionally be of any value?
`kubectl get po -n kube-system  --field-selector="status.phase!=Running"`
which should return
`No resources found.`

I wouldn't fret too much on it, if its working nevertheless; we would anyways start to see test failures if the cluster isn't healthy.

font: I think this would work if all pods were launched. But there will be pods that aren't launched yet so their status may not even exist.

pmorie:
I kicked the travis build so we can be sure this still works. This looks like a very fine start and is LGTM to me.

PR #186, opened 2018-07-24
Serve DNS endpoints from CRD

shashidharatd:
This pr adds a new CRD `DNSEndpoint` in `multiclusterdns.federation.k8s.io` group which contains list of `DNS endpoints` in status. These DNS endpoints have the same structure as defined in `external-dns`. https://github.com/kubernetes-incubator/external-dns/blob/master/endpoint/endpoint.go#L111

A new controller `servicednsendpoint` is also added which monitors `MultiClusterServiceDNSRecord` objects and creates/updates/deletes the corresponding `DNSEndpoint` objects and is responsible for implementing the `cross-cluster service discovery` feature by putting out the DNS records for the federated service.

Further to this external-dns needs to have a generic source which can list the `DNSEndpoint` CRD objects and extract DNS Endpoints. (Note: there is no source level dependency bw federation-v2 and external-dns repos). I have already implemented this `generic source` and will raise a pr to external-dns.

Note: We will not be using the DNS provider library from federation v1 and instead delegate writing DNS names to specific providers using `external-dns`.

/cc @pmorie @marun @kubernetes-sigs/federation-wg 

p.s: I am adding tests for the code and have raised this pr to get comments on overall direction.


shashidharatd:
Added the test cases and its ready for review. @pmorie, @marun, PTAL Thanks !

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
```diff
@@ -283,7 +283,6 @@ var (
 			Scope: "Namespaced",
 			Validation: &v1beta1.CustomResourceValidation{
 				OpenAPIV3Schema: &v1beta1.JSONSchemaProps{
-					Type: "object",
```
irfanurrehman: Apologies for my ignorance, but why is this change needed as part of this PR?

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
```diff
@@ -1397,7 +1396,6 @@ var (
...
```
irfanurrehman: As above!

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	)
+
+	d.minRetryDelay = minRetryDelay
+	d.maxRetryDelay = maxRetryDelay
+	if minimizeLatency {
+		d.minimizeLatency()
+	}
+	d.queue = workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(d.minRetryDelay, d.maxRetryDelay), userAgent)
+
+	netWrapper = &NetWrapperDefaultImplementation{}
```
irfanurrehman: 1. The declaration of variable netWrapper, should ideally be before this statement; it took me couple of minutes to search where this variable appeared here from.
2. It seems like u just want it initialized once, would `init()` be a more suitable location?

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+
+	for _, clusterDNS := range dnsObject.Status.DNS {
+		globalTargets = append(globalTargets, ExtractLoadBalancerTargets(clusterDNS.LoadBalancer)...)
+	}
+
+	return zoneTargets, regionTargets, globalTargets
+}
+
+// Abstracting away the internet
+type NetWrapper interface {
```
irfanurrehman: Are you expecting multiple implementations for this interface? If the answer is 'NO' "for now", probably this interface is not needed.

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
shashidharatd: The generated code by kubebuilder for types containing status sub-resource has issue during validation. So the work around is to skip that validation for now by manually fixing. this issue should go away once the pr https://github.com/kubernetes/kubernetes/pull/65357 is cherry-picked to 1.11 branch or we move to 1.12 branch. 
here https://github.com/kubernetes/kubernetes/pull/65357#issuecomment-401919242 is more details on that conversation

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: Agree with your comments. updated accordingly. PTAL.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: This is for test purpose, you can see the unit tests for its usage.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	<-stopCh
+}
+
+func (d *ServiceDNSEndpointController) needsUpdate(oldObject, newObject *feddnsv1a1.MultiClusterServiceDNSRecord) bool {
+	if !reflect.DeepEqual(oldObject.Spec, newObject.Spec) {
+		return true
+	}
+	if !reflect.DeepEqual(oldObject.Status, newObject.Status) {
```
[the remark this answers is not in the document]
shashidharatd: The objects we are comparing here is `MultiClusterServiceDNSRecord` and not the `DNSEndpoint`. 
`MultiClusterServiceDNSRecord` does have fields in status.

shashidharatd:
@irfanurrehman, updated and answered the comments. PTAL, Thanks !

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
irfanurrehman: Probably a comment here would make this confusion go away.

shashidharatd: have already added the below comment in code.
```
// Abstracting away the internet for testing purposes
type NetWrapper interface {
	LookupHost(host string) (addrs []string, err error)
}
```

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+				},
+			}
+
+			_, err = d.client.MulticlusterdnsV1alpha1().DNSEndpoints(namespace).Create(newDNSEndpointObject)
+		}
+		return err
+	}
+
+	dnsEndpointObject.Spec.Endpoints = dnsEndpoints
+	_, err = d.client.MulticlusterdnsV1alpha1().DNSEndpoints(namespace).Update(dnsEndpointObject)
```
irfanurrehman: Does it make sense to update only if there is a difference in existing and newly computed endpoints?

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	return err
+}
+
+// GetEndpointsForServiceDNSObject returns endpoint objects for each MultiClusterServiceDNSRecord object that should be processed.
+func GetEndpointsForServiceDNSObject(dnsObject *feddnsv1a1.MultiClusterServiceDNSRecord) ([]*feddnsv1a1.Endpoint, error) {
+	endpoints, err := getNamesForDNSObject(dnsObject, dnsObject.Spec.FederationName, dnsObject.Spec.DNSSuffix)
```
irfanurrehman: Do you really need to have 3 arguments to the function; it effectively seems that `dnsobject` is the only actual argument.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	// Sort endpoints by DNSName
+	sort.Slice(endpoints, func(i, j int) bool {
+		return endpoints[i].DNSName < endpoints[j].DNSName
+	})
+
+	// Remove the endpoint with no targets/ empty targets
+	for i := 0; i < len(endpoints); {
+		for j := 0; j < len(endpoints[i].Targets); {
+			if endpoints[i].Targets[j] == "" {
+				endpoints[i].Targets = append(endpoints[i].Targets[:j], endpoints[i].Targets[j+1:]...)
```
irfanurrehman: I dont think you intend to preserve order here; if that is the case, you can use a much more efficient version as listed [here](https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order)

irfanurrehman:
@shashidharatd few more minor ones. Apologies for not giving all of them in one shot.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	endpoints, err := getNamesForDNSObject(dnsObject, dnsObject.Spec.FederationName, dnsObject.Spec.DNSSuffix)
+	if err != nil {
+		return nil, err
+	}
+
+	return DedupeAndMergeEndpoints(endpoints), nil
+}
+
+// Merge and remove duplicate endpoints
+func DedupeAndMergeEndpoints(endpoints []*feddnsv1a1.Endpoint) (result []*feddnsv1a1.Endpoint) {
```
marun: (No action required) Consider making this function package local since it is not used outside of the package.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+		}
+		i++
+	}
+	return targets
+}
+
+func getNamesForDNSObject(dnsObject *feddnsv1a1.MultiClusterServiceDNSRecord, federation, dnsSuffix string) ([]*feddnsv1a1.Endpoint, error) {
+	endpoints := []*feddnsv1a1.Endpoint{}
+
+	commonPrefix := strings.Join([]string{dnsObject.Name, dnsObject.Namespace, federation, "svc"}, ".")
```
marun: (No action required) Consider extracting construction of the common prefix into a utility function since it appears once here and 9 times in the tests.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+// getHealthyTargets returns the hostnames and/or IP addresses of healthy endpoints for the service, at a zone, region and global level (or an error)
+func getHealthyTargets(zone, region string, dnsObject *feddnsv1a1.MultiClusterServiceDNSRecord) (zoneTargets, regionTargets, globalTargets feddnsv1a1.Targets) {
+	// If federated dnsObject is deleted, return empty endpoints, so that DNS records are removed
+	if dnsObject.DeletionTimestamp != nil {
+		return zoneTargets, regionTargets, globalTargets
+	}
+
+	for _, clusterDNS := range dnsObject.Status.DNS {
```
marun: (No action required) Code golf (collect all targets in a single loop)

```golang
for _, clusterDNS := range dnsObject.Status.DNS {
  lbTargets := ExtractLoadBalancerTargets(clusterDNS.LoadBalancer)
  if clusterDNS.Zone == zone {
    zoneTargets = append(zoneTargets, lbTargets...)
  }
  if clusterDNS.Region == region {
    ...
  }
  ...
}

File: pkg/controller/servicednsendpoint/servicednsendpoint_controller.go [hunk 12]
```diff
@@ -0,0 +1,464 @@
...
+		ep.Targets = []string{uplevelCname}
+		ep.RecordType = RecordTypeCNAME
+	}
+
+	return ep, nil
+}
+
+// getResolvedTargets performs DNS resolution on the provided slice of endpoints (which might be DNS names
+// or IPv4 addresses) and returns a list of IPv4 addresses.  If any of the endpoints are neither valid IPv4
+// addresses nor resolvable DNS names, non-nil error is also returned (possibly along with a partially
```
marun: (No action required) Why is a partially complete list of resolved endpoints returned if the result is thrown away on error by the only caller?

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+			// It's not a valid IP address, so assume it's a DNS name, and try to resolve it,
+			// replacing its DNS name with its IP addresses in expandedEndpoints
+			// through an interface abstracting the internet
+			ipAddrs, err := netWrapper.LookupHost(target)
+			if err != nil {
+				glog.Errorf("Failed to resolve %s, err: %v", target, err)
+				return resolvedTargets.List(), err
+			}
+			for _, ip := range ipAddrs {
+				resolvedTargets = resolvedTargets.Union(sets.NewString(ip))
```
marun: Assuming you're intending to insert this new item into the set, is there a reason to avoid `Insert`?

```golang
resolvedTargets.Insert(ip)
```

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+			}
+		} else {
+			resolvedTargets = resolvedTargets.Union(sets.NewString(target))
```
marun: ditto re: `Insert`

pkg/controller/servicednsendpoint/servicednsendpoint_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+				{DNSName: c2RegionDNSName, Targets: []string{lb2}, RecordType: RecordTypeA, RecordTTL: minDNSTTL},
+				{DNSName: c2ZoneDNSName, Targets: []string{lb2}, RecordType: RecordTypeA, RecordTTL: minDNSTTL},
+			},
+			expectError: false,
+		},
+	}
+
+	for testName, tc := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			endpoints, err := GetEndpointsForServiceDNSObject(&tc.dnsObject)
```
marun: (No action required) I think this testing could be both simpler and more comprehensive if each non-trivial function were tested from the bottom up rather than only validating the behavior of `GetEndpointsForServiceDNSObject`.  Please consider applying this approach to future testing efforts.

test/e2e/servicedns.go
```diff
@@ -90,60 +92,147 @@ var _ = Describe("MultiClusterServiceDNS", func() {
...
-		By("Create the ServiceDNS object")
-		serviceDNSObj := common.NewServiceDNSObject(baseName, namespace)
-		serviceDNS, err := fedClient.MulticlusterdnsV1alpha1().MultiClusterServiceDNSRecords(namespace).Create(serviceDNSObj)
-		framework.ExpectNoError(err, "Error creating MultiClusterServiceDNS object %v", serviceDNS)
-		name := serviceDNS.Name
+		It("ServiceDNS object status should be updated, when service and endpoint are created in member cluster", func() {
+			namespace := f.TestNamespaceName()
 
-		serviceDNSStatus := dnsv1a1.MultiClusterServiceDNSRecordStatus{DNS: []dnsv1a1.ClusterDNS{}}
+			By("Create the ServiceDNS object")
```
marun: (No action required) I think the tense of this may be incorrect?  (e.g. ServiceDNS object status should be updated BY Creating (vs Created) the ServiceDNS object).

test/e2e/servicedns.go
```diff
@@ -90,60 +92,147 @@ var _ = Describe("MultiClusterServiceDNS", func() {
...
+			serviceDNSStatus = createClusterServiceAndEndpoints(f, name, namespace, serviceDNSStatus)
 
-			loadbalancerStatus := apiv1.LoadBalancerStatus{Ingress: []apiv1.LoadBalancerIngress{{IP: clusterLb}}}
-			serviceDNSStatus.DNS = append(serviceDNSStatus.DNS, dnsv1a1.ClusterDNS{Cluster: clusterName, LoadBalancer: loadbalancerStatus})
+			serviceDNS.Status = *serviceDNSStatus
 
-			// Ensure the test namespace exists in the target cluster.
-			_, err = client.CoreV1().Namespaces().Create(&apiv1.Namespace{
+			By("Wait for the ServiceDNS object to have correct status")
+			common.WaitForObject(tl, namespace, name, objectGetter, serviceDNS, framework.PollInterval, wait.ForeverTestTimeout)
```
marun: Am I missing something, or does the first part of this test duplicate the previous test?  If so, consider removing the previous test.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: good catch. missed it earlier. updated now.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: agree. updated.

pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
```diff
@@ -0,0 +1,68 @@
...
+type Labels map[string]string
+
+// Endpoint is a high-level way of a connection between a service and an IP
+type Endpoint struct {
+	// The hostname of the DNS record
+	DNSName string `json:"dnsName,omitempty"`
+	// The targets the DNS record points to
+	Targets Targets `json:"targets,omitempty"`
+	// RecordType type of record, e.g. CNAME, A, SRV, TXT etc
+	RecordType string `json:"recordType,omitempty"`
```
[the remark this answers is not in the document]
shashidharatd: i am trying to use the exact same structure as used by external-dns https://github.com/kubernetes-incubator/external-dns/blob/master/endpoint/endpoint.go#L111

pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// TTL for the record
+	RecordTTL TTL `json:"ttl,omitempty"`
```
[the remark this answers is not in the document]
shashidharatd: Agree, updated.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+
+func StartController(config *restclient.Config, stopChan <-chan struct{}, minimizeLatency bool) error {
+	restclient.AddUserAgent(config, userAgent)
+	fedClient := fedclientset.NewForConfigOrDie(config)
+
+	controller, err := newServiceDNSEndpointController(fedClient, minimizeLatency)
+	if err != nil {
+		return err
+	}
+	go controller.Run(stopChan, 2)
```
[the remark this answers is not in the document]
shashidharatd: Updated.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (pkgruntime.Object, error) {
+				return client.MulticlusterdnsV1alpha1().MultiClusterServiceDNSRecords(metav1.NamespaceAll).List(options)
+			},
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return client.MulticlusterdnsV1alpha1().MultiClusterServiceDNSRecords(metav1.NamespaceAll).Watch(options)
+			},
+		},
+		&feddnsv1a1.MultiClusterServiceDNSRecord{},
+		0,
```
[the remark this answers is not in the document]
shashidharatd: Done

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: d.enqueueObject,
+			UpdateFunc: func(old, cur interface{}) {
+				oldObj, ok1 := old.(*feddnsv1a1.MultiClusterServiceDNSRecord)
+				curObj, ok2 := cur.(*feddnsv1a1.MultiClusterServiceDNSRecord)
+				if ok1 && ok2 && d.needsUpdate(oldObj, curObj) {
```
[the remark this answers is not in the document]
shashidharatd: Updated now.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: I was under the impression that Insert would just insert even if it is a duplicate. Updated now to use Insert.

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: done

test/e2e/servicedns.go
shashidharatd: Updated.

test/e2e/servicedns.go
shashidharatd: yes, you were right except for very small difference, but it does not make much impact so removing the earlier test case.

shashidharatd:
@irfanurrehman, @marun, addressed the review comments. ptal. Thanks !

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
marun: A property of a set is that it does not contain duplicates, so adding a value that already exists is effectively a no-op.

marun:
@shashidharatd I suggest updating the tense of all `By` statements in the e2e test (e.g. Create->Creating and Wait->Waiting), but otherwise lgtm.  I'll let @irfanurrehman approve.

marun:
Hmm, travis job is failing on unrelated e2e test (crd).  Kicking it again in case it's a flake.

shashidharatd:
> @shashidharatd I suggest updating the tense of all By statements in the e2e test (e.g. Create-Creating and Wait->Waiting), but otherwise lgtm. I'll let @irfanurrehman approve.

Done.

irfanurrehman:
Thanks @shashidharatd.
/lgtm

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,475 @@
...
+}
+
+// Merge and remove duplicate endpoints
+func DedupeAndMergeEndpoints(endpoints []*feddnsv1a1.Endpoint) (result []*feddnsv1a1.Endpoint) {
+	// Sort endpoints by DNSName
+	sort.Slice(endpoints, func(i, j int) bool {
+		return endpoints[i].DNSName < endpoints[j].DNSName
+	})
+
+	// Remove the endpoint with no targets/ empty targets
```
font: In what cases is this true? It seems like there would always be a valid string in the Targets string array as there are checks for that before inserting values.

pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
```diff
@@ -20,21 +20,37 @@ import (
...
+	RecordTTL TTL `json:"recordTTL,omitempty"`
+	// Labels stores labels defined for the Endpoint
+	Labels Labels `json:"labels,omitempty"`
+}
 
 // DNSEndpointSpec defines the desired state of DNSEndpoint
 type DNSEndpointSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "kubebuilder generate" to regenerate code after modifying this file
+	Endpoints []*Endpoint `json:"endpoints,omitempty"`
```
font: Is there a reason to make this a pointer?

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
shashidharatd: This may happen if there are no endpoints for a service in any zone or region for the global service. we would end up with 3 CNAME records, and the last CNAME record (global level) has empty target to point to nowhere. In such case we don't write the last CNAME record. here is the test case for it https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/controller/servicednsendpoint/servicednsendpoint_controller_test.go#L192

pkg/controller/servicednsendpoint/servicednsendpoint_controller.go
```diff
@@ -0,0 +1,464 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	feddnsv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/multiclusterdns/v1alpha1"
+	fedclientset "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned"
+)
+
+const (
+	// minDNSTTL is the minimum safe DNS TTL value to use (in seconds).  We use this as the TTL for all DNS records.
+	minDNSTTL = 180
```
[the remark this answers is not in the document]
shashidharatd: I would add a field for `TTL` in MultiClusterServiceDNSRecord API in a follow up. have added a issue to track it https://github.com/kubernetes-sigs/federation-v2/issues/201

PR #191, opened 2018-07-27
Update kubebuilder binary version to 1.0.0

marun:
cc: @font

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -13,7 +13,7 @@ is a detailed list of binaries required.
 The federation deployment depends on `kubebuilder`, `etcd`, `kubectl`, and
 `kube-apiserver` >= v1.11 being installed in the path. The `kubebuilder`
 release packages all of these dependencies together, but the latest version of
-`kubebuilder` (v0.1.12 as of this writing) only includes binaries for
+`kubebuilder` (v1.0.0 as of this writing) only includes binaries for
 kubernetes 1.10. Once `kubebuilder` bundles binaries >= 1.11, it will
 only require downloading the `kubebuilder` release.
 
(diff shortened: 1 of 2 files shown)


scripts/download-binaries.sh
```diff
@@ -45,7 +45,7 @@ root_dir="$(cd "$(dirname "$0")/.." ; pwd)"
 dest_dir="${root_dir}/bin"
 mkdir -p "${dest_dir}"
 
-kb_version="0.1.12"
+kb_version="1.0.0"
```
font: Does it include kube v1.11 binaries?

gyliu513: I opened an issue at https://github.com/kubernetes-sigs/kubebuilder/issues/339 to request kubebuilder upgrade kubernetes to 1.11.

irfanurrehman:
@marun, can you please update the version in the userguide also in this PR, as pointed out by @gyliu513 so that we can merge this. Thanks!

marun:
Updated userguide reference as requested.

docs/userguide.md
```diff
@@ -31,8 +31,8 @@ Or you can install them manually yourself using the guidelines provided below.
 
 This repo depends on
 [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder)
-to generate code and build binaries. Download the [v0.1.12
-release](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v0.1.12)
+to generate code and build binaries. Download the [v1.0.0
```
gyliu513: image

gyliu513:
@marun I think `hack/install-latest.yaml` should also be updated as the kubebuilder version changed. Shall we create another PR to handle this?

marun:
@gyliu513 There is zero functional change to the yaml between 0.1.12 and 1.0.0, just the tagged version change.  I'd rather not update it until there is an actual need to do so.

gyliu513:
@marun yes, only tag version change and does not impact function, but without the update, people may feel confused if they compare the `hack/install.yaml` generated by themselves with the `hack/install-latest.yaml` in fed v2 repo.

PR #194, opened 2018-08-02
Refactor scheduling

irfanurrehman:
Updated and refactored to enable the same controller usage for other schedulers (am working on JobSchedulingPreferences)

### DIFF
--- a/pkg/schedulingtypes/replicascheduler.go
+++ b/pkg/schedulingtypes/replicascheduler.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package scheduler
+package schedulingtypes
 
 import (
 	"bytes"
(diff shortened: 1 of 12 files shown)


pkg/schedulingtypes/resources.go
```diff
@@ -0,0 +1,75 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
gyliu513: How about move this to `schedulingtypes/utils`?

pkg/schedulingtypes/typeregistry.go
```diff
@@ -0,0 +1,51 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
gyliu513: How about move this to `schedulingtypes/utils`?

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -47,7 +47,7 @@ type ReplicaScheduler struct {
 	podInformer FederatedInformer
 }
 
-func NewReplicaScheduler(fedClient fedclientset.Interface, kubeClient kubeclientset.Interface, crClient crclientset.Interface, federationEventHandler, clusterEventHandler func(pkgruntime.Object), handlers *ClusterLifecycleHandlerFuncs) *Scheduler {
+func NewReplicaScheduler(fedClient fedclientset.Interface, kubeClient kubeclientset.Interface, crClient crclientset.Interface, federationEventHandler, clusterEventHandler func(pkgruntime.Object), handlers *ClusterLifecycleHandlerFuncs) Scheduler {
```
gyliu513: Why not return a scheduler pointer?

pkg/schedulingtypes/resources.go
```diff
@@ -0,0 +1,59 @@
...
+	if t.Kind() != reflect.Ptr {
+		panic("All types must be pointers to structs.")
+	}
+
+	t = t.Elem()
+	return t.Name()
+}
+
+func GetPluralName(name string) string {
+	return strings.ToLower(name) + "s"
```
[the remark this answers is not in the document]
irfanurrehman: Example please. Meanwhile as far as I understand this is the scheme followed for all k8s (or k8s like, in our case) resources.

pkg/schedulingtypes/resources.go
irfanurrehman: This is not a utility file, but inherent component of the main functionality.

pkg/schedulingtypes/typeregistry.go
irfanurrehman: This is not a utility file, but inherent component of the main functionality.

pkg/schedulingtypes/resources.go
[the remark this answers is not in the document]
gyliu513: One example is I was creating a `federatedquery` resources, and the `kubctl get crd` return `federatedqueries`:

```
root@gyliu-dev1:~# kubectl get crds | grep quer
federatedqueries.core.federation.k8s.io                           2018-08-06T05:59:54Z
```

pkg/schedulingtypes/typeregistry.go
```diff
@@ -0,0 +1,51 @@
...
+
+func SchedulingTypes() map[string]SchedulingType {
+	result := make(map[string]SchedulingType)
+	for key, value := range typeRegistry {
+		result[key] = value
+	}
+	return result
+}
+
+func GetSchedulerFactory(typ string) SchedulerFactory {
```
marun: (No action required) Consider using a more descriptive name (e.g. schedulerType) rather than a potentially opaque short form.

Issue #197, opened 2018-08-09
Control plane only supports hard-coded system namespaces


marun:
The fedv2 control plane currently sources `FederatedCluster` resources and their associated secrets from the `federation-system` namespace and `Cluster` resources from the `kube-multicluster-public` namespace.  It would be preferable to allow the names of these namespaces to vary to allow for test isolation on shared clusters and give operators flexibility in how they deploy federation.

Tasks:
- [x] Allow the federation namespace to be configured in testing (#198)
- [x] Update FederatedTypeConfig to be namespace-scoped (#209)
- [x] Update FederatedNamespacePlacement to be namespace-scoped (#210)
- [x] Allow the cluster registry namespace to vary in testing (#211)
- [x] Enable the federation control plane to target a single namespace (#212)
- [x] Enable kubefed2 configuration of federation and cluster namespaces (#213)
- [x] Enable kubefed2 joining of a cluster for only a single namespace (#215)
- [x] Support deploying namespaced control plane (#216)
- [x] Ensure control plane can run with only authz for the namespace it runs in (#230)
- [x] Update docs to reflect option to deploy control plane namespaced (#264)

marun:
Initial implementation complete!

PR #198, opened 2018-08-09
Allow the federation system namespace to be configured in testing

marun:
This is the first step in fixing #197.  The system namespace is no longer hardcoded, and the value is passed in from the controller manager (hard-coded for now but configurable via downward api in a future PR) or the tests (dynamically determined for managed/integration and configurable for unmanaged and defaulted to the previous value).


marun:
Not sure why the travis check isn't showing up.  Tests pass locally, so I don't think that should block review, but I've added a hold label to prevent accidental merge.

shashidharatd:
The changes LGTM. the only issue though would be there is no trace of the CI test. https://travis-ci.org/kubernetes-sigs/federation-v2/pull_requests. Don't know how to trigger one.

marun:
@shashidharatd Travis is working again and is passing.

PR #148, opened 2018-07-05
Kubebuilder 0.1.12

gyliu513:
/cc @marun @shashidharatd @onyiny-ang

PR #120, opened 2018-06-19
Transition to kubebuilder + CRDs

marun:
The remaining tasks are related to manual verification of e2e testing against an unmanaged deployment of kubernetes, and shouldn't block review.

Tasks:
- [x] Integration tests passing
- [x] Binaries building
- [x] Updated pre-commit script
- [x] Update deployment script 
- [x] Run e2e against unmanaged deployment

Note: `kube-apiserver` < 1.11 is not compatible with fedv2, which means testing requires pre-release binaries (or built from master).  This is due to support for crd subresources being broken in 1.10.

PR #200, opened 2018-08-14
Added back github.com/kubernetes/repo-infra to Gopkg.toml

gyliu513:
/cc @marun @pmorie

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,6 +1,8 @@
 
 # Users add deps lines here
 
+required = ["github.com/kubernetes/repo-infra/verify/boilerplate/test"]
+
 [prune]
   go-tests = true
   #unused-packages = true


marun:
Why is this necessary?  A boilerplate test is already being invoked by travis.  For future PRs, please ensure that you provide enough information for reviewers to understand your intent.

gyliu513:
@marun The reason I create this PR is because after I checkout code and then run `dep ensure`, all of the `test-infra` files under `vendor` will be deleted, so here I was adding it back. Can you please elaborate more for how did you run `dep ensure` in your local cluster? Thanks.

marun:
@gyliu513 Why are you running `dep ensure`?  This should only be required if you are adding/changing dependencies, which is not the case in this PR.

gyliu513:
This is a issue that I encounter when submit the PR https://github.com/kubernetes-sigs/federation-v2/pull/148#commits-pushed-e2037ff , after I update the kubebuilder to v1.0.0 in `Gopkg.toml`, then run `dep ensure`, all of the `test-infra` files under `vendor` will be deleted, that's why I want to add back `test-infra` in this PR. With this PR, when people update `Gopkg.toml`, they will not have issues related with `test-infra`.

I found that in this commit https://github.com/kubernetes-sigs/federation-v2/pull/120/commits/d4a3519c0172438363016b6487164c87a83b94fe#diff-836546cc53507f6b2d581088903b1785 , you removed `required = ["github.com/kubernetes/repo-infra/verify/boilerplate/test"]`, not sure how did you make sure `github.com/kubernetes/repo-infra` in `vendor` when you run `dep ensure`?

marun:
@gyliu513 Looks like I retained the old vendored version during the transition to kubebuilder but not the `Gopkg.toml` entry.  The fix here is to add the entry, though, not to update deps unnecessarily.

marun:
Adding WIP to avoid merging before the dep updates are removed.

gyliu513:
Sorry @marun  I was confused. 

Do you mean that I only need to update `Gopkg.toml` as follows?

```
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,6 +1,8 @@
 
 # Users add deps lines here
 
+required = ["github.com/kubernetes/repo-infra/verify/boilerplate/test"]
+
 [prune]
   go-tests = true
   #unused-packages = true
```

But as we are using master branch and there are indeed some change for the master branch of `github.com/kubernetes/repo-infra/verify/boilerplate/test`, so seems we still need to call `dep ensure` to use latest code in `repo-infra`?

gyliu513:
If do not run `dep ensure`, then after I add the entry to `Gopkg.toml`, when others add new dependency package, and run `dep ensure`, it will also generate those code, so why not update the deps at this time? What is your concern of this?

marun:
/approve
/lgtm

Thank you!

PR #202, opened 2018-08-14
Add more debug log for sync to clusters.

gyliu513:


marun:
Why is this desirable?  It might be acceptable in the context of other work, but on its own I don't think it merits a PR.  The logic behind placement is really simple right now - the placement resource is authoritative.

quinton-hoole:
This seems useful to me.  It would be difficult (but not impossible) after-the-fact to determine what the state of the relevant placement resource was at the time that propagation was attempted?

/lgtm

marun:
@quinton-hoole Why do you think adding this logging detail will be useful?  Lots of things are potentially useful, but I think it's preferable to be able to justify log additions on the basis of actual utility.  The current implementation of placement is trivial - there is no possibility of ambiguity - and every operation invoked on underlying clusters (create/update/delete) is already logged as an event.

marun:
@gyliu513 Travis is reporting a syntax error.

pkg/controller/sync/controller.go
```diff
@@ -590,7 +590,8 @@ func (s *FederationSyncController) syncToClusters(selectedClusters, unselectedCl
 	templateKind := s.typeConfig.GetTemplate().Kind
 	key := util.NewQualifiedName(template).String()
 
-	glog.V(3).Infof("Syncing %s %q in underlying clusters", templateKind, key)
+	glog.V(3).Infof("Syncing %s %q in underlying clusters, selected clusters are: %s, unselected clusters are: %s",
+		+templateKind, key, selectedClusters, unselectedClusters)
```
marun: `s/+templateKind/templateKind/`

gyliu513: Done

PR #203, opened 2018-08-14
Remove duplicate import.

gyliu513:


marun:
This is what I would call a pointless PR.  This is alpha code - there are so much cleanup that will occur before beta that doing one-line non-functional fixes is just a waste of time.

quinton-hoole:
@marun I'll review this if you don't have the time. LGTM.

marun:
@gyliu513 @quinton-hoole has convinced me of the error of my ways.  Please add your branch back so it can be merged.  

For future PRs, and as I've mentioned in the past, I would prefer to avoid cleanup that is not related to functional changes for now.  When it comes time to clean things up, we can use tools (e.g. linter) to detect problems like this and resolve them all at once rather than in a piecemeal fashion.

gyliu513:
@marun as I was reviewing code and there are some things that I can fix, so just create the PR. OK, cleanups can be done in future. 

Thanks @quinton-hoole @marun

PR #205, opened 2018-08-14
Bump e2e poll timeout to account for unmanaged fixture on travis

marun:
The previous `SingleCallTimeout` value of `30s` appears to be too low, as evidenced by the increasing incidence of CI jobs failing with timeout failures since e2e tests have started executing against unmanaged fixture.


marun:
Manually merging to unblock other PRs.

Issue #201, opened 2018-08-14
Make TTL configurable by user in MultiClusterServiceDNSRecord API


shashidharatd:
Currently the DNS records created have hard coded TTL. There is a desire to make it configurable by user. See the comment https://github.com/kubernetes-sigs/federation-v2/pull/186#discussion_r206324729
If user does not care to configure it, then default value would be used for TTL.

shashidharatd:
Closed by #207

PR #207, opened 2018-08-15
Make DNS record TTL configurable by user

shashidharatd:
Addresses https://github.com/kubernetes-sigs/federation-v2/issues/201

/cc @font @kubernetes-sigs/federation-wg

### DIFF
--- a/pkg/apis/multiclusterdns/v1alpha1/multiclusterservicednsrecord_types.go
+++ b/pkg/apis/multiclusterdns/v1alpha1/multiclusterservicednsrecord_types.go
@@ -27,6 +27,8 @@ type MultiClusterServiceDNSRecordSpec struct {
 	FederationName string `json:"federationName,omitempty"`
 	// DNSSuffix is the suffix (domain) to append to DNS names
 	DNSSuffix string `json:"dnsSuffix,omitempty"`
+	// RecordTTL is the TTL in seconds for DNS records created for this Service, if omitted a default would be used
+	RecordTTL TTL `json:"recordTTL,omitempty"`
 }
 
 // MultiClusterServiceDNSRecordStatus defines the observed state of MultiClusterServiceDNSRecord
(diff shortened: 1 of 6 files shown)


marun:
lgtm, but I'll leave it to @font to approve.

pkg/apis/multiclusterdns/v1alpha1/multiclusterservicednsrecord_types.go
```diff
@@ -27,6 +27,8 @@ type MultiClusterServiceDNSRecordSpec struct {
 	FederationName string `json:"federationName,omitempty"`
 	// DNSSuffix is the suffix (domain) to append to DNS names
 	DNSSuffix string `json:"dnsSuffix,omitempty"`
+	// RecordTTL is the TTL for DNS records created for this Service, if omitted a default would be used
```
font: Can we add the unit of time measurement for this field either as a comment or as part of the field name itself e.g. `RecordTTLSeconds`?

Also, I just wanted to point out that we use the `TTL` type in the `EndPoint` API but not here. I don't necessarily have a preference either way.

gyliu513:
@shashidharatd it would be great to add some user guide for multicluster service DNS test this as well, so that people can try this.

pkg/apis/multiclusterdns/v1alpha1/multiclusterservicednsrecord_types.go
shashidharatd: Have updated the comment to specify the unit is in Seconds. Also updated the type to TTL to be in sync.

shashidharatd:
@font, have updated this pr to address your comments. PTAL.
@gyliu513, updating the user guide is on the way. some functional part is still in progress in external-dns in https://github.com/kubernetes-incubator/external-dns/pull/657. Would update the user guide as soon as it is merged.

PR #213, opened 2018-08-17
Enable kubefed2 config of federation and cluster namespaces

marun:
Targets #197

### DIFF
--- a/pkg/kubefed2/options/options.go
+++ b/pkg/kubefed2/options/options.go
@@ -29,6 +29,7 @@ type SubcommandOptions struct {
 	ClusterName         string
 	HostClusterContext  string
 	FederationNamespace string
+	ClusterNamespace    string
 	Kubeconfig          string
 	DryRun              bool
 }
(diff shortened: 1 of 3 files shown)


marun:
@gyliu513 Please avoid /lgtm while self-approval remains enabled (https://github.com/kubernetes/test-infra/pull/9041) to ensure other reviewers can review before merge.  The only thing stopping this from merging is the failing tide job.

shashidharatd:
Assuming others won't have any comments as its a small pr and already couple of guys have reviewed it. :)
/approve

pkg/kubefed2/options/options.go
```diff
@@ -37,6 +38,10 @@ type SubcommandOptions struct {
 func (o *SubcommandOptions) CommonBind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
 	flags.StringVar(&o.HostClusterContext, "host-cluster-context", "", "Host cluster context")
+	flags.StringVar(&o.FederationNamespace, "federation-namespace", util.DefaultFederationSystemNamespace,
+		"Namespace in the host cluster where the federation system components are installed")
+	flags.StringVar(&o.ClusterNamespace, "cluster-namespace", util.MulticlusterPublicNamespace,
```
font: The use of `cluster-namespace` immediately makes me think of the joining cluster. How about something like `registry-namespace` since we already have `add-to-registry`?

marun: Sure, I'll add a follow-up PR.

As an aside, the concept of 'cluster registry' no longer makes sense to me now that the resource is namespaced.  All we have is a cluster api resource rather than the central concept of a registry.

font: I see your point. In contrast, the concept of a central registry could be cluster wide or namespace wide, depending on who's using it. Can we come up with a better name? I just wanted to avoid the overused term `cluster`.

marun: Yeah, that reminds me of that thread on the ml about cr naming (i.e. maybe something like `ApiEndpoint` is more appropriate).

PR #215, opened 2018-08-20
Enable kubefed to join a cluster for only the federation namespace

marun:
Targets #197

### DIFF
--- a/pkg/kubefed2/join.go
+++ b/pkg/kubefed2/join.go
@@ -70,6 +70,7 @@ type joinFederationOptions struct {
 	clusterContext string
 	secretName     string
 	addToRegistry  bool
+	limitedScope   bool
 }
 
 // Bind adds the join specific arguments to the flagset passed in as an
@@ -81,6 +82,8 @@ func (o *joinFederationOptions) Bind(flags *pflag.FlagSet) {
 		"Name of the secret where the cluster's credentials will be stored in the host cluster. This name should be a valid RFC 1035 label. If unspecified, defaults to a generated name containing the cluster name.")
 	flags.BoolVar(&o.addToRegistry, "add-to-registry", false,
 		"Add the cluster to the cluster registry that is aggregated with the kubernetes API server running in the host cluster context.")
+	flags.BoolVar(&o.limitedScope, "limited-scope", false,
+		"Whether the federation namespace (configurable via --federation-namespace) will be the only target for federation.  If true, join will add a service account with access only to the federation namespace in the target cluster.")
 }
 
 // NewCmdJoin defines the `join` command that joins a cluster to a
@@ -125,9 +128,9 @@ func (j *joinFederation) Complete(args []string) error {
 		j.clusterContext = j.ClusterName
 	}
 
-	glog.V(2).Infof("Args and flags: name %s, host: %s, host-system-namespace: %s, cluster-namespace: %s, kubeconfig: %s, cluster-context: %s, secret-name: %s, dry-run: %v",
+	glog.V(2).Infof("Args and flags: name %s, host: %s, host-system-namespace: %s, cluster-namespace: %s, kubeconfig: %s, cluster-context: %s, secret-name: %s, limited-scope: %s, dry-run: %v",
 		j.ClusterName, j.HostClusterContext, j.FederationNamespace, j.ClusterNamespace, j.Kubeconfig, j.clusterContext,
-		j.secretName, j.DryRun)
+		j.secretName, j.limitedScope, j.DryRun)
 
 	return nil
 }
@@ -149,7 +152,7 @@ func (j *joinFederation) Run(cmdOut io.Writer, config util.FedConfig) error {
 	}
 
 	err = JoinCluster(hostConfig, clusterConfig, j.FederationNamespace, j.ClusterNamespace,
-		j.HostClusterContext, j.ClusterName, j.secretName, j.addToRegistry, j.DryRun)
+		j.HostClusterContext, j.ClusterName, j.secretName, j.addToRegistry, j.limitedScope, j.DryRun)
 	if err != nil {
 		return err
 	}
@@ -160,7 +163,7 @@ func (j *joinFederation) Run(cmdOut io.Writer, config util.FedConfig) error {
 // JoinCluster performs all the necessary steps to join a cluster to the
 // federation provided the required set of parameters are passed in.
 func JoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, clusterNamespace,
-	hostClusterContext, joiningClusterName, secretName string, addToRegistry, dryRun bool) error {
+	hostClusterContext, joiningClusterName, secretName string, addToRegistry, limitedScope, dryRun bool) error {
 	hostClientset, err := util.HostClientset(hostConfig)
 	if err != nil {
 		glog.V(2).Infof("Failed to get host cluster clientset: %v", err)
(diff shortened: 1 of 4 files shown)


pkg/kubefed2/join.go
```diff
@@ -534,6 +552,116 @@ func createClusterRoleBinding(clusterClientset client.Interface, saName, namespa
...
+		},
+		{
+			Verbs:           []string{"Get"},
+			NonResourceURLs: []string{"/healthz"},
+		},
+	}
+
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: roleName,
```
gyliu513: It is better to add namespace here for metadata. I know that it is not required, but adding namespace can make the logic clear.

pkg/kubefed2/util/util.go
```diff
@@ -113,9 +113,9 @@ func ClusterServiceAccountName(joiningClusterName, hostClusterName string) strin
...
 }
 
-// ClusterRoleName returns the name of a ClusterRole and its associated
-// ClusterRoleBinding that are used to allow the service account to
-// access necessary resources on the cluster.
-func ClusterRoleName(serviceAccountName string) string {
+// RoleName returns the name of a Role or ClusterRole and its
+// associated RoleBinding or ClusterRoleBinding that are used to allow
+// the service account to access necessary resources on the cluster.
+func RoleName(serviceAccountName string) string {
```
gyliu513: Shall we enhance this a bit by adding namespace to this as well?

Perhaps adding two util functions maybe better? One is for `RoleName` with `namespace` and another is `ClusterRoleName` without `namespace`?

```
func RoleName(serviceAccountName, namespace string) string {
}
```
```
func ClusterRoleName(serviceAccountName string) string {
}
```

pkg/kubefed2/join.go
```diff
@@ -534,6 +552,116 @@ func createClusterRoleBinding(clusterClientset client.Interface, saName, namespa
 	return &clusterRoleBinding, nil
 }
 
+// createRoleBinding creates an RBAC role and binding that allows the
+// service account identified by saName to access all resources in the
+// specified namespace.
+func createRoleBinding(clusterClientset client.Interface, saName, namespace,
```
irfanurrehman: [Suggestion: Implementation not necessary in this PR to merge] Some portion of code in `createRoleBinding()` and `createClusterRoleBinding()` is common, which can be pushed to a common function and reused in both.

irfanurrehman:
@marun this seems fine to me. There's only one suggestion, which is not necessary for this PR. I will leave   it for @font to have a look and approve.

pkg/kubefed2/join.go
marun: There is a single namespace parameter in this function, I'm not sure how adding it redundantly to the type is going to 'make the logic clear'.

pkg/kubefed2/util/util.go
marun: I call YAGNI on both counts.  Having the same name for different resources is fine, they are entirely different contexts.  If/when we see a use case for clusters participating in multiple federations, I can imagine a single cluster role allowing access to `healthz` being defined once for the cluster by an administrator rather than relying on a tool like kubefed.

pkg/kubefed2/join.go
marun: I've posted a new PR that factors things out as you suggest: #222

pkg/kubefed2/join.go
gyliu513: Usually for a Role, we always have `namespace` specified in `metadata` as follows:

```
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  namespace: default
  name: pod-reader
rules:
- apiGroups: [""] # "" indicates the core API group
  resources: ["pods"]
  verbs: ["get", "watch", "list"]
```

For ClusterRole, do not specify namespace.
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  # "namespace" omitted since ClusterRoles are not namespaced
  name: secret-reader
rules:
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["get", "watch", "list"]
```

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#api-overview

pkg/kubefed2/join.go
```diff
@@ -81,6 +82,8 @@ func (o *joinFederationOptions) Bind(flags *pflag.FlagSet) {
 		"Name of the secret where the cluster's credentials will be stored in the host cluster. This name should be a valid RFC 1035 label. If unspecified, defaults to a generated name containing the cluster name.")
 	flags.BoolVar(&o.addToRegistry, "add-to-registry", false,
 		"Add the cluster to the cluster registry that is aggregated with the kubernetes API server running in the host cluster context.")
+	flags.BoolVar(&o.limitedScope, "limited-scope", false,
+		"Whether the federation namespace will be the only target for federation.  If true, join will add a service account with access only to the federation namespace in the target cluster.")
```
font: Consider adding a reference to the `--federation-namespace` flag to this flag description.

Also, the description for the `federation-namespace` flag probably needs to be updated to reflect that it can be the namespace to target in the joining cluster.

pkg/kubefed2/join.go
```diff
@@ -81,6 +82,8 @@ func (o *joinFederationOptions) Bind(flags *pflag.FlagSet) {
 		"Name of the secret where the cluster's credentials will be stored in the host cluster. This name should be a valid RFC 1035 label. If unspecified, defaults to a generated name containing the cluster name.")
 	flags.BoolVar(&o.addToRegistry, "add-to-registry", false,
 		"Add the cluster to the cluster registry that is aggregated with the kubernetes API server running in the host cluster context.")
+	flags.BoolVar(&o.limitedScope, "limited-scope", false,
```
font: (no action required) This implies that the user will only ever want to federate resources into the `--federation-namespace` they specified when they originally joined the cluster. What if they want to add more namespaces? This would limit the ability to specify either all namespaces or just one, but not X number of namespaces.

I think we may need to eventually provide the ability to specify multiple namespaces initially upon join. Then if they later want to add or reduce the number of namespaces, we could eventually provide an option to do that or just have them unjoin/rejoin the cluster as an expedient, but that could be catastrophic to their existing workload.

In any case, this is good enough for now unless we want to consider using a flag to list multiple namespaces and limit it to one for now.

marun: An informer can be configured to target a single namespace or all namespaces, but not multiple namespaces.  Supporting multiple namespaces is possible, but would require refactoring the sync controller to abstract use of the informers and to create an informer per targeted namespace.  Because of this complexity, and the lack of users (let alone ones calling for the feature), I call YAGNI.  If a user wants to target multiple namespaces, they can deploy a controller manager per namespace (this is a common pattern for namespaced controllers).

pkg/kubefed2/join.go
marun: Done.

The thought exercise of considering the relationship between the 2 args led to the thought that `kubefed2` should be discovering control plane configuration (the cluster registry namespace and whether the control plane is targeting a single namespace) rather than requiring it to be provided on every invocation.

pkg/kubefed2/join.go
marun: I get that it's possible to specify the namespace for a namespaced-scoped resource, and not possible for a cluster-scoped resource.  It's not necessary to specify the namespace for a namespace-scoped resource where the namespace will be specified in a creation call (e.g. `kubectl -n foo create -f myresource.yaml` or `MyResources("foo").Create()`).  

I disagree that there is value in including it 'for clarity' in this case, since the definition and creation of the resource are both performed in the context of a simple function.

pkg/kubefed2/join.go
font: Yes, agreed! Perhaps that same mechanism can be used to configure the federation control plane.

pkg/kubefed2/join.go
font: Yeah that gets complicated. YAGNI sgtm.

PR #220, opened 2018-08-21
Adds .idea to .gitignore

danehans:
Updates .gitignore to ignore Intellij IDE project settings
stored in a .idea directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /config
 /hack/install.yaml
 /images/federation-v2/controller-manager
+.idea


marun:
@danehans @onyiny-ang I would rather not accept ignore changes that are environment specific.   I think it's preferable for developers to configure local ignores:

```
Run git config --global core.excludesfile ~/.gitignore and add patterns to your ~/.gitignore. 
This option applies if you want to ignore certain patterns across all trees. 
I use this for .pyc and .pyo files, for example.
```
Source: https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

font:
I agree with @marun and were my thoughts exactly. We avoided the same thing in cluster registry for the same reasons.

danehans:
@marun that worked for me. Thanks for the git pro tip :-)

You can close the issue if you would like.

gyliu513:
You can also update `.git/info/exclude` to exclude specified files based on your environment.

Issue #219, opened 2018-08-21
Incomplete Pluralization of Types


danehans:
53566d9fc6d4062f9c15504af7dda5bf5c42e81b pluralizes federated types ending with "s", such as `Ingress`. However, adding "es" to federated types should be extended beyond suffixes ending with "s".

I will submit a PR to fix this issue following the proper pluralization [guidelines](https://howtospell.co.uk/adding-es-plural-rule).

marun:
Fixed by #221

PR #221, opened 2018-08-21
Extends Pluralization Support

danehans:
Previously, the pluralName function would only add 'es' to
strings with an 's' suffix. Adding 'es' should be used for
additional suffixes. This will prevent improper pluralization
of future supported types. See:

https://howtospell.co.uk/adding-es-plural-rule

Fixes Issue: #219

### DIFF
--- a/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
+++ b/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
@@ -99,12 +99,12 @@ func SetFederatedTypeConfigDefaults(obj *FederatedTypeConfig) {
 		group := nameParts[1]
 		setStringDefault(&obj.Spec.Target.Group, group)
 	}
-	setStringDefault(&obj.Spec.Template.PluralName, pluralName(obj.Spec.Template.Kind))
-	setStringDefault(&obj.Spec.Placement.PluralName, pluralName(obj.Spec.Placement.Kind))
+	setStringDefault(&obj.Spec.Template.PluralName, PluralName(obj.Spec.Template.Kind))
+	setStringDefault(&obj.Spec.Placement.PluralName, PluralName(obj.Spec.Placement.Kind))
 	setStringDefault(&obj.Spec.Placement.Group, obj.Spec.Template.Group)
 	setStringDefault(&obj.Spec.Placement.Version, obj.Spec.Template.Version)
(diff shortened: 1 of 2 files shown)


pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -119,11 +119,13 @@ func setStringDefault(value *string, defaultValue string) {
...
 
-// PluralNameForKind naively computes the plural name from the kind by
-// lowercasing and suffixing with 's'.
-func pluralName(kind string) string {
+// PluralName computes the plural name from the kind by
+// lowercasing and suffixing with 's' or `es`.
+func PluralName(kind string) string {
 	lowerKind := strings.ToLower(kind)
-	if strings.HasSuffix(lowerKind, "s") {
+	if strings.HasSuffix(lowerKind, "s") || strings.HasSuffix(lowerKind, "x") ||
```
pmorie: Is this list of suffixes inspired by a list written by people who have thought a lot of about pluralizing? (I have not :))

pmorie: Whoops, missed the link in your PR description.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -119,11 +119,13 @@ func setStringDefault(value *string, defaultValue string) {
 	*value = defaultValue
 }
 
-// PluralNameForKind naively computes the plural name from the kind by
-// lowercasing and suffixing with 's'.
-func pluralName(kind string) string {
+// PluralName computes the plural name from the kind by
+// lowercasing and suffixing with 's' or `es`.
```
gyliu513: @danehans do we need to handle the case if the word ends up with `y`? Such as `library` should be `libraries`?

danehans: Probably a good idea.

marun: I call YAGNI - let's focus on stuff we actually have a need for.  I think it's premature to optimize any one piece while we're still in alpha.

danehans: :-)

Issue #188, opened 2018-07-26
`unjoin` should succeed if some resources remove failed.


gyliu513:
This is a follow up issue based on comments at https://github.com/kubernetes-sigs/federation-v2/pull/137#issuecomment-406447550

From @marun :  I am still stuck on error handling. Removal of any resource would ideally not fail if not found, ensuring that cleanup of partial configuration is possible.

/assign

gyliu513:
@xunpan can you help on this?

raffaelespazzoli:
from the #410 that was marked as a duplicate of this, I think that join should succeed if some of the resources that it would have created already exist. Join should update these resources to the desired state when possible. In order to do this join should use an `apply` strategy rather than a `create` strategy.

PR #222, opened 2018-08-21
kubefed: clean up role and binding creation

marun:
This PR cleans up role and binding creation by separating out the common healthz rule into a new role.  The existing methods are cleaned up for consistency.

This is in response to a comment from @irfanurrehman: https://github.com/kubernetes-sigs/federation-v2/pull/215#discussion_r211552117

### DIFF
--- a/pkg/kubefed2/unjoin.go
+++ b/pkg/kubefed2/unjoin.go
@@ -317,8 +317,7 @@ func deleteRBACResources(hostClusterClientset, unjoiningClusterClientset client.
 	glog.V(2).Infof("Deleting cluster role binding for service account: %s in unjoining cluster: %s",
 		saName, unjoiningClusterName)
 
-	err := deleteClusterRoleAndBinding(unjoiningClusterClientset, saName, namespace,
-		unjoiningClusterName, dryRun)
+	err := deleteClusterRoleAndBinding(unjoiningClusterClientset, saName, namespace, dryRun)
 	if err != nil {
 		glog.V(2).Infof("Error deleting cluster role binding for service account: %s in unjoining cluster: %v",
 			saName, err)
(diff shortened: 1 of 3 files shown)


marun:
Whoops, closure was accidental.

pkg/kubefed2/unjoin.go
```diff
@@ -381,33 +380,32 @@ func deleteServiceAccount(clusterClientset client.Interface, saName,
...
-	if err != nil && !errors.IsNotFound(err) {
-		glog.V(2).Infof("Could not delete cluster role binding for service account: %s in unjoining cluster: %v",
-			saName, err)
-		return err
-	}
+	for _, name := range []string{roleName, healthzRoleName} {
+		err := clusterClientset.RbacV1().ClusterRoleBindings().Delete(name, &metav1.DeleteOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			glog.V(2).Infof("Could not delete cluster role binding %q in unjoining cluster: %v", name, err)
+			return err
```
irfanurrehman: Nit: Would it rather make sense to continue and try delete/remove remaining roles/bindings if an error occurs on one. I know this is a tricky bit in unjoin flow, this however makes more sense to me.

We can have a separate discussion on this outside this particular PR. I always felt that unjoin should ideally try and delete/remove all the resources sequentially, only logging failures, and always finishing the sequence, which does not seem to be the case now. If that is the case, unjoin command can be triggered again in case a failure is seen (NotFound is obviously not a failure).

irfanurrehman:
@marun Apologies for visiting this PR late.
Seems fine to me except one nit, which probably needs a discussion outside this PR. Will create an issue if that seems like a logical thing to do. Merging this anyways, but please do post your opinion on [this comment](https://github.com/kubernetes-sigs/federation-v2/pull/222/files#r214798913).

/lgtm

pkg/kubefed2/unjoin.go
marun: I agree that unjoin should be resilient to the case of missing resources.  It wasn't my intent to improve the error handling here (just simplify role handling as you requested in a previous PR), but it is intended to be done as per #188.

irfanurrehman: Ack.

PR #224, opened 2018-08-22
Add godoc for FederatedCluster

pmorie:
what it sounds like

### DIFF
--- a/pkg/apis/core/v1alpha1/federatedcluster_types.go
+++ b/pkg/apis/core/v1alpha1/federatedcluster_types.go
@@ -25,7 +25,8 @@ import (
 
 // FederatedClusterSpec defines the desired state of FederatedCluster
 type FederatedClusterSpec struct {
-	// Name of the cluster resource from which to source api endpoints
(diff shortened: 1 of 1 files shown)


pkg/apis/core/v1alpha1/federatedcluster_types.go
```diff
@@ -61,7 +63,10 @@ type FederatedClusterStatus struct {
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// FederatedCluster
+// FederatedCluster configures federation to be aware of a Kubernetes cluster
+// from the cluster-registry and provides a Kubeconfig for federation to use to
```
gyliu513: Do we need to mention this `provides a Kubeconfig` at here again? I think it already been mentioned in `FederatedClusterSpec`?

irfanurrehman: Does not seem like a dead repetition.

PR #225, opened 2018-08-22
Rename --cluster-namespace to --registry-namespace

marun:
This is for consistency with --add-to-registry as per a request by @font: https://github.com/kubernetes-sigs/federation-v2/pull/213#discussion_r211833990

### DIFF
--- a/pkg/kubefed2/options/options.go
+++ b/pkg/kubefed2/options/options.go
@@ -40,7 +40,7 @@ func (o *SubcommandOptions) CommonBind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.HostClusterContext, "host-cluster-context", "", "Host cluster context")
 	flags.StringVar(&o.FederationNamespace, "federation-namespace", util.DefaultFederationSystemNamespace,
 		"Namespace in the host cluster where the federation system components are installed.  This namespace will also be the target of propagation if the controller manager is configured with --limited-scope and clusters are joined with --limited-scope.")
-	flags.StringVar(&o.ClusterNamespace, "cluster-namespace", util.MulticlusterPublicNamespace,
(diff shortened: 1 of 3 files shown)


pkg/kubefed2/options/options.go
```diff
@@ -40,7 +40,7 @@ func (o *SubcommandOptions) CommonBind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.HostClusterContext, "host-cluster-context", "", "Host cluster context")
 	flags.StringVar(&o.FederationNamespace, "federation-namespace", util.DefaultFederationSystemNamespace,
 		"Namespace in the host cluster where the federation system components are installed.  This namespace will also be the target of propagation if the controller manager is configured with --limited-scope and clusters are joined with --limited-scope.")
-	flags.StringVar(&o.ClusterNamespace, "cluster-namespace", util.MulticlusterPublicNamespace,
+	flags.StringVar(&o.ClusterNamespace, "registry-namespace", util.MulticlusterPublicNamespace,
```
gyliu513: How about s/ClusterNamespace/RegistryNamespace ?

marun: I don't see the point of rippling the change through the codebase while things remain in flux.  The intent of this fix is to ensure consistency of UX.

marun: I'm hesitant to want to ripple the change through the code at this time.  If the namespace is the same as the federation namespace (i.e. localized rather than cluster-wide), I don't think 'registry' is a good description.  Even for the user I don't think 'registry' makes that much sense - it's just an API resource at this point - but I agree with @font that UX consistency is a good idea.  I don't agree that the developer experience has to exactly match the UX - those are 2 different contexts.

gyliu513: Thanks @marun , sgtm

PR #230, opened 2018-08-24
Ensure control plane can run with only authz for the federation namespace

marun:
This PR limits the permissions of the federation service account to a role binding when run in namespaced mode.  For some reason the `cluster-admin` roles is required - `admin` or `edit` do not grant sufficient access.

Supporting limited authz required rewriting the controller for `FederatedTypeConfig`.  The `kubebuilder` controller framework didn't appear to have any provision for limiting informer scope, so this PR rewrites the controller to be consistent with all the other federation controllers.

Targets #197

### DIFF
--- a/pkg/controller/federatedtypeconfig/controller.go
+++ b/pkg/controller/federatedtypeconfig/controller.go
@@ -18,38 +18,36 @@ package federatedtypeconfig
 
 import (
 	"fmt"
-	"log"
 	"sync"
+	"time"
 
-	"github.com/kubernetes-sigs/kubebuilder/pkg/controller"
-	"github.com/kubernetes-sigs/kubebuilder/pkg/controller/types"
-	"k8s.io/apimachinery/pkg/api/errors"
+	"github.com/golang/glog"
+
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/record"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/flowcontrol"
+	"k8s.io/client-go/util/workqueue"
 
 	corev1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+	fedclientset "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned"
 	corev1alpha1client "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned/typed/core/v1alpha1"
-	corev1alpha1informer "github.com/kubernetes-sigs/federation-v2/pkg/client/informers/externalversions/core/v1alpha1"
-	corev1alpha1lister "github.com/kubernetes-sigs/federation-v2/pkg/client/listers/core/v1alpha1"
 	synccontroller "github.com/kubernetes-sigs/federation-v2/pkg/controller/sync"
-
-	"github.com/kubernetes-sigs/federation-v2/pkg/inject/args"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 )
 
-const (
-	finalizer string = "core.federation.k8s.io/federated-type-config"
-)
+const finalizer string = "core.federation.k8s.io/federated-type-config"
 
-type FederatedTypeConfigController struct {
-	lister corev1alpha1lister.FederatedTypeConfigLister
+// Controller manages the FederatedTypeConfig objects in federation.
+type Controller struct {
 	client corev1alpha1client.CoreV1alpha1Interface
-	// recorder is an event recorder for recording Event resources to the
-	// Kubernetes API.
-	recorder record.EventRecorder
 
 	// Reference to config used to start new controllers
 	config *rest.Config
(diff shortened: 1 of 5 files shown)


pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -66,83 +64,214 @@ type FederatedTypeConfigController struct {
...
+}
+
+type reconciliationStatus int
 
-	gc := &controller.GenericController{
-		Name:             name,
-		Reconcile:        c.Reconcile,
-		InformerRegistry: arguments.ControllerManager,
+const (
+	statusAllOK reconciliationStatus = iota
```
font: (no action required) we should externalize these to be re-used in each controller.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -173,22 +301,22 @@ func (c *FederatedTypeConfigController) startController(tc *corev1a1.FederatedTy
 		close(stopChan)
 		return fmt.Errorf("Error starting sync controller for %q: %v", kind, err)
 	}
-	log.Printf("Started sync controller for %s \n", kind)
+	glog.Infof("Started sync controller for %q \n", kind)
```
font: (no action required) we should remove these benign newlines.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -66,83 +64,214 @@ type FederatedTypeConfigController struct {
...
+	// Informer for the FederatedTypeConfig objects
+	controller cache.Controller
+
+	// Work queue allowing parallel processing of resources
+	workQueue workqueue.Interface
+
+	// Backoff manager
+	backoff *flowcontrol.Backoff
+
+	reviewDelay time.Duration
```
font: `reviewDelay` and `smallDelay` are not being used.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -66,83 +64,214 @@ type FederatedTypeConfigController struct {
...
+		util.NewTriggerOnAllChanges(func(obj pkgruntime.Object) {
+			c.deliverObj(obj, 0, false)
+		}),
+	)
+
+	return c, nil
+}
+
+// minimizeLatency reduces delays and timeouts to make the controller more responsive (useful for testing).
+func (c *Controller) minimizeLatency() {
```
font: Without the use of `reviewDelay` and `smallDelay`, this method is not needed.

scripts/deploy-federation.sh
```diff
@@ -117,11 +117,13 @@ else
...
 
 # Create a permissive clusterrolebinding to allow federation controllers to run.
-# TODO(marun) Make this more restrictive.
-# TODO(marun) Use a rolebinding when namespaced once the
-# federaatedtypeconfig controller can limit informer scope to the
-# namespace.
-kubectl create clusterrolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
+if [[ "${NAMESPACED}" ]]; then
+  # TODO(marun) Investigate why cluster-admin is required to view cluster registry clusters in a namespace
+  kubectl -n "${NS}" create rolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
```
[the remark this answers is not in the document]
marun: Filed #265

scripts/deploy-federation.sh
```diff
@@ -117,11 +117,13 @@ else
 fi
 
 # Create a permissive clusterrolebinding to allow federation controllers to run.
-# TODO(marun) Make this more restrictive.
-# TODO(marun) Use a rolebinding when namespaced once the
-# federaatedtypeconfig controller can limit informer scope to the
-# namespace.
-kubectl create clusterrolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
+if [[ "${NAMESPACED}" ]]; then
+  # TODO(marun) Investigate why cluster-admin is required to view cluster registry clusters in a namespace
```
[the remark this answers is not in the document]
marun: Should cluster registry clusters be labeled as `system:aggregate-to-admin` or `system:aggregate-to-edit`?  It looks like label selection is [not supported](https://github.com/kubernetes/kubernetes/issues/44703) in rbac directly, so I think we'll want to include federation resources in the `edit` or `admin` role role to avoid having to maintain explicit rbac rules as the list of federation types grows over time.

pkg/controller/federatedtypeconfig/controller.go
marun: That is [already done](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/controller/util/constants.go#L33), but this controller only needs a subset of the enum so I defined just what was needed here.

pkg/controller/federatedtypeconfig/controller.go
marun: Done.

pkg/controller/federatedtypeconfig/controller.go
marun: Done

scripts/deploy-federation.sh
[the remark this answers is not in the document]
pmorie: I believe the issue here is that we are only installing the cluster registry and not making any changes to any existing rbac entity to grant users the ability to use it. `cluster-admin` works as a rolebinding because it grants */*/* permissions.

For now, using cluster-admin in this way is fine. As a follow-up, we should have a more narrowly-scoped role for the federation controller and a role binding to the service account the controller runs under. This is consistent with norms around least-privilege elsewhere in the ecosystem. For example, when you deploy service-catalog, there are service accounts with distinct least-privilege rbac permissions for each process (in that case, apiserver and controller).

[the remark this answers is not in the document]
marun: I don't think namespaced cluster admin is a problem for now.  I'm not clear we'll want to have to maintain rbac roles with explicit type references if the control plane should have access to everything in the namespace, that seems like an operational hassle.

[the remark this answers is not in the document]
font: No action was required on this original comment - only spurring discussion. I think `cluster-admin` is good for now and we can do a follow-up.

PR #235, opened 2018-08-28
Reduce federated informer error messages

pmorie:
Reduces confusing error messages from FederatedInformer


onyiny-ang:
Thanks for clarifying this message! LGTM

PR #237, opened 2018-08-28
Add godoc for FederatedCluster controller

pmorie:
It was using "log"


pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -238,7 +239,7 @@ func (np *NamespacePredicate) HandleCreate(obj interface{}) bool {
 func (np *NamespacePredicate) objInNamespace(obj interface{}) bool {
 	metaObj, ok := obj.(metav1.Object)
 	if !ok {
-		log.Printf("Cannot handle %T because obj is not an Object: %v\n", obj, obj)
+		glog.Infof("Cannot handle %T because obj is not an Object: %v\n", obj, obj)
```
danehans: Should `\n` be removed here?

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -104,7 +105,7 @@ func ProvideController(arguments args.InjectArgs, fedNamespace, clusterNamespace
 }
 
 func (c *FederatedTypeConfigController) Reconcile(k types.ReconcileKey) error {
-	log.Printf("Running reconcile FederatedTypeConfig for %s\n", k.Name)
+	glog.Infof("Reconciling FederatedTypeConfig %v/%v", k.Namespace, k.Name)
```
font: (no action required) We can just print out `k` here with one format specifier as it has a `String` method that prints it as `namespace/name` if it contains a namespace. See the example on line 113 below.

gyliu513: @pmorie I think we also need to unify the log level in all controllers, you may see that we are using different log levels of `glog.V(4).Infof` at https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/controller/schedulingpreference/controller.go#L268

irfanurrehman: @gyliu513 how about opening an issue for this to handle it across.

irfanurrehman:
@pmorie few nits here too. Let me know if you intend to update?

marun:
FYI, the use of log vs glog is due to being based a kubebuilder-generated controller.  Even post 1.0 code generated by kubebuilder continues to use log instead of glog.

That said, the controller in question is completely rewritten by #230 (and no longer uses log), so I suggest closing this PR.

pkg/controller/federatedtypeconfig/controller.go
gyliu513: Thanks @irfanurrehman , I'm +1 for this, and I also like to use `glog.Infof()` as @pmorie is doing here.

@marun do you think it is reasonable to create a PR to unify the log level for all controllers for `Reconcile`?

marun: I'm not sure of the value of synchronizing log level everywhere without clear criteria for doing so.  I think @irfanurrehman's suggestion of filing an issue to discuss is a good next step.

gyliu513: Done, I opened an issue at https://github.com/kubernetes-sigs/federation-v2/issues/251 for discussion.

PR #239, opened 2018-08-29
Update install-latest.yaml to work with latest image

### LINKED ISSUE #232: User Guide example fails
When trying to run through the federation-v2 tutorial using the suggested minikube CI build, the status of resources (from this point onward: https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/userguide.md#check-status-of-resources) does not appear to exist in any of the clusters. The only thing that does appear in the first `get` call is:
```
------------ cluster1 ------------

NAME                  TYPE                                  DATA      AGE
default-token-wfkzm   kubernetes.io/service-account-token   3         26m
```

font:
Reverted the update that made `FederatedTypeConfig` namespace scoped in order to continue working with the latest stable released version of fedv2 (v0.0.1). This allows the userguide to continue working as expected.

Fixes #232.

/cc @pmorie

### DIFF
--- a/hack/install-latest.yaml
+++ b/hack/install-latest.yaml
@@ -930,7 +930,7 @@ spec:
   names:
     kind: FederatedTypeConfig
     plural: federatedtypeconfigs
-  scope: Namespaced
+  scope: Cluster
   validation:
     openAPIV3Schema:
       properties:


pmorie:
/lgtm

Thanks a lot for hunting this down.

PR #240, opened 2018-08-29
Update docs on using master canary or latest stable version image

font:
/cc @pmorie

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -256,7 +256,7 @@ Follow the [cleanup instructions in the user guide](userguide.md#cleanup).
 
 ## Test Your Changes
 
-In order to test your changes on your own kubernetes cluster, you'll need
+In order to test your changes on your kubernetes cluster, you'll need
 to build an image and a deployment config.
 
 ### Automated Deployment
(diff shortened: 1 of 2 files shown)


font:
@pmorie The dev workflow currently instructs you to build your own image and config for testing. I think we should keep it that way instead of using canary that way users can test their own changes. I think the canary can be used as an extension of what's in the user's guide, if users are eager to try out some new feature before it's ready to be released. WDYT?

irfanurrehman:
> @pmorie The dev workflow currently instructs you to build your own image and config for testing. I think we should keep it that way instead of using canary that way users can test their own changes. I think the canary can be used as an extension of what's in the user's guide, if users are eager to try out some new feature before it's ready to be released. WDYT?

how about only adding an explicit note in the dev workflow (and probably user guide too), for now, that its possible to use `scripts/deploy-federation.sh <image name> <other args>`, where other args are same as in `scripts/deploy-federation-latest.sh`. The image to use could either be `quay.io/kubernetes-multicluster/federation-v2:latest`, `quay.io/kubernetes-multicluster/federation-v2:canary` or a user specific image location. We can as well ignore this and deprecate these scripts in favour of a chart if its landing soon (There's already a pr for it).

I also think that the `latest` tag on our image is a little confusing. I think `latest` should always be put on the latest built/pushed (master) image not on the latest released image. @onyiny-ang.

pmorie:
+1 to what you said @irfanurrehman

font:
@pmorie @irfanurrehman PTAL.

docs/development.md
```diff
@@ -307,23 +307,38 @@ Run the following command to build the deployment config `hack/install.yaml`
...
-    --controller-image "<containerregistry>/<username>/federation-v2:test" \
-    --name federation
+INSTALL_YAML="hack/install.yaml"
+IMAGE_NAME="<containerregistry>/<username>/federation-v2:test"
+INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-install-yaml.sh
 ```
 
-Once the installation YAML config `hack/install.yaml` is created, we need to
-update it to modify a few fields.
+Once the installation YAML config `hack/install.yaml` is created, you are able
```
irfanurrehman: Nit: My english is not that good, so I might be wrong also, but `you would be able to apply....` in place of `you are able to apply.. ` sounds more right somehow.

docs/development.md
```diff
@@ -307,23 +307,38 @@ Run the following command to build the deployment config `hack/install.yaml`
...
 ```bash
-sed -i 's/memory: 20Mi/memory: 64Mi/' hack/install.yaml
-sed -i 's/memory: 30Mi/memory: 128Mi/' hack/install.yaml
+INSTALL_YAML="hack/install.yaml"
+IMAGE_NAME="quay.io/kubernetes-multicluster/federation-v2:canary"
+INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-install-yaml.sh
 ```
 
-Once the installation YAML config `hack/install.yaml` is updated, you are able
+Once the installation YAML config `hack/install.yaml` is created, you are able
```
irfanurrehman: As previous comment.

irfanurrehman:
Seems quite fine to me, except one English nit. Thanks for the changes @font.

docs/development.md
font: @irfanurrehman Either works, but using *would* implies future tense, whereas it's written up in the present tense as if the user is currently performing the steps.

docs/development.md
font: See previous comment.

PR #244, opened 2018-09-01
Fixes for namespace-scoped federatednamespaceplacement

font:
`FederatedNamespacePlacement` becoming namespace-scoped has resulted in the user guide example not working. Specifically, the resources all get created successfully, but when you try to remove `cluster2` from the `FederatedNamespacePlacement` object, the resources in the `test-namespace` are not removed because the `test-namespace` is never removed.

This was caused by mismatching keys to retrieve `Namespace` and `FederatedNamespacePlacement` objects from the cache. This is a temporary fix as `Namespaces` are the only resources we currently support whose template is cluster-scoped while the placement is namespace-scoped. A proper long-term solution should be investigated, especially when we begin to support more cluster-scoped resources.

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -321,8 +321,8 @@ successfully verified a working federation-v2 deployment.
 To cleanup the example simply delete the namespace and its placement:
 
 ```bash
+kubectl -n test-namespace delete federatednamespaceplacement test-namespace
 kubectl delete ns test-namespace
-kubectl delete federatednamespaceplacement test-namespace
 ```
 
 ## Cleanup
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -420,6 +420,7 @@ func (s *FederationSyncController) reconcile(qualifiedName util.QualifiedName) u
 
 	templateKind := s.typeConfig.GetTemplate().Kind
 	key := qualifiedName.String()
+	placementKey := key
 	namespace := qualifiedName.Namespace
 
 	targetKind := s.typeConfig.GetTarget().Kind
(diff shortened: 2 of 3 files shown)


pkg/controller/sync/controller.go
```diff
@@ -431,6 +432,24 @@ func (s *FederationSyncController) reconcile(qualifiedName util.QualifiedName) u
...
+		// template or placement objects from the cache by using the proper
+		// cluster-scoped or namespace-scoped key.
+		if qualifiedName.Namespace == "" {
+			qualifiedName.Namespace = qualifiedName.Name
+			placementKey = qualifiedName.String()
+			glog.V(4).Infof("Received Namespace update for %v. Using placement key %v", key, placementKey)
+		} else if qualifiedName.Namespace != "" {
+			qualifiedName.Namespace = ""
+			key = qualifiedName.String()
+			glog.V(4).Infof("Received FederatedNamespacePlacement update for %v. Using template key %v", placementKey, key)
```
gyliu513: Can you please show more detail why did the reconcile namespace can handle `FederatedNamespacePlacement` object?

docs/userguide.md
```diff
@@ -321,8 +321,8 @@ successfully verified a working federation-v2 deployment.
 To cleanup the example simply delete the namespace and its placement:
 
 ```bash
+kubectl -n test-namespace delete federatednamespaceplacement test-namespace
```
gyliu513: do we still need this step? I think `kubectl delete ns test-namespace` will delete all resources under this namespace including the namespace placement.

marun:
Hmm, good catch.  I don't understand the failure mode yet, but this seems like something that would ideally be validated in the tests.  Can you think of a way to accomplish this?

docs/userguide.md
font: This is intended to work with v0.0.1 as well as the latest master. In v0.0.1, `federatednamespaceplacement` is cluster-scoped so this deletion still needs to happen and the `test-namespace` option will be ignored.

pkg/controller/sync/controller.go
font: I'm not sure I understand your question. Can you restate?

marun:
Ah, it's the namespace removal that is the problem. I didn't read the comment very carefully.  Yeah, sounds like multiple clusters is the only way to uncover this.

pkg/controller/sync/controller.go
gyliu513: I'm clear for the logic now, `FederatedNamespacePlacement` is namespace scoped now, and the `qualifiedName` will include `Namespace`.

docs/userguide.md
gyliu513: > This is intended to work with v0.0.1 as well as the latest master. 
Yes, but I think the document here is intend for 0.0.2 and future versions, why still need to consider 0.0.1?

font: The document is still targeting 0.0.1 because the `install-latest.yaml` is still referencing the latest tag. So until we update the config to use the latest master, this step should remain.

pkg/controller/sync/controller.go
```diff
@@ -429,8 +430,27 @@ func (s *FederationSyncController) reconcile(qualifiedName util.QualifiedName) u
...
+		// Namespace template and placement keys depending on if we are
+		// reconciling due to a Namespace or FederatedNamespacePlacement
+		// update. This insures we will successfully retrieve the Namespace
+		// template or placement objects from the cache by using the proper
+		// cluster-scoped or namespace-scoped key.
+		if qualifiedName.Namespace == "" {
+			qualifiedName.Namespace = qualifiedName.Name
+			placementKey = qualifiedName.String()
+			glog.V(4).Infof("Received Namespace update for %v. Using placement key %v", key, placementKey)
+		} else if qualifiedName.Namespace != "" {
```
marun: (No action required) Consider using `else` here.

PR #245, opened 2018-09-03
Helm chart for federation v2.

gyliu513:
/cc @pmorie @shashidharatd @marun @font @irfanurrehman

### DIFF
--- a/charts/federation-v2/Chart.yaml
+++ b/charts/federation-v2/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+appVersion: 0.0.2
+description: Kubernetes Federation V2 helm chart
+name: federation-v2
+version: 0.0.2
--- a/charts/federation-v2/README.md
+++ b/charts/federation-v2/README.md
@@ -1,0 +1,102 @@
+# Kubernetes Federation V2
+
+Kubernetes Federation V2 is a Kubernetes Incubator project. It builds on the sync controller
+(a.k.a. push reconciler) from [Federation v1](https://github.com/kubernetes/federation/)
+to iterate on the API concepts laid down in the [brainstorming
+doc](https://docs.google.com/document/d/159cQGlfgXo6O4WxXyWzjZiPoIuiHVl933B43xhmqPEE/edit#)
+and further refined in the [architecture
+doc](https://docs.google.com/document/d/1ihWETo-zE8U_QNuzw5ECxOWX0Df_2BVfO3lC4OesKRQ/edit#). 
+Access to both documents is available to members of the
+[kubernetes-sig-multicluster google
+group](https://groups.google.com/forum/#!forum/kubernetes-sig-multicluster).
+
+## Prerequisites
+
+- Kubernetes 1.11+
+- Helm 2.10+
+
+## Installing the Chart
+
+First you'll need to create the reserved namespace for registering clusters with the
+cluster registry:
+
+```bash
+kubectl create ns kube-multicluster-public
+```
+
+Install the chart with the release name `federation-v2`:
+
+```bash
+$ helm install charts/federation-v2 --name federation-v2 --namespace federation-system
+```
+
+If you already have clusterregistry installed, skip installing it during federation-v2 installation
+as follows:
+
+```bash
+$ helm install charts/federation-v2 --name federation-v2 --namespace federation-system --set clusterregistry.enabled=false
+```
+
+## Uninstalling the Chart
+
+Due to this helm [issue](https://github.com/helm/helm/issues/4440), the CRDs cannot be deleted
+when delete helm release, so before delete the helm release, we need first delete all
+of the CR and CRDs for federation v2 release.
+
+Delete all federation v2 `FederatedTypeConfig`:
+
+```bash
+kubectl -n federation-system delete FederatedTypeConfig --all
+```
+
+Delete all federation v2 CRDs:
+
+```bash
+kubectl delete crd $(kubectl get crd | grep -E 'federation.k8s.io' | awk '{print $1}')
+```
+
+If you want to delete `clusters.clusterregistry.k8s.io` as well, do it as follows:
+
+```bash
+kubectl delete crd clusters.clusterregistry.k8s.io
+```
+
+Then you can uninstall/delete the `federation-v2` release:
+
+```bash
+$ helm delete --purge federation-v2
+```
+
+The command above removes all the Kubernetes components associated with the chart
+and deletes the release.
+
+Delete the reserved namespace for registering clusters:
+
+```bash
+kubectl delete ns kube-multicluster-public
+```
+
+## Configuration
+
+The following tables lists the configurable parameters of the Federation V2
+chart and their default values.
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| controllermanager.replicaCount | Number of replica for federation v2 controller manager | 1 |
+| controllermanager.repository | Repo of the federation v2 image | quay.io/kubernetes-multicluster |
+| controllermanager.image | Name of the federation v2 image | federation-v2 |
+| controllermanager.tag | Tag of the federation v2 image | latest |
+| controllermanager.pullPolicy | Image pull policy | IfNotPresent |
+| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2 | true |
+| federatedtypes.enabled | Specifies whether to install the `FederatedTypeConfig` in federation v2 | true |
+
+Specify each parameter using the `--set key=value[,key=value]` argument to
+`helm install`.
+
+Alternatively, a YAML file that specifies the values for the parameters can be
+provided while installing the chart. For example:
+
+```bash
+$ helm install . --name federation-v2 --namespace federation-system --values values.yaml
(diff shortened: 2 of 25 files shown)


shashidharatd:
Better to follow prescribed RBAC best practices for chart https://github.com/helm/helm/blob/master/docs/chart_best_practices/rbac.md

charts/federation-v2/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: 0.0.2
```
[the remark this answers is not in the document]
gyliu513: As 0.0.1 already released, shall we start from 0.0.2?

charts/federation-v2/templates/statefulset.yaml
```diff
@@ -0,0 +1,42 @@
...
+        kubebuilder.k8s.io: 1.0.0
+    spec:
+      containers:
+      - args:
+        - --install-crds=false
+        command:
+        - /root/controller-manager
+        image: "{{ .Values.controllerManager.repository }}/{{ .Values.controllerManager.image }}:{{ .Values.controllerManager.tag }}"
+        name: controller-manager
+        resources:
```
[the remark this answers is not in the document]
gyliu513: Will fix this.

charts/federation-v2/values.yaml
```diff
@@ -0,0 +1,12 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+## Configuration values for federation v2 controllerManager statefulset.
+##
+controllerManager:
+  replicaCount: 1
+  repository: quay.io/kubernetes-multicluster
+  image: federation-v2
+  tag: latest
```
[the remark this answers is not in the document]
gyliu513: How about use `canary` here, and developer can update the `tag` based on their own image?

charts/federation-v2/templates/crds.yaml
```diff
@@ -0,0 +1,1293 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    "helm.sh/hook": crd-install
+  labels:
+    api: ""
+    kubebuilder.k8s.io: 0.1.10
+  name: clusters.clusterregistry.k8s.io
```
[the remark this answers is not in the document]
gyliu513: Done, I was creating a separate chart for this.

gyliu513:
> Also i missed creating federatedtypeconfigs in this chart. Are they included?

Done, I was adding a separated chart to install federatedtypeconfigs.

> Better to follow prescribed RBAC best practices for chart https://github.com/helm/helm/blob/master/docs/chart_best_practices/rbac.md

How about fix this in future PR, as federation v2 statefulset do not have serviceAccount yet and I was just following current YAML template.

shashidharatd:
> Done, I was adding a separated chart to install federatedtypeconfigs.

Separate chart for federatedtypeconfigs does not seem to be right. `subcharts` cannot depend on parent chart and the reverse is possible. federatedtypeconfigs depends on the `FederatedTypeConfig` CRD to exist before applying them. @gyliu513, have you tested this?

charts/federation-v2/templates/statefulset.yaml
```diff
@@ -0,0 +1,39 @@
+---
```
shashidharatd: (Query), Not sure why we chose StatefulSet for controller-manager, maybe because that's the default. IMHO, We don't have any requirement for it to be a StatefulSet, instead Deployment would have worked better for HA. Is the reason behind StatefulSet because controller-manager installs CRDs? @marun wdyt?

charts/federation-v2/templates/crds.yaml
```diff
@@ -0,0 +1,1200 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    "helm.sh/hook": crd-install
```
shashidharatd: (Suggestion) As discussed in wg meeting, it would be better additionally to add a script to generate CRD definitions from the code and add customised stuff like adding annotations.

gyliu513: Yes, I will follow up this task.

gyliu513:
With helm chart, if customer want to install with YAML template, they can run the command as follows to get the YAML template:

```yaml
helm template federation-v2/ > federation-v2.yaml
```

charts/federation-v2/templates/statefulset.yaml
gyliu513: @marun @pmorie comments for this? I think we can change the federation v2 controller to deployment instead of sts, as we do not have state need to persist and also do not need to do discovery via headless service.

shashidharatd:
@gyliu513, i fetched this pr locally and tried to install the chart. Below is the error i am getting.
```
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2# helm install --namespace federation-system --name myfed charts/federation-v2
Error: apiVersion "core.federation.k8s.io/v1alpha1" in federation-v2/charts/federatedtypes/templates/namespace.yaml is not available
```
It would be good, if you can come up with a script similar to `scripts/deploy-federation.sh` with all the things it is doing except that fcp is installed via helm chart.

charts/federation-v2/templates/statefulset.yaml
marun: The rational for StatefulSet (iaccording to feedback I received on `#kubebuilder`) was that deployments don't guarantee only a single instance will be running at a time, and running more than one controller (in the absence of leader election) could cause problems.

gyliu513: > deployments don't guarantee only a single instance will be running at a time

Not clear for what does this mean, if set `replica` as 1, can not we guarantee only a single instance?

marun: My understanding is that while setting `replicas: 1` will ensure that eventually the number of replicas will converge on 1, when configuration is changed such that a redeploy occurs, there may be more than 1 instance running.  StatefulSets is more careful about ensuring only a single instance is running at a time.  I encourage you to confirm this perspective with the kubebuilder devs.

charts/federation-v2/values.yaml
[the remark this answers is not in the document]
shashidharatd: yes, please use `canary` as default instead of `latest`

charts/federation-v2/README.md
```diff
@@ -0,0 +1,101 @@
...
+kubectl create ns kube-multicluster-public
+```
+
+Install the chart with the release name `federation-v2`:
+
+```bash
+$ helm install . --name federation-v2 --namespace federation-system
+```
+
+If you already have clusterregistry installed, you can just ignore the `clusterregistry` chart:
```
shashidharatd: Replace "you can just ignore the `clusterregistry` chart" with
"skip installing it during `federation-v2` installation as follows:"

charts/federation-v2/README.md
```diff
@@ -0,0 +1,101 @@
...
+## Uninstalling the Chart
+
+Due to this helm [issue](https://github.com/helm/helm/issues/4440), the CRDs cannot be deleted
+when delete helm release, so before delete the helm release, we need first delete all
+of the CR and CRDs for federation v2 release.
+
+Delete all federation v2 `FederatedTypeConfig`:
+
+```bash
+kubectl delete FederatedTypeConfig --all
```
shashidharatd: "kubectl delete FederatedTypeConfig --all" should be "kubectl -n federation-system delete FederatedTypeConfig --all"

charts/federation-v2/README.md
```diff
@@ -0,0 +1,101 @@
...
+cluster registry:
+
+```bash
+kubectl create ns kube-multicluster-public
+```
+
+Install the chart with the release name `federation-v2`:
+
+```bash
+$ helm install . --name federation-v2 --namespace federation-system
```
shashidharatd: this command fails when executed from root of the repo as follows:
```
root@shashi-laptop:~/gopath/src/github.com/kubernetes-sigs/federation-v2# helm install . --name federation-v2 --namespace federation-system
Error: no Chart.yaml exists in directory "/root/gopath/src/github.com/kubernetes-sigs/federation-v2
```
So the command should be
```
helm install charts/federation-v2 --name federation-v2 --namespace federation-system
```

gyliu513: Done. Actually, if switch to `charts/federation-v2`, then we will be able to install it with `.`, but using whole dir is more easy to understand.

charts/federation-v2/README.md
gyliu513: Done

charts/federation-v2/README.md
gyliu513: Done

charts/federation-v2/values.yaml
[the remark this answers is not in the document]
gyliu513: Done

shashidharatd:
Thanks @gyliu513, please squash the commits.

gyliu513:
@shashidharatd done. ;-)

Issue #246, opened 2018-09-04
Federated Status for Federated types


shashidharatd:
Status for federated types can be of two forms
- status containing elaborate details using which we can know resource status in a particular cluster (this could be essentially a list of resource status in all cluster).
- aggregated status which could be compatible with the corresponding resource status.

The first type of status could be useful to the controllers as well and they need not look into all member clusters to get the status.

p.s: This is as discussed in federation wg meeting. please update with any opinions or thoughts on this topic.

Issue #60, opened 2018-04-24
API driven federation of arbitrary types


pmorie:
Related to #34; assuming that we have a CRD-based way to extend federation by creating some new CRD resources for:

- template
- override
- placement

...then there should be an API resource to have these CRDs created automatically. Example:

```yaml
kind: FederatedAPI
metadata:
  name: statefulset
spec:
  basis:
    group: apps
    version: v1
    kind: StatefulSet
  target:
    group: federation.k8s.io
    version: v1alpha1
```

A controller would list/watch this API resource and ensure that the correct CRDs were created for the named `basis` in the API group/version described by `target. In this example those would be:

- `FederatedStatefulSet` (possibly creating the OpenAPI schema and validations based on the OpenAPI for the basis type)
- `FederatedStatefulSetOverride`
- `FederatedStatefulSetPlacement`

In order to have automatic federation of all the registered types in an API server, there could also be a controller that examines the discovery endpoint for a cluster and ensures a `FederatedAPI` exists for the right API services. There are a number of ways a controller might be configured, like:

- configmap holding explicit configuration
- annotations on `APIService` and `CustomResourceDefinition` resources

idea formulated with @marun 

**Work Items**

- [x] Factor propagated version handling out of the sync controller (#273)
- [x] Factor type-specific version manager code into adapter (#276)
- [x] Add api generation to kubefed (??)

danehans:
@pmore do you know what remains to close this issue?

pmorie:
@danehans the remaining bit is actually generating the federation APIs themselves. For example, imagine if creating a FederatedTypeConfig actually caused the federation APIs to be generated.

pmorie:
Maybe we should split that piece into a separate issue.

danehans:
@pmorie OK. Is anyone working on this remaining bit?

danehans:
@pmorie I created [this diagram](https://docs.google.com/drawings/d/1Glq-pI79ZYbb7h4GBqWwTUFavMilwg_ch_LWkThOIrM/edit?usp=sharing) to illustrate what you described [above](https://github.com/kubernetes-sigs/federation-v2/issues/60#issue-317369700). Let me know if I captured it correctly.

marun:
@danehans I'm working on this.

marun:
[...]

PR #247, opened 2018-09-04
Federated Status Implementation for federated types

shashidharatd:
This pr implements a generic federated status for federated types by extending the sync controller. The federated status is a list of resource status in member cluster, as outlined in the first alternative in issue https://github.com/kubernetes-sigs/federation-v2/issues/246

Later the controllers depending upon the status in member clusters can make use of this without monitoring the resource in individual clusters.

/cc @onyiny-ang @pmorie @marun @kubernetes-sigs/federation-wg


marun:
@shashidharatd Am I missing something, or is the entirety of status being collected?  I wasn't aware that chatty fields like `conditions` were useful?

pkg/apis/core/v1alpha1/federatedingress_types.go
```diff
@@ -29,6 +29,13 @@ type FederatedIngressSpec struct {
 
 // FederatedIngressStatus defines the observed state of FederatedIngress
 type FederatedIngressStatus struct {
+	ClusterStatuses []FederatedIngressClusterStatus `json:"clusterStatuses,omitempty"`
+}
+
+// FederatedDeploymentClusterStatus is the observed status for a named cluster
```
gyliu513: s/FederatedDeploymentClusterStatus/FederatedIngressClusterStatus

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
```diff
@@ -303,7 +303,6 @@ var (
 			Scope: "Namespaced",
 			Validation: &v1beta1.CustomResourceValidation{
 				OpenAPIV3Schema: &v1beta1.JSONSchemaProps{
-					Type: "object",
```
gyliu513: I think this change will not needed if the PR https://github.com/kubernetes-sigs/federation-v2/pull/250 got merged before this PR.

pkg/apis/core/typeconfig/interface.go
```diff
@@ -31,4 +31,5 @@ type Interface interface {
 	GetPlacement() metav1.APIResource
 	GetOverride() *metav1.APIResource
 	GetOverridePath() []string
+	GetCollectStatus() bool
```
gyliu513: Can you move this commit ahead of https://github.com/kubernetes-sigs/federation-v2/pull/247/commits/3f94daf2256e73b4bc1014d9441f68dd93767f75

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -89,6 +89,35 @@ type FederatedTypeConfig struct {
 	Status FederatedTypeConfigStatus `json:"status,omitempty"`
 }
 
+// FederatedResource is a generic resource representing any federated type
```
gyliu513: What is the target of this resource type? Does the intend is want to handle arbitrary resource types?

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
shashidharatd: Agree. if it lands before. i will remove it.

pkg/apis/core/typeconfig/interface.go
shashidharatd: its already in the same order as you asked (if you look in git). github may have messed up the order of commits).

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
shashidharatd: yes, as the comment suggest FederatedResource type generically represent any arbitrary federated type.

shashidharatd: The sync controller is already capable of handling propogation of arbitrary types. here is the e2e test case which does that, https://github.com/kubernetes-sigs/federation-v2/blob/master/test/e2e/crud.go#L89 and there is no need to define an API for that. #60 is some additional feature to automatically create template, override & placement CRDs for a federated type.

The type `FederatedResource` which i defined in this PR represents a generic federated type which is used by the controller to Marshall/Unmarshall Json. We still need specific federated types defined like `FederatedDeployment`, `FederatedIngress` etc.. because they help validating user inputs and offer type safety.

gyliu513: Thanks @shashidharatd , clear now. But still have more question, I think we do not have `placement` policy for generic resources, what is the plan for this?

test/common/crudtester.go
```diff
@@ -364,6 +365,29 @@ func (c *FederatedTypeCrudTester) CheckPropagation(template, placement, override
...
+				resource := fedv1a1.FederatedResource{}
+				err = json.Unmarshal(content, &resource)
+				if err != nil {
+					c.tl.Fatalf("Failed to json unmarshal federated resource object %s %q: %v", targetKind, qualifiedName, err)
+				}
+
+				statusFound := false
+				for _, resourceClusterStatus := range resource.Status.ClusterStatuses {
+					if resourceClusterStatus.ClusterName == clusterName {
+						statusFound = true
```
font: break

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -89,6 +89,35 @@ type FederatedTypeConfig struct {
...
+// +k8s:deepcopy-gen=false
+type FederatedResource struct {
```
font: I'm not sure I understand the need for this generic `FederatedResource` API type. I don't see it being persisted; it looks like it's just used to capture the status into this generic object and then just copied into the specified `Federated<Type>` resource for persistence. Why not work directly with the status for each individual `Federated<Type>` resource that you've added in this PR? Am I missing something?

test/common/crudtester.go
shashidharatd: updated.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
shashidharatd: Agree with your argument. Updated accordingly. PTAL. Thanks!

shashidharatd:
> I'm reviewing this.

cool, waiting for your inputs. Thanks !

shashidharatd:
Closing this in favour of https://github.com/kubernetes-sigs/federation-v2/pull/291

PR #248, opened 2018-09-04
Makes Namespaces Conditional in Tests

danehans:
Sets a `namespace` in a template only if it exists in the type config. This is required for resources that are not ns-scoped.

### DIFF
--- a/test/common/testobjects.go
+++ b/test/common/testobjects.go
@@ -114,7 +114,9 @@ func GetPlacementTestObject(typeConfig typeconfig.Interface, namespace string, c
 		return nil, err
 	}
 	placementAPIResource := typeConfig.GetPlacement()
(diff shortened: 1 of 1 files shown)


test/common/testobjects.go
```diff
@@ -40,7 +40,9 @@ func NewTestObjects(typeConfig typeconfig.Interface, namespace string, clusterNa
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	template.SetNamespace(namespace)
+	if typeConfig.GetNamespaced() {
+		template.SetNamespace(namespace)
```
gyliu513: Based on Kubernetes, if the resource is not namespace scoped, then even if we create the resource with namespace, the namespace will have no effect. But +1 to @danehans , it is always good to add those check.

LGTM

Issue #214, opened 2018-08-20
Flake in FederatedDeployment e2e test


shashidharatd:
`FederatedDeployment` e2e test seems to be flaking. here is one such failure. https://travis-ci.org/kubernetes-sigs/federation-v2/builds/417463481?utm_source=github_status&utm_medium=notification

```
Federated types "FederatedDeployment" resources 
  should be created, read, updated and deleted successfully
  /home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/crud.go:73
STEP: Obtaining a list of federated clusters
STEP: Checking readiness of cluster "cluster1"
Aug 17 22:07:51.416: INFO: Cluster cluster1 is Ready
Aug 17 22:07:51.423: INFO: Creating new FederatedDeployment in namespace "e2e-tests-federated-types-slg5f"
Aug 17 22:07:51.429: INFO: Created new FederatedDeployment "e2e-tests-federated-types-slg5f/test-crud-6gxhn"
Aug 17 22:07:51.429: INFO: Creating new FederatedDeploymentPlacement in namespace "e2e-tests-federated-types-slg5f"
Aug 17 22:07:51.433: INFO: Created new FederatedDeploymentPlacement "e2e-tests-federated-types-slg5f/test-crud-6gxhn"
Aug 17 22:07:51.433: INFO: Creating new FederatedDeploymentOverride in namespace "e2e-tests-federated-types-slg5f"
Aug 17 22:07:51.438: INFO: Created new FederatedDeploymentOverride "e2e-tests-federated-types-slg5f/test-crud-6gxhn"
Aug 17 22:07:51.438: INFO: [Resource versions for e2e-tests-federated-types-slg5f/test-crud-6gxhn: template "713", placement "714", override "715"]
Aug 17 22:07:53.445: INFO: Waiting for Deployment "e2e-tests-federated-types-slg5f/test-crud-6gxhn" in cluster "cluster1"
Aug 17 22:07:53.447: INFO: Updating FederatedDeployment "e2e-tests-federated-types-slg5f/test-crud-6gxhn"
Aug 17 22:07:57.473: INFO: Waiting for Deployment "e2e-tests-federated-types-slg5f/test-crud-6gxhn" in cluster "cluster1"
Aug 17 22:07:57.475: INFO: Updating FederatedDeploymentPlacement "e2e-tests-federated-types-slg5f/test-crud-6gxhn"
[...]

PR #253, opened 2018-09-07
Improve e2e reliability

marun:
2 commits target #214

- Fix crudtester check of placement change when the host cluster is joined to itself.
- Fix case of placement field `clusterNames`

The remaining commits speed up test execution:

- Change replicaset and deployment e2e fixture to use replicas=1 with an override of replicas=2 (previously was 3 and override of 5).
- Change replicaset, deployment and job e2e fixture to use <2mb `busybox` image (previously was 100mb `nginx` image).
- Reduce poll and wait intervals

### DIFF
--- a/pkg/controller/util/placement.go
+++ b/pkg/controller/util/placement.go
@@ -20,11 +20,13 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+const clusterNamesField = "clusterNames"
+
 func GetClusterNames(placement *unstructured.Unstructured) ([]string, error) {
-	clusterNames, _, err := unstructured.NestedStringSlice(placement.Object, "spec", "clusternames")
+	clusterNames, _, err := unstructured.NestedStringSlice(placement.Object, "spec", clusterNamesField)
 	return clusterNames, err
 }
 
 func SetClusterNames(placement *unstructured.Unstructured, clusterNames []string) error {
-	return unstructured.SetNestedStringSlice(placement.Object, clusterNames, "spec", "clusternames")
+	return unstructured.SetNestedStringSlice(placement.Object, clusterNames, "spec", clusterNamesField)
 }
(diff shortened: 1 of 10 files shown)


test/common/crudtester.go
```diff
@@ -224,7 +224,9 @@ func (c *FederatedTypeCrudTester) CheckPlacementChange(template, placement, over
 		if err != nil {
 			c.tl.Fatalf("Error retrieving cluster names: %v", err)
 		}
-		clusterNames = c.deleteOneNonPrimaryCluster(clusterNames)
+		// Removing the resource from an arbitrary cluster (even the
+		// primary one) is safe for any resource but namespaces.
```
font: Seems like we need a check for namespaces here.

Also, I'm not understanding how this commit message relates to these code changes. Am I missing something?

test/common/crudtester.go
```diff
@@ -499,21 +501,6 @@ func (c *FederatedTypeCrudTester) getPrimaryClusterName() string {
 	return ""
 }
 
-func (c *FederatedTypeCrudTester) deleteOneNonPrimaryCluster(clusterNames []string) []string {
-	primaryClusterName := c.getPrimaryClusterName()
-
-	for i, name := range clusterNames {
-		if name == primaryClusterName {
```
gyliu513: I think we can keep this function but adding a new check for namespace?

```
if name == primaryClusterName && c.typeConfig.GetTarget().Kind == util.NamespaceKind {
}
```

pkg/controller/util/placement.go
```diff
@@ -20,11 +20,13 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+const clusterNamesField = "clusterNames"
+
 func GetClusterNames(placement *unstructured.Unstructured) ([]string, error) {
-	clusterNames, _, err := unstructured.NestedStringSlice(placement.Object, "spec", "clusternames")
+	clusterNames, _, err := unstructured.NestedStringSlice(placement.Object, "spec", clusterNamesField)
```
font: (no action required) we may want to consider returning the boolean indicating whether the value of the field was found. The caller can choose to do what it wants with it. In the case of the crudtester, it can emit a fatal error if it's expecting a value for this field to be there and it's not.

pmorie:
Bouncing to see if it triggers travis

PR #257, opened 2018-09-10
Added scripts to generate crds for federation v2.

gyliu513:
Those crds will be used by helm chart.

/cc @pmorie @shashidharatd @marun @kubernetes-sigs/federation-wg


scripts/generate-install-crds-yaml.sh
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+INSTALL_CRDS_YAML="${INSTALL_CRDS_YAML:-hack/install-crds-latest.yaml}"
+
+kubebuilder create config --crds --output "${INSTALL_CRDS_YAML}"
```
irfanurrehman: Testing the presence of `kubebuilder` and exiting with an appropriate message if not found might be useful.

gyliu513: Shell script will check the existence of `kuebbuilder`, also how about add the checking in the script which call this script? We can check the return value is 0 or not to decide if this script succeeds or not. Comments?

Same issue also exist for this script https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/generate-install-yaml.sh#L24 , seems we may need some checking for the return value of the script as well.

gyliu513: Thanks @pmorie , +1 for this.

@irfanurrehman @pmorie any more comments for other part of this script? Can we get this merged?

Issue #98, opened 2018-05-30
Higher level Job Preferences API and controller


irfanurrehman:
k8s jobs has already been federated using the simple federation resources (template, placement and override) for jobs. This however does not provide higher level job scheduling logic, which can provide an automated distribution of parallelism and completions for the given job across federated clusters.
This particular use case seems quite valuable and calls for a higher level job scheduling API (aka `JobSchedulingPreferences`) and an associated controller logic.
This also provides feature parity with federation V1.

sdminonne:
@irfanurrehman  I tried to have a look at federation V1 code but I couldn't find `JobSchedulingPreferences`. Instead, I found something concerning
`ReplicaAllocationPreferences` as an annotation with key `federation.kubernetes.io/job-preferences`.
May you give more details? Thanks

irfanurrehman:
In V1 there is a single set of preference annotations. The definition was originally created for deployments and replicasets. The same was later reused for jobs controller, as the intent was first to create an alpha prototype, and most of the fields of existing preferences sort of did fit the usage pattern. The preferences however needed a key name specific to the resource, as you mentioned `federation.kubernetes.io/job-preferences` for jobs. In the case of jobs, many of the preferences fields were meanigless and unused.

In V2 our plan is to have a job specific definition of the API named `JobSchedulingPreferences`. 
@shashidharatd signed up to do this and very recently started work on the same. 
What is your intent digging this information. Are you interested in taking this up? If so, we can sync to fill you in with info, we did already brainstorm about the same.

sdminonne:
@irfanurrehman thanks.
I was having a look at the list of issues and, having some interests in jobs, this one caught my eyes.
I could help on this but I'm not sure to estimate the effort. In mid-July, I'll be in PTO for a couple of weeks and if @shashidharatd already started I don't if and hand-over it's worth.
Eventually, if it's not too work for you guys we can imagine chatting a little before the end of this week.
Thanks

PR #259, opened 2018-09-11
Simplified high level Job Scheduling

irfanurrehman:
***Notes to reviewers:***
1. This scheduler is not useful for representing Federated Work Queue jobs as pod instances coordinate among each other (Cross cluster pod communication is currently not in federation purview).
2. The success of the jobs scheduled needs to be monitored independently using FederatedJobStatus.

Partially implements https://github.com/kubernetes-sigs/federation-v2/issues/98.

***Next step:***
Work on extended algorithm, where the controller will monitor the existing jobs in federated clusters and redistribute jobs if needed such that it does not change completions from existing jobs. This is beneficial for scenarios where a cluster joins or leaves federation (probably user/admin intent) after creation of a job, or user changes the scheduled job parallelism and/or completions in JobSchedulingPreferences.


irfanurrehman:
@sdminonne you can also take a stab at this. Thanks! I was not able to assign you as a reviewer.

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,144 @@
...
+	return (a[i].Weight > a[j].Weight) || (a[i].Weight == a[j].Weight && a[i].hash < a[j].hash)
+}
+
+func NewJobPlanner(preferences *fedschedulingv1a1.JobSchedulingPreference) *JobPlanner {
+	return &JobPlanner{
+		preferences: preferences,
+	}
+}
+
+// This assumes that there is atleast one available cluster and distributes
```
gyliu513: s/atleast/at least

pkg/controller/jobschedulingpreference/controller.go
```diff
@@ -0,0 +1,88 @@
+/*
```
gyliu513: All of the controller files can be deleted.

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,144 @@
...
+			} else {
+				// Clusters which do not have explicit weights given, while
+				// some others have, are treated equivalent to 0 weight.
+				plan[cluster] = int32(0)
+			}
+		}
+	}
+	sort.Sort(ClustersByWeight(weightedList))
+
+	//distributeInThisLoop := toDistribute
```
gyliu513: kill this

pkg/schedulingtypes/adapters/deployment.go
```diff
@@ -108,7 +109,12 @@ func (d *FederatedDeploymentAdapter) ReconcilePlacement(fedClient fedclientset.I
 	return nil
 }
 
-func (d *FederatedDeploymentAdapter) ReconcileOverride(fedClient fedclientset.Interface, qualifiedName util.QualifiedName, result map[string]int64) error {
+func (d *FederatedDeploymentAdapter) ReconcileOverride(fedClient fedclientset.Interface, qualifiedName util.QualifiedName, untypedResult interface{}) error {
+	result, ok := untypedResult.(map[string]int64)
+	if !ok {
+		return fmt.Errorf("Wrong result type passed in FederatedDeploymentAdapter for %s", qualifiedName)
```
gyliu513: Question here: For which case we will get wrong type here?

pkg/schedulingtypes/resources.go
```diff
@@ -43,6 +45,16 @@ var PodResource = &metav1.APIResource{
...
+	Name:       GetPluralName(Job),
+	Group:      batchv1.SchemeGroupVersion.Group,
+	Version:    batchv1.SchemeGroupVersion.Version,
+	Kind:       Job,
+	Namespaced: true,
+}
+
+// The resource name key is the corresponding federated resource name
+// for the given k8s resource definition.
 var ReplicaSechedulingResources = map[string]metav1.APIResource{
```
sohankunkerkar: There's a typo while assigning a variable. It should be **ReplicaSchedulingResources**.

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,147 @@
...
+	} else {
+		// No weights, default distribution
+		distributionStep := toDistribute / int32(numClusters)
+		if distributionStep > 0 {
+			for _, clusterName := range availableClusters {
+				plan[clusterName] = distributionStep
+			}
+		}
+
+		quotient := toDistribute - (distributionStep * int32(numClusters))
```
pmorie: Isn't `quotient` just the modulus of `toDistribute` and the number of clusters? It should be an integer between zero and `numClusters-1`

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,147 @@
...
+		if quotient == 1 {
```
pmorie: I think you could erase the `if` block here and just put the body of `else` as the body of `if quotient > 0`

test/integration/integration_test.go
```diff
@@ -66,5 +66,6 @@ func TestIntegration(t *testing.T) {
 		t.Run("TestCrud", TestCrud)
 		t.Run("TestReplicaSchedulingPreference", TestReplicaSchedulingPreference)
 		t.Run("TestMultiClusterServiceDNS", TestMultiClusterServiceDNS)
+		t.Run("TestJobSchedulingPreference", TestJobSchedulingPreference)
```
marun: Consider rewriting as an e2e test.

Integration tests can only run against managed fixture, but e2e tests can target both managed and unmanaged (i.e. deployed clusters) fixture.  Only tests that can run with unmanaged fixture can validate interaction with the kubernetes control plane (in this case the job controller).

test/integration/common.go
```diff
@@ -0,0 +1,36 @@
+/*
```
marun: The federation integration fixture already has a `ClusterNames` method that returns the same information as the method in this file, as does the e2e framework.

test/integration/jobschedulingpreference_test.go
```diff
@@ -0,0 +1,264 @@
...
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+
+	return err
+}
+
+func getFederatedJobTemplate(namespace string) *fedv1a1.FederatedJob {
```
marun: Is it necessary to define the template in code vs reusing the fixture used for crud testing?

test/integration/jobschedulingpreference_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	kubeClient := FedFixture.KubeApi.NewClient(tl, "jsp")
+	testNamespace := framework.CreateTestNamespace(tl, kubeClient, "jsp-")
+
+	testCases := map[string]struct {
+		jspSpec  fedschedulingv1a1.JobSchedulingPreferenceSpec
+		expected map[string]adapters.JobSchedulingResult
+	}{
+		"Distribution happens equally in clusters, with no explicit per cluster weights": {
```
marun: What is tested in these 3 cases that isn't already tested in the unit tests?  Consider limiting this test to a single golden-path check that the output of the scheduler is correctly translated into the expected placement and override resources.

test/integration/jobschedulingpreference_test.go
```diff
@@ -0,0 +1,264 @@
...
+	return t.Name, nil
+}
+
+func waitForMatchingJobPlacement(fedClient clientset.Interface, name, namespace string, expected map[string]adapters.JobSchedulingResult) error {
+	err := wait.Poll(framework.DefaultWaitInterval, wait.ForeverTestTimeout, func() (bool, error) {
+		clusterNames := []string{}
+		p, err := fedClient.CoreV1alpha1().FederatedJobPlacements(namespace).Get(name, metav1.GetOptions{})
+		clusterNames = p.Spec.ClusterNames
+		if err != nil {
+			return false, nil
```
marun: Why is it desirable to ignore this error (other than `IsNotFound`)?  Consider logging relevant errors ato aid in debugging.

test/integration/jobschedulingpreference_test.go
```diff
@@ -0,0 +1,264 @@
...
+	})
+
+	return err
+}
+
+func waitForMatchingJobOverride(fedClient clientset.Interface, name, namespace string, expected map[string]adapters.JobSchedulingResult) error {
+	err := wait.Poll(framework.DefaultWaitInterval, wait.ForeverTestTimeout, func() (bool, error) {
+		override, err := fedClient.CoreV1alpha1().FederatedJobOverrides(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
```
marun: Why is it desirable to ignore this error?

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
```diff
@@ -303,6 +303,7 @@ var (
 			Scope: "Namespaced",
 			Validation: &v1beta1.CustomResourceValidation{
 				OpenAPIV3Schema: &v1beta1.JSONSchemaProps{
+					Type: "object",
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go
```diff
@@ -303,7 +303,6 @@ var (
 			Scope: "Namespaced",
 			Validation: &v1beta1.CustomResourceValidation{
 				OpenAPIV3Schema: &v1beta1.JSONSchemaProps{
-					Type: "object",
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,144 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/controller/util/planner/job_planner.go
irfanurrehman: Updated!

pkg/controller/jobschedulingpreference/controller.go
irfanurrehman: Updated!

pkg/schedulingtypes/adapters/deployment.go
irfanurrehman: Not necessary though, used this as safe code, the method is implementing an interface.

pkg/apis/scheduling/v1alpha1/jobschedulingpreference_types.go
```diff
@@ -0,0 +1,64 @@
...
+	// and completions will be done using weightN/sumWeights.
+	// If omitted for a particular cluster(s), cluster(s) without explicit weight will not
+	// have any jobs scheduled.
+	// +optional
+	ClusterWeights map[string]int32 `json:"clusterWeights,omitempty"`
+}
+
+// JobSchedulingPreferenceStatus defines the observed state of JobSchedulingPreference
+type JobSchedulingPreferenceStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
```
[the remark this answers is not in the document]
irfanurrehman: No for now, but one the common status PR lands, the status from individual clusters might be useful.

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,147 @@
...
+			weightedList = append(weightedList, named(cluster, weight))
+			weightSum += weight
+		} else {
+			if weight, found := p.preferences.Spec.ClusterWeights["*"]; found {
+				weightedList = append(weightedList, named(cluster, weight))
+				weightSum += weight
+			} else {
+				// Clusters which do not have explicit weights given, while
+				// some others have, are treated equivalent to 0 weight.
+				// TODO(irfanurrehman): fix up a strategy about creating jobs
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,147 @@
...
+	// weights specified in the preferences.
+	weightedList := make([]*clusterWithWeight, 0, numClusters)
+	weightSum := int32(0)
+
+	for _, cluster := range availableClusters {
+		if weight, found := p.preferences.Spec.ClusterWeights[cluster]; found {
+			weightedList = append(weightedList, named(cluster, weight))
+			weightSum += weight
+		} else {
+			if weight, found := p.preferences.Spec.ClusterWeights["*"]; found {
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/controller/util/planner/job_planner.go
irfanurrehman: Thanks for pointing this out, I was wrongly using `quotient` for `remainder`. Updated!

pkg/controller/util/planner/job_planner.go
irfanurrehman: Updated!

hack/sample/jobschedulingpreference.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: scheduling.federation.k8s.io/v1alpha1
```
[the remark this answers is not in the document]
irfanurrehman: Done!

pkg/schedulingtypes/resources.go
```diff
@@ -43,6 +45,16 @@ var PodResource = &metav1.APIResource{
 	Namespaced: true,
 }
 
+var JobResource = &metav1.APIResource{
```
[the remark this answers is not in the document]
irfanurrehman: Done for Jobs. I see that the same can be used for some other resources also, created https://github.com/kubernetes-sigs/federation-v2/issues/325 for the same.

pkg/apis/core/v1alpha1/federatedjoboverride_types.go
```diff
@@ -31,6 +31,14 @@ type FederatedJobClusterOverride struct {
...
 	ClusterName string `json:"clusterName,omitempty"`
 	Parallelism *int32 `json:"parallelism,omitempty"`
+	// Completions value cannot be updated in a k8s job once created.
+	// This is added here with a disclaimer that a user (or a high level
+	// scheduler) will ensure to create a federate job override before
+	// creating a federated job template. Also once created, the same
+	// will not be updated.
+	// TODO(irfanurrehman): This needs to be reflected in documentation
+	// somewhere.
+	Completions *int32 `json:"completions,omitempty"`
```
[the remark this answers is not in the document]
irfanurrehman: I am working on the same. As talked in previous fed WG sync, I would prefer this done as another PR (and same for the e2e tests). Also a generic way of overriding any field is also desirable. I would push something which can override the additional needed field(s) and a generic way could be done, later, when the real need arises. I am fine if you feel that this needs to be blocked until the additional override and the e2e tests are completed. Let me know your comments.

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("Failed to get cluster list: %v", err))
+		return StatusError
+	}
+	if len(clusterNames) == 0 {
+		// no joined clusters, nothing to do
+		return StatusAllOK
+	}
+
+	if !j.plugin.TemplateExists(qualifiedName.String()) {
```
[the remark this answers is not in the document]
irfanurrehman: That's not the prerequisite anymore. The order of the creation, and/or steps needed at users end however needs to be documented separately. 
I find the resource readiness as suggested to be an useful feature too. I would prefer we keep the discussion outside this PR, as additional brainstorming might be needed for the same.

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,175 @@
...
+
+type JobScheduler struct {
+	plugin    *Plugin
+	fedClient fedclientset.Interface
+}
+
+func NewJobScheduler(fedClient fedclientset.Interface, kubeClient kubeclientset.Interface, crClient crclientset.Interface, fedNamespace, clusterNamespace, targetNamespace string, federationEventHandler, clusterEventHandler func(pkgruntime.Object), handlers *ClusterLifecycleHandlerFuncs) Scheduler {
+	scheduler := &JobScheduler{}
+	scheduler.fedClient = fedClient
+	adapter := adapters.NewFederatedJobAdapter(fedClient)
```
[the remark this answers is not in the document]
irfanurrehman: The use of the Adapter is to necessarily not duplicate some portion of code which is common among ReplicaScheduler and JobScheduler wrt to their target federated types.

test/integration/integration_test.go
irfanurrehman: Agree, I have reduced the integration tests to a single test to validate the path; would be submitting another PR for e2e tests.

test/integration/common.go
irfanurrehman: Updates to use the already existing function.

test/integration/jobschedulingpreference_test.go
irfanurrehman: This one is tricky. It would need some amount of update in the crud setup to allow getting only a template for a given type as a utility function. I think its not necessary for this PR and can be taken as an improvement.

irfanurrehman:
@marun @pmorie @gyliu513 updated. PTAL.
There are some actions which are needed as followup. I am actively working on them.

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -21,15 +21,18 @@ import (
 
 	fedschedulingv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/scheduling/v1alpha1"
 	fedclientset "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 	. "github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util/planner"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
xunpan: This "k8s.io/apimachinery/pkg/apis/meta/v1" is duplicated.

irfanurrehman: Thanks for pointing that out; updated!

test/integration/jobschedulingpreference_test.go
irfanurrehman: Fixed!

test/integration/jobschedulingpreference_test.go
irfanurrehman: Fixed!

test/integration/jobschedulingpreference_test.go
irfanurrehman: Have reduced the test to a single test validation the integration flow!

pkg/controller/util/planner/job_planner.go
```diff
@@ -0,0 +1,143 @@
...
+package planner
+
+import (
+	"hash/fnv"
+	"sort"
+
+	fedschedulingv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/scheduling/v1alpha1"
+)
+
+// JobPlanner decides how should to distribute jobs into federated clusters,
```
marun: nit: s/should//

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,177 @@
...
+type JobScheduler struct {
+	plugin    *Plugin
+	fedClient fedclientset.Interface
+}
+
+func NewJobScheduler(fedClient fedclientset.Interface, kubeClient kubeclientset.Interface, crClient crclientset.Interface, fedNamespace, clusterNamespace, targetNamespace string, federationEventHandler, clusterEventHandler func(pkgruntime.Object), handlers *ClusterLifecycleHandlerFuncs) Scheduler {
+	scheduler := &JobScheduler{}
+	scheduler.fedClient = fedClient
+	adapter := adapters.NewFederatedJobAdapter(fedClient)
+	tc, err := fedClient.CoreV1alpha1().FederatedTypeConfigs(util.DefaultFederationSystemNamespace).Get(JobTypeName, metav1.GetOptions{})
```
marun: Using the default system namespace will preclude the scheduler starting when the conrol plane is deployed to an arbitrary namespace.

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		glog.Errorf("Failed to get APIResource for %s: %v", JobTypeName, err)
```
marun: If this error occurs, the scheduler will error out on line 59.  I think this highlights the fragility of the current mechanism for enabling scheduling types (all or nothing).  If a type config for a type targeted by a given scheduler is either missing or not enabled, consider failing scheduler start.  

A future refinement could be adding scheduler management to the federation type config controller (or adding a similar controller dedicated to scheduling) so that a given scheduler could be started/stopped in response to the addition/deletion/modification of federated type configs that scheduling depends on.

test/integration/jobschedulingpreference_test.go
```diff
@@ -0,0 +1,245 @@
...
+		})
+	}
+}
+
+func initJSPTest(tl common.TestLogger, fedFixture *framework.FederationFixture) (*framework.ControllerFixture, clientset.Interface) {
+	config := fedFixture.KubeApi.NewConfig(tl)
+	restclient.AddUserAgent(config, "rsp-test")
+	kubeClient := fedFixture.KubeApi.NewClient(tl, "rsp-test")
+	fedClient := clientset.NewForConfigOrDie(config)
+	err := common.EnableType(kubeClient, fedClient, schedulingtypes.JobTypeName, util.DefaultFederationSystemNamespace)
```
marun: I don't think it should be necessary to create the type config in the API like this.  Consider passing the typeconfig into the scheduler constructor so that tests can pass the typeconfig directly and whatever starts the scheduler in the controller manager can be responsible for retrieving the appropriate type config from the API.

irfanurrehman:
> I can understand your hesitation to add e2e testing in this PR (it may be a considerable amount of work), but I'm concerned at the prospect of merging code that hasn't been tested. Especially in light of our desire to start releasing more often, I'd like to either have the changes you are proposing in this PR be hidden behind feature flags or that the PR merge only once e2e tests exist. Otherwise we risk releasing versions of federation that contain potentially broken functionality.

I understand your concern and am ok to push this out of the 0.0.2 release. Infact I have removed it from the 0.0.2 release milestone. I would be submitting a PR for multiple overrides soon and would follow it up with an e2e updated in this PR.

> Separately, the switch to retrieving the type config from the API has prompted me to think about how scheduler start could be made to better reflect system configuration. If a type config is missing or is not enabled for propagation, should a scheduler be enabled? I'm not suggesting this needs to be handled in this PR, just food for thought.

Yes, this seems doable; I have created https://github.com/kubernetes-sigs/federation-v2/issues/349 for the same. We can discuss more of over there, if need be.

tghartland:
There should be some way to accomodate the use case of a continuous stream of independant single completion jobs which are unaware of each other but part of a logical group (i.e same container and placement, but each with slightly different data/configuration).

The planning logic here is copied from the RSP planner, which is fine when you have a known (and large) amount of objects to distribute. But for the above example it would simply fill each cluster to capacity one by one, placing each new job in the highest weighted available cluster, effectively ignoring the desired weighting.

It may be possible to make the planner aware of the logical group through metadata, but it would be much simpler to have some placement option to say that for single jobs they should be distributed using the cluster weights to make a weighted random choice of the available clusters. With enough jobs this will produce the same outcome eventually.

The planning logic being non-deterministic should not be a problem because as long as the number of completions is left as 1 it should never try to re-plan, yes? Preference should be given to leaving a running job where it is, but this means that changes in the placement weightings will not be reflected until old jobs start finishing. And to remove a cluster from the placement it would be set to weight 0 on new jobs and you would then have to wait for all the old jobs to complete.

Re-planning jobs on clusters that go unhealthy is probably fine, but moving jobs to a cluster that becomes healthy again isn't. Is that an issue or is it possible to do one but not the other?

pkg/apis/scheduling/v1alpha1/jobschedulingpreference_types.go
```diff
@@ -0,0 +1,67 @@
...
+// JobSchedulingPreferenceSpec defines the desired state of JobSchedulingPreference
+type JobSchedulingPreferenceSpec struct {
+	// Specifies the maximum desired number of pods this FederatedJob should
+	// run at any given time.
+	// Parallelism specified in the spec for target job template will be
+	// discarded/overridden when job scheduling preferences are specified.
+	// // TODO(irfanurrehman): If needed fix up a strategy about creating jobs
+	// into federated clusters with parallelism as 0. Right now we do not
+	// create jobs in those clusters which have 0 weightage in ClusterWeights.
+	TotalParallelism int32 `json:"totalParallellism"`
```
tghartland: The json tag has an extra L in parallelism here. I Ctrl-F for this but only see it in generated code, so this should be the only place to change.

irfanurrehman: Updated!

irfanurrehman:
> There should be some way to accomodate the use case of a continuous stream of independant single completion jobs which are unaware of each other but part of a logical group (i.e same container and placement, but each with slightly different data/configuration).

As discussed in the last workgroup sync, we can work incrementally to accommodate your use cases.



> The planning logic here is copied from the RSP planner, which is fine when you have a known (and large) amount of objects to distribute. But for the above example it would simply fill each cluster to capacity one by one, placing each new job in the highest weighted available cluster, effectively ignoring the desired weighting.

The Planning algorithm is quite different from the replicaPlanner; the idea to implement planner and use of the same while scheduling however is similar. The weight is not across the jobs, but used to distribute from TotalParallelism and TotalCompletions.

> It may be possible to make the planner aware of the logical group through metadata, but it would be much simpler to have some placement option to say that for single jobs they should be distributed using the cluster weights to make a weighted random choice of the available clusters. With enough jobs this will produce the same outcome eventually.

This is more akin to weighted placement which can be another scheduler applicable to generic resource set; this particular PR deals with scheduling jobs from a common job configuration.



> The planning logic being non-deterministic should not be a problem because as long as the number of completions is left as 1 it should never try to re-plan, yes? Preference should be given to leaving a running job where it is, but this means that changes in the placement weightings will not be reflected until old jobs start finishing. And to remove a cluster from the placement it would be set to weight 0 on new jobs and you would then have to wait for all the old jobs to complete.

I am still leaving the replan option as future enhancement. As discussed in the workgroup sync, this PR is to distribute a set of jobs with a common Federated template from a total Parallelism and total Completions.  What I have in mind for replan is to be able to redistribute the two values with only incremental changes possible in placement (for example placement can be updated from set of 2 clusters to a superset say of 3 clusters including the older 2). The values of completions in the jobs sent to the older clusters are not changed, but newer clusters in the set can get additional completions. We can work together to arrive at a suitable algorithm (keeping the immutability of completions in mind) fitting your use cases.




> Re-planning jobs on clusters that go unhealthy is probably fine, but moving jobs to a cluster that becomes healthy again isn't. Is that an issue or is it possible to do one but not the other?

I think what you meant is replanning jobs away from clusters which go unhealthy is fine. This PR would do that.  In this implementation pass a job once created in a cluster wont simply be updated until that cluster is there. This ideally works for your use case of Parallelism=1 Completions=1 cases.
The steps to achieve that would be:
- create JobSchedulingPreference with totalP=1 and totalC=1
- create FederatedJob (template) to reflect the spec you want for your job.
The algorithm will create a placement with precisely once cluster chosen at random.

Now following cases can happen:
1. The cluster in which the job is placed remains healthy:
  Job completes, All good.
2. The cluster in which the job was placed goes unhealthy and never comes back:
  The job is created into another cluster after a timeout and all is good.
3. The cluster in which the job was placed goes unhealthy and comes  back after a timeout:
  There would be 2 copies of the same job in 2 different clusters and the controller won't update either, both run to completion. Is this a problem?

@tghartland

irfanurrehman:
@marun I have updated and rebased this on the latest code. Have updated one e2e test also for validation as you requested. PTAL!

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		overrides = append(overrides, overridesMap)
+	}
+	obj.Object[util.SpecField] = map[string]interface{}{
+		util.OverridesField: overrides,
+	}
+}
+
+// TODO (irfanurrehman): in this pass we ignore updates for jobs
+func (r *JobScheduleResult) OverrideUpdateNeeded(typeConfig typeconfig.Interface, obj *unstructured.Unstructured) bool {
```
marun: Given that this method appears to have been implemented in the fake method for the test, and why should this method remain unimplemented in this PR?

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,229 @@
...
+	plugin *Plugin
+
+	fedClient fedclientset.Interface
+
+	controllerConfig *ControllerConfig
+
+	eventHandlers SchedulerEventHandlers
+}
+
+func NewJobScheduler(controllerConfig *ControllerConfig, eventHandlers SchedulerEventHandlers) (Scheduler, error) {
```
marun: (No action required) I'm not sure I understand the rational for an entirely new scheduler.  Plugins are already a thing - why can't a single controller delegate to type-specific plugins?

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,229 @@
...
+		controllerConfig: controllerConfig,
+		eventHandlers:    eventHandlers,
+		fedClient:        fedClient,
+	}
+
+	return scheduler, nil
+}
+
+func (j *JobScheduler) Kind() string {
+	return RSPKind
```
marun: I'm assuming this should be `JSPKind`?

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (j *JobScheduler) StartPlugin(typeConfig typeconfig.Interface, stopChan <-chan struct{}) error {
+	kind := typeConfig.GetTemplate().Kind
+	plugin, err := NewPlugin(j.controllerConfig, j.eventHandlers, typeConfig)
+	if err != nil {
+		return fmt.Errorf("Failed to initialize replica scheduling plugin for %q: %v", kind, err)
```
marun: More copy-pasta: `replica`

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	plugin.Start(stopChan)
+	j.plugin = plugin
+
+	return nil
+}
+
+func (j *JobScheduler) ObjectType() pkgruntime.Object {
+	return &fedschedulingv1a1.JobSchedulingPreference{}
```
marun: (No action required) This repo now vendors `controller-runtime`, so generic client/informer/lister is a thing.  Consider migrating away from the legacy generated stuff in future iterations.

pkg/schedulingtypes/jobscheduler.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (j *JobScheduler) FedList(namespace string, options metav1.ListOptions) (pkgruntime.Object, error) {
+	return j.fedClient.SchedulingV1alpha1().JobSchedulingPreferences(j.controllerConfig.TargetNamespace).List(options)
+}
+
+func (j *JobScheduler) FedWatch(namespace string, options metav1.ListOptions) (watch.Interface, error) {
+	return j.fedClient.SchedulingV1alpha1().JobSchedulingPreferences(j.controllerConfig.TargetNamespace).Watch(options)
+}
+
+func (j *JobScheduler) Start(stopChan <-chan struct{}) {
```
marun: (No action required) I find it pretty confusing that `Start` is a method of the interface that isn't used.  It suggests a future refactor.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -273,26 +273,103 @@ func (s *ReplicaScheduler) GetSchedulingResult(rsp *fedschedulingv1a1.ReplicaSch
...
+
+func (r *ReplicaScheduleResult) clusterNames() []string {
+	clusterNames := []string{}
+	for name := range r.result {
+		clusterNames = append(clusterNames, name)
+	}
+	return clusterNames
+}
+
+func (r *ReplicaScheduleResult) SetPlacementSpec(obj *unstructured.Unstructured) {
```
marun: I don't think this method and the one that follows should exist in a type-specific implementation.   Placement is common across all federated types.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -273,26 +273,103 @@ func (s *ReplicaScheduler) GetSchedulingResult(rsp *fedschedulingv1a1.ReplicaSch
...
+			replicasField:         replicas,
+		}
+		overrides = append(overrides, overridesMap)
+	}
+	obj.Object[util.SpecField] = map[string]interface{}{
+		util.OverridesField: overrides,
+	}
+}
+
+func (r *ReplicaScheduleResult) OverrideUpdateNeeded(typeConfig typeconfig.Interface, obj *unstructured.Unstructured) bool {
```
marun: The bulk of  this method is not type-specific, which suggests to me factoring out the common functionality into a type-agnostic method that can be used by both job and replica scheduling.

test/e2e/jobscheduling.go
```diff
@@ -0,0 +1,225 @@
...
+					}
+
+					err = waitForMatchingOverride(tl, typeConfig, kubeConfig, name, namespace, expected)
+					if err != nil {
+						tl.Fatalf("Failed waiting for matching overrides: %v", err)
+					}
+				})
+			}
+		})
+		// This test runs only for one type/kind
```
marun: The replica scheduling logic is already a bit funky in looping over all the type configs, and I think this break statement takes it too far.  Please separate detection of the target type configs from the test definition to avoid this hackiness.

test/e2e/jobscheduling.go
```diff
@@ -0,0 +1,225 @@
+/*
```
marun: (No action required) Most of this file is duplicative for the tests of replica scheduling, and I'd appreciate seeing the 2 converge in subsequent PRs.

test/e2e/framework/managed/controller.go
```diff
@@ -99,12 +99,11 @@ func NewClusterControllerFixture(config *util.ControllerConfig) *ControllerFixtu
 }
 
 // NewRSPControllerFixture initializes a new RSP controller fixture.
-func NewRSPControllerFixture(tl common.TestLogger, config *util.ControllerConfig, typeConfigs []typeconfig.Interface) *ControllerFixture {
+func NewSchedulingControllerFixture(tl common.TestLogger, schedulingKind string, config *util.ControllerConfig, typeConfigs []typeconfig.Interface) *ControllerFixture {
```
marun: Please move the `schedulingKind` argument to the end of the argument list for consistency with the other fixture factory methods.

tghartland:
@irfanurrehman 

> The Planning algorithm is quite different from the replicaPlanner; the idea to implement planner and use of the same while scheduling however is similar. The weight is not across the jobs, but used to distribute from TotalParallelism and TotalCompletions.

I was referring specifically to [this](https://github.com/kubernetes-sigs/federation-v2/blob/5a62ec5bd062eaee2a55de60cb683afa8716d0bc/pkg/controller/util/planner/job_planner.go#L105) logic where the highest weighted cluster always gets distributed to first, which is where the problem of distributing single objects comes from.

> This is more akin to weighted placement which can be another scheduler applicable to generic resource set; this particular PR deals with scheduling jobs from a common job configuration.

The parallelism/completions of a job do justify it having a specialised scheduler, and ignoring those by setting them both to one may be a better fit for a more general scheduler where a random weighted placement makes more sense.

While I'm thinking about it, the idea of some kind of "Group" resource comes to mind. When using a single cluster it's convenient to put several resource types in one .yaml file to create them with a single kubectl command, but doing so in a random weighted federation will end up with them all in different places. If there could be some kind of group resource which can be randomly placed on a single cluster and then all following federated resources which specify that group in their metadata will go to the same location, only the group scheduling preference would be required. The more I think about it the more complex it gets though, so it's probably not a good idea. Maybe I'm just thinking of a randomly placed namespace, but without the isolation that namespaces have.

> > Re-planning jobs on clusters that go unhealthy is probably fine, but moving jobs to a cluster that becomes healthy again isn't. Is that an issue or is it possible to do one but not the other?

> I think what you meant is replanning jobs away from clusters which go unhealthy is fine.

That is what I meant, yes. As for the issue that this brings with it:

> The cluster in which the job was placed goes unhealthy and comes back after a timeout:
There would be 2 copies of the same job in 2 different clusters and the controller won't update either, both run to completion. Is this a problem?

I think this is a problem even if the cluster does not come back to a healthy state, assuming that the job still continues running and is uploading results to S3 storage or something similar, and the unhealthy state is not bad enough to stop that happening (i.e the cluster is unreachable from the control plane, but the job itself still has full networking capabilities).

The burden for checking that the job has not already been completed somewhere else might just have to be passed to the job itself. And for that I'm assuming a job structure in which it always waits until the end to transfer out its results. If it does many smaller transfers while still processing and you get a mix of results then that's a bigger problem, especially if it's something like a random simulation where the output of the two jobs is expected to be similar but not be the same.

The only solution I can think of that could be implemented as part of federation would be to periodically check for duplicates and simply delete the older jobs, since you don't know what might have gone wrong with them while the cluster was unhealthy.

I played around with the RSP a while ago, and tried disabling and re-enabling the kube-api service on one of the clusters in the federation. The pods that were on the unhealthy cluster were recognised  when the kube-api came back up, and adopted by the FederatedDeployment again (I think, they might have been deleted and remade but I can't remember). Will this implementation of JobSchedulingPreferences do the same thing or are the job pods that are on the cluster that went unhealthy abandoned since the total replicas/parallelism can't change and the job pods were already recreated somewhere else? Or does it adopt the job pods and just accept that it has more than there should be?

marun:
@irfanurrehman This PR hasn't been updated in some time.   Are you still intending to work on it?

irfanurrehman:
> @irfanurrehman This PR hasn't been updated in some time. Are you still intending to work on it?

I do intend to work on the feature; but its not higher on priority list compared to the items we have at hand. Closing it for now.

Issue #255, opened 2018-09-07
Overrides are not being applied in testing


marun:
While debugging #214, I discovered that overrides are not actually being applied in e2e testing.  Since resources are not directly compared (only versions), no test is catching this problem.

The crudtester created `FederatedDeployment` with replicas of 3, and `FederatedDeploymentOverride` with replicas of 5.

What resulted: A `Deployment` was created with replicas of 3.

What should have happened: A `Deployment` was created with replicas of 5.

marun:
This bug is the result of a partial rename of `clusternames` to `clusterNames`.  Before this issue is closed, testing for the application of overrides needs to be added.

marun:
Fixed by #266

PR #260, opened 2018-09-11
s/clustername/clusterName/ everywhere

marun:
#253 only partially fixed a renaming of `clusternames` to `clusterNames` (the actual json field name), and this caused #255.  This PR fixes the remaining references to fix the breakage that resulted from the partial rename.  As part of #255, a follow-on PR will add testing to prevent this from recurring.

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -957,7 +957,7 @@ func (s *FederationSyncController) objectForCluster(template, override *unstruct
 	overrideField := overridePath[len(overridePath)-1]
 	for _, overrideInterface := range overrides {
 		clusterOverride := overrideInterface.(map[string]interface{})
-		if clusterOverride["clustername"] != clusterName {
+		if clusterOverride[util.ClusterNameField] != clusterName {
 			continue
 		}
 		data, ok := clusterOverride[overrideField]
--- a/pkg/controller/util/constants.go
+++ b/pkg/controller/util/constants.go
@@ -24,8 +24,9 @@ import (
 const (
 	NoResyncPeriod time.Duration = 0 * time.Second
 
-	NamespaceKind = "Namespace"
-	ServiceKind   = "Service"
+	NamespaceKind     = "Namespace"
+	ServiceKind       = "Service"
+	ClusterNamesField = "clusterNames"
 )
 
 type ReconciliationStatus int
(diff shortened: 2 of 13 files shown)


pkg/controller/sync/controller.go
```diff
@@ -957,7 +957,7 @@ func (s *FederationSyncController) objectForCluster(template, override *unstruct
 	overrideField := overridePath[len(overridePath)-1]
 	for _, overrideInterface := range overrides {
 		clusterOverride := overrideInterface.(map[string]interface{})
-		if clusterOverride["clustername"] != clusterName {
+		if clusterOverride["clusterName"] != clusterName {
```
gyliu513: If you update here, the `federated<Type>-override.yaml` at https://github.com/kubernetes-sigs/federation-v2/tree/master/example/sample1 should also be updated.

```
  spec:
    overrides:
    - clustername: cluster2 <<<<
      data:
        foo: bar
```

The commit message is also not accurate, you are not only changing `clusternames` but also `clustername`.

Another point is it is terrible that the test `**override.yaml` are using `clusterName`, but the template in example are using `clustername`, why not keep the files in `sample1` and `fixtures` keep synced and use same file?

marun: Good catch on both clustername and clusternames needing to be fixed.  I disagree that the tests need to match the samples, they serve different use case.  The goal of ensuring consistent field name usage can be accomplished in code as per the new commit in this PR.

pkg/controller/util/constants.go
```diff
@@ -26,6 +26,7 @@ const (
 
 	NamespaceKind     = "Namespace"
 	ServiceKind       = "Service"
+	ClusterNameField  = "clusterName"
```
gyliu513: This is good, but I still have some concerns if we do not sync up the YAML templates for example and test.

1) People add a new federated resource, and add some test cases YAML template with `override` policy, and use correct `clusterName` in YAML, CI test passed.
2) In example file, they use `clustername`, but this will not be detected if people do not test the example in a federation environment.

If we can keep the YAML templates in example and test keep synced, then we will not have such issues. Comments @marun

marun: I agree that the walkthrough needs to be tested, but I think that requirement is orthogonal to other forms of testing.  #104 is intended to track the need to automate the walkthrough so it can be tested in CI (cc: @font).  Also, #60 (my next work item) will remove the need to define primitives in code (they will exist only in the api).

Issue #262, opened 2018-09-12
Add more examples for 0.0.2 release


gyliu513:
Follow-up from September 12 working group meeting.

In the coming 0.0.2 release, we introduced some new features for federation v2 but we do not have examples for those new features, it is better add examples for those features for end user to have a try for the new features.

Some features in my mind is as follows:
- Replica Scheduling Preference
- Job Scheduling Preference
- Federated Ingress
- Federated Jobs
- ...

@irfanurrehman @marun @kubernetes-sigs/federation-wg

gyliu513:
@shashidharatd would you please add a test example for federated ingress?

@irfanurrehman as you are working for Job Scheduling at https://github.com/kubernetes-sigs/federation-v2/pull/259 , I assume you will add example for this ;-)

marun:
@gyliu513 Given that 0.0.2 was released some time ago, I'm closing this issue.  Please file a separate issue for a future release if there are outstanding doc requirements not captured elsewhere.

gyliu513:
@marun all of the examples are now there and it is ok to close this.

PR #263, opened 2018-09-13
Added example for federatedJob.

gyliu513:
This fixed part of https://github.com/kubernetes-sigs/federation-v2/issues/262

/cc @irfanurrehman @pmorie @marun

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -216,7 +216,7 @@ kubectl apply -R -f example/sample1
 Check the status of all the resources in each cluster by running:
 
 ```bash
-for r in configmaps secrets service deployment serviceaccount; do
+for r in configmaps secrets service deployment serviceaccount job; do
     for c in cluster1 cluster2; do
         echo; echo ------------ ${c} resource: ${r} ------------; echo
         kubectl --context=${c} -n test-namespace get ${r}
@@ -251,7 +251,7 @@ kubectl -n test-namespace edit federatednamespaceplacement test-namespace
 Then wait to verify all resources are removed from `cluster2`:
 
 ```bash
-for r in configmaps secrets service deployment serviceaccount; do
+for r in configmaps secrets service deployment serviceaccount job; do
     for c in cluster1 cluster2; do
         echo; echo ------------ ${c} resource: ${r} ------------; echo
         kubectl --context=${c} -n test-namespace get ${r}
@@ -274,7 +274,7 @@ kubectl -n test-namespace edit federatednamespaceplacement test-namespace
 Then wait and verify all resources are added back to `cluster2`:
 
 ```bash
-for r in configmaps secrets service deploy; do
+for r in configmaps secrets service deploy job; do
     for c in cluster1 cluster2; do
         echo; echo ------------ ${c} ------------; echo
         kubectl --context=${c} -n test-namespace get ${r}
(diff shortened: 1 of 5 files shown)


marun:
I'm concerned that deploying all workload types to the existing userguide workflow will make it less comprehensible.  Ideally the userguide walkthrough introduces federation concepts, and I don't see how employing all workload types is an improvement.  I think a new jobs-specific walkthrough could be useful, since its behavior differs from a deployment.  I'm less convinced of the value of a replicaset walkthrough, since replicaset behavior is a subset of deployments. 

cc: @font

example/sample1/federateddeployment-template.yaml
```diff
@@ -23,5 +23,4 @@ items:
           spec:
             containers:
             - image: nginx
-              imagePullPolicy: Always
```
marun: (No action required) Do we want to continue using nginx for the walkthrough?  I switched the test fixture to use busybox because it is much lighter weight (2mb vs 100mb).

cc: @font

example/sample1/federatedreplicaset-override.yaml
```diff
@@ -0,0 +1,12 @@
...
+items:
+- apiVersion: core.federation.k8s.io/v1alpha1
+  kind: FederatedReplicaSetOverride 
+  metadata:
+    name: test-replicaset
+    namespace: test-namespace
+  spec:
+    overrides:
+    - clusterName: cluster2
+      replicas: 5
```
marun: (No action required) Same comment about replicas, the test fixtures now use the minimum (`replicas: 1` for template, `replicas: 2` for override).

example/sample1/federateddeployment-template.yaml
font: We should probably keep using `nginx` for deployments *only*, as we rely on it to verify `FederatedService` works by curling the web server using the minikube IP and `NodePort`.

example/sample1/federatedjob-template.yaml
```diff
@@ -0,0 +1,24 @@
...
+      template:
+        metadata:
+          labels:
+            app: nginx
+        spec:
+          terminationGracePeriodSeconds: 0
+          restartPolicy: Never
+          containers:
+            - name: nginx
+              image: nginx
```
font: Please use `busybox` if all we're doing is looping for a while to keep the container running.

docs/userguide.md
```diff
@@ -251,7 +251,7 @@ kubectl -n test-namespace edit federatednamespaceplacement test-namespace
 Then wait to verify all resources are removed from `cluster2`:
 
 ```bash
-for r in configmaps secrets service deploy; do
+for r in configmaps secrets service deploy rs job; do
```
font: I'm not sure that adding replicasets to the example really adds that much value as we already have deployments. It becomes a little repetitive and requires even more resources to run the example. We could just add a note about them, or list/point to the set of resources we support. I'm okay with jobs though.

font:
I agree with @marun here. I don't think replicasets adds that much value over deployments.

docs/userguide.md
gyliu513: Done, deleted rs.

example/sample1/federatedjob-template.yaml
gyliu513: done

marun:
LGTM, will leave to @font to approve.

PR #264, opened 2018-09-13
Add doc for deployment of namespaced federation

marun:
Targets #197

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -351,6 +351,27 @@ on both the host and member clusters.
 It may be useful to supply `FEDERATION_NAMESPACE=test-namespace` to
 allow the examples to work unmodified.
 
+### Joining Clusters
+
+Joining additional clusters to a namespaced federation requires
+providing additional arguments to `kubefed2 join`:
+
+- `--federation-namespace=<namespace>` to ensure the cluster is joined
+  to the federation running in the specified namespace
+- `--registry-namespace=<namespace>` if using `--add-to-registry` to
+  ensure the cluster is registered in the correct namespace
+- `--limited-scope=true` to ensure that the scope of the service account created in
+  the target cluster is limited to the federation namespace
+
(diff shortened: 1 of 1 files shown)


docs/userguide.md
```diff
@@ -323,3 +323,40 @@ The argument(s) used is/are the list of context names of the additional clusters
 unjoined from this federation control plane. Clarifying, say the `host-cluster-context` used is
 `cluster1`, then on successful completion of the script used in example, both `cluster1` and `cluster2`
 will be unjoined from the deployed federation control plane.
+
+## Namespaced Federation
```
gyliu513: Do we need to update the tile for above to highlight it is `Cluster Level Federation`?

Including title for `Automated Deployment` and `Manual Deployment`?

marun: I don't think I want to rewrite the other sections just yet.  Likely that will happen eventually, but initially the goal is making this feature available more than changing general perception of cluster-scoped federation.

gyliu513: I think we do not need to rewrite the other sections, but just updating the title as follows should be good enough currently:

Automated Deployment -> Cluster Level Federation Automated Deployment
Manual Deployment -> Cluster Level Federation Manual Deployment

marun:
I've addressed @font's feedback and added a new section on how to join a cluster to a namespaced federation.

docs/userguide.md
```diff
@@ -323,3 +323,60 @@ The argument(s) used is/are the list of context names of the additional clusters
...
+plane to.  The control plane will only have permission to access this
+on both the host and member clusters.
+
+It may be useful to supply `FEDERATION_NAMESPACE=test-namespace` to
+allow the examples to work unmodified.
+
+### Joining Clusters
+
+Joining additional clusters to a namespaced federation requires
+providing additional arguments to `kubefed join`:
```
pmorie: nit: `kubefed2`

PR #266, opened 2018-09-14
Add check of application of overrides to the crudtester

marun:
Previously there was no testing of overrides being applied in integration and e2e testing.  As part of this PR, the marshaling of overrides was refactored for reuse.

Targets #255

### DIFF
--- a/pkg/controller/util/overrides.go
+++ b/pkg/controller/util/overrides.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


marun:
Adding wip - intend to unify marshaling between the sync controller and crudtester.

marun:
Rebased

pkg/controller/util/overrides.go
```diff
@@ -0,0 +1,79 @@
...
+	return &ClusterOverrides{
+		Path:      path,
+		Overrides: overrides,
+	}, nil
+}
+
+func marshallOverrides(typeConfig typeconfig.Interface, override *unstructured.Unstructured) (map[string]interface{}, []string, error) {
+	overrideMap := make(map[string]interface{})
+	overridePath := []string{}
+	if typeConfig.GetOverride() != nil && override != nil {
```
font: (no action required) Consider performing the nil check first and returning immediately.

pkg/controller/util/overrides.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &ClusterOverrides{
+		Path:      path,
+		Overrides: overrides,
+	}, nil
+}
+
+func marshallOverrides(typeConfig typeconfig.Interface, override *unstructured.Unstructured) (map[string]interface{}, []string, error) {
```
font: (no action required) is marshall the right term here? It seems like we're deserializing the unstructured data into a meaningful map and string slice.

Issue #75, opened 2018-05-17
Simplify placement


marun:
Resource placement currently requires resource-specific placement.  To be propagated to a given cluster, a template must have that cluster listed in its associated placement resource, and the containing namespace must also have a placement resource listing that cluster.

It would be desirable to make namespace-based placement the default.  If the placement for a namespace specified a given cluster, all federated resources in that namespace would be propagated to that cluster if resource-specific placement was not present.

Where more fine-grained control is required, resource-specific placement could be employed.  Where both namespace and resource-specific placement resources existed, placement could be determined either additively (namespace clusters + resource clusters) or exclusively (resource clusters only).  Both namespace and resource-specific placement resources could indicate via a boolean whether propagation should be restricted (e.g. `propagate: true|false`).

marun:
Moving out of v1alpha1, this will be targeted at a subsequent alpha release.

shashidharatd:
@marun, this feature is implemented and available. is there something pending for this, otherwise shall we close this issue?

marun:
@shashidharatd I don't believe this feature is implemented yet.  The only option for placement is resource-specific.

PR #272, opened 2018-09-18
Add godoc for FederatedNamespacePlacement

pmorie:
/wip

### DIFF
--- a/pkg/apis/core/v1alpha1/federatednamespaceplacement_types.go
+++ b/pkg/apis/core/v1alpha1/federatednamespaceplacement_types.go
@@ -33,7 +33,9 @@ type FederatedNamespacePlacementStatus struct {
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// FederatedNamespacePlacement
+// FederatedNamespacePlacement is not implemented yet and should not be
+// used at this time.
+//
 // +k8s:openapi-gen=true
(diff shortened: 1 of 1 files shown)


pkg/apis/core/v1alpha1/federatednamespaceplacement_types.go
```diff
@@ -33,7 +33,9 @@ type FederatedNamespacePlacementStatus struct {
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// FederatedNamespacePlacement
+// FederatedNamespacePlacement is not implemented yet and should not be
+// used at this time.
```
gyliu513: Why this not implemented? I saw we already have an example https://github.com/kubernetes-sigs/federation-v2/blob/master/example/sample1/federatednamespace-placement.yaml

Also if this intend to be wip, it is better update PR title by adding `WIP` as keyword.

marun: Namespace placement is implemented.  It is slated for revision in #75, since currently namespaces are propagated by default and that complicates a selective (not all namespaces intended to be propagated) cluster-scoped deployment of the federation control plane.

gyliu513: I see, thanks @marun , so here I think mentioning `FederatedNamespacePlacement is not implemented` may not that accurate, can we explain this in detail and also link #75 in the comment?

PR #281, opened 2018-09-21
Changes to support releasing v0.0.2-rc.1

font:


font:
@gyliu513 The image will be updated when we create a new release tag.

PR #283, opened 2018-09-24
Added more parameters for federation chart.

gyliu513:
/cc @shashidharatd

### DIFF
--- a/charts/federation-v2/README.md
+++ b/charts/federation-v2/README.md
@@ -21,7 +21,7 @@ First you'll need to create the reserved namespace for registering clusters with
 cluster registry:
 
 ```bash
-kubectl create ns kube-multicluster-public
+$ kubectl create ns kube-multicluster-public
 ```
 
 Install the chart with the release name `federation-v2`:
@@ -46,19 +46,19 @@ of the CR and CRDs for federation v2 release.
 Delete all federation v2 `FederatedTypeConfig`:
 
 ```bash
-kubectl -n federation-system delete FederatedTypeConfig --all
+$ kubectl -n federation-system delete FederatedTypeConfig --all
 ```
 
 Delete all federation v2 CRDs:
 
 ```bash
-kubectl delete crd $(kubectl get crd | grep -E 'federation.k8s.io' | awk '{print $1}')
+$ kubectl delete crd $(kubectl get crd | grep -E 'federation.k8s.io' | awk '{print $1}')
 ```
 
 If you want to delete `clusters.clusterregistry.k8s.io` as well, do it as follows:
 
 ```bash
-kubectl delete crd clusters.clusterregistry.k8s.io
+$ kubectl delete crd clusters.clusterregistry.k8s.io
 ```
 
 Then you can uninstall/delete the `federation-v2` release:
@@ -73,7 +73,7 @@ and deletes the release.
 Delete the reserved namespace for registering clusters:
 
 ```bash
-kubectl delete ns kube-multicluster-public
+$ kubectl delete ns kube-multicluster-public
 ```
 
 ## Configuration
@@ -83,13 +83,17 @@ chart and their default values.
 
 | Parameter | Description | Default |
 |-----------|-------------|---------|
-| controllermanager.replicaCount | Number of replica for federation v2 controller manager | 1 |
-| controllermanager.repository | Repo of the federation v2 image | quay.io/kubernetes-multicluster |
-| controllermanager.image | Name of the federation v2 image | federation-v2 |
-| controllermanager.tag | Tag of the federation v2 image | latest |
-| controllermanager.pullPolicy | Image pull policy | IfNotPresent |
-| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2 | true |
-| federatedtypes.enabled | Specifies whether to install the `FederatedTypeConfig` in federation v2 | true |
+| controllermanager.replicaCount | Number of replica for federation v2 controller manager. | 1 |
+| controllermanager.repository | Repo of the federation v2 image. | quay.io/kubernetes-multicluster |
+| controllermanager.image | Name of the federation v2 image. | federation-v2 |
+| controllermanager.tag | Tag of the federation v2 image. | latest |
+| controllermanager.pullPolicy | Image pull policy. | IfNotPresent |
+| controllermanager.featureGates | Feature gates are a set of `key=value` pairs that describe alpha or experimental features. An administrator can use the `--feature-gates` command line flag on each component to turn a feature on or off. | PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true |
+| controllermanager.federationNamespace | The namespace the federation control plane is deployed in. | federation-system |
+| controllermanager.registryNamespace | The cluster registry namespace. | kube-multicluster-public |
+| controllermanager.limitedScope | Whether the federation namespace will be the only target for federation. If set to true, the value set for `controllermanager.registryNamespace` and `controllermanager.registryNamespace` will be ignored. | false |
+| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2. | true |
+| federatedtypes.enabled | Specifies whether to install the `FederatedTypeConfig` in federation v2. | true |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to
 `helm install`.
@@ -98,5 +102,5 @@ Alternatively, a YAML file that specifies the values for the parameters can be
 provided while installing the chart. For example:
 
 ```bash
-$ helm install . --name federation-v2 --namespace federation-system --values values.yaml
+$ helm install charts/federation-v2 --name federation-v2 --namespace federation-system --values values.yaml
 ```
--- a/charts/federation-v2/templates/role.yaml
+++ b/charts/federation-v2/templates/role.yaml
@@ -1,12 +1,13 @@
-{{- if not .Values.controllermanager.limitedScope }}
+{{- if .Values.controllermanager.limitedScope }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
   labels:
     api: federation
     kubebuilder.k8s.io: 1.0.0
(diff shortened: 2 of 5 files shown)


charts/federation-v2/README.md
```diff
@@ -83,13 +83,17 @@ chart and their default values.
...
-| controllermanager.pullPolicy | Image pull policy | IfNotPresent |
-| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2 | true |
-| federatedtypes.enabled | Specifies whether to install the `FederatedTypeConfig` in federation v2 | true |
+| controllermanager.replicaCount | Number of replica for federation v2 controller manager. | 1 |
+| controllermanager.repository | Repo of the federation v2 image. | quay.io/kubernetes-multicluster |
+| controllermanager.image | Name of the federation v2 image. | federation-v2 |
+| controllermanager.tag | Tag of the federation v2 image. | latest |
+| controllermanager.pullPolicy | Image pull policy. | IfNotPresent |
+| controllermanager.featureGates | Feature gates are a set of `key=value` pairs that describe alpha or experimental features. An administrator can use the `--feature-gates` command line flag on each component to turn a feature on or off. | PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true |
+| controllermanager.fedNamespace | The namespace the federation control plane is deployed in. | federation-system |
```
marun: s/fedNamespace/federationNamespace/

charts/federation-v2/README.md
```diff
@@ -83,13 +83,17 @@ chart and their default values.
...
+| controllermanager.clusterNamespace | The cluster registry namespace. | kube-multicluster-public |
+| controllermanager.clusterNamespace | Whether the federation namespace will be the only target for federation. | false |
```
marun: s/clusterNamespace/limitedScope/

charts/federation-v2/README.md
```diff
@@ -83,13 +83,17 @@ chart and their default values.
...
-| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2 | true |
-| federatedtypes.enabled | Specifies whether to install the `FederatedTypeConfig` in federation v2 | true |
+| controllermanager.replicaCount | Number of replica for federation v2 controller manager. | 1 |
+| controllermanager.repository | Repo of the federation v2 image. | quay.io/kubernetes-multicluster |
+| controllermanager.image | Name of the federation v2 image. | federation-v2 |
+| controllermanager.tag | Tag of the federation v2 image. | latest |
+| controllermanager.pullPolicy | Image pull policy. | IfNotPresent |
+| controllermanager.featureGates | Feature gates are a set of `key=value` pairs that describe alpha or experimental features. An administrator can use the `--feature-gates` command line flag on each component to turn a feature on or off. | PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true |
+| controllermanager.fedNamespace | The namespace the federation control plane is deployed in. | federation-system |
+| controllermanager.clusterNamespace | The cluster registry namespace. | kube-multicluster-public |
```
marun: s/clusterNamespace/registryNamespace/

charts/federation-v2/README.md
```diff
@@ -83,13 +83,17 @@ chart and their default values.
...
+| controllermanager.clusterNamespace | Whether the federation namespace will be the only target for federation. | false |
+| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2. | true |
+| federatedtypes.enabled | Specifies whether to install the `FederatedTypeConfig` in federation v2. | true |
```
marun: s/federatedtypes/installfederatedtypes/ ?

I think it would be desirable to be explicit about what behavior is being enabled, which in this case is enabling the installation of federated types rather than enabling federated types themselves (which is accomplished via the installation of the FederatedTypeConfig CRD).

charts/federation-v2/values.yaml
```diff
@@ -18,6 +18,20 @@ controllermanager:
     requests:
       cpu: 100m
       memory: 64Mi
+  ## Feature gates are a set of key=value pairs that describe alpha or
+  ## experimental features. An administrator can use the --feature-gates
+  ## command line flag on each component to turn a feature on or off.
+  ## Federation V2 now have the following default feature gates:
+  ## PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true
```
marun: (No action required) Ugh, this is gross.  I think we need to transition to versioned component config to ensure that federation config can be backwards compatible or risk complicating upgrades.

cc: @pmorie

charts/federation-v2/templates/statefulset.yaml
```diff
@@ -28,6 +28,18 @@ spec:
...
+{{- if .Values.controllermanager.featureGates }}
+        - --feature-gates={{ .Values.controllermanager.featureGates }}
+{{- end }}
+{{- if .Values.controllermanager.fedNamespace }}
+        - --federation-namespace={{ .Values.controllermanager.fedNamespace }}
+{{- end }}
+{{- if .Values.controllermanager.clusterNamespace }}
+        - --registry-namespace={{ .Values.controllermanager.clusterNamespace }}
+{{- end }}
+{{- if .Values.controllermanager.limitedScope }}
```
marun: I think enabling limited scope implies setting the federation and registry namespaces to the namespace the controller manager is deployed in (e.g. [1]).  Also, if limitedScope is true, the service account that the controller manager runs under should be restricted to the federation namespace.


1: https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/generate-namespaced-install-yaml.sh#L39

charts/federation-v2/README.md
```diff
@@ -83,13 +83,17 @@ chart and their default values.
...
+| controllermanager.repository | Repo of the federation v2 image. | quay.io/kubernetes-multicluster |
+| controllermanager.image | Name of the federation v2 image. | federation-v2 |
+| controllermanager.tag | Tag of the federation v2 image. | latest |
+| controllermanager.pullPolicy | Image pull policy. | IfNotPresent |
+| controllermanager.featureGates | Feature gates are a set of `key=value` pairs that describe alpha or experimental features. An administrator can use the `--feature-gates` command line flag on each component to turn a feature on or off. | PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true |
+| controllermanager.fedNamespace | The namespace the federation control plane is deployed in. | federation-system |
+| controllermanager.clusterNamespace | The cluster registry namespace. | kube-multicluster-public |
+| controllermanager.clusterNamespace | Whether the federation namespace will be the only target for federation. | false |
+| clusterregistry.enabled | Specifies whether to enable the clusterregistry in federation v2. | true |
```
marun: What does this flag imply?  Since 'cluster registry' is primarily about installing the CRD, can't this be done implicitly (i.e. if crd isn't present, install it)?  Maybe the only flag should be for creating the `kube-multicluster-public` namespace?

charts/federation-v2/templates/statefulset.yaml
gyliu513: I see, let me update.

charts/federation-v2/README.md
marun: If the CRD is installed, great.  If it is not installed, it needs to be installed.  I think this behavior should be automatic, and not controlled by a flag.

For a cluster-scoped federation deployment, `kube-multicluster-public` is the default registry namespace and federation will fail to work if it does not exist.  I think this implies that the chart should create it.

charts/federation-v2/README.md
marun: I think the folder name should change, then.  As per my previous comment, 'enabling' federated types is an ambiguous action.

gyliu513: @marun I was treating the `federatedtypes` as a sub chart and helm usually use noun for a chart name, but rename the folder to `installfederatedtypes` as sub chart name seems a bit strange?

charts/federation-v2/README.md
gyliu513: @marun If the CRD already exist, and when the helm chart include the CRD and if install the CRD again, the `helm install` will be failed, so seems we need a flag to control this? How to make this behavior automatic?

I will try to see if I can create the namespace in helm chart, but how about handle the namespace create in helm chart in another PR?

charts/federation-v2/templates/statefulset.yaml
gyliu513: Done @marun

charts/federation-v2/README.md
marun: I'm hoping helm is capable enough to be able to check for the CRD and install if needed rather than forcing the user to perform that mechanical action.  Is that not the case?

I'm fine with the namespace create in a separate PR.

charts/federation-v2/README.md
marun: Nevermind, I don't think this will matter in light of #60, which implies that installation of federated types is likely to go away in favor of automation.

marun:
I'm concerned that this PR modifies a helm chart for which we have no CI testing.  At what point will this chart be being integrated into the deployment script?

charts/federation-v2/README.md
gyliu513: After some check, I think helm do not have such capability now, so here I was using some condition checking. But I will check more for this in helm community.

gyliu513:
> I'm concerned that this PR modifies a helm chart for which we have no CI testing. At what point will this chart be being integrated into the deployment script?

I will fix this after 0.0.2 got released. For now, I want to make sure the helm chart works fine with 0.0.2 release.

gyliu513:
@shashidharatd @marun all comments are now addressed, can you please help check if we can have this merged?

charts/federation-v2/templates/role.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.controllermanager.limitedScope }}
```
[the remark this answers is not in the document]
marun: Fixed in #299.

marun:
Please squash the fixup PR and then I think this is ready.

gyliu513:
@marun done.

PR #216, opened 2018-08-20
WIP: Install namespaced control plane

marun:
**NOTE**: This incorporates the outstanding PRs from #197 and updates the deployment scripting on top of them.  It is not intended for review at this time.

Targets #197

PR #285, opened 2018-09-24
Fix controller manager to only accept args from command line

marun:
As per a [PR comment](https://github.com/kubernetes-sigs/federation-v2/pull/216#discussion_r212668792) from @pmorie, the controller manager should accept arguments set via the downward api from the command line rather than as environment variables.

### DIFF
--- a/scripts/generate-namespaced-install-yaml.sh
+++ b/scripts/generate-namespaced-install-yaml.sh
@@ -32,8 +32,11 @@ sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
 # Convert rbac from cluster- to namespace-scoped
 sed -i -e 's/ClusterRole/Role/' "${INSTALL_YAML}"
 
-# Add --limited-scope arg to container
-sed -i -e 's/\(\s*- \)\(--install-crds=false\)/\1\2\n\1--limited-scope=true/' "${INSTALL_YAML}"
+# Add args to container
+sed -i -e '/--install-crds=false/a\
+        - --limited-scope=true\
+        - --federation-namespace=$(FEDERATION_NAMESPACE)\
+        - --registry-namespace=$(CLUSTER_REGISTRY_NAMESPACE)' "${INSTALL_YAML}"
 
 # Add namespace env args to container
  sed -i -e '/terminationGracePeriodSeconds/i\
--- a/scripts/generate-namespaced-install-yaml.sh
+++ b/scripts/generate-namespaced-install-yaml.sh
@@ -36,7 +36,8 @@ sed -i -e 's/ClusterRole/Role/' "${INSTALL_YAML}"
 sed -i -e 's/\(\s*- \)\(--install-crds=false\)/\1\2\n\1--limited-scope=true/' "${INSTALL_YAML}"
 
 # Add namespace env args to container
- sed -i -e '/terminationGracePeriodSeconds/i\        env:\
+ sed -i -e '/terminationGracePeriodSeconds/i\
+        env:\
         - name: FEDERATION_NAMESPACE\
           valueFrom:\
             fieldRef:\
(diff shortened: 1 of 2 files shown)


scripts/generate-namespaced-install-yaml.sh
```diff
@@ -32,11 +32,14 @@ sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
 # Convert rbac from cluster- to namespace-scoped
 sed -i -e 's/ClusterRole/Role/' "${INSTALL_YAML}"
 
-# Add --limited-scope arg to container
-sed -i -e 's/\(\s*- \)\(--install-crds=false\)/\1\2\n\1--limited-scope=true/' "${INSTALL_YAML}"
+# Add args to container
+sed -i -e '/--install-crds=false/a\
+        - --federation-namespace=FEDERATION_NAMESPACE\
```
pmorie: Should this be `$FEDERATION_NAMESPACE` ?

scripts/generate-namespaced-install-yaml.sh
```diff
@@ -32,11 +32,15 @@ sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
...
+sed -i -e '/--install-crds=false/a\
+        - --limited-scope=true\
+        - --federation-namespace=$(FEDERATION_NAMESPACE)\
+        - --registry-namespace=$(CLUSTER_REGISTRY_NAMESPACE)' "${INSTALL_YAML}"
 
 # Add namespace env args to container
- sed -i -e '/terminationGracePeriodSeconds/i\        env:\
+ sed -i -e '/terminationGracePeriodSeconds/i\
+        env:\
         - name: FEDERATION_NAMESPACE\
```
gyliu513: Why still update env for `FEDERATION_NAMESPACE` and `CLUSTER_REGISTRY_NAMESPACE` to `install-namespaced.yaml`?

scripts/generate-namespaced-install-yaml.sh
```diff
@@ -32,11 +32,15 @@ sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
 # Convert rbac from cluster- to namespace-scoped
 sed -i -e 's/ClusterRole/Role/' "${INSTALL_YAML}"
 
-# Add --limited-scope arg to container
-sed -i -e 's/\(\s*- \)\(--install-crds=false\)/\1\2\n\1--limited-scope=true/' "${INSTALL_YAML}"
+# Add args to container
+sed -i -e '/--install-crds=false/a\
+        - --limited-scope=true\
+        - --federation-namespace=$(FEDERATION_NAMESPACE)\
```
gyliu513: I think https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/deploy-federation.sh#L130 also need some update to set those two variables:  `FEDERATION_NAMESPACE` and `CLUSTER_REGISTRY_NAMESPACE `.

scripts/generate-namespaced-install-yaml.sh
marun: The goal is setting these environment variables to the namespace the pod is running in (i.e. via the downward api).

Reference: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables

scripts/generate-namespaced-install-yaml.sh
marun: These variables are being set from the pod's namespace, and the pod's namespace is determined by the [namespace the deployment script applies the yaml to](https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/deploy-federation.sh#L139).

gyliu513: I see, thanks @marun

PR #286, opened 2018-09-24
Use namespace sync controller in the dns tests

marun:
Previously the dns tests created (and attempted to cleanup) the test namespace in member clusters.  This commit updates the tests to rely instead on the namespace sync controller to ensure resiliency.

This change was motivated by test failures noticed while working on #284.

### DIFF
--- a/test/e2e/crud.go
+++ b/test/e2e/crud.go
@@ -42,23 +42,7 @@ var _ = Describe("Federated types", func() {
 
 	tl := framework.NewE2ELogger()
 
-	typeConfigs, err := common.FederatedTypeConfigs()
-	if err != nil {
-		tl.Fatalf("Error loading type configs: %v", err)
-	}
-
-	// Find the namespace typeconfig to be able to start its sync
-	// controller as needed.
-	var namespaceTypeConfig typeconfig.Interface
-	for _, typeConfig := range typeConfigs {
-		if typeConfig.GetTarget().Kind == util.NamespaceKind {
-			namespaceTypeConfig = typeConfig
-			break
-		}
-	}
-	if namespaceTypeConfig == nil {
-		tl.Fatalf("Unable to find namespace type config")
-	}
+	typeConfigs := common.TypeConfigsOrDie(tl)
 
 	for i, _ := range typeConfigs {
 		// Bind the type config inside the loop to ensure the ginkgo
@@ -86,7 +70,7 @@ var _ = Describe("Federated types", func() {
 				testObjectFunc := func(namespace string, clusterNames []string) (template, placement, override *unstructured.Unstructured, err error) {
 					return common.NewTestObjects(typeConfig, namespace, clusterNames)
 				}
-				validateCrud(f, tl, typeConfig, namespaceTypeConfig, testObjectFunc)
+				validateCrud(f, tl, typeConfig, testObjectFunc)
 			})
 		})
 	}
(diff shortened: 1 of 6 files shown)


test/e2e/crud.go
```diff
@@ -42,23 +42,8 @@ var _ = Describe("Federated types", func() {
...
-		if typeConfig.GetTarget().Kind == util.NamespaceKind {
-			namespaceTypeConfig = typeConfig
-			break
-		}
-	}
-	if namespaceTypeConfig == nil {
-		tl.Fatalf("Unable to find namespace type config")
-	}
+	typeConfigs := common.TypeConfigsOrDie(tl)
+	namespaceTypeConfig := common.NamespaceTypeConfigOrDie(tl)
```
font: It looks like this line is no longer needed since `validateCrud()` no longer sets up the namespace controller fixture directly.

test/e2e/crud.go
```diff
@@ -260,13 +245,6 @@ func waitForCrd(pool dynamic.ClientPool, tl common.TestLogger, apiResource metav
 func validateCrud(f framework.FederationFramework, tl common.TestLogger, typeConfig, namespaceTypeConfig typeconfig.Interface, testObjectFunc testObjectAccessor) {
```
font: We can also remove the `namespaceTypeConfig` parameter.

test/e2e/framework/framework.go
```diff
@@ -176,8 +178,20 @@ func createNamespace(client kubeclientset.Interface, baseName string) (string, e
...
+}
+
+func (f *frameworkWrapper) setUpNamespaceControllerFixture(typeConfig typeconfig.Interface) {
+	if TestContext.LimitedScope || (typeConfig != nil && typeConfig.GetTarget().Kind == util.NamespaceKind) {
+		return
+	}
+	// The namespace controller is required to ensure namespaces
+	// are created as needed in member clusters in advance of
+	// propagation of other namespaced types.
+	f.SetUpControllerFixture(common.NamespaceTypeConfigOrDie(NewE2ELogger()))
```
font: Does this pose implications on the managed e2e testing? It seems like we also need a check for `TestContext.TestManagedFederation` as namespaces using a test managed federation don't work without a k8s namespace controller.

marun: Given that e2e tests run fine with managed fixture, and namespace sync controllers were being started for non-namespace sync controllers prior to this PR, it doesn't appear that there is a problem.  

Crudtester validation of namespace deletion doesn't work for managed fixture, as would be expected when the k8s namespace controller is missing.  But for managed fixture, I don't think namespaces need to be deleted since the fixture is not persistent.

test/e2e/framework/framework.go
```diff
@@ -176,8 +178,20 @@ func createNamespace(client kubeclientset.Interface, baseName string) (string, e
...
+	f.setUpNamespaceControllerFixture(nil)
 }
 
 func (f *frameworkWrapper) SetUpIngressDNSControllerFixture() {
 	f.framework().SetUpIngressDNSControllerFixture()
+	f.setUpNamespaceControllerFixture(nil)
+}
+
+func (f *frameworkWrapper) setUpNamespaceControllerFixture(typeConfig typeconfig.Interface) {
+	if TestContext.LimitedScope || (typeConfig != nil && typeConfig.GetTarget().Kind == util.NamespaceKind) {
```
font: (no action required) consider adding a comment that this check helps prevent infinite recursion.

PR #290, opened 2018-09-26
Fine-grained deploy for cluster scoped federation controller.

gyliu513:
This PR enables end user can select the namespace that want to use
when deploying a cluster scoped federation controller.

/cc @marun

### DIFF
--- a/scripts/delete-federation.sh
+++ b/scripts/delete-federation.sh
@@ -36,9 +36,9 @@ else
   USE_LATEST=
 fi
 
-KF_NS_ARG=
+KF_NS_ARG="--federation-namespace=${NS} "
 if [[ "${NAMESPACED}" ]]; then
-  KF_NS_ARG="--federation-namespace=${NS} --registry-namespace=${NS}"
+  KF_NS_ARG+="--registry-namespace=${NS}"
 fi
 
 # Unjoin clusters by removing objects added by kubefed2.
@@ -65,13 +65,14 @@ done
 if [[ ! "${USE_LATEST}" ]]; then
   if [[ "${NAMESPACED}" ]]; then
     ${KCD} -n "${NS}" -f hack/install-namespaced.yaml
-    ${KCD} ns "${NS}"
   else
     ${KCD} -f hack/install.yaml
   fi
 else
   ${KCD} -f hack/install-latest.yaml
 fi
+
+${KCD} ns "${NS}"
 
 # Remove permissive rolebinding that allows federation controllers to run.
 if [[ "${NAMESPACED}" ]]; then
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -61,9 +61,9 @@ else
   INSTALL_YAML=hack/install.yaml
 fi
 
-KF_NS_ARGS=
+KF_NS_ARGS="--federation-namespace=${NS} "
 if [[ "${NAMESPACED}" ]]; then
-  KF_NS_ARGS="--federation-namespace=${NS} --registry-namespace=${NS} --limited-scope=true"
+  KF_NS_ARGS+="--registry-namespace=${NS} --limited-scope=true"
   INSTALL_YAML=hack/install-namespaced.yaml
 fi
 
@@ -106,11 +106,11 @@ if [[ ! "${USE_LATEST}" ]]; then
   ${DOCKER_PUSH_CMD}
 fi
 
-if [[ "${NAMESPACED}" ]]; then
-  if ! kubectl get ns "${NS}" > /dev/null 2>&1; then
-    kubectl create ns "${NS}"
-  fi
-else
+if ! kubectl get ns "${NS}" > /dev/null 2>&1; then
+  kubectl create ns "${NS}"
+fi
+
+if [[ ! "${NAMESPACED}" ]]; then
   if ! kubectl get ns "${PUBLIC_NS}" > /dev/null 2>&1; then
     kubectl create ns "${PUBLIC_NS}"
   fi
@@ -129,7 +129,7 @@ if [[ ! "${USE_LATEST}" ]]; then
   if [[ "${NAMESPACED}" ]]; then
     INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-namespaced-install-yaml.sh
   else
-    INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-install-yaml.sh
+    INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" FEDERATION_NAMESPACE="${NS}" scripts/generate-install-yaml.sh
   fi
 fi
 
--- a/scripts/generate-install-yaml.sh
+++ b/scripts/generate-install-yaml.sh
@@ -20,6 +20,7 @@ set -o pipefail
 
 INSTALL_YAML="${INSTALL_YAML:-hack/install-latest.yaml}"
 IMAGE_NAME="${IMAGE_NAME:-quay.io/kubernetes-multicluster/federation-v2:latest}"
+FEDERATION_NAMESPACE="${FEDERATION_NAMESPACE:-""}"
 
 kubebuilder create config --controller-image "${IMAGE_NAME}" \
             --name federation --output "${INSTALL_YAML}"
@@ -27,3 +28,19 @@ kubebuilder create config --controller-image "${IMAGE_NAME}" \
 # Increase memory request and limit to avoid OOM issues.
 sed -i 's/memory: 20Mi/memory: 64Mi/' "${INSTALL_YAML}"
 sed -i 's/memory: 30Mi/memory: 128Mi/' "${INSTALL_YAML}"
+
+# Remove namespace resource from the top of the file
+sed -i -e '/---/,$!d' "${INSTALL_YAML}"
+
+# Remove namespace fields
+sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
+
+if [[ "${FEDERATION_NAMESPACE}" ]]; then
+  # Add namespace to ClusterRoleBinding
+  sed -i -e '/- kind: ServiceAccount/a\
+  namespace: '${FEDERATION_NAMESPACE}'' "${INSTALL_YAML}"
+
+  # Add args to container
+  sed -i -e '/--install-crds=false/a\
+        - --federation-namespace='${FEDERATION_NAMESPACE}'' "${INSTALL_YAML}"
+fi
--- a/scripts/generate-namespaced-install-yaml.sh
+++ b/scripts/generate-namespaced-install-yaml.sh
@@ -21,13 +21,7 @@ set -o pipefail
 INSTALL_YAML="${INSTALL_YAML:-hack/install-namespaced.yaml}"
 IMAGE_NAME="${IMAGE_NAME:-quay.io/kubernetes-multicluster/federation-v2:latest}"
 
-INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-install-yaml.sh
-
-# Remove namespace resource from the top of the file
-sed -i -e '/---/,$!d' "${INSTALL_YAML}"
-
-# Remove namespace fields
-sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
+INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" FEDERATION_NAMESPACE= scripts/generate-install-yaml.sh
 
 # Convert rbac from cluster- to namespace-scoped
 sed -i -e 's/ClusterRole/Role/' "${INSTALL_YAML}"


gyliu513:
Did some test for both cluster scoped and namespace scoped federation.

- Cluster Scoped
```
NAMESPACED=y FEDERATION_NAMESPACE=test-namespace scripts/deploy-federation.sh  gyliu/federation-v2:canary cluster2
NAMESPACED=y FEDERATION_NAMESPACE=test-namespace ./scripts/delete-federation.sh
```

- Namespace Scoped
```
FEDERATION_NAMESPACE=test-namespace scripts/deploy-federation.sh  gyliu/federation-v2:canary cluster2
FEDERATION_NAMESPACE=test-namespace ./scripts/delete-federation.sh
```

scripts/generate-install-yaml.sh
```diff
@@ -27,3 +27,29 @@ kubebuilder create config --controller-image "${IMAGE_NAME}" \
...
+
+# Remove namespace resource from the top of the file
+sed -i -e '/---/,$!d' "${INSTALL_YAML}"
+
+# Remove namespace fields
+sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
+
+# Add namespace to ClusterRoleBinding
+sed -i -e '/- kind: ServiceAccount/a\
+  namespace: ${NS}' "${INSTALL_YAML}"
```
gyliu513: I need to update here to make the namespace can be replaced, will update later.

scripts/generate-install-yaml.sh
```diff
@@ -27,3 +27,31 @@ kubebuilder create config --controller-image "${IMAGE_NAME}" \
...
+  sed -i -e '/- kind: ServiceAccount/a\
+  namespace: '${NS}'' "${INSTALL_YAML}"
+fi
+
+# Add args to container
+sed -i -e '/--install-crds=false/a\
+        - --federation-namespace=$(FEDERATION_NAMESPACE)' "${INSTALL_YAML}"
+
+# Add namespace env args to container
+sed -i -e '/terminationGracePeriodSeconds/i\
```
marun: Why is it desirable to add these args to the non-namespaced install yaml?

scripts/generate-install-yaml.sh
```diff
@@ -27,3 +27,31 @@ kubebuilder create config --controller-image "${IMAGE_NAME}" \
...
 sed -i 's/memory: 20Mi/memory: 64Mi/' "${INSTALL_YAML}"
 sed -i 's/memory: 30Mi/memory: 128Mi/' "${INSTALL_YAML}"
+
+# Remove namespace resource from the top of the file
+sed -i -e '/---/,$!d' "${INSTALL_YAML}"
+
+# Remove namespace fields
+sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
+
+if [ "${NS}" ]; then
```
marun: Please define this at the top of the script and set a default, as per the example of `INSTALL_YAML` and `IMAGE_NAME`, to ensure that this script outputs a reasonable default when called directly.  Also, avoid an abbreviation like `NS` for a variable that is passed in from the environment rather than used internally (i.e. use `FEDERATION_NAMESPACE` as the name of the environment variable as per the deployment script).

scripts/generate-install-yaml.sh
gyliu513: Fixed

scripts/generate-install-yaml.sh
gyliu513: As we need `FEDERATION_NAMESPACE` to be specified in L45, so that when a cluster join federation controller, it can join to different namespaces and federation controller can get those federationclusters from this namespace.

We do not need `CLUSTER_REGISTRY_NAMESPACE` here, will move it to `generate-namespaced-install-yaml.sh`.

scripts/generate-install-yaml.sh
gyliu513: Will fix in update.

gyliu513: @marun I think https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/deploy-federation.sh#L50 also need update?

```
NS="${FEDERATION_NAMESPACE:-federation-system}"
```

Do you think we also need to update above as following? If so, I intend to take the re-name in another PR.

```
FEDERATION_NAMESPACE="${FEDERATION_NAMESPACE:-federation-system}"
```

marun: It's fine to use the `NS` name internally so long as the external name (`FEDERATION_NAMESPACE`) is verbose.

scripts/generate-install-yaml.sh
marun: The only reason for these args is to allow the controller args to be set by the downward api.  In the non-namespaced case, `--federation-namespace` can be hardcoded (e.g. `--federation-namespace=${NS}`).

gyliu513: Done.

One question @marun , can you please show some use case for the downward api? Why we need this for namespace scoped federation controller? Thanks.

scripts/delete-federation.sh
```diff
@@ -56,16 +57,17 @@ done
 # Remove federation CRDs, namespace, RBAC and deployment resources.
 if [[ "${NAMESPACED}" ]]; then
   ${KCD} -n "${NS}" -f hack/install-namespaced.yaml
-  ${KCD} ns "${NS}"
 else
   ${KCD} -f hack/install-latest.yaml
```
gyliu513: Here need some update, we cannot always delete `hack/install-latest.yaml`, if we are not using `latest` but customer build docker image, we should delete `hack/install.yaml`, I will handle this in a separate PR.

scripts/generate-install-yaml.sh
marun: When the control plane is deployed namespaced, the downward api can be used to retrieve the namespace that the control manager pod is running in so the sync controllers can be configured to target that namespace.  Without this capability, it would not be possible for static yaml to deploy a namespaced control plane to an arbitrary namespace.

scripts/delete-federation.sh
```diff
@@ -28,15 +28,16 @@ NS="${FEDERATION_NAMESPACE:-federation-system}"
...
 if [[ "${NAMESPACED}" ]]; then
-  KF_NS_ARG="--federation-namespace=${NS} --registry-namespace=${NS}"
+  KF_NS_ARG+="--registry-namespace=${NS}"
 fi
 
 # Unjoin clusters by removing objects added by kubefed2.
 HOST_CLUSTER="$(kubectl config current-context)"
 JOINED_CLUSTERS="$(kubectl -n "${NS}" get federatedclusters -o=jsonpath='{range .items[*]}{.metadata.name}{" "}{end}')"
 for c in ${JOINED_CLUSTERS}; do
+  echo ./bin/kubefed2 unjoin "${c}" --host-cluster-context "${HOST_CLUSTER}" --remove-from-registry --v=2 ${KF_NS_ARG}
```
marun: Even if this were desirable (and I don't think it is - a user can use `bash -x` to invoke the script if they want to debug), duplicating a command entirely like this represents a maintenance burden.  Please remove.

scripts/deploy-federation.sh
```diff
@@ -122,14 +122,14 @@ if [[ "${NAMESPACED}" ]]; then
...
-  kubectl create clusterrolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
+  kubectl create clusterrolebinding federation-admin-${NS} --clusterrole=cluster-admin --serviceaccount="${NS}:default"
 fi
 
 if [[ ! "${USE_LATEST}" ]]; then
   if [[ "${NAMESPACED}" ]]; then
     INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-namespaced-install-yaml.sh
   else
-    INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-install-yaml.sh
+    INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" NS="${NS}" scripts/generate-install-yaml.sh
```
marun: Just like in any programming langauge, it's fine to use a short name internally but the API exposed to the user should be appropriately descriptive.  That suggests `FEDERATION_NAMESPACE` as the environment variable the generation script accepts, not `NS`.

scripts/deploy-federation.sh
```diff
@@ -154,9 +154,10 @@ done
 # Join the host cluster
 go build -o bin/kubefed2 cmd/kubefed2/kubefed2.go
 CONTEXT="$(kubectl config current-context)"
+echo ./bin/kubefed2 join "${CONTEXT}" --host-cluster-context "${CONTEXT}" --add-to-registry --v=2 ${KF_NS_ARGS}
```
marun: ditto

scripts/deploy-federation.sh
```diff
@@ -154,9 +154,10 @@ done
...
 ./bin/kubefed2 join "${CONTEXT}" --host-cluster-context "${CONTEXT}" --add-to-registry --v=2 ${KF_NS_ARGS}
 
 for c in ${JOIN_CLUSTERS}; do
+  echo ./bin/kubefed2 join "${c}" --host-cluster-context "${CONTEXT}" --add-to-registry --v=2 ${KF_NS_ARGS}
```
marun: ditto

scripts/deploy-federation.sh
```diff
@@ -154,9 +154,10 @@ done
...
   ./bin/kubefed2 join "${c}" --host-cluster-context "${CONTEXT}" --add-to-registry --v=2 ${KF_NS_ARGS}
-
```
marun: Unrelated non-functional change, please remove.

scripts/generate-install-yaml.sh
```diff
@@ -20,10 +20,33 @@ set -o pipefail
...
+
+# Remove namespace resource from the top of the file
+sed -i -e '/---/,$!d' "${INSTALL_YAML}"
+
+# Remove namespace fields
+sed -i -e '/^  namespace: federation-system$/d' "${INSTALL_YAML}"
+
+if [ "${FEDERATION_NAMESPACE}" ]; then
+  # Update name for ClusterRole
+  sed -i -e 's/name: federation-role$/name: federation-role-'${FEDERATION_NAMESPACE}'/' "${INSTALL_YAML}"
```
marun: Why is it desirable to name the cluster role and role binding with the namespace?  A cluster-scoped federation can only be deployed once to a cluster, so finding a name clash would be a good sign of cleanup required rather than a problem.

scripts/generate-install-yaml.sh
```diff
@@ -20,10 +20,33 @@ set -o pipefail
...
+
+  # Update name for ClusterRoleBinding
+  sed -i -e 's/name: federation-rolebinding$/name: federation-rolebinding-'${FEDERATION_NAMESPACE}'/' "${INSTALL_YAML}"
+
+  # Add namespace to ClusterRoleBinding
+  sed -i -e '/- kind: ServiceAccount/a\
+  namespace: '${FEDERATION_NAMESPACE}'' "${INSTALL_YAML}"
```
marun: Why is this desirable for a namespaced deployment?  The removal of lines 29-30 in `scripts/generate-namespaced-install-yaml.sh` will preclude this field from being removed.

scripts/generate-install-yaml.sh
```diff
@@ -20,10 +20,33 @@ set -o pipefail
...
+
+  # Add args to container
+  sed -i -e '/--install-crds=false/a\
+        - --federation-namespace='${FEDERATION_NAMESPACE}'' "${INSTALL_YAML}"
```
marun: The addition of this arg suggests not adding the script again in the namespaced generation script.

scripts/generate-install-yaml.sh
gyliu513: Deleted.

scripts/generate-install-yaml.sh
gyliu513: This is because clusterrolebindings request a namespace, please refer to https://github.com/kubernetes-sigs/federation-v2/pull/146

scripts/delete-federation.sh
gyliu513: done

scripts/generate-install-yaml.sh
gyliu513: This is only for cluster scoped federation.

marun:
Ah, I missed the point of the namespaced generation script not supplying the `FEDERATION_NAMESPACE` arg to the non-namespaced generation script, so that that block wouldn't execute.

I think this will be ready once the namespace suffix on the cluster role and binding are removed.

marun:
Also, please squash the 'fixed' commits.

gyliu513:
@marun  done.

scripts/delete-federation.sh
```diff
@@ -56,16 +56,17 @@ done
...
 fi
 
+${KCD} ns "${NS}"
+
 # Remove permissive rolebinding that allows federation controllers to run.
 if [[ "${NAMESPACED}" ]]; then
   ${KCD} -n "${NS}" rolebinding federation-admin
 else
-  ${KCD} clusterrolebinding federation-admin
+  ${KCD} clusterrolebinding federation-admin-${NS}
```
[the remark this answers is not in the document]
gyliu513: Done

scripts/deploy-federation.sh
```diff
@@ -122,14 +122,14 @@ if [[ "${NAMESPACED}" ]]; then
   kubectl -n "${NS}" create rolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
 else
   # TODO(marun) Make this more restrictive.
-  kubectl create clusterrolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
+  kubectl create clusterrolebinding federation-admin-${NS} --clusterrole=cluster-admin --serviceaccount="${NS}:default"
```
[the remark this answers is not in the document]
gyliu513: Done

scripts/deploy-federation.sh
```diff
@@ -122,14 +122,14 @@ if [[ "${NAMESPACED}" ]]; then
...
 fi
 
 if [[ ! "${USE_LATEST}" ]]; then
   if [[ "${NAMESPACED}" ]]; then
     INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-namespaced-install-yaml.sh
   else
-    INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" scripts/generate-install-yaml.sh
+    INSTALL_YAML="${INSTALL_YAML}" IMAGE_NAME="${IMAGE_NAME}" FEDERATION_NAMESPACE="${NS}" NAMESPACED="${NAMESPACED}" scripts/generate-install-yaml.sh
```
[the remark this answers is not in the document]
gyliu513: This command is already `!${NAMESPACED}`, the reason I need this is because the script `scripts/generate-namespaced-install-yaml.sh` also need to call `scripts/generate-install-yaml.sh`, so in `scripts/generate-install-yaml.sh`, I need to make sure the `--federation-namespace` was only added when `${FEDERATION_NAMESPACE}` set and  `!${NAMESPACED}`, make sense?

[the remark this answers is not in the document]
marun: Since the namespaced install does not set an explicit namespace (it is intended to be applied with `-n <namespace>`), the namespaced generation script will never supply `FEDERATION_NAMESPACE` to this script.  That suggests that supplying `NAMESPACED` is redundant.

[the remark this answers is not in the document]
gyliu513: Yes, the namespaced install yaml does not set an explicit namespace (it is intended to be applied with `-n <namespace>`), but the `-n <namespace>` was set by `FEDERATION_NAMESPACE` in `deploy-federation.sh`, am I missing anything? https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/userguide.md#automated-deployment-1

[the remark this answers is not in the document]
marun: The script I was referring to was `scripts/generate-install-yaml.sh` not `scripts/deploy-federation.sh`.  I'm suggesting that `scripts/generate-install-yaml.sh` should not accept `NAMESPACED`, and assume that if `FEDERATION_NAMESPACE` is supplied, that the configuration is not for a namespaced deployment.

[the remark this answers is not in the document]
gyliu513: If I remove `NAMESPACED` from `scripts/generate-install-yaml.sh`, then the following command will generate two `- --federation-namespace='${FEDERATION_NAMESPACE}'`.

```
NAMESPACED=y FEDERATION_NAMESPACE=<namespace> scripts/deploy-federation.sh <image name>
```

As `scripts/generate-namespaced-install-yaml.sh` also need to call `scripts/generate-install-yaml.sh`, if I do not check `NAMESPACED` in `scripts/generate-install-yaml.sh`, it will generate two `- --federation-namespace='${FEDERATION_NAMESPACE}'` in federation controller statefulset.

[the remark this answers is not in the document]
marun: When I apply [this diff](https://gist.github.com/marun/333e4b640f8f35adeaa4e4e14e03bb24), the generated yaml does not does not contain duplicate `--federation-namespace` arguments.  Am I missing something?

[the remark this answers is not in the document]
gyliu513: This is exactly the same diff that I want you test. Please run the command as follows, you will see there are two `- --federation-namespace=`.

1.
```
NAMESPACED=y FEDERATION_NAMESPACE=test-namespace scripts/deploy-federation.sh  gyliu/federation-v2:canary cluster2
```
2. 
After cluster2 and cluster1 joined, run 
```
kubectl edit sts -n test-namespace
```

You will see follows:
```
  spec:
      containers:
      - args:
        - --install-crds=false
        - --limited-scope=true
        - --federation-namespace=$(FEDERATION_NAMESPACE) <<<<<<
        - --registry-namespace=$(CLUSTER_REGISTRY_NAMESPACE)
        - --federation-namespace=test-namespace <<<<<<
        command:
```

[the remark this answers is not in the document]
marun: Ah, I was testing the generation separate from the deploy script.  I forgot that environment variables passed on the command line are inherited by child processes, so that the `FEDERATION_NAMESPACE` passed to the deploy script would end up in the yaml generation script 3 calls deep.  I would rather see the call in the namespaced yaml generation script [clear the env variable](https://gist.github.com/marun/333e4b640f8f35adeaa4e4e14e03bb24#file-ns-diff-L44) rather than adding the `NAMESPACE` control variable.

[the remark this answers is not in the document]
gyliu513: @marun thanks, updated.

marun:
Thank you!

/approve
/lgtm

Issue #271, opened 2018-09-18
We should publish a versioned kubefed2 binary


pmorie:
Currently, we rely on users building the `kubefed2` binary themselves. Long term, as we begin releasing additional versions, we should publish a version-specific binary somehow. We could publish a distinct container image for kubefed2 or add the kubefed2 binary to the existing federation-v2 image.

gyliu513:
I'm +1 to add the kubefed2 binary to the existing federation-v2 image, as this simplified the release, and end user can `kubectl cp` to get the kubefed2 from federation controller.

marun:
@gyliu513 Given that it's trivial to add a cli binary like kubefed2 to a github release, and that's the way other kube client binaries (e.g. kubectl) are released, I'd hold off on putting kubefed2 into an image just yet.  If/when that becomes a desirable thing, I think it would make sense to ship a separate image to avoid bloating the server image (since kubefed2 is inexplicably 40mb).

gyliu513:
@marun I'm fine with this. The reason that I prefer shipping the binary into a docker image because this will help release federation v2 with only one docker image.

We already have some best practise with [IBM Cloud Private](https://www.ibm.com/cloud/private), here I want to show [an example](https://www.ibm.com/support/knowledgecenter/SSBS6K_2.1.0.3/app_center/create_helm_cli.html) here: We are also shipping `helm` in a docker image and end user can get the `helm` binary via `docker run` command like
```
docker run -e LICENSE=accept --net=host -v /usr/local/bin:/data ibmcom/icp-helm-api:1.0.0 cp /usr/src/app/public/cli/linux-amd64/helm /data
```

Agree with what you have mentioned,  it's trivial to add a cli binary like kubefed2 to a github release, so we can leave this as it is for now, and try to see if we need to ship with a docker image in future.

marun:
A kubefed2 binary is included in each release, which I think satisfies the main requirement in this issue.  If/when it becomes desirable to ship the client binary in the server image that issue can be raised separately.

PR #294, opened 2018-09-27
WIP Added kubefed2 to federation docker image.

gyliu513:
Fixed part of https://github.com/kubernetes-sigs/federation-v2/issues/271

/cc @pmorie @marun 

User guide also need some update to enable customer get `kubefed2` from federation controller container via `kubectl cp`.


marun:
@gyliu513 Please close, as per discussion on #271.

PR #295, opened 2018-09-27
Enabled delete federation can delete hack/install.yaml

gyliu513:
There are now three types for install federation:
- install-latest.yaml
- install.yaml
- install-namespace.yaml

We need to make sure the delete script can handle all of the cases.

/cc @marun @font 

This is a follow up for PR https://github.com/kubernetes-sigs/federation-v2/pull/290#pullrequestreview-159342381

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -325,14 +325,12 @@ Run the following command to perform a cleanup of the cluster registry and
 federation deployments:
 
 ```bash
-./scripts/delete-federation.sh cluster2
+./scripts/delete-federation.sh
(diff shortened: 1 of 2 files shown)


marun:
The deletion script should not need to know the image name.  Consider an alternate approach for determining what mode of deletion is appropriate.

gyliu513:
> The deletion script should not need to know the image name. Consider an alternate approach for determining what mode of deletion is appropriate.

I was now using `kubectl get sts` to get the image being used and compare with latest to check.

PR #296, opened 2018-09-27
Refactor all controllers to use the same async worker code

marun:
This was motivated by needing the same code for #276.  Likely there will be further convergence in the future if not a switch to kubebuilder-based controllers if we can work with them to get the features we need.


shashidharatd:
This looks clean and good from my side. would wait for others to lgtm.

Issue #297, opened 2018-09-28
Namespace sync controller is run when --limited-scope=true


font:
With `--limited-scope=true` (namespace-scoped deployment) the federation control plane operates on a single namespace. As a result, the `Namespace` sync controller is not necessary as the creation and deletion of the namespace is not up to the federation controller, nor does it have the permissions necessary for it. There is a check for this in the testing code, but a check needs to be added to the main path.

gyliu513:
@font I think we can close this one as namespace sync controller will not run after PR #299

font:
Fixed by #299.

PR #299, opened 2018-09-28
Avoid starting a namespace sync controller for namespaced federation

marun:
@font (offline) and @gyliu513 (in a [comment](https://github.com/kubernetes-sigs/federation-v2/pull/283#discussion_r221222183)) both noticed that a sync controller was being launched for namespaces when the control plane was running namespaced.  This isn't desirable since a namespaced control plane shouldn't require permission to list namespaces.

Previously the only check against running a sync controller for namespaces was in the deployment script not applying the type config for namespaces if deploying a namespaced control plane.  Since helm deployment is now an option, it's necessary to move the check to code so it can be shared.

### DIFF
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -144,10 +144,6 @@ kubectl apply --validate=false -f vendor/k8s.io/cluster-registry/cluster-registr
 
 # Enable available types
 for filename in ./config/federatedtypes/*.yaml; do
-  # The namespace controller is not required when the control plane is namespaced
-  if [[ "${NAMESPACED}" && "${filename}" == *namespace.yaml ]]; then
-    continue
-  fi
   kubectl -n "${NS}" apply -f "${filename}"
 done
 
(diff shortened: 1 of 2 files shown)


scripts/deploy-federation.sh
```diff
@@ -144,10 +144,6 @@ kubectl apply --validate=false -f vendor/k8s.io/cluster-registry/cluster-registr
 
 # Enable available types
 for filename in ./config/federatedtypes/*.yaml; do
-  # The namespace controller is not required when the control plane is namespaced
-  if [[ "${NAMESPACED}" && "${filename}" == *namespace.yaml ]]; then
```
font: Should we leave this in? It may help avoid confusing users thinking there's namespace push reconciliation support for namespace-scoped deployment without having to look at the logs.

font:
Fix for #297.

scripts/deploy-federation.sh
gyliu513: +1 to leave this in, I will also update the helm chart by adding some condition check to avoid creating namespace controller in namespaced scope federation controller.

marun: DRY - I don't think it's reasonable to require that every deployment mechanism know not to apply a federated type config for namespaces when deploying namespaced federation.  If/when user confusion becomes an issue we can revisit.

PR #300, opened 2018-09-29
Move controller manager to a sub chart.

gyliu513:
This PR did two tasks:
1) Move controller manager to a sub chart.
2) Move out all of the federated types to the main chart, so that
we do not need a parameter to enable it. This resolved @marun's
concern of changing chart name to `installfederatedtype`.

/cc @marun @shashidharatd @kubernetes-sigs/federation-wg 

Any comments to this new topology?

NOTE: This needs to be rebased after https://github.com/kubernetes-sigs/federation-v2/pull/283 merged.


marun:
I appreciate your desire to address my concern with the variable name, but I'm not sure making the controller manager optional is desirable.  When would it ever be desirable to not install the controller manager, given that it is currently the primary component of federation?

gyliu513:
@marun Yes, when I propose this PR, I was also thinking the same. For now, there is no case that the end user can set controller manager as disabled, but we already set this as enabled by default.

Also for the current `federatedtypes` chart, the federation will not work if we do not install this chart? So seems `federatedtypes` and `controllermanager` are in same position: both of them are required when we want to enable federation v2.

I know that we are planning to deprecate federation types at #60 , but I think putting `controllermanager` as a sub chart should not impact much. I also take a reference for istio helm chart at https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/requirements.yaml#L20-L22 , `pilot` is a must install component for Istio, but it was also in a sub chart and controlled by a the `enabled` parameter, I think as long as we document this in README, this should be ok?

marun:
federatedtypes is not likely to go away, but it's use will be changing.

I don't think it's desirable to make a mandatory component optional in the chart without good reason.  Does istio have a good reason?

gyliu513:
@marun Sorry, I'm  not clear the history of why Istio putting the required component into a sub chart, but I'm proposing this PR mainly for some discussion and comments, if you agree the current direction, I'm OK to close this PR.

charts/federation-v2/charts/controllermanager/templates/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+    control-plane: controller-manager
+    kubebuilder.k8s.io: 1.0.3
+  name: federation-controller-manager-service
+  namespace: {{ .Release.Namespace }}
+spec:
+  clusterIP: None
+  selector:
+    api: federation
+    control-plane: controller-manager
+    kubebuilder.k8s.io: 1.0.3
```
pmorie: Is the kubebuilder version a relevant part of this service's spec?

charts/federation-v2/README.md
```diff
@@ -83,13 +83,13 @@ chart and their default values.
 
 | Parameter | Description | Default |
 |-----------|-------------|---------|
+| controllermanager.enabled | Specifies whether to enable the controller manager in federation v2 | true |
```
pmorie: @marun 

A use-case for not installing the controller-manager might be that you like the federation APIs but want to write your own controller, for example if you were prototyping a new propagation mechanism. I don't think it's harmful to allow this in the chart, since doing so does not drastically effect the PR and the default is to enable it.

charts/federation-v2/charts/controllermanager/templates/service.yaml
gyliu513: Yes, I just copy the service spec from the generated `install-latest.yaml`.

charts/federation-v2/README.md
gyliu513: @pmorie thanks for the comments, this is exactly a use-case for the helm chart refactor.

@font Yes, I want to put controller manager chart and federation API CRDs into same chart, but due to there maybe same name files, like `service.yaml`, I intend to put them to different charts.

So there are three options here for the helm chart

- Original Topology:

federation-v2
   |
   | -- charts
   |           |
   |           | -- clusterregistry
   |           |
   |           | -- federatedtypes
   |
   | -- controllermanager

- New Topology 1:

federation-v2
   |
   | -- charts
   |           |
   |           | -- clusterregistry
   |           |
   |           | -- controllermanager
   |
   | -- federatedtypes

- New Topology 2:

federation-v2
   |
   | -- charts
   |           |
   |           | -- clusterregistry
   |           |
   |           | -- controllermanager
   |           |
   |           | -- federatedtypes

Any comments for the above proposal?

gyliu513:
@marun @font @pmorie 

New PR updated based on topology as follows:

```
federation-v2
   |
   | -- charts
   |           |
   |           | -- clusterregistry
   |           |
   |           | -- controllermanager
   |
   | -- federatedtypes
```

shashidharatd:
I don't have any concerns regarding this pr. I see that there are couple of comments from @marun. So i would request him to do lgtm. Thanks !

PR #301, opened 2018-09-29
Expand on foundational concepts in README

pmorie:
This PR adds additional context for the concepts diagram. My intent is to make this part of the readme slightly more useful. As a next step, I'll expand this into a doc in the docs directory in a subsequent PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,23 +15,38 @@ Access to both documents is available to members of the
 [kubernetes-sig-multicluster google
 group](https://groups.google.com/forum/#!forum/kubernetes-sig-multicluster).
 
-<p align="center"><img src="docs/images/concepts.png" width="711"></p>
-
 # Concepts
 
-The following abstractions support the propagation of a logical
-federated type:
+<p align="center"><img src="docs/images/concepts.png" width="711"></p>
 
-- Template: defines the representation of the resource common across clusters
-- Placement: defines which clusters the resource is intended to appear in
-- Override: optionally defines per-cluster field-level variation to apply to the template
+Federation is configured with two types of information:
 
-These 3 abstractions provide a concise representation of a resource
-intended to appear in multiple clusters.  Since the details encoded by
-the abstractions are the minimum required for propagation, they are
-well-suited to serve as the glue between any given propagation
-mechanism and higher-order behaviors like policy-based placement and
+- **Type configuration** declares which API types federation should handle
+- **Cluster configuration** declares which clusters federation should target
+
+**Propagation** refers to the mechanism that distributes resources to federated
+clusters.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,24 +15,45 @@ Access to both documents is available to members of the
...
 group](https://groups.google.com/forum/#!forum/kubernetes-sig-multicluster).
 
+# Concepts
+
 <p align="center"><img src="docs/images/concepts.png" width="711"></p>
 
-# Concepts
+Federation is configured with two types of information:
+
+- **Type configuration** declares which API types federation should handle
```
gyliu513: Period to the end, ditto for others.

How about s/Type configuration/Type Configuration to keep consistent with the diagram?

README.md
```diff
@@ -15,24 +15,45 @@ Access to both documents is available to members of the
...
+- **Cluster configuration** declares which clusters federation should target
```
gyliu513: How about s/Cluster configuration/Cluster Configuration to keep consistent with the diagram?

README.md
```diff
@@ -15,24 +15,45 @@ Access to both documents is available to members of the
...
+clusters.
 
-The following abstractions support the propagation of a logical
-federated type:
+Type configuration has three fundamental concepts:
 
-- Template: defines the representation of the resource common across clusters
-- Placement: defines which clusters the resource is intended to appear in
-- Override: optionally defines per-cluster field-level variation to apply to the template
+- **Template types** define the representation of a resource common across clusters
```
gyliu513: How about remote `types` but only keep `Template` to keep consistent with diagram?

Ditto for others.

README.md
```diff
@@ -15,24 +15,45 @@ Access to both documents is available to members of the
...
+These three abstractions provide a concise representation of a resource intended
+to appear in multiple clusters. They encode the minimum information required for
+**propagation** and are well-suited to serve as the glue between any given
+propagation mechanism and higher-order behaviors like policy-based placement and
 dynamic scheduling.
 
+These fundamental concepts provide building blocks that can be used by
+higher-level APIs:
+
+**Status** collects the status of resources distributed by federation across all
```
gyliu513: - **Status** collects the status of resources distributed by federation across all
federated clusters.
- **Policy** determines which subset of clusters a resource is allowed to be
distributed to.
- **Scheduling** refers to a decision-making capability that can decide how
workloads should be spread across different clusters similar to how a human
operator would. Schedulers use status and policy information to make decisions
about how to implement high-level requirements from users about where workloads
should go.

README.md
gyliu513: Got it, thanks @pmorie

Issue #302, opened 2018-10-01
Add missing go tools.


poothia:
This issue is to track to have various go tools and their automated scripts.

/sig multicluster

marun:
I'd like to see an explanation and discussion of why these go tools are necessary or desirable.   Is there a mandate to impose these tools from outside the sig, or some other rational you'd like to share?

marun:
@poothia Can this issue be closed?

poothia:
@marun I think it can be closed and individual issues can be opened to add missing go scripts checks, that would be more specific.
what do you think?

PR #303, opened 2018-10-01
Adding Go-fmt tools and fixing go-fmt errors 

poothia:
Adding verify-gofmt.sh and update-gofmt.sh to verify and automate fixing of gofmt errors.

part of #302

### DIFF
--- a/hack/go-tools/verify-gofmt.sh
+++ b/hack/go-tools/verify-gofmt.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_ROOT}/../.." && pwd)"
+pushd ${REPO_ROOT} > /dev/null
+
+find_files() {
+  find . -not \( \
+      \( \
+        -wholename '*/vendor/*' \
+        -o -wholename '*/*kubebuilder.go' \
+        -o -wholename '*/*deepcopy.go' \
+        -o -wholename '*/doc.go' \
+      \) -prune \
+    \) -name '*.go'
+}
+
+GOFMT="gofmt -s"
+bad_files=$(find_files | xargs $GOFMT -l)
+popd > /dev/null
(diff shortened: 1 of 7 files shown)


gyliu513:
/cc @font 

How about adding the scripts and steps from https://github.com/kubernetes/cluster-registry/blob/master/docs/development.md#verify-go-source-files to federation v2?

pmorie:
Looks like the build flaked.

hack/go-tools/verify-gofmt.sh
```diff
@@ -0,0 +1,38 @@
...
+
+SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_ROOT}/../.." && pwd)"
+pushd ${REPO_ROOT} > /dev/null
+
+find_files() {
+  find . -not \( \
+      \( \
+        -wholename '*/vendor/*' \
+      \) -prune \
```
pmorie: We should exclude generated files here, since those are likely not to be compliant with gofmt.

hack/go-tools/update-gofmt.sh
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eou pipefail
+
+SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_ROOT}/../.." && pwd)"
+pushd ${REPO_ROOT} > /dev/null
+
+find . -name "*.go" | grep -v "\/vendor\/" | xargs gofmt -s -w
```
pmorie: We should exclude generated files here; reformatting them is likely to be more trouble than it's worth.

pmorie:
One other point, we should include the verify-gofmt script in the checks run by CI. Suggest closing #304, rolling the changes (to non-generated files) into this PR, and adding the verify script to the travis job.

hack/go-tools/verify-gofmt.sh
poothia: @pmorie thanks for reviewing this, what should i exclude apart from vendor? i thought all generated and third-party files are in vendors directory.

poothia: @pmorie  i have added some more checks as you suggested, please review this again.

poothia:
@gyliu513 should I add the scripts from https://github.com/kubernetes/cluster-registry/blob/master/docs/development.md#verify-go-source-files in this PR only? or create separate PR to do the same.

gyliu513:
@poothia I think we have some tasks as follows, but those tasks can be created in follow up PRs.

1) Add more checking files like cluster-registry at https://github.com/kubernetes/cluster-registry/tree/master/hack
2) Update https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md by adding some steps like https://github.com/kubernetes/cluster-registry/blob/master/docs/development.md#verify-go-source-files to tell developer how to use those scripts.

hack/go-tools/verify-gofmt.sh
pmorie: Ack, I think that should be sufficient.

marun:
Echoing @pmorie, this check should be integrated into the CI script as part of this PR and the changes from #304 should be added to this PR.

poothia:
please refer to #361

PR #304, opened 2018-10-01
fixing gofmt errors in the files

poothia:
it fixes go-fmt errors in the files.

### DIFF
--- a/pkg/apis/multiclusterdns/doc.go
+++ b/pkg/apis/multiclusterdns/doc.go
@@ -1,4 +1,3 @@
-
 /*
 Copyright 2018 The Kubernetes Authors.
 
@@ -15,11 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-
-
 // +k8s:deepcopy-gen=package,register
 // +groupName=multiclusterdns.federation.k8s.io
 
 // Package api is the internal version of the API.
 package multiclusterdns
-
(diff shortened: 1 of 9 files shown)


marun:
Please remove changes to generated files.

Also, I don't think there is much value in mechanical changes like this one.  If we care about this kind of thing it should be in an automated check.

gyliu513:
@marun the same issue also exist for `cluster-registry`, when I run https://github.com/kubernetes/cluster-registry/blob/master/docs/development.md#verify-go-source-files , some auto generated files will also be updated.

/cc @font may have more comments for this.

marun:
@gyliu513 Those file should be excluded from update.  There is zero value in running govet or gofmt against files that are mechanically maintained.

shashidharatd:
@poothia, please exclude generated files and retain only the changes for non-generated code.

poothia:
@shashidharatd @font @marun please review this again, i have removed the generated files from here, it was my mistake to add them in the first place, because i thought all the files except vendor folder belongs to project.

pkg/apis/multiclusterdns/doc.go
```diff
@@ -15,11 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-
-
```
shashidharatd: This is generated by kubebuilder. And all the subsequent changes in doc.go. please exclude them.

PR #311, opened 2018-10-04
Improve godoc for MultiClusterServiceDNSRecord and DNSEndpoint

pmorie:
Adds basic information about service DNS to the API godoc. Since ingress is very similar, I'll make this for DNS only for now, get the verbiage correct in this PR, and then do a follow-up PR to add godoc for ingress as well.

### DIFF
--- a/pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
+++ b/pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
@@ -30,17 +30,17 @@ type TTL int64
 // it is then stored in a persistent storage via serialization
 type Labels map[string]string
 
-// Endpoint is a high-level way of a connection between a service and an IP
+// Endpoint is a high-level association between a service and an IP.
 type Endpoint struct {
-	// The hostname of the DNS record
+	// The FQDN of the DNS record.
 	DNSName string `json:"dnsName,omitempty"`
-	// The targets the DNS record points to
+	// The targets that the DNS record points to.
 	Targets Targets `json:"targets,omitempty"`
-	// RecordType type of record, e.g. CNAME, A, SRV, TXT etc
+	// RecordType type of record, e.g. CNAME, A, SRV, TXT etc.
 	RecordType string `json:"recordType,omitempty"`
-	// TTL for the record in seconds
+	// TTL for the record in seconds.
 	RecordTTL TTL `json:"recordTTL,omitempty"`
-	// Labels stores labels defined for the Endpoint
+	// Labels stores labels defined for the Endpoint.
 	// +optional
 	Labels Labels `json:"labels,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


pmorie:
I'll follow this PR up with similar doc for ingress

pkg/apis/multiclusterdns/v1alpha1/multiclusterservicednsrecord_types.go
```diff
@@ -51,7 +52,31 @@ type ClusterDNS struct {
...
+//
+// For the example service mentioned earlier this would yield:
+//
+// test-service.test-namespace.(spec.federationName).svc
+//
+// Additionally, subdomains of this DNS name will be programmed for endpoints on
+// a region and zone basis:
+//
+// (status.DNS[*].region).test-service.test-namespace.(spec.federationName).svc
+// (status.DNS[*].zone).(status.DNS[*].region).test-service.test-namespace.(spec.federationName).svc
```
shashidharatd: Thanks @pmorie for this godoc.
This is slightly incorrect. There will be 3 names per cluster service instance.
and naming scheme is as follows:
1. Zone Level - (metadata.name).(metadata.namespace).(spec.federationName).svc.(status.DNS[\*].zone).(status.DNS[*].region).(spec.DNSSuffix)
2. Region Level - (metadata.name).(metadata.namespace).(spec.federationName).svc.(status.DNS[*].region).(spec.DNSSuffix)
3. Global Level - (metadata.name).(metadata.namespace).(spec.federationName).svc.(spec.DNSSuffix)

pmorie:
@danehans 

> Is there a reason why this PR does not cover `multiclusteringressdnsrecord_types.go`?

Since they're very similar, I wanted to get the verbiage correct here and then make another PR for ingress.

danehans:
@pmorie I created Issue #317 to track the follow-on effort for godoc'ing `MultiClusterIngressDNSRecord`

pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
```diff
@@ -30,17 +30,17 @@ type TTL int64
 // it is then stored in a persistent storage via serialization
 type Labels map[string]string
 
-// Endpoint is a high-level way of a connection between a service and an IP
+// Endpoint is a high-level connection between a service and an IP.
```
danehans: I would consider changing `connection` to `association` or `mapping`. When I see connection, I think of network plumbing.

pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
```diff
@@ -30,17 +30,17 @@ type TTL int64
...
 type Endpoint struct {
-	// The hostname of the DNS record
+	// The hostname of the DNS record.
```
danehans: I thought this is the FQDN of the DNS resource record?

pkg/apis/multiclusterdns/v1alpha1/dnsendpoint_types.go
pmorie: Good point

pmorie:
@danehans incorporated your feedbacks and rebased

cc @shashidharatd

shashidharatd:
/lgtm
Thanks @pmorie !

Issue #313, opened 2018-10-08
Version information in controller log is not correct


pmorie:
Running `latest` image, controller shows the following log messages:

```
I1004 20:54:16.093347       1 main.go:77] Version: {Version:v0.0.1-alpha.0 GitCommit:unknown GitTreeState:unknown BuildDate:unknown GoVersion:go1.10.1 Compiler:gc Platform:linux/amd64}
```

...which are not correct.

font:
Fixed by #315.

PR #315, opened 2018-10-09
Use Makefile to build binaries with version info

font:
Fix for #313.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ GOTARGET = github.com/kubernetes-sigs/$(TARGET)
 REGISTRY ?= quay.io/kubernetes-multicluster
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
+BIN_DIR := bin
 DOCKER ?= docker
 
 GIT_VERSION ?= $(shell git describe --always --dirty)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,6 +18,7 @@ GOTARGET = github.com/kubernetes-sigs/$(TARGET)
 REGISTRY ?= quay.io/kubernetes-multicluster
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
+BIN_DIR := bin
 DOCKER ?= docker
 
 GIT_VERSION ?= $(shell git describe --always --dirty)
```
gyliu513: Do we need to update here to use `tags`?

```
root@gyliu-dev1:~/go/src/github.com/kubernetes-sigs/federation-v2# git describe --tags --always --dirty
v0.0.2-rc.1-49-gbb7f952-dirty
```
```
root@gyliu-dev1:~/go/src/github.com/kubernetes-sigs/federation-v2# git describe --always --dirty
v0.0.1-283-gbb7f952-dirty
```

gyliu513: Seems we also re-tag the image with `--tags`, so we will push two images to dockerhub? https://github.com/kubernetes-sigs/federation-v2/blob/master/Makefile#L93-L101

font: > Seems we also re-tag the image with `--tags`, so we will push two images to dockerhub? https://github.com/kubernetes-sigs/federation-v2/blob/master/Makefile#L93-L101

We are not currently using the Makefile push target in any scripts yet. But it potentially pushes up to 3 images, the `GIT_VERSION` one is always pushed, and then if it's an official git tag it will push the tag version itself, along with a `latest` tag. You can specify any registry to push to, but we use `quay.io` not dockerhub for our official images.

PR #323, opened 2018-10-12
Adds Multi-Cluster Ingress DNS documentation

danehans:
Adds documentation explaining federated ingress dns and how to use the feature.

/cc @shashidharatd @gyliu513


danehans:
@gyliu513 I'll make the naming changes that you suggest in your review with a slight variation. Instead of specifically referencing Goog Cloud DNS, I'm going to reference the ExternalDNS project. This is because a) some sections are not specific to goog cloud dns and b) a new section in this doc can be added to support provider specific implementation steps.

docs/ingressdns.md
```diff
@@ -0,0 +1,226 @@
...
+[providers](https://github.com/kubernetes-incubator/external-dns#the-latest-release-v05). If needed, create a domain
+name with one of the supported providers or delegate a DNS subdomain to use for ExternalDNS. Reference your DNS
+provider documentation on how to create a domain or delegate a subdomain. The `example` subdomain of domain name
+`example.com` is used by this guide.
+
+Deploy the ExternalDNS controller. __NOTE:__ [Google Cloud DNS](https://cloud.google.com/dns) is used as the provider
+and `external.example.com` is used as the `domain-filter`. The filter will make ExternalDNS see only the hosted zones
+matching `external.example.com`. Omit this flag to process all available hosted zones.
+```bash
+cat <<EOF | kubectl create -f -
```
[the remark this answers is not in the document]
danehans: @shashidharatd I reviewed the fed-v3 helm [readme](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/README.md) and I don't see where/how it deploys external-dns or a sample deployment/service/ingress/multiclusteringressdnsrecord. It would also be nice to simplify the deps/steps for the time being by not using Helm and get user feedback first.

docs/ingressdns-with-externaldns.md
```diff
@@ -77,43 +77,37 @@ EOF
 Deploy test `FederatedDeployment`, `FederatedService` and `FederatedIngress` resources:
 ```bash
 cat <<EOF | kubectl create -f -
-apiVersion: v1
```
gyliu513: I found that most of the YAML templates in `example/sample1` is in such format, how about delete following content for all of the examples? @marun @font 

```
apiVersion: v1
kind: List
items:
```

docs/ingressdns-with-externaldns.md
```diff
@@ -0,0 +1,237 @@
+# Federated Ingress DNS with ExternalDNS Guide
+
+Federated Ingress DNS (FIDNS) provides the ability to programmatically manage DNS records of Ingress objects.
+FIDNS is not meant to replace in-cluster DNS providers such as [CoreDNS](https://coredns.io/). Instead, FIDNS works with
+systems such as [ExternalDNS](https://github.com/kubernetes-incubator/external-dns) that are responsible for managing
```
pmorie: I would rephrase this as "FIDNS leverages ExternalDNS to enable reprogramming many different DNS providers", as is it sounds like external-dns is one of many things that can be used, when it's the only way :)

docs/ingressdns-with-externaldns.md
```diff
@@ -0,0 +1,237 @@
...
+  hosts:
+  - ing.example.com
+  recordTTL: 300
+EOF
+```
+
+The DNS Endpoint controller uses the addresses from each `Ingress` to populate the `targets` field of the `DNSEndpoint`
+object after creating a `MultiClusterIngressDNSRecord`. For example:
+```bash
+kubectl get dnsendpoints -o yaml
```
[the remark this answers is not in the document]
danehans: ack

danehans:
@pmorie @gyliu513 @shashidharatd are you OK with me using `Federated` instead of `Multi-Cluster` in this document? I know the code and other docs refer to this feature as multi cluster service dns.

docs/ingressdns-with-externaldns.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Concepts
+
+<p align="center"><img src="./images/ingressdns-with-externaldns.png" width="711"></p>
+
+The above diagram illustrates MCIDNS. A typical MCIDNS workflow consists of:
+
+1. Creating `FederatedDeployment`, `FederatedService` and `FederatedIngress` objects. The Federation sync
+controller propagates the corresponding objects to target clusters.
+2. Creating a `MultiClusterIngressDNSRecord` object that identifies the intended domain name and other optional resource
```
gyliu513: s/MultiClusterIngressDNSRecord/IngressDNSRecord

danehans:
@pmorie yes, I was able to get serviceDNS working last week. I should have time before Thanksgiving break to update this PR.

danehans:
I am going to wait for https://github.com/kubernetes-sigs/federation-v2/pull/326 to merge before refactoring this doc.

example/sample1/federatedingress-template.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: core.federation.k8s.io/v1alpha1
+kind: FederatedIngress
+metadata:
+  name: test-ingress
+  namespace: test-namespace
+spec:
+  template:
+    spec:
+      rules:
+      - host: ingress.example.com
```
gyliu513: Should here be `ing.example.com`?

danehans: Good catch. Yes. I'll update the PR.

danehans: I am going to update the .md to use ingress instead of ing so the yaml will be correct.

Issue #305, opened 2018-10-03
Consider minor API renames for DNS API


pmorie:
Our API names in the area of DNS run on a bit; I wonder if we should rename:

- API group: multiclusterdns.federation.k8s
- API type: `MultiClusterServiceDNSRecord`

since the group already contains `multicluster`, how about changing the type to just `ServiceDNS` ? and similar for ingress.

WDYT @shashidharatd

irfanurrehman:
Seems good to me. API name of `ServiceDNS` should also be fine; however I remember that the `Record` was appended to the API name, because the bootstrapping and subsequent auto-generation of clients did not work as expected if the name did end with `S or s`. I am guessing it was a bug then, which might have been rectified now.

sohankunkerkar:
@pmorie Sure. I will dig into it.

sohankunkerkar:
I've created a new resource of the type "ServiceDNS" using kubebuilder for a test-project and verfied that `ServiceDNS` resource is abiding the appropriate plurization.
here's my finding:
```
$ kubebuilder create api --group multiclusterdns --version v1beta1 --kind ServiceDNS
Create Resource under pkg/apis [y/n]?
y
Create Controller under pkg/controller [y/n]?
y
Writing scaffold for you to edit...
pkg/apis/multiclusterdns/v1beta1/servicedns_types.go
pkg/apis/multiclusterdns/v1beta1/servicedns_types_test.go
pkg/controller/servicedns/servicedns_controller.go
pkg/controller/servicedns/servicedns_controller_test.go
Running make...
go generate ./pkg/... ./cmd/...
go fmt ./pkg/... ./cmd/...
go vet ./pkg/... ./cmd/...
go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
CRD manifests generated under '/home/skunkerk/golang/src/test-project/config/crds' 
RBAC manifests generated under '/home/skunkerk/golang/src/test-project/config/rbac' 
go test ./pkg/... ./cmd/... -coverprofile cover.out
?   	test-project/pkg/apis	[no test files]
?   	test-project/pkg/apis/multiclusterdns	[no test files]
ok  	test-project/pkg/apis/multiclusterdns/v1beta1	13.739s	coverage: 23.9% of statements
?   	test-project/pkg/controller	[no test files]
ok  	test-project/pkg/controller/servicedns	14.736s	coverage: 67.6% of statements
[...]

PR #331, opened 2018-10-17
Renaming DNS API

sohankunkerkar:
As discussed in the SIG meeting, I've renamed `MultiClusterServiceDNSRecord` to `ServiceDNSRecord` and `MultiClusterIngressDNSRecord` to `IngressDNSRecord`.


sohankunkerkar:
cc/ @font @shashidharatd @pmorie

pmorie:
/approve

Looks like what we agreed to in the WG today

gyliu513:
@sohankunkerkar does all of the code updated manually or you are also using `kubebuilder` to generate some code?

If you are also using `kubebuilder` to generate some code, then I suggest that you u se two commits in this PR, one is manually updated code, and another is auto generated code.

sohankunkerkar:
@gyliu513 I did not use `kubebuilder` in this case. The code is updated manually. Also, it's a temporary solution, we still need to figure out a way to rename these APIs using the `kubebuilder` tool. Correct me If I'm wrong, the final solution here is to rename `MultiClusterServiceDNSRecord` to `ServiceDNS` and `MultiClusterIngressDNSRecord` to `IngressDNS`.

gyliu513:
@sohankunkerkar I think you mean  rename `MultiClusterServiceDNSRecord` to `ServiceDNSRecord` and `MultiClusterIngressDNSRecord` to `IngressDNSRecord`. ;-)

I recalled that @marun also renamed some APIs in the PR https://github.com/kubernetes-sigs/federation-v2/pull/120 , not sure if those files are auto-generated or updated manually, let us get some comments from @marun for this.

pmorie:
#120 was a pretty fundamental re-cast of the repo - since we introduced the use of kubebuilder in this project, I believe there was a lot of regeneration.

I don't think renaming of APIs is something that kubebuilder has a feature for - it's more focused on optimizing the creation of new APIs.

marun:
@sohankunkerkar This PR looks fine and the tests run, but if it were to merge in its current form any subsequent invocations of `kubebuilder generate` would result in a broken tree due to kubebuilder generating new files for clientsets/listers/informers (i.e. `servicednsrecord.go` and `ingressdnsrecord.go`) that duplicate the contents of the files that you changed with your search and replace (i.e. `multiclusterdnsrecord.go` and `multiclusteringressdnsrecord.go`).   When modifying generated files, it's necessary to run `kubebuilder generate` followed by a build to ensure that the manual modifications do not conflict with kubebuilder output.

To fix (from repo root):
 - `kubebuilder generate`
 - `find -name 'multicluster*record.go' -exec git rm {} \;`
 - `go build ./cmd/controller-manager` (to verify fix)

The fact that this kind of error is possible suggests that `kubebuilder generate` should be invoked in ci to ensure that merging a PR will not result in this problem.

marun:
Added a check for kubebuilder generate in #337.  I'd like to see that PR merge and then CI re-run against this PR before merge.

danehans:
@pmorie I would like to see my https://github.com/kubernetes-sigs/federation-v2/issues/305#issuecomment-431086479 addressed before merging.

sohankunkerkar:
I'm closing this pull request and opening it again for triggering Travis CI build.

marun:
@sohankunkerkar For future, a simpler way to rekick travis is to invoke `git ci --amend` and then `git push`.  Even an amend without changes to the commit text will work.

pmorie:
@danehans 

> @pmorie I would like to see my #305 (comment) addressed before merging.

I agree we should address the name of the API group, but it's probably better for now to move forward on what we've agreed to so far - renaming the resource - and tackle renaming the API group as a separate task. For now I think renaming the resource to something more concise is movement in the right direction.

danehans:
#347 has been created to track renaming the DNS API group.

Issue #306, opened 2018-10-03
Add Support for Cluster-Scoped Objects 


danehans:
Federation-v2 currently supports namespace-scoped Kubernetes objects (i.e. ConfigMap). Cluster-scoped objects (i.e. ClusterRole) can not be [added](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md#adding-a-new-api-type) due to how fed-v2 version propagation currently works. The UX for adding cluster-scoped objects was addressed by https://github.com/kubernetes-sigs/federation-v2/issues/242.

danehans:
@marun thanks for all your hard work in getting this issue resolved!

PR #334, opened 2018-10-18
Add ClusterPropagatedVersion

marun:
This is the first of a series of PRs that add propagation of cluster-scoped resources (#306).


pmorie:
Actually, I had one question that I thought of before I tag with LGTM:

I noticed that the installation story hasn't change re: installing this new resource. Is that intentional?

PR #336, opened 2018-10-18
Fix typo in the federatedbar_crd.yaml example

roytman:


roytman:
The changes will be done as part of another PR, with code examples.

PR #337, opened 2018-10-18
Add ci check for 'kubebuilder generate' making no changes

marun:
Prompted by a [comment](https://github.com/kubernetes-sigs/federation-v2/pull/331) on #331.

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -65,6 +65,28 @@ function run-namespaced-e2e-tests() {
   return ${rc}
 }
 
+function check-kubebuilder-output() {
+  echo "Checking initial state of working tree"
+  if ! check-git-state; then
+    return 1
+  fi
+  ./bin/kubebuilder generate
+  echo "Checking state of working tree after running 'kubebuilder generate'"
+  check-git-state
+}
+
+function check-git-state() {
+  local output
+  if output=$(git status --porcelain) && [ -z "${output}" ]; then
+    return
+  fi
+  echo "ERROR: the working tree is dirty:"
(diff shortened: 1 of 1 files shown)


scripts/pre-commit.sh
```diff
@@ -75,12 +83,24 @@ cd "$base_dir" || {
...
 rc=0
 
-echo "Building federation binaries"
-build-binaries
-
 echo "Downloading test dependencies"
 ./scripts/download-binaries.sh
 
+echo "Checking that 'kubebuilder generate' is up-to-date"
+if output=$(git status --porcelain) && [ -z "$output" ]; then
```
font: Missing `{` and `}` in `$output`.

font: Can we instead encapsulate this logic in a function e.g. `run-kubebuilder-generate`?

marun: Done

marun: Done

scripts/pre-commit.sh
```diff
@@ -65,6 +65,30 @@ function run-namespaced-e2e-tests() {
   return ${rc}
 }
 
+function check-kubebuilder-output() {
+  echo "Checking initial state of working tree"
+  if check-git-state; then
+    ./bin/kubebuilder generate
+    echo "Checking state of working tree after running 'kubebuilder generate'"
+    if check-git-state; then
+      exit 0
```
font: We should not exit the script here. Simply returning a zero or non-zero return code should do the right thing since we have `errexit` set.

PR #339, opened 2018-10-18
Reduce flakes in e2e testing

marun:
Recently merged test additions have increased the incidence of flakes considerably.  This PR will attempt to identify and fix those flakes.

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -36,8 +36,6 @@ function run-integration-tests() {
   export TEST_ASSET_ETCD="${TEST_ASSET_PATH}/etcd"
   export TEST_ASSET_KUBE_APISERVER="${TEST_ASSET_PATH}/kube-apiserver"
   go test -v ./test/integration
-  rc=$((rc || $?))
-  return ${rc}
 }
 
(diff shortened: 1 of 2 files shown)


scripts/pre-commit.sh
```diff
@@ -55,14 +53,14 @@ function launch-minikube-cluster() {
...
-  return ${rc}
 }
 
 function run-namespaced-e2e-tests() {
-  ${E2E_TEST_CMD} -federation-namespace=foo -registry-namespace=foo -limited-scope=true
-  rc=$((rc || $?))
-  return ${rc}
+  local namespaced_e2e_test_cmd="${E2E_TEST_CMD} -federation-namespace=foo -registry-namespace=foo -limited-scope=true"
+  # Run the placement test separately to avoid crud failures if
+  # teardown doesn't remove namespace placement.
```
font: The `Placement` test explicitly removes the namespace placement and fails the test if it doesn't remove it. Do you know why teardown isn't removing the namespace placement?

font: Ah ok. I assumed this PR was reducing some flakes in e2e testing.

PR #340, opened 2018-10-18
Add 'kubefed2 federate' command

marun:
`kubefed2 federate` is intended to create the CRDs and FederatedTypeConfig necessary to federate a given target type.  This should remove the need to commit code to the repo to support types that are not expected to be referred to by controller code.

It will still likely make sense to include sample test fixture and examples in the tree, just not CRDs and FederatedTypeConfigs.

This code has been tested to work, but at present the CRDs are being generated without a validation schema (TODO).  Propagation still works without this.

A first pass of #60.

### DIFF
--- a/pkg/kubefed2/federate.go
+++ b/pkg/kubefed2/federate.go
@@ -1,0 +1,435 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubefed2
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextv1b1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+
+	apicommon "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
+	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
+	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+	ctlutil "github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/options"
+	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/util"
+)
+
+var (
+	federate_long = `
+	Federate enables a Kubernetes API type (including a CRD) to be
+	propagated to members of a federation.  Federation primitives will be
+	generated as CRDs and a FederatedTypeConfig will be created to
+	configure a sync controller.
+
+    Current context is assumed to be a Kubernetes cluster
+    hosting the federation control plane. Please use the
+    --host-cluster-context flag otherwise.`
+
+	federate_example = `
+	# Enable federation of a Kubernetes type by specifying the name,
+	# shortname, or kind of the target type. The context of the
+	# federation control plane's host cluster must be supplied if it
+	# is not the current context.
+	kubefed2 federate NAME --host-cluster-context=bar`
+)
+
+type federateType struct {
+	options.SubcommandOptions
+	federateTypeOptions
+}
+
+type federateTypeOptions struct {
+	targetName         string
+	rawComparisonField string
+	comparisonField    apicommon.VersionComparisonField
+	rawOverridePaths   string
+	overridePaths      []string
+	templateVersion    string
+	templateGroup      string
+	useExistingCRDs    bool
+}
+
+// Bind adds the join specific arguments to the flagset passed in as an
+// argument.
+func (o *federateTypeOptions) Bind(flags *pflag.FlagSet) {
+	flags.StringVar(&o.rawComparisonField, "comparison-field", string(apicommon.ResourceVersionField),
+		fmt.Sprintf("The field in the target type to compare for equality. Valid values are %q (default) and %q.",
+			apicommon.ResourceVersionField, apicommon.GenerationField,
+		),
+	)
+	flags.StringVar(&o.rawOverridePaths, "override-paths", "", "A common-separated list of dot-separated paths (e.g. spec.completions,spec.parallelism).")
+	flags.StringVar(&o.templateGroup, "template-group", "generated.federation.k8s.io", "The name of the API group of the target API type.")
+	flags.StringVar(&o.templateVersion, "template-version", "v1alpha1", "The API version of the target API type.")
+	flags.BoolVar(&o.useExistingCRDs, "use-existing-crds", false, "Whether to reuse existing primitive CRDs.")
+}
+
+// NewCmdFederate defines the `federate` command that enables
+// federation of a Kubernetes API type.
+func NewCmdFederate(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	opts := &federateType{}
+
+	cmd := &cobra.Command{
+		Use:     "federate NAME",
+		Short:   "Enable propagation of a Kubernetes API type",
+		Long:    federate_long,
+		Example: federate_example,
+		Run: func(cmd *cobra.Command, args []string) {
+			err := opts.Complete(args)
+			if err != nil {
+				glog.Fatalf("error: %v", err)
+			}
+
+			err = opts.Run(cmdOut, config)
+			if err != nil {
+				glog.Fatalf("error: %v", err)
+			}
+		},
+	}
+
+	flags := cmd.Flags()
+	opts.CommonBind(flags)
+	opts.Bind(flags)
+
+	return cmd
+}
+
+// Complete ensures that options are valid and marshals them if necessary.
+func (j *federateType) Complete(args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("NAME is required")
+	}
+	j.targetName = args[0]
+
+	if j.rawComparisonField == string(apicommon.ResourceVersionField) ||
+		j.rawComparisonField == string(apicommon.GenerationField) {
+		j.comparisonField = apicommon.VersionComparisonField(j.rawComparisonField)
+	} else {
+		return fmt.Errorf("--comparison-field must be %q or %q",
+			apicommon.ResourceVersionField, apicommon.GenerationField,
+		)
+	}
+	if len(j.rawOverridePaths) > 0 {
+		j.overridePaths = strings.Split(j.rawOverridePaths, ",")
+	}
+	if len(j.templateGroup) == 0 {
+		return fmt.Errorf("--template-group is a mandatory parameter ")
+	}
+	if len(j.templateVersion) == 0 {
+		return fmt.Errorf("--template-version is a mandatory parameter")
+	}
+
+	return nil
+}
+
+// Run is the implementation of the `federate` command.
+func (j *federateType) Run(cmdOut io.Writer, config util.FedConfig) error {
+	hostConfig, err := config.HostConfig(j.HostClusterContext, j.Kubeconfig)
+	if err != nil {
+		return fmt.Errorf("Failed to get host cluster config: %v", err)
+	}
+
+	_, err = EnableFederation(hostConfig, j.FederationNamespace, j.targetName, j.templateGroup,
+		j.templateVersion, j.comparisonField, j.overridePaths, j.useExistingCRDs, j.DryRun)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TODO(marun) Allow updates to the configuration for a type that has
+// already been enabled for federation.  This would likely involve
+// updating the version of the target type and the validation of the schema.
+func EnableFederation(config *rest.Config, federationNamespace, key, templateGroup,
+	templateVersion string, comparisonField apicommon.VersionComparisonField,
+	overridePaths []string, useExisting, dryRun bool) (typeconfig.Interface, error) {
+
+	apiResource, err := lookupAPIResource(config, key)
+	if err != nil {
+		return nil, err
+	}
+	glog.V(2).Infof("Found resource %q", resourceKey(*apiResource))
+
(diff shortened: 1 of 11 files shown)


pkg/kubefed2/federate.go
```diff
@@ -0,0 +1,292 @@
...
+    hosting the federation control plane. Please use the
+    --host-cluster-context flag otherwise.`
+
+	federate_example = `
+	# Enable federation of a Kubernetes type by specifying the
+	# Kind, version and group (optional for core types) of the
+	# target type. The context of the federation control plane's
+	# host cluster must be supplied if it is not the current
+	# context.
+	kubefed2 federate --group= --version=v1 --kind=Secret --host-cluster-context=bar`
```
danehans: Should `federate` be able to support different versions of the same group/kind? See:
```
$ ./bin/kubefed2 federate --namespaced=false --group=autoscaling --version=v1 --kind=HorizontalPodAutoscaler
$ kubectl get crd | grep hor
federatedhorizontalpodautoscalerplacements.generated.federation.k8s.io   2018-10-19T18:02:52Z
federatedhorizontalpodautoscalers.generated.federation.k8s.io            2018-10-19T18:02:52Z
$ ./bin/kubefed2 federate --namespaced=false --group=autoscaling --version=v2beta1 --kind=HorizontalPodAutoscaler
F1019 18:04:24.862496    3656 federate.go:113] error: Error creating CRD "federatedhorizontalpodautoscalers.generated.federation.k8s.io": customresourcedefinitions.apiextensions.k8s.io "federatedhorizontalpodautoscalers.generated.federation.k8s.io" already exists
```

danehans: OK. I'm trying to think through lifecycle management of federated resources using fed-v2. Here is the use case:

1. I have an older federated resource (i.e.  `FederatedHorizontalPodAutoscaler` v1beta1).
2. I need to upgrade the resource (i.e. v1, along with other attributes).
3. I want to deploy the new resource version while still running the old version.
4. I use the `FederatedHorizontalPodAutoscalerPlacement` to deploy the federated resource to a single cluster (test/dev). After all goes well, I update the `FederatedHorizontalPodAutoscalerPlacement` to include all clusters.
5. After the new version runs for a period of time, I remove the older version of `FederatedHorizontalPodAutoscaler`.

I think the above use case will be common. Without supporting different resource versions, I don't think fed-v2 can support this use case. Am I missing something or how would you go about supporting the above use case?

pkg/kubefed2/federate.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	// TODO(marun) Perform this check with the discovery client.
+	targetClient, err := controllerutil.NewResourceClientFromConfig(hostConfig, &apiResource)
+	if err != nil {
+		return fmt.Errorf("Error creating client for target resource %q: %v", apiResource.Kind, err)
+	}
+	_, err = targetClient.Resources(j.FederationNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("Unable to verify that %q is a valid target type: %v", apiResource.Kind, err)
```
danehans: I am using `kubefed2 federate` to federate resource `MutatingWebhookConfiguration` and I get the following error:
```
$ ./bin/kubefed2 federate --namespaced=true --group=admissionregistration.k8s.io --version=v1beta1 --kind=MutatingWebhookConfiguration
F1019 19:08:11.706428   25028 federate.go:113] error: Unable to verify that "MutatingWebhookConfiguration" is a valid target type: the server could not find the requested resource
```

The proper admission controllers are enabled on kube-apiserver and the necessary api-version is available:
```
root     30494 30466  4 Oct18 ?        00:54:38 kube-apiserver --authorization-mode=Node,RBAC --enable-admission-plugins=Initializers,NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,NodeRestriction,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota --advertise-address=10.138.0.2 --allow-privileged=true --client-ca-file=/var/lib/localkube/certs/ca.crt --disable-admission-plugins=PersistentVolumeLabel --enable-bootstrap-token-auth=true --etcd-cafile=/var/lib/localkube/certs/etcd/ca.crt --etcd-certfile=/var/lib/localkube/certs/apiserver-etcd-client.crt --etcd-keyfile=/var/lib/localkube/certs/apiserver-etcd-client.key --etcd-servers=https://127.0.0.1:2379 --insecure-port=0 --kubelet-client-certificate=/var/lib/localkube/certs/apiserver-kubelet-client.crt --kubelet-client-key=/var/lib/localkube/certs/apiserver-kubelet-client.key --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname --proxy-client-cert-file=/var/lib/localkube/certs/front-proxy-client.crt --proxy-client-key-file=/var/lib/localkube/certs/front-proxy-client.key --requestheader-allowed-names=front-proxy-client --requestheader-client-ca-file=/var/lib/localkube/certs/front-proxy-ca.crt --requestheader-extra-headers-prefix=X-Remote-Extra- --requestheader-group-headers=X-Remote-Group --requestheader-username-headers=X-Remote-User --secure-port=8443 --service-account-key-file=/var/lib/localkube/certs/sa.pub --service-cluster-ip-range=10.96.0.0/12 --tls-cert-file=/var/lib/localkube/certs/apiserver.crt --tls-private-key-file=/var/lib/localkube/certs/apiserver.key

$ kubectl api-versions | grep admissionregistration.k8s.io/v1beta1
admissionregistration.k8s.io/v1beta1
```

pkg/kubefed2/federate.go
marun: If I create `FederatedHorizontalPodAutoscaler` `foo` in namespace `bar` with version `v1alpha`, trying to create that same resource with same name with version 'v1' will fail because the resource already exists.  That's because the version doesn't affect the storage of the resource, only how the API endpoint receives requests (e.g. allowing new fields or values at a newer endpoint).

If a member of the federation is upgraded to a newer version of Kubernetes that supports a newer version of `HorizontalPodAutoscaler`, propagating resources to that cluster automatically get  upgraded behavior regardless of the version of the API they are targeting.  Because of the unified storage and backwards compatibility guarantees, a `v1beta1` request to an API that supports `v1` will automatically apply defaults to added fields to ensure `v1` compatibility.  The HPA controller won't know what version of the API was used to create the resource.

That's not to say that the version of the target resource defined in the FederatedTypeConfig isn't important.  If a given version of HPA supports field `foo`, the version of the target resource must be set to that version or greater for the sync controller to be able to successfully deliver that field to a member cluster.

pkg/kubefed2/federate.go
marun: Consider specifying `--namespaced=false`

```
NAME                                     SHORTNAMES   APIGROUP                            NAMESPACED   KIND
mutatingwebhookconfigurations                         admissionregistration.k8s.io        false        MutatingWebhookConfiguration
```

marun: The requirement to specify `--namespaced` is temporary.  I'm working on looking up as much as possible from the host cluster to minimize what can go wrong.

pkg/kubefed2/federate.go
danehans: Lets say the use case is: I have a 3 cluster federation. Each cluster runs k8s v1.11. I need to upgrade to 1.12. I start off upgrading my test/dev cluster. All goes well, but v1.12 upgrades resource X from `v1beta2` to `v1`. I need fed-v2 to support both of these versions so resources can be managed across 1.11 and 1.12 clusters until all 3 clusters are upgraded to 1.12.

danehans: BTW, thanks for all the feedback above.

pkg/kubefed2/federate.go
danehans: That did the trick. Sorry for not trying that before pinging you. It looks like the Istio manifest incorrectly states a ns for the resource:
```
# Source: istio/charts/sidecarInjectorWebhook/templates/mutatingwebhook.yaml
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  name: istio-sidecar-injector
  namespace: istio-system
<SNIP>
```

danehans:
FYI.. I had to remove the `Deployment` type to use an older version. When I created `Deployment`, it did not create the `FederatedDeploymentOverride` CRD when using the `federate` command even though it existed before I removed the newer version.

pkg/kubefed2/federate.go
marun: Because of the backwards compatibility guarantees that Kubernetes provides for APIs, upgrading a member cluster to 1.12 would enable `v1` for type X while retaining support for `v1beta2`.  Since the sync controller for type X would need to target the version common to all members of the federation, the controller would need to continue targeting `v1beta2` until all clusters were updated to support `v1`.

marun:
As per a comment in the code, generation of the override CRD is not yet implemented.  This PR is still WIP, so there are gaps yet to be filled.

marun:
@danehans The current iteration of this PR enables a simpler `federate` command:

```bash
kubefed2 federate <type/kind/short name>
```

group/version/namespaced are now determined automatically based on the servers preferred resources (i.e. latest).

danehans:
> group/version/namespaced are now determined automatically based on the servers preferred resources (i.e. latest).

Do I still have the ability to override `group` and/or `version`? The reason I ask is b/c I am in the process of federating Istio. As you can see from [this manifest](https://raw.githubusercontent.com/danehans/fedv2-manifests/master/istio/v0.8.0/src/istio.yaml), older versions of some resources are being used even-though my k8s cluster is v1.11.

marun:
@danehans While it could be possible to choose an older API version, why would that be desirable?  

A `FederatedTypeConfig` defines the version of the API endpoint that will be used in propagating resources to member clusters.  If a resource is provided in a template with an older version (i.e. missing fields found in newer versions as per the Kubernetes convention of backwards compatibility), the default behavior would be provided for those fields regardless of which endpoint was used to create it.

marun:
- `federate` now supports the `enable` and `disable` subcommands
 - `federate enable` supports overrides (via `--override-paths`)
 - crd validation is in progress

danehans:
> While it could be possible to choose an older API version, why would that be desirable?

@marun l am in the process of federating Istio v0.8 and 1.x. I want to support both versions because I intend on demonstrating upgrades. The v0.8 manifest that I shared in my previous comment uses older API versions. I could see others doing the same.

danehans:
I am trying to federate the Istio custom resource `stdio` and I get the following error:
```
F1030 17:55:45.006244   67587 federate.go:113] error: Unable to verify that "stdio" is a valid target type: the server could not find the requested resource (get stdioes.config.istio.io)
```

Here is the api resource:
```
$ kubectl api-resources | grep std
stdios                                                         config.istio.io                     true         stdio
```

As you can see, the federate command is looking for the incorrect pluralization of `stdio`.

Grammer reference:
https://www.grammarly.com/blog/spelling-plurals-with-s-es/

I believe I ran into this same issue with ingress a while back.

danehans:
Same issue with `logentry`:
```
$ kubectl api-resources | grep logentry
logentries                                                     config.istio.io                     true         logentry
$ kubefed2 federate --namespaced=true --group=config.istio.io --version=v1alpha2 --kind=logentry
F1030 18:10:36.478587   69257 federate.go:113] error: Unable to verify that "logentry" is a valid target type: the server could not find the requested resource (get logentrys.config.istio.io)
```

marun:
As per a slack conversation with @danehans, the above reported errors were using a stale version of this PR and do not represent failures with its current state.

pkg/kubefed2/federate/util.go
```diff
@@ -0,0 +1,127 @@
...
+	if apiResource.Version == "" {
+		version = "v1"
+	} else {
+		version = apiResource.Version
+	}
+	return fmt.Sprintf("%s.%s/%s", apiResource.Name, group, version)
+}
+
+func groupQualifiedName(apiResource metav1.APIResource) string {
+	if len(apiResource.Group) == 0 {
```
font: (no action required) Just noticed we are using `len` to determine empty string and in the function right above we use `""`.

pkg/kubefed2/federate/util.go
```diff
@@ -0,0 +1,127 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: groupQualifiedName(apiResource),
+		},
+		Spec: apiextv1b1.CustomResourceDefinitionSpec{
+			Group:   apiResource.Group,
+			Version: apiResource.Version,
+			Scope:   scope,
+			Names: apiextv1b1.CustomResourceDefinitionNames{
+				Plural:   apiResource.Name,
+				Singular: apiResource.SingularName,
```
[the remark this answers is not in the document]
marun: Good catch.  I've removed it, which should ensure that the default value of `lower(kind)` is used.

pkg/kubefed2/federate/enable.go
```diff
@@ -0,0 +1,264 @@
...
+		to members of a federation.  Federation primitives will be
+		generated as CRDs and a FederatedTypeConfig will be created to
+		configure a sync controller.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the federation control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	enable_example = `
+		# Enable federation of Services with service type overrideable
```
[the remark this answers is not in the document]
marun: Good catch!

marun:
@font Thank you for the review, much appreciated!  My apologies for the size of this PR.

irfanurrehman:
@marun apologies on running late on the reviews; When I got to this now it was already merged.
I had few observations/questions which I would point out anyways.

- I believe some documentation around this would be warranted soon.
Would it be a good idea to clear it out in help that this should be run against a cluster which has a minimal federation API (say TypeConfig CRD) installed. I mean one can go ahead and run it against a pristine cluster also.
The help line probably does not have that info.
`Current context is assumed to be a Kubernetes cluster hosting the federation control plane`.
I am fine if there is a plan of documenting it otherwise soon.

- Does it make sense to have `An optional` subtext to arguments description where they are optional

- Plan around having a same name Federated CRD installed under different gv? Does it make sense to federate same target type under different gv?

marun:
@irfanurrehman 

1) `kubefed2` has a dependency on the federation API being installed, not just the added command.  I think you're correct that documentation of this requirement is suggested, and would ask that you add an issue to track it.

2) All flags are optional (`name` is an argument, and mandatory), so I'm not sure it makes sense to add gratuitous text to each argument description.

3) I don't think it makes sense for a given target to have more than a single set of primitives (template, placement, override).  My intention is for all primitives to be created in the `primitives.federation.k8s.io` group, and only users with a need we haven't anticipated should be changing `--primitive-group`.  So long as a given deployment uses the same group for all primitives, it won't be possible to define a set of primitives more than once for a given target since CRDs are cluster scoped and the generated name is `<target plural name>.<target group>`.

Version I'm less clear on, since versioning of CRDs is not something I understand very well.  At this time I would expect the primitives for a given type to be updated with the validation schema of a newer version of the target rather than supplanted by new primitives under a different version.

irfanurrehman:
1. Created https://github.com/kubernetes-sigs/federation-v2/issues/383 for doc.
2. Ack.
3. 
> 3\. So long as a given deployment uses the same group for all primitives, it won't be possible to define a set of primitives more than once for a given target since CRDs are cluster scoped and the generated name is `<target plural name>.<target group>`.

Probably I was not completely clear the first time. So currently I can federate say `Services` under different groups as many number of times as I want. Is this expected? Is different users (or same user) federating the same type under different groups a valid use case? (I haven't come across the use of the same so far).
Lets leave the version discussion for a later date.. :)

marun:
There is an added check against federating services multiple times under many different groups for a given federation control plane - the `FederatedTypeConfig` is also named `<target plural name>.<target group>`.  Thus, for a given federation, it is only possible to configure propagation/scheduling/(soon status collection) once for a given target type.

It would be possible for user A to generate federated type CRDs under group `A`, and  userB to generate CRDs under group `B`, and so long as each of them ran namespaced federation they could have each have `FederatedTypeConfig` instances for the target types pointing to their desired primitive CRDs.

Issue #342, opened 2018-10-19
Add ObservedGeneration to FederatedTypeConfig status


pmorie:
`ObservedGeneration` is used in many API statuses to indicate whether the controller backing a resource has 'seen' the latest version of that resource's spec. We should have an `ObservedGeneration` field in the status of FederatedTypeConfig to allow users to determine this information for that resource. This will increase understandability of the behavior of the system - for example, if the controller crashes, it will be clear that `ObservedGeneration` does not match the current `metadata.generation` of the resource.

pmorie:
@sohankunkerkar is interested in working on this but we cannot assign him in github yet, so noting here.

sohankunkerkar:
I've created a PR #345 for this issue.

PR #345, opened 2018-10-21
 Added ObservedGeneration field in the status of FederatedTypeConfig

sohankunkerkar:


gyliu513:
Fixed https://github.com/kubernetes-sigs/federation-v2/issues/342

/ok-to-test

marun:
@sohankunkerkar You'll want to add the status subresource tag to ensure that the status of the generated crd can be updated independently of the rest of the resource:

 - [status tag on PropagatedVersion](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/apis/core/v1alpha1/propagatedversion_types.go#L58)
- [generated subresource field](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/apis/core/v1alpha1/zz_generated.kubebuilder.go#L1644)

marun:
@sohankunkerkar As per the CI failure, you'll need to update `hack/install-latest.yaml`:

```bash
FEDERATION_NAMESPACE=federation-system INSTALL_YAML=./hack/install-latest.yaml ./scripts/generate-install-yaml.sh quay.io/kubernetes-multicluster/federation-v2:latest
```

gyliu513:
@sohankunkerkar The CI is reporting `hack/install-latest.yaml` is not up-to-date, please re-generate this file and update the PR. @marun just added this check in CI https://github.com/kubernetes-sigs/federation-v2/pull/344

hack/install-latest.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   creationTimestamp: null
   labels:
     api: federation
-    kubebuilder.k8s.io: 1.0.3
+    kubebuilder.k8s.io: 1.0.4
```
gyliu513: Please use kubebuilder 1.0.3, we did not upgrade kubebuilder to 1.0.4 yet.

sohankunkerkar:
@marun @gyliu513 done!

pmorie:
@sohankunkerkar There are a couple pieces that you should do in a follow-up:

- Client for FederatedTypeConfig needs to be regenerated to have `UpdateStatus` method
- FederatedTypeConfig controller should set the `observedGeneration` in its control loop

sohankunkerkar:
@pmorie I could see `UpdateStatus` method which is already present here 
(https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/client/clientset/versioned/typed/core/v1alpha1/federatedtypeconfig.go#L127)
Not sure if I'm missing any information.

@gyliu513 for #358, I will be following the above two pieces mentioned by @pmorie

PR #357, opened 2018-10-24
Support multiple field overrides

irfanurrehman:
cc @marun @shashidharatd


marun:
@irfanurrehman You'll need to update `hack/download-binaries.sh` to download `kubebuilder` `1.0.4` for ci to pass.

irfanurrehman:
> @irfanurrehman You'll need to update `hack/download-binaries.sh` to download `kubebuilder` `1.0.4` for ci to pass

@marun thanks for the tip. Updated!

irfanurrehman:
For some weird reason, however, the integration test passes for me locally, which it should not.
I think I will need to skip check for `spec.completions` update in Jobs crud test (the update fails because `spec.completions` is immutable. 
@marun, you have a better suggestion?

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -77,6 +77,14 @@ type APIResource struct {
 	PluralName string `json:"pluralName,omitempty"`
 }
 
+// OverridePath is used to specify the name:path pair for a given override config.
+type OverridePath struct {
+	// Name of the field in the override type.
```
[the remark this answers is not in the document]
marun: I think name is desirable.  Not to disambiguate the definition in the type config, since as you point out the full path is available.  Rather, when a user defines an override they should be able to provided a name instead of the path.  By default the name would default to the last entry in the path, but in the case where last entries (not full paths) are duplicated a name could be explicitly provided:

```yaml
# No disambiguation required
overridePaths:
- path: spec/foo
- path: spec/bar

overrides: 
- clusterName: cluster1
  foo: "value"
  bar: "value"

# Provide name to disambiguate
overridePaths:
- name: fooBar
  path: spec/foo/bar
- name: barBar
  path: spec/bar/bar

overrides: 
- clusterName: cluster1
  fooBar: "value"
  barBar: "value"
```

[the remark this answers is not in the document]
irfanurrehman: Got it, Updated! Thanks for explaining.

irfanurrehman:
@marun Updated! FYI, I did also add a field in the default fixture yamls relevant to `Deployment` to enable testing out multiple overrides.

config/federatedtypes/deployment.yaml
```diff
@@ -17,6 +17,6 @@ spec:
     kind: FederatedDeploymentPlacement
   override:
     kind: FederatedDeploymentOverride
-  overridePath:
-    - spec
-    - replicas
+  overridePaths:
+    - path: spec.replicas
+    - path: spec.revisionHistoryLimit
```
marun: I appreciate your desire to test multiple overrides with the crud tester, but this type config will be deployed (it's not test-only).  Once #340 lands it will be trivial to generate a type config + crds to enable this testing, can the testing wait until then?

irfanurrehman: Sounds ok to me, Updated!
Meanwhile, I am working my way to submit tests for job overrides and service overrides (multiple fields) but I can wait to use `kubefed federated` (which I did look up after you mentioned it.. :) ). As far as I understood some additional test yamls would still be needed though (stuff currently in test/common/fixtures). The ones already there might be good for default crud test; What would be your suggestion on updating them vs using newer ones for specific tests?

marun: It should be possible to use the crd test in #340 to test multiple overrides against a generated crd rather than requiring test fixture.  I should have something for overrides up later today now that this is ready to merge.

marun:
/lgtm

Thank you!

PR #358, opened 2018-10-25
Added observedGeneration to FederatedCluster status

sohankunkerkar:
Similar to `FederatedTypeConfig`, adding `ObservedGeneration` field in the status of `FederatedCluster` to allow users to determine whether the controller backing a resource has 'seen' the latest version of that resource's spec. This will increase understandability of the behavior of the system.

### DIFF
--- a/hack/install-latest.yaml
+++ b/hack/install-latest.yaml
@@ -150,6 +150,9 @@ spec:
                 - status
                 type: object
               type: array
+            observedGeneration:
+              format: int64
+              type: integer
             region:
               type: string
             zone:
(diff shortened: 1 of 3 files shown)


sohankunkerkar:
cc/ @pmorie @marun

marun:
@sohankunkerkar I think this PR should ensure the controller populates the field.

sohankunkerkar:
@marun I'm working on adding this field in the controller for both `FederatedCluster` and `FederatedTypeConfig`. Should I track this under different PR?

marun:
@sohankunkerkar I think the value in adding the field is populating it, and that adding the field+populating it for each of the types in question is the order I would choose vs adding the field for both and then populating it.

sohankunkerkar:
@marun makes complete sense. I will not create a separate PR for these changes instead add it in the existing PR itself.

sohankunkerkar:
Closing this PR after having some discussion with @pmorie

PR #366, opened 2018-10-26
Run controllers based on system configuration

### LINKED ISSUE #349: Run controllers based on system configuration
Currently, a typeconfig controller watches for typeconfig resources and starts sync controllers per federated types if a typeconfig resource with propagation enabled is created for that federated type.
Other controllers, (specifically scheduling controllers) which depend on the target federated types availability when the given controller starts can also be run only if the given target federated type(s) is/are enabled in typeconfig.
The mechanism can be similar to (or inside) typeconfig controller to start/stop the relevant (scheduling) controller(s) when the target type is enabled/disabled.
This allows for more fine grained automatic configuration of the controllers.

xunpan:
Fix #349.

Make scheduler plugins with correct adapters created when corresponding
federated resource enabled in federatedtypeconfig.

/cc @irfanurrehman @gyliu513

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -38,11 +38,10 @@ import (
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/federatedcluster"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/federatedtypeconfig"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/ingressdns"
-	"github.com/kubernetes-sigs/federation-v2/pkg/controller/schedulingpreference"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/schedulingmanager"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/servicedns"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 	"github.com/kubernetes-sigs/federation-v2/pkg/features"
-	"github.com/kubernetes-sigs/federation-v2/pkg/schedulingtypes"
 	"github.com/kubernetes-sigs/federation-v2/pkg/version"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	flagutil "k8s.io/apiserver/pkg/util/flag"
@@ -125,12 +124,7 @@ func main() {
 	federatedcluster.StartClusterController(controllerConfig, stopChan, clusterMonitorPeriod)
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulerPreferences) {
-		for kind, schedulingType := range schedulingtypes.SchedulingTypes() {
-			err = schedulingpreference.StartSchedulingPreferenceController(controllerConfig, stopChan, kind, schedulingType.SchedulerFactory)
-			if err != nil {
-				glog.Fatalf("Error starting schedulingpreference controller for %q : %v", kind, err)
-			}
-		}
+		schedulingmanager.StartSchedulerController(controllerConfig, stopChan)
 	}
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.CrossClusterServiceDiscovery) {
--- a/pkg/controller/schedulingmanager/controller.go
+++ b/pkg/controller/schedulingmanager/controller.go
@@ -1,0 +1,165 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package schedulingmanager
+
+import (
+	"fmt"
+
+	"github.com/golang/glog"
+	corev1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+	fedclientset "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned"
+	corev1alpha1client "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned/typed/core/v1alpha1"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/schedulingpreference"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+	"github.com/kubernetes-sigs/federation-v2/pkg/schedulingtypes"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+var SchedulingRegistry = map[string]string{
+	"FederatedDeployment": schedulingtypes.RSPKind,
+	"FederatedReplicaSet": schedulingtypes.RSPKind,
+}
+
+type SchedulerController struct {
+	// Store for the FederatedTypeConfig objects
+	store cache.Store
+	// Informer for the FederatedTypeConfig objects
+	controller cache.Controller
+
+	worker util.ReconcileWorker
+
+	scheduler map[string]schedulingtypes.Scheduler
+
+	config   *util.ControllerConfig
+	stopChan <-chan struct{}
+}
+
+func StartSchedulerController(config *util.ControllerConfig, stopChan <-chan struct{}) {
+
+	userAgent := "SchedulerController"
+	kubeConfig := config.KubeConfig
+	restclient.AddUserAgent(kubeConfig, userAgent)
+	client := fedclientset.NewForConfigOrDie(kubeConfig).CoreV1alpha1()
+
+	controller := newController(config, client, stopChan)
+
+	glog.Infof("Starting scheduler controller")
+	controller.Run(stopChan)
+}
+
+func newController(config *util.ControllerConfig, client corev1alpha1client.CoreV1alpha1Interface, stopChan <-chan struct{}) *SchedulerController {
+	c := &SchedulerController{
+		scheduler: make(map[string]schedulingtypes.Scheduler),
+		config:    config,
+		stopChan:  stopChan,
+	}
+
+	fedNamespace := config.FederationNamespace
+	c.worker = util.NewReconcileWorker(c.reconcile, util.WorkerTiming{})
+
+	c.store, c.controller = cache.NewInformer(
+		&cache.ListWatch{
+			// Only watch the federation namespace to ensure
+			// restrictive authz can be applied to a namespaced
+			// control plane.
+			ListFunc: func(options metav1.ListOptions) (pkgruntime.Object, error) {
+				return client.FederatedTypeConfigs(fedNamespace).List(options)
+			},
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return client.FederatedTypeConfigs(fedNamespace).Watch(options)
+			},
+		},
+		&corev1a1.FederatedTypeConfig{},
+		util.NoResyncPeriod,
+		util.NewTriggerOnAllChanges(c.worker.EnqueueObject),
+	)
+
+	return c
+}
+
+// Run runs the Controller.
+func (c *SchedulerController) Run(stopChan <-chan struct{}) {
+	go c.controller.Run(stopChan)
+
+	// wait for the caches to synchronize before starting the worker
+	if !cache.WaitForCacheSync(stopChan, c.controller.HasSynced) {
+		runtime.HandleError(fmt.Errorf("Timed out waiting for cache to sync"))
+		return
+	}
+
+	c.worker.Run(stopChan)
+}
+
+func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.ReconciliationStatus {
+	key := qualifiedName.String()
+
+	glog.Infof("Running reconcile FederatedTypeConfig for %q", key)
+
+	cachedObj, exist, err := c.store.GetByKey(key)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("Failed to query FederatedTypeConfig store for %q: %v", key, err))
+		return util.StatusError
+	}
+	if !exist {
+		return util.StatusAllOK
+	}
+	typeConfig := cachedObj.(*corev1a1.FederatedTypeConfig)
+
+	deleted := typeConfig.DeletionTimestamp != nil
+	if deleted {
+		// Do not support deletion yet
+		return util.StatusAllOK
+	}
+
+	// set name and group for the type config target
+	corev1a1.SetFederatedTypeConfigDefaults(typeConfig)
+	apiResource := typeConfig.GetTarget()
+	templateKind := typeConfig.GetTemplate().Kind
+
+	kind, ok := SchedulingRegistry[templateKind]
+	if !ok {
+		// No any scheduler supported for this resource
+		return util.StatusAllOK
+	}
+	schedulingType, _ := schedulingtypes.SchedulingTypes()[kind]
+
+	// Scheduling preference controller is started on demand
+	_, ok = c.scheduler[kind]
+	if !ok {
+		scheduler, err := schedulingpreference.StartSchedulingPreferenceController(c.config, c.stopChan, kind, schedulingType.SchedulerFactory)
+		if err != nil {
+			runtime.HandleError(fmt.Errorf("Error starting schedulingpreference controller for %q : %v", kind, err))
+			return util.StatusError
+		}
+		c.scheduler[kind] = scheduler
+	}
+
+	glog.Infof("Start plugin with kind %s for scheduling type %s", templateKind, kind)
+	err = c.scheduler[kind].StartPlugin(templateKind, &apiResource, c.stopChan)
+	if err != nil {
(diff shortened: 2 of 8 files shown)


xunpan:
I signed it!

xunpan:
I test the case after install fresh federated clusters based on user guide.

``
$ kubectl apply -f example/sample1/federatednamespace-template.yaml -f example/sample1/federatednamespace-placement.yaml
$ kubectl apply -f example/sample1/federateddeployment-override.yaml -f example/sample1/federateddeployment-template.yaml -f example/sample1/federateddeployment-placement.yaml
``
After a while, 3 deployments in cluster1 and 5 deployments in cluster2

``
$ kubectl apply -f example/scheduling/federateddeployment-rsp.yaml
``
After a while, 4 deployments in cluster1 and 6 deployments in cluster2

xunpan:
I finished the test part and passed local integration test. e2e case has unrelated error. So, remove WIP to make CI have a try.

pkg/controller/schedulingpreference/scheduler.go
```diff
@@ -0,0 +1,148 @@
+/*
```
marun: I think this controller belongs in its own package (e.g. `pkg/controller/schedulingmanager`).

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -108,6 +108,38 @@ func (s *ReplicaScheduler) Kind() string {
...
+	default:
+		return
+	}
+
+	if _, ok := s.plugins[kind]; ok {
+		return
+	}
+
+	glog.Infof("Kind %s is registered to the scheduler plugin", kind)
+	s.plugins[kind] = NewPlugin(
```
marun: Consider having the plugin accept the `pluginArgs` struct.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -48,43 +48,43 @@ func init() {
...
-	plugins         map[string]*Plugin
+	plugins map[string]*Plugin
+	pluginArgs
+
 	fedClient       fedclientset.Interface
 	podInformer     FederatedInformer
 	targetNamespace string
 }
 
 func NewReplicaScheduler(fedClient fedclientset.Interface, kubeClient kubeclientset.Interface, crClient crclientset.Interface, fedNamespace, clusterNamespace, targetNamespace string, federationEventHandler, clusterEventHandler func(pkgruntime.Object), handlers *ClusterLifecycleHandlerFuncs) Scheduler {
```
marun: (No action required) Consider having this method accept the `pluginArgs` struct.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -48,43 +48,43 @@ func init() {
...
+	targetNamespace        string
+	federationEventHandler func(pkgruntime.Object)
+	clusterEventHandler    func(pkgruntime.Object)
+	handlers               *ClusterLifecycleHandlerFuncs
+}
+
 type ReplicaScheduler struct {
-	plugins         map[string]*Plugin
+	plugins map[string]*Plugin
+	pluginArgs
```
marun: Why is it desirable to embed this type?

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -108,6 +108,38 @@ func (s *ReplicaScheduler) Kind() string {
...
 
+func (s *ReplicaScheduler) RegisterPlugins(kind string, apiResource *metav1.APIResource, stopChan <-chan struct{}) {
+	var adapter adapters.Adapter
+	switch kind {
+	case FederatedDeployment:
+		adapter = adapters.NewFederatedDeploymentAdapter(s.fedClient)
+	case FederatedReplicaSet:
+		adapter = adapters.NewFederatedReplicaSetAdapter(s.fedClient)
+	default:
+		return
```
marun: I think returning an error is suggested if the kind is not recognized, so that the caller knows they are attempting to do something that is not supported.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -108,6 +108,38 @@ func (s *ReplicaScheduler) Kind() string {
...
+		s.pluginArgs.kubeClient,
+		s.pluginArgs.crClient,
+		s.pluginArgs.fedNamespace,
+		s.pluginArgs.clusterNamespace,
+		s.pluginArgs.targetNamespace,
+		s.pluginArgs.federationEventHandler,
+		s.pluginArgs.clusterEventHandler,
+		s.pluginArgs.handlers,
+	)
+	s.plugins[kind].Start(stopChan)
```
marun: Given that this PR ensures that plugins are only enabled by calling this method, the `Start` method of the controller no longer needs to attempt to start plugins.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -178,7 +210,12 @@ func (s *ReplicaScheduler) Reconcile(obj pkgruntime.Object, qualifiedName Qualif
 		return StatusNeedsRecheck
 	}
 
-	if !s.plugins[kind].TemplateExists(qualifiedName.String()) {
+	v, ok := s.plugins[kind]
```
marun: Please prefer a descriptive variable name to a single-character names where possible.  Acceptable uses include loop indexes and receiver variables (e.g. `s` in this function)

pkg/controller/schedulingpreference/scheduler.go
```diff
@@ -0,0 +1,148 @@
...
+		// Do not support deletion yet
+		return util.StatusAllOK
+	}
+
+	// set name and group for the type config target
+	corev1a1.SetFederatedTypeConfigDefaults(typeConfig)
+	apiResource := typeConfig.GetTarget()
+	kind := typeConfig.GetTemplate().Kind
+
+	for name, _ := range schedulingtypes.SchedulingTypes() {
```
marun: I don't think it should be valid to attempt to register plugins for all type configs.  Consider adding a mapping of kind to scheduling type and only register kind plugins with an appropriate scheduler.  This will also aid in discovering which scheduler to stop a plugin for when the `FederatedTypeConfig` goes away or has propagation disabled.

@irfanurrehman Should scheduling be undertaken for a type even if propagation is disabled?  Is a separate 'schedulingEnabled' option suggested?

pkg/schedulingtypes/interface.go
```diff
@@ -36,6 +36,8 @@ type Scheduler interface {
 	HasSynced() bool
 	Stop()
 	Reconcile(obj pkgruntime.Object, qualifiedName QualifiedName) ReconciliationStatus
+
+	RegisterPlugins(kind string, apiResource *metav1.APIResource, stopChan <-chan struct{})
```
marun: I think a more appropriate name for this method is `StartPlugin`.  I don't think it makes sense for the method name to be plural, since it is intended to only start a single plugin, and it will likely be necessary to add a `StopPlugin` equivalent at some point.

pkg/controller/schedulingpreference/scheduler.go
```diff
@@ -0,0 +1,148 @@
...
+	typeConfig := cachedObj.(*corev1a1.FederatedTypeConfig)
+
+	deleted := typeConfig.DeletionTimestamp != nil
+	if deleted {
+		// Do not support deletion yet
+		return util.StatusAllOK
+	}
+
+	// set name and group for the type config target
+	corev1a1.SetFederatedTypeConfigDefaults(typeConfig)
```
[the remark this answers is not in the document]
xunpan: During my debug for this place if corev1a1.SetFederatedTypeConfigDefaults(typeConfig) is not called. The returned apiResource (from typeConfig.GetTarget()) does not have 'name' and 'group' specified. It makes federated informer failed in retrieve resource from member clusters.
```E1029 03:13:56.960870       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Unable to understand list result ...```


I check related code and did below things:
1. add 'name' and 'groups' in config/federatedtypes/deployment.yaml
    It still fails. In debug message, 'name' is not returned from typeConfig.GetTarget()
2. I check related code and found SetFederatedTypeConfigDefaults() sets correct values from FederatedTypeConfig.Name. So, I reuse this logic.

File: pkg/controller/schedulingpreference/scheduler.go (hunk 8 above)
Comment (irfanurrehman) [reply pkg/controller/schedulingpreference/scheduler.go]: I don't think there is a need of another option. There already is a feature flag `features.SchedulerPreferences`, enabling which enables the high level scheduling. If there is a need in future to allow users to enable/disable schedulers at more fine grained level, we can break this into sub features. 
It is quite appropriate to start/stop a given high level scheduler based on the propagation enabled/disabled for the target type. As you suggested a target kind to a scheduler map would be useful. This also indicates that an overall `Stop` routine for the scheduler wrt a given type is necessary right now (which seems to be missing).

We can consider another additional point (may be in some future PR, might be an overkill for now) is to check for the target type CRD being installed/not installed and then start controllers in response.

File: pkg/controller/schedulingpreference/scheduler.go (hunk 1 above)
Comment (irfanurrehman) [reply pkg/controller/schedulingpreference/scheduler.go]: I also think that `schedulingManager` is a better name and pkg rather then `schedulingPreference`.

File: cmd/controller-manager/main.go [hunk 11]
```diff
@@ -108,11 +107,9 @@ func main() {
 	federatedcluster.StartClusterController(config, fedNamespace, clusterNamespace, stopChan, clusterMonitorPeriod)
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulerPreferences) {
-		for kind, schedulingType := range schedulingtypes.SchedulingTypes() {
-			err = schedulingpreference.StartSchedulingPreferenceController(kind, schedulingType.SchedulerFactory, config, fedNamespace, clusterNamespace, targetNamespace, stopChan, true)
-			if err != nil {
-				glog.Fatalf("Error starting schedulingpreference controller for %q : %v", kind, err)
-			}
+		err = schedulingpreference.StartSchedulerController(config, fedNamespace, clusterNamespace, targetNamespace, stopChan, true)
```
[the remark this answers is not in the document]
irfanurrehman: This mistake was discussed with @tghartland and already fixed in https://github.com/kubernetes-sigs/federation-v2/pull/371. @xunpan you should also ensure to make the value `false` in real controller runs.

pkg/controller/schedulingpreference/scheduler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	worker util.ReconcileWorker
+
+	scheduler *map[string]schedulingtypes.Scheduler
+	stopChan  <-chan struct{}
+}
+
+func StartSchedulerController(config *restclient.Config, fedNamespace, clusterNamespace, targetNamespace string, stopChan <-chan struct{}, minimizeLatency bool) error {
+
+	s := make(map[string]schedulingtypes.Scheduler)
```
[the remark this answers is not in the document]
irfanurrehman: Consider a scenario like below:
- When the controller manager is deployed and the container starts `SchedulerManager` is started. There are no `TypeConfig` resources created so far.
- An admin (or another user who has enough access) creates a `TypeConfig` resource to enable propagation of configmaps.
- This controller (`SchedulerManager`/The one that you are introducing in this PR) which is watching for `TypeConfig` resources gets an event and in reconcile gets this new TypeConfig resource (with details about FederatedConfigMaps and relevant resources).
- In the reconcile of this controller logic it looks for an associated scheduler for the target type (configmap) in the `type:scheduler` map (a predefined map, with available/registered schedulers for target types). It does not find an associated scheduler, thus the reconcile is a no-op.
- An admin (or another user who has enough access) creates a `TypeConfig` resource to enable propagation of `FederatedReplicaSets`.
- Same steps as earlier, but this time the logic finds an associated scheduler (`SchedulingPreference`) for the target type (`ReplicaSet/FederatedReplicaSet`), it starts the `SchedulingPreference Controller` (the one that exists in the current code) if one is not already running. (This ideally should also check for propagation enabled/disabled, the controller should not be started if the propagation of target resource is disabled).
- If later either the propagation is disabled or the `TypeConfig` resource is removed for `FederatedReplicaSets`, the `SchedulingManager` can stop the already running `SchedulingPreference Controller`.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -0,0 +1,176 @@
...
+					c.scheduler[kind] = scheduler
+				}
+
+				glog.Infof("Start plugin with kind %s for scheduling type %s", templateKind, kind)
+				err := c.scheduler[kind].StartPlugin(templateKind, &apiResource, c.stopChan)
+				if err != nil {
+					runtime.HandleError(fmt.Errorf("Error starting plugin for %q : %v", templateKind, err))
+					return util.StatusError
+				}
+				break
```
marun: Why is it desirable to break out of this loop but not the outer one?

pkg/controller/schedulingmanager/controller.go
```diff
@@ -0,0 +1,176 @@
...
+	apiResource := typeConfig.GetTarget()
+	templateKind := typeConfig.GetTemplate().Kind
+
+	for kind, schedulingType := range schedulingtypes.SchedulingTypes() {
+		for _, fedKind := range SchedulingRegistry[kind] {
+			if fedKind == templateKind {
+				// Scheduling preference controller is started on demand
+				_, ok := c.scheduler[kind]
+				if !ok {
+					scheduler, err := schedulingpreference.StartSchedulingPreferenceController(kind, schedulingType.SchedulerFactory, c.config, c.fedNamespace, c.clusterNamespace, c.targetNamespace, c.stopChan, util.DefaultClusterAvailableDelay, util.DefaultClusterUnavailableDelay, c.minimizeLatency)
```
marun: As per a previous comment, `clusterAvailableDelay` and `clusterUnavailableDelay` should use passed-in values rather than being defaulted.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -0,0 +1,176 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+var SchedulingRegistry = map[string][]string{
```
marun: Would it ever be desirable for a federated type to have more than one scheduler targeting it?  If not, this mapping should probably be template kind to scheduling kind.

cc: @irfanurrehman

pkg/controller/schedulingmanager/controller.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return err
+	}
+	glog.Infof("Starting scheduler controller")
+	controller.Run(stopChan)
+
+	return nil
+}
+
+func newController(client corev1alpha1client.CoreV1alpha1Interface, config *restclient.Config, fedNamespace, clusterNamespace, targetNamespace string, stopChan <-chan struct{}, minimizeLatency bool) (*SchedulerController, error) {
```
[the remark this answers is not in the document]
xunpan: I found several controllers, e.g.servicedns.StartController, federatedtypeconfig.StartController, have similar code. I guess it is there for further enhancement because main() logic has error checking for related fail condition too.

pkg/controller/schedulingmanager/controller.go
xunpan: Currently, we have "ReplicaSchedulingPreference" scheduler.  schedulingtypes.RegisterSchedulingType() support multiple schedulers. 

If a federated resource can be registered into more than 1 scheduler, break out of inner loop should be fine. If the resource cannot be registered into multiple scheduler (may introduce any conflict), it will not be defined in our SchedulingRegistry. It is also fine break from inner loop.

Of course, if we can sure there is no multiple registration case, break out of outside loop is fine.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -0,0 +1,176 @@
...
+		return util.StatusError
+	}
+	if !exist {
+		return util.StatusAllOK
+	}
+	typeConfig := cachedObj.(*corev1a1.FederatedTypeConfig)
+
+	deleted := typeConfig.DeletionTimestamp != nil
+	if deleted {
+		// Do not support deletion yet
```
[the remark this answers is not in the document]
xunpan: issue [386](https://github.com/kubernetes-sigs/federation-v2/issues/386) is logged to trace deletion issue.

pkg/controller/schedulingmanager/controller.go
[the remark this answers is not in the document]
marun: This is an example of YAGNI (You Ain't Gonna Need It).  If no error is currently being returned, the signature should reflect that, and an error can be added back in the future _if_ it becomes necessary.

pkg/controller/schedulingmanager/controller.go
marun: I value consistency, and I think either both loops should be broken out of or none should to avoid confusing future maintainers.

marun:
The test failure is indeed a flake, sorry about that.  I'm hoping to get @irfanurrehman's feedback on a [question](
https://github.com/kubernetes-sigs/federation-v2/pull/366#discussion_r230461325) of mine before merge.

pkg/controller/schedulingmanager/controller.go
irfanurrehman: Apologies, missed this particular comment earlier.
As far as I know and can think, I can't see a scenario where a single federated type is controlled by more then 1 scheduler at the same time. It would obviously result in scheduling conflicts. I can think of a scenario where users don't like the high level scheduler that's packed with open source release and they write their custom scheduler to replace it.
On the other hand a single scheduler can target and/or manipulate multiple federated types (as is the case right now for FederatedDeployments and FederatedReplicaSets being targeted by FederatedRSP).

marun: @irfanurrehman Thanks!

@xunpan As per this exchange, please switch to mapping federated type to the appropriate scheduler.  This should allow scheduler start to be simplified.

xunpan:
@marun  These 3 commits are squashed. Thanks for your helpful comments in understanding related logic better. I'm glad to contribute to this community.

marun:
Thank you!

/approve
/lgtm

PR #369, opened 2018-10-27
Fix validation

roytman:
Fix validation verification and remove `--validate=false` from `kubectl` commands

### DIFF
--- a/docs/federating_crds.md
+++ b/docs/federating_crds.md
@@ -33,8 +33,9 @@ spec:
         metadata:
           type: object
         spec:
-          data:
-            type: string
+          properties:
+            data:
+              type: string
         status:
           type: object
   names:
@@ -48,10 +49,10 @@ spec:
 Make sure to install target CRDs on all member clusters otherwise this example will not work as expected.
 
 ```shell
-$ kubectl --validate=false apply -f ./example/crd/bar_crd.yaml --context=cluster1
+$ kubectl apply -f ./example/crd/bar_crd.yaml --context=cluster1
 customresourcedefinition.apiextensions.k8s.io/bars.example.io created
 
-$ kubectl --validate=false apply -f ./example/crd/bar_crd.yaml --context=cluster2
+$ kubectl apply -f ./example/crd/bar_crd.yaml --context=cluster2
 customresourcedefinition.apiextensions.k8s.io/bars.example.io created
 ```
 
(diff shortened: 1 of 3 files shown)


docs/federating_crds.md
```diff
@@ -158,29 +150,22 @@ spec:
...
                     type: string
                   data:
                     type: string
-                type: object
-              type: array
-          type: object
         status:
           type: object
   version: v1alpha1
-status:
```
gyliu513: Why removing this? I found that all of the CRDs in `hack/install-latest.yaml` do have the status field.

poothia: @roytman so i guess removing https://github.com/kubernetes/cluster-registry/blob/f7a48a6cb15ee4c9ba02c368cf0428c4c920bfdf/examples/cluster.yaml#L13 and https://github.com/kubernetes/cluster-registry/blob/f7a48a6cb15ee4c9ba02c368cf0428c4c920bfdf/examples/cluster-two.yaml#L13

should solve https://github.com/kubernetes/cluster-registry/issues/255 ??

roytman: Actually, removing lines 89-93 from https://github.com/kubernetes/cluster-registry/blob/f7a48a6cb15ee4c9ba02c368cf0428c4c920bfdf/cluster-registry-crd.yaml#L89 
solves the validation issue. 
This example when the lines are not commented out : 
```
kubectl apply -f crd/clusterregistry.yaml
error: error validating "crd/clusterregistry.yaml": error validating data: [ValidationError(CustomResourceDefinition.status): missing required field "conditions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus, ValidationError(CustomResourceDefinition.status): missing required field "storedVersions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionStatus]; if you choose to ignore these errors, turn validation off with --validate=false
```
Now I commented them out:
kubectl apply -f crd/clusterregistry.yaml
customresourcedefinition.apiextensions.k8s.io/clusters.clusterregistry.k8s.io created

This example from service-registry, but it is relevant to this PR as well.

pmorie:
@sohankunkerkar 

I missed that these example CRDs contain a status. Status should not be part of a resource fed to kubectl create.

gyliu513:
> I missed that these example CRDs contain a status. Status should not be part of a resource fed to kubectl create.

@pmorie I found that the `status` field in CRD was generated by `kubebuilder`, should this be fixed in kubebuider?

PR #371, opened 2018-10-29
Add flags to controller-manager to set cluster availability delays

### LINKED ISSUE #356: Make RSP clusterUnavailableDelay and clusterAvailableDelay configurable
Now, when I enable RSP, if I did a test with the rebalance feature. After deploying on two clusters, I deliberately brought down one of the clusters, then I need to wait at least 1min for all of my pods balanced to the running cluster, it is better to make the clusterUnavailableDelay/clusterAvailableDelay configurable, so that customer can configure those values based on their cluster scale.

Please refer to https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/controller/schedulingpreference/controller.go#L99 for the hard-code clusterUnavailableDelay and clusterAvailableDelay.

/cc @irfanurrehman can you please add `good-first-issue` for this issue? Thanks.

tghartland:
I have tested using negative values for the delay arguments, it works just as if the delay were set to 0 seconds.

Closes #356.

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -147,7 +147,7 @@ func main() {
 	}
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.PushReconciler) {
-		err = federatedtypeconfig.StartController(config, fedNamespace, clusterNamespace, targetNamespace, stopChan)
+		err = federatedtypeconfig.StartController(config, fedNamespace, clusterNamespace, targetNamespace, stopChan, clusterAvailableDelay, clusterUnavailableDelay)
 		if err != nil {
 			glog.Fatalf("Error starting federated type config controller: %v", err)
 		}
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -54,8 +54,8 @@ var featureGates map[string]bool
 var fedNamespace string
 var clusterNamespace string
 var limitedScope bool
-var clusterAvailableDelay int
-var clusterUnavailableDelay int
+var clusterAvailableDelay time.Duration
+var clusterUnavailableDelay time.Duration
 var installCRDs = flag.Bool("install-crds", true, "install the CRDs used by the controller as part of startup")
 
 // Controller-manager main.
@@ -66,8 +66,8 @@ func main() {
 	flag.StringVar(&fedNamespace, "federation-namespace", util.DefaultFederationSystemNamespace, "The namespace the federation control plane is deployed in.")
 	flag.StringVar(&clusterNamespace, "registry-namespace", util.MulticlusterPublicNamespace, "The cluster registry namespace.")
 	flag.BoolVar(&limitedScope, "limited-scope", false, "Whether the federation namespace will be the only target for federation.")
-	flag.IntVar(&clusterAvailableDelay, "cluster-available-delay", util.DefaultClusterAvailableDelay, "Time in seconds to wait before reconciling on a healthy cluster.")
-	flag.IntVar(&clusterUnavailableDelay, "cluster-unavailable-delay", util.DefaultClusterUnavailableDelay, "Time in seconds to wait before giving up on an unhealthy cluster.")
+	flag.DurationVar(&clusterAvailableDelay, "cluster-available-delay", util.DefaultClusterAvailableDelay, "Time to wait before reconciling on a healthy cluster.")
+	flag.DurationVar(&clusterUnavailableDelay, "cluster-unavailable-delay", util.DefaultClusterUnavailableDelay, "Time to wait before giving up on an unhealthy cluster.")
 
 	flag.Parse()
 	if *verFlag {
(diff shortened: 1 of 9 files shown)


cmd/controller-manager/main.go
```diff
@@ -64,6 +66,8 @@ func main() {
 	flag.StringVar(&fedNamespace, "federation-namespace", util.DefaultFederationSystemNamespace, "The namespace the federation control plane is deployed in.")
 	flag.StringVar(&clusterNamespace, "registry-namespace", util.MulticlusterPublicNamespace, "The cluster registry namespace.")
 	flag.BoolVar(&limitedScope, "limited-scope", false, "Whether the federation namespace will be the only target for federation.")
+	flag.IntVar(&clusterAvailableDelay, "cluster-available-delay", util.DefaultClusterAvailableDelay, "Time in seconds to wait before reconciling on a healthy cluster.")
```
marun: Please use `flag.DurationVar`

cmd/controller-manager/main.go
```diff
@@ -94,6 +98,8 @@ func main() {
 
 	glog.Infof("Federation namespace: %s", fedNamespace)
 	glog.Infof("Cluster registry namespace: %s", clusterNamespace)
+	glog.Infof("Cluster available delay: %d", clusterAvailableDelay)
```
marun: Do these need to be logged?  The namespaces are logged for troubleshooting purposes since they may be set from either args or metadata and getting it wrong will render the control plane unusable.

cmd/controller-manager/main.go
```diff
@@ -109,15 +115,15 @@ func main() {
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulerPreferences) {
 		for kind, schedulingType := range schedulingtypes.SchedulingTypes() {
-			err = schedulingpreference.StartSchedulingPreferenceController(kind, schedulingType.SchedulerFactory, config, fedNamespace, clusterNamespace, targetNamespace, stopChan, true)
+			err = schedulingpreference.StartSchedulingPreferenceController(kind, schedulingType.SchedulerFactory, config, fedNamespace, clusterNamespace, targetNamespace, stopChan, clusterAvailableDelay, clusterUnavailableDelay, false)
```
marun: Good catch on not minimizing latency for non-test execution!

test/integration/framework/controller.go
```diff
@@ -54,7 +54,7 @@ func NewServiceDNSControllerFixture(tl common.TestLogger, config *restclient.Con
 	f := &ControllerFixture{
 		stopChan: make(chan struct{}),
 	}
-	err := servicedns.StartController(config, fedNamespace, clusterNamespace, targetNamespace, f.stopChan, true)
+	err := servicedns.StartController(config, fedNamespace, clusterNamespace, targetNamespace, f.stopChan, 20, 60, true)
```
marun: These and all other test-invoked controllers should be using the default constants rather than magic numbers.

cmd/controller-manager/main.go
gyliu513: +1 to have the log here, as it will help for debugging.

cmd/controller-manager/main.go
```diff
@@ -109,15 +115,15 @@ func main() {
...
 			if err != nil {
 				glog.Fatalf("Error starting schedulingpreference controller for %q : %v", kind, err)
 			}
 		}
 	}
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.CrossClusterServiceDiscovery) {
-		err = servicedns.StartController(config, fedNamespace, clusterNamespace, targetNamespace, stopChan, false)
+		err = servicedns.StartController(config, fedNamespace, clusterNamespace, targetNamespace, stopChan, clusterAvailableDelay, clusterUnavailableDelay, false)
```
irfanurrehman: @shashidharatd please refer [this comment](https://github.com/kubernetes-sigs/federation-v2/issues/356#issuecomment-433157872), and confirm if its ok to have common delays for these three controllers. I have a feeling that dns based stuff might need larger timeouts. Although it might be ok to set them to common values, but probably the documentation of this flag provides a min value suitable to dns related code (for example, `setting the value lesser then xxxx might hinder with correct name resolution`)

tghartland:
@marun thanks for the feedback. Regarding the sync controller, it was not suggested in the issue that it be changed (and I had actually not seen that it should use the values). It is already different from the others in that it does not have the minimizeLatency value passed through to it, should this be added as well to make it consistent with the other controllers?

marun:
@tghartland I would like to see the sync controller and type config controller updated to accept the new values to ensure consistent behavior across all controllers.  Otherwise, looks good.

tghartland:
The delay values are now passed through the federatedtypeconfig controller to get to the sync controller, as requested.

marun:
LGTM.  Merge is pending @shashidharatd response to @irfanurrehman's [comment](
https://github.com/kubernetes-sigs/federation-v2/pull/371#discussion_r229226015).

cmd/controller-manager/main.go
shashidharatd: I think using common flags makes more sense to me. These flags mainly control when the reconciliation triggered and all the controllers are similar

cmd/controller-manager/main.go
shashidharatd: @tghartland, Please fix this and i think its ready to be merged.

PR #372, opened 2018-10-29
Move the package comments above package

mooncak:
1.Move the package comments above package
2.Delete the duplication "imports"

### DIFF
--- a/hack/imports.go
+++ b/hack/imports.go
@@ -14,10 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package hack
-
 /*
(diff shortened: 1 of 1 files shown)


marun:
@mooncak This is a generated file.  I don't think there's value in manually cleaning it up - I suggest targeting your efforts at [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) which is the source of the generated code.

mooncak:
> @mooncak This is a generated file. I don't think there's value in manually cleaning it up - I suggest targeting your efforts at [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) which is the source of the generated code.

ok, I will change the source files, thanks.

PR #380, opened 2018-11-01
Add docs/OWNERS

pmorie:
Adds `docs/OWNERS`, which most notably:

- makes @gyliu513 an approver in `docs/`
- makes @danehans a reviewer in `docs/`

Thanks a lot to both @gyliu513 and @danehans for their contributions to this area!

### DIFF
--- a/docs/OWNERS
+++ b/docs/OWNERS
@@ -1,0 +1,13 @@
+reviewers:
+  - gyliu513
+  - font
+  - marun
+  - irfanurrehman
+  - danehans
+approvers:
+  - font
+  - irfanurrehman
+  - marun
+  - pmorie
+  - shashidharatd
+  - gyliu513


pmorie:
> Is it necessary to add everyone to the nested OWNERS vs inheriting from the root one?

Good question; I did some searching and couldn't find a place where the exact semantics of OWNERS files re: precedence and nesting are defined. I've asked a question in sig-contribex channel; we should make sure the semantics are written down in a specific place explicitly.

marun:
Ugh, the namespace placement test is really flakey. :(

marun:
LGTM.  Waiting for @irfanurrehman or @shashidharatd to approve.

irfanurrehman:
@pmorie Lets order the list and merge this. You can figure out the nested `OWNERS` file stuff when you can later.

shashidharatd:
AFAIK previously, you don't have to copy the OWNERS in base directory, just adding new OWNERS to the sub-directory is fine.
/lgtm

PR #382, opened 2018-11-02
Refactor all controllers to accept a common config struct

marun:
This is intended to simplify maintenance of controllers.  Inspired by an in-progress refactor of the rsp testing.

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -88,10 +88,10 @@ func main() {
 
 	stopChan := signals.SetupSignalHandler()
 
-	config := configlib.GetConfigOrDie()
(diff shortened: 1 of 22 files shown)


shashidharatd:
The changes looks good to me. As a subsequent action it would be good to see all the arguments to controller be stored in a structure called ControllerConfig instead of global variables as they are today. It would also facilitate moving towards Component Config which is used in all the main controllers of k8s.

marun:
@shashidharatd I agree, component config is overdue, and would simplify integration with 3rd party federation controllers.  I'll add that shortly.

PTAL - I've made some naming changes that I think are an improvement.

shashidharatd:
/lgtm
Thanks @marun !

PR #387, opened 2018-11-05
Add missing weight to user guide RSP example

### LINKED ISSUE #374: User guide RSP example missing weight value
In the [final RSP example](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/userguide.md#distribute-replicas-evenly-in-all-clusters-however-not-more-then-20-in-c) no weight is specified for cluster C. The weight defaults to 0 so the following scenarios are impossible as C would never be scheduled on. ```weight: 1``` should be added.

tghartland:
Fixes #374.

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -527,6 +527,7 @@ spec:
       weight: 1
     "C":
       maxReplicas: 20
+      weight: 1
 ```
 Possible scenarios
 


docs/userguide.md
```diff
@@ -527,6 +527,7 @@ spec:
       weight: 1
     "C":
       maxReplicas: 20
+      weight: 1
```
gyliu513: Should not the following already covered "C"'s weight?

```
"*":
  weight: 1
```

tghartland: That's not what I find when testing it.

https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/controller/util/planner/planner.go#L88
"C" exists so it doesn't even consider "*", which means C's weight gets the zero value of an int.

tghartland: We could change it to inherit values from "*", but then what about maxReplicas? Once it's set in "*" there's no way to unset it for any other clusters (other than setting it to 99999 and assuming that's enough).

gyliu513: Thanks @tghartland , I see, yes, we need to set `weight` here.

PR #390, opened 2018-11-06
Fix gofmt errors

marun:
Not sure  how these slipped through, but I'm going to self-approve because master is now broken.

### DIFF
--- a/test/e2e/crd.go
+++ b/test/e2e/crd.go
@@ -47,7 +47,7 @@ var _ = Describe("Federated CRD resources", func() {
 		true,
 		false,
 	}
-	for i, _ := range namespaceScoped {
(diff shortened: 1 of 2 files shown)


marun:
@poothia I'm guessing there was code added after your PR was last tested.  Not a great start to checking gofmt, breaking CI for non-functional formatting issues.

poothia:
@marun yes you are right, i checked my branch, it did not have these changes, they must be added after the PR was tested.

so i think we can manually run the scripts once in a while and commit changes, or we can revert https://github.com/kubernetes-sigs/federation-v2/pull/393 so that new commits fail at CI, if they do not adhere to gofmt.

PR #391, opened 2018-11-06
Remove test-specific setup from the e2e framework

marun:
Previously the e2e framework was responsible for setting and tearing down controller fixture even when it wasn't common to more than one test.  This PR moves the responsibility for setup and teardown to the tests to keep the framework simple and focused on shared utility.


shashidharatd:
This is way forward and LGTM.

marun:
Rebased

PR #393, opened 2018-11-06
Disable gofmt check until it can be made reliable

marun:
cc: @poothia

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -104,8 +104,10 @@ echo "Downloading test dependencies"
 echo "Checking initial state of working tree"
 check-git-state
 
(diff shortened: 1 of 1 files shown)


poothia:
@marun i think we should revert this... ref: https://github.com/kubernetes-sigs/federation-v2/pull/390#issuecomment-436140394
thoughts??

marun:
@poothia I'm open to re-enabling the gofmt check.

poothia:
@marun ok, i'll open a PR to re-enable it... and fix any changes done, that might have not been gofmt compliant, but we need to get it merged before some other non-gofmt changes leak into master, and start breaking the CI again.

Issue #208, opened 2018-08-15
Migrate code structure to kubebuilder v1.0.x


shashidharatd:
- kubebuilder 1.0.0 introduces big architectural change in the project/code structure. 1.0.0 projects uses client and controller libraries developed under controller-runtime repo. It uses tools from controller-tools repo for scaffolding and manifests (CRD/RBAC) generation.
- 1.0.0 will continue to support projects generated using older version (pre 1.0.0 version) of Kubebuilder, but we would highly recommend migrating to newer version. Follow instructions for migration your project to 1.0.0 version.

I know this is not a urgent issue for us as of now. just adding it to backlog.
/cc @marun @gyliu513 @kubernetes-sigs/federation-wg

gyliu513:
Thanks @shashidharatd .

There is a WIP PR here https://github.com/kubernetes-sigs/federation-v2/pull/148 , hope we can get more comments from @marun to see when we can do the migration.

marun:
@gyliu513 Upgrading the binaries is fine.  Migrating to the new client-less framework should wait until after #60, which I'll be tackling after #216 is complete.

gyliu513:
@marun yes, I only created PR https://github.com/kubernetes-sigs/federation-v2/pull/250 to update binaries based on your comments.

marun:
I think the next steps for this issue are: 

 - Replacing all calls to the generated clients with equivalent calls to the `controller-runtime` generic client
 - Removing the generated clients (and the `+genclient` annotation on the types to prevent regeneration)

Here is an example of a call to a generated client:

```golang
fedClient := fedclientset.NewForConfigOrDie(kubeConfig).
typeConfig, err := fedClient.CoreV1alpha1().FederatedTypeConfigs(namespace).Get(name, metav1.GetOptions{})
```

Here is an equivalent call with the generic client:

```golang
// Assumes import of: 
//    - "context"
//    - "sigs.k8s.io/controller-runtime/pkg/client"

genericClient, err := client.New(kubeConfig, client.Options{})
if err != nil {
...
}
typeConfig := &fedv1a1.FederatedTypeConfig{}
key := client.ObjectKey{Name: name, Namespace: namespace}
err := genericClient.Get(context.Background(), key, typeConfig)
```

shashidharatd:
This is addressed with https://github.com/kubernetes-sigs/federation-v2/pull/716

PR #394, opened 2018-11-06
Enable LeaderElect for federation controller

fisherxu:
Enable LeaderElect for federation controller, then can deployed as multiple instances.
/cc @marun @font @irfanurrehman @shashidharatd

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -757,7 +757,7 @@
   version = "kubernetes-1.12.4"
 
 [[projects]]
-  digest = "1:e29b9d05002abc1a54d1a0f392041e8bec05bcdd861a81f3b87d7e3a4eae9398"
+  digest = "1:1bec3939c715c6a9cda0225353d2fad4b6b95932f0a628319e8916052d3ad8de"
   name = "k8s.io/apimachinery"
   packages = [
     "pkg/api/errors",
@@ -794,6 +794,7 @@
     "pkg/util/runtime",
     "pkg/util/sets",
     "pkg/util/strategicpatch",
+    "pkg/util/uuid",
     "pkg/util/validation",
     "pkg/util/validation/field",
     "pkg/util/wait",
@@ -820,7 +821,7 @@
   revision = "92634e4423af1b6ad84ab94b43948b0e504a3edd"
 
 [[projects]]
-  digest = "1:b619f9388b3b3f7ce5364c18995e45c951c50946f072bbafda2d17c9645755df"
+  digest = "1:845b33393a6dbcbf40198317b7c8f342ee4794eae8fa88bc1ff57b41bc4afce9"
   name = "k8s.io/client-go"
   packages = [
     "discovery",
@@ -951,6 +952,8 @@
     "tools/clientcmd/api",
     "tools/clientcmd/api/latest",
     "tools/clientcmd/api/v1",
+    "tools/leaderelection",
+    "tools/leaderelection/resourcelock",
     "tools/metrics",
     "tools/pager",
     "tools/record",
@@ -1072,6 +1075,7 @@
     "k8s.io/apimachinery/pkg/util/net",
     "k8s.io/apimachinery/pkg/util/runtime",
     "k8s.io/apimachinery/pkg/util/sets",
+    "k8s.io/apimachinery/pkg/util/uuid",
     "k8s.io/apimachinery/pkg/util/wait",
     "k8s.io/apimachinery/pkg/util/yaml",
     "k8s.io/apimachinery/pkg/watch",
@@ -1090,6 +1094,8 @@
     "k8s.io/client-go/tools/cache",
     "k8s.io/client-go/tools/clientcmd",
     "k8s.io/client-go/tools/clientcmd/api",
+    "k8s.io/client-go/tools/leaderelection",
+    "k8s.io/client-go/tools/leaderelection/resourcelock",
     "k8s.io/client-go/tools/record",
     "k8s.io/client-go/util/cert",
     "k8s.io/client-go/util/flowcontrol",
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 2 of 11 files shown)


charts/federation-v2/values.yaml
```diff
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-## Configuration values for federation v2 controllermanager statefulset.
+## Configuration values for federation v2 controllermanager deployment.
```
gyliu513: Please check https://github.com/kubernetes-sigs/federation-v2/pull/245#discussion_r218828597 for why we are using `sts` here.

/cc @marun

fisherxu: thanks for reminding @gyliu513 :) I have checked the comment, and that's in the absence of leader election. We need deploy multi instances with leader election, and to the multi instances of controller, deployment should be fine like kube-controller-manager.

marun:
I think this kind of feature deserves some discussion, maybe even a design document, before implementation is proposed.

fisherxu:
>  I'd like to have the use cases this is intended to target clearly documented 

@marun Has raised the issue #402 to describe some use cases, and we can have some discussion there :)

cmd/controller-manager/main.go
```diff
@@ -119,44 +151,93 @@ func main() {
...
+	}
+
+	leaderelection.RunOrDie(leaderelection.LeaderElectionConfig{
+		Lock:          rl,
+		LeaseDuration: leaseDuration,
+		RenewDeadline: renewDeadline,
+		RetryPeriod:   retryPeriod,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: run,
+			OnStoppedLeading: func() {
```
marun: The leader election callback would appear to ensure that a master is started, but how will a master be stopped if it loses the leadership?  It appears that a master once started would never be stopped since there is nothing that closes the stop channel.

cmd/controller-manager/main.go
```diff
@@ -119,44 +151,93 @@ func main() {
...
+				glog.Fatalf("leaderelection lost")
```
pmorie: On loss of leadership, it seems like it would be sufficient to send on the stop channel passed to run.

cmd/controller-manager/main.go
marun: That suggests that the stop channel for the non-leader elect path be created in the `if !enableLeaderElection` so that it's clear that it won't be used outside of that block.

marun:
@fisherxu Do you intend to follow up on this PR?  It's been dormant for over a month.

fisherxu:
@marun Sorry for the delay, and I think this PR should depend on this PR [#57932](https://github.com/kubernetes/kubernetes/pull/57932) to pass stop channel to run function.
And this PR is merged in client-go v1.12.0, so now the client-go we used(v1.10.1) really can't pass the stop channel.
Do we have any plane to upgrade the client-go? :)

marun:
@fisherxu Thank you for following up!  

I'm ok with upgrading to the latest client-go if it's possible.  I think you mentioned on slack that upgrading client-go may depend on kubebuilder upgrading?

fisherxu:
@marun I have checked today.
the code(pkg/client/clientset/versioned/typed/GROUP/VERSION/fake) generated by kubebuilder(v1.0.4) is not compatible with client-go(kubernetes-1.13.0), but is  compatible with client-go(kubernetes-1.12.0).

the controller-runtime (only used in e2e now) use the v0.1.1 version which is generated by kubebuilder(v1.0.4) , the controller-runtime(v0.1.1) is not compatible with client-go(kubernetes-1.12.0), so can we upgrade it to v0.1.8 manuall as in #530 , through it's generated by kubebuilder(v1.0.4)?

If we upgrade the kubebuilder to 1.0.5 or .6, project maybe will have big changes... and kubebuilder(v1.0.4) can only support up to  client-go(kubernetes-1.12.0).

marun:
@fisherxu What changes to the project do you expect with kubebuilder > 1.04?  fedv2 doesn't use any kb-generated controllers, we use it to bootstrap new types and maintain the generated deepcopy, clients/listers/informers, and crd yaml.  @shashidharatd has agreed to work on removing the need for generated clients/listers/informers, so only the deepcopy and crd yaml will remain.

fisherxu:
@marun @shashidharatd The func `NewPatchSubresourceAction` in `pkg/client/clientset/versioned/typed/GROUP/VERSION/fake` generayed by kubebuilder(v1.0.4) [(line for example)](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/client/clientset/versioned/typed/core/v1alpha1/fake/fake_federatedtypeconfig.go#L134) don't pass the PatchType parameter. But in the latest client-go[1.13.+ this func](https://github.com/kubernetes/client-go/blob/kubernetes-1.13.0/testing/actions.go#L191) need the PatchType parameter.  So upgrade client-go to 1.13 will make the CI failed.  Have any suggestion here? :)

marun:
@fisherxu Does that mean that running `kubebuilder generate` with a version of kubebuilder > 1.04 will fix the problem?  If it's more complicated than that, it may be desirable to wait until the [generated client is removed from the tree](
https://github.com/kubernetes-sigs/federation-v2/issues/208#issuecomment-450881311).

cmd/controller-manager/main.go
fisherxu: @marun Have updated, now we can pass the stopChan through the ctx to the run func. I have checked in my environment, it works fine :) PTAL.

fisherxu:
> If it's more complicated than that, it may be desirable to wait until the generated client is removed from the tree.

@marun  Agree,  also desirable to wait until the generated client is removed from the tree.

marun:
@fisherxu Regarding testing the changes in this PR, I suggest writing a managed (fixture is in-process and test-managed) e2e test in which 2 controller managers are started in-process (using etcd fixture), one is killed, and the test validates that the second controller is elected leader.

marun:
Also, removal of the generated client is complicated by the apparent lack of support of the `Watch` method in controller-runtime's generic client.  Do you know of an alternative way of watching without a generated client (without having to adopt the new and largely untested controller scheme of the newer versions of kubebuilder)?  Or would it be possible to update the generated clients to be compatible?

shashidharatd:
Hi @fisherxu, Shall we close this pr in favour of https://github.com/kubernetes-sigs/federation-v2/pull/632 ?

fisherxu:
Sure, Let's close this pr in favour of #632.

shashidharatd:
Thanks @fisherxu

PR #397, opened 2018-11-06
install.yaml: fix have coherent namespace on tuto

mikefaille:
`federation-system` should be used on install.yaml since there is a statefullset inside this config where it should having a correct permission for RBAC.


marun:
I think I'd rather see the 'manual' section simply removed, since it represents a non-executable maintenance burden vs the deployment script and helm options.  Would you be willing to update this PR accordingly?

PR #399, opened 2018-11-07
Migrate ReplicaSchedulingPreferences controller to the dynamic client

marun:
This PR migrates the RSP controller to use the dynamic client both in anticipation of generic overrides and the removal of primitive CRDs from the codebase in favor of generating them with `kubefed2 federate`.


marun:
@xunpan @irfanurrehman Please take note of the commit named `Refactor type registration for RSP`.

marun:
@irfanurrehman I would really appreciate your giving this PR a thorough review.  I want to make sure I was able to maintain your intent for how this controller evolves.

irfanurrehman:
> @irfanurrehman I would really appreciate your giving this PR a thorough review. I want to make sure I was able to maintain your intent for how this controller evolves.

@marun Apologies for not being able to look at PRs this week. I was on business travel to kubecon china and could not spend time at all to look at PRs. I will be spending time on this tomorrow for sure.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -164,7 +163,7 @@ func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.R
 	}
 
 	glog.Infof("Start plugin with kind %s for scheduling type %s", templateKind, kind)
-	err = c.scheduler[kind].StartPlugin(templateKind, &apiResource, c.stopChan)
+	err = c.scheduler[kind].StartPlugin(typeConfig, c.stopChan)
```
irfanurrehman: not necessarily relevant to this PR; however came across this thought while looking at this code wrt to your changes in https://github.com/kubernetes-sigs/federation-v2/pull/427.
I think its possible to create multiple `TypeConfigs` for the same target kind, (perhaps a possible mistake by whoever creates them). For example a `TypeConfig` was created with name `deployments.apps` for `deployment` as target as part of the control plane deployment automation. Later an admin creates a `TypeConfig` with name `deployments` for same target. Multiple schedulers would be created here, (and possibly multiple sync controllers for same target). Does it make sense to handle it in some way, or this seems to be a remote possibility, or .... ?

pkg/schedulingtypes/plugin.go
```diff
@@ -129,13 +123,28 @@ func (p *Plugin) HasSynced() bool {
...
+	}
+
 	clusters, err := p.targetInformer.GetReadyClusters()
 	if err != nil {
 		runtime.HandleError(fmt.Errorf("Failed to get ready clusters: %v", err))
 		return false
 	}
 
-	return p.targetInformer.GetTargetStore().ClustersSynced(clusters)
+	if !p.targetInformer.GetTargetStore().ClustersSynced(clusters) {
```
irfanurrehman: Is this change for readability or am missing something?

irfanurrehman:
@marun Only few queries as comments; The PR seems to be achieving the same functionality, as to what it was earlier, but with cleaner code. 
Thanks for doing this.

pkg/controller/schedulingmanager/controller.go
marun: Good catch.  I've internalized the expectation that type configs are named `<target plural name>.<target group name>`, such that there will never be more than one type config for a given target for a given federation deployment.  When I checked that assumption, though, I realized that no validation exists to preclude this possibility.  I've added #433 to track resolution.

pkg/schedulingtypes/plugin.go
marun: Yes. When I was adding the checks for the template/placement/override controllers having synced, I thought it made sense for every failed check to return false and then a default true be returned so that no matter the ordering or changes to the checks the true would always be the default.  I'm happy to revert if you'd like.

irfanurrehman: Makes more sense this way. Thanks for explaining.

pkg/controller/schedulingmanager/controller.go
irfanurrehman: Ack.

irfanurrehman:
@marun thanks for this PR.
/lgtm
/approve

Issue #314, opened 2018-10-08
Make it easier to understand when the target type is not present in all federated clusters


pmorie:
Currently, there is not an easy way to determine when a target type is not present in one or more federated clusters. Example controller logs:

```console
I1008 13:33:49.657514       1 controller.go:212] Running reconcile FederatedTypeConfig for "federation-system/bars.example.io"
I1008 13:33:49.668942       1 controller.go:125] Starting sync controller for "FederatedBar"
I1008 13:33:49.668992       1 controller.go:290] Started sync controller for "FederatedBar"
I1008 13:33:49.669008       1 controller.go:212] Running reconcile FederatedTypeConfig for "federation-system/bars.example.io"
I1008 13:33:49.689848       1 federated_informer.go:211] Cluster federation-system/cluster1 is ready
I1008 13:33:49.697323       1 federated_informer.go:211] Cluster federation-system/cluster2 is ready
E1008 13:33:49.700713       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Failed to list <nil>: the server could not find the requested resource (get bars.example.io)
E1008 13:33:50.703301       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Failed to list <nil>: the server could not find the requested resource (get bars.example.io)
E1008 13:33:51.710217       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Failed to list <nil>: the server could not find the requested resource (get bars.example.io)
E1008 13:33:52.712783       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Failed to list <nil>: the server could not find the requested resource (get bars.example.io)
E1008 13:33:53.715131       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Failed to list <nil>: the server could not find the requested resource (get bars.example.io)
E1008 13:33:54.775163       1 reflector.go:205] github.com/kubernetes-sigs/federation-v2/pkg/controller/util/federated_informer.go:430: Failed to list <nil>: the server could not find the requested resource (get bars.example.io)
```

Should this be an element of status in `FederatedTypeConfig`, perhaps?

gyliu513:
[...]

PR #403, opened 2018-11-07
Set observedGeneration in the federatedTypeConfig controller

sohankunkerkar:
Following up #345 
 FederatedTypeConfig controller should set the observedGeneration in its control loop.

### DIFF
--- a/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
+++ b/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
@@ -95,11 +95,27 @@ type OverridePath struct {
 	Path string `json:"path"`
 }
 
+// ControllerStatus defines the current state of the controller
+type ControllerStatus string
+
+const (
+	// ControllerStatusRunning means controller is in "running" state
+	ControllerStatusRunning ControllerStatus = "Running"
+	// ControllerStatusNotRunning means controller is in "notrunning" state
+	ControllerStatusNotRunning ControllerStatus = "NotRunning"
+)
+
 // FederatedTypeConfigStatus defines the observed state of FederatedTypeConfig
 type FederatedTypeConfigStatus struct {
 	// ObservedGeneration is the generation as observed by the controller consuming the FederatedTypeConfig.
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// PropagationController tracks the status of the sync controller.
+	// +optional
+	PropagationController ControllerStatus `json:"propagationController,omitempty"`
+	// StatusController tracks the status of the status controller.
+	// +optional
+	StatusController ControllerStatus `json:"statusController,omitempty"`
 }
 
 // +genclient
--- a/pkg/controller/federatedtypeconfig/controller.go
+++ b/pkg/controller/federatedtypeconfig/controller.go
@@ -145,6 +145,15 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	limitedScope := c.controllerConfig.TargetNamespace != metav1.NamespaceAll
 	if limitedScope && syncEnabled && !typeConfig.GetNamespaced() {
 		glog.Infof("Skipping start of sync & status controller for cluster-scoped resource %q.  It is not required for a namespaced federation control plane.", typeConfig.GetTemplate().Kind)
+
+		typeConfig.Status.ObservedGeneration = typeConfig.Generation
+		typeConfig.Status.PropagationController = corev1a1.ControllerStatusNotRunning
+		typeConfig.Status.StatusController = corev1a1.ControllerStatusNotRunning
+		_, err = c.client.FederatedTypeConfigs(typeConfig.Namespace).UpdateStatus(typeConfig)
+		if err != nil {
+			runtime.HandleError(fmt.Errorf("Could not update status fields of the CRD: %q: %v", key, err))
+			return util.StatusError
+		}
 		return util.StatusAllOK
 	}
 
@@ -160,6 +169,7 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 		if statusRunning {
 			c.stopController(statusKey, statusStopChan)
 		}
+
 		err := c.removeFinalizer(typeConfig)
 		if err != nil {
 			runtime.HandleError(fmt.Errorf("Failed to remove finalizer from FederatedTypeConfig %q: %v", key, err))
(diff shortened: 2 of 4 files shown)


pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -145,6 +145,15 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	limitedScope := c.controllerConfig.TargetNamespace != metav1.NamespaceAll
 	if limitedScope && syncEnabled && !typeConfig.GetNamespaced() {
 		glog.Infof("Skipping start of sync & status controller for cluster-scoped resource %q.  It is not required for a namespaced federation control plane.", typeConfig.GetTemplate().Kind)
+
+		typeConfig.Status.ObservedGeneration = typeConfig.Generation
+
+		_, err = c.client.FederatedTypeConfigs(typeConfig.Namespace).UpdateStatus(typeConfig)
+		if err != nil {
+
+			runtime.HandleError(fmt.Errorf("Could not update ObservedGeneration of the CRD: %q: %v", key, err))
```
marun: This block should return `util.StatusError` to ensure the event is processed again in a timely fashion.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -266,7 +278,7 @@ func (c *Controller) ensureFinalizer(tc *corev1a1.FederatedTypeConfig) error {
 	}
 	finalizers.Insert(finalizer)
 	accessor.SetFinalizers(finalizers.List())
-	_, err = c.client.FederatedTypeConfigs(tc.Namespace).Update(tc)
+	_, err = c.client.FederatedTypeConfigs(tc.Namespace).UpdateStatus(tc)
```
marun: This change will prevent the finalizer from being saved.  Everything but status is written with `Update`, and only status is written with `UpdateStatus`.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -281,6 +293,6 @@ func (c *Controller) removeFinalizer(tc *corev1a1.FederatedTypeConfig) error {
 	}
 	finalizers.Delete(finalizer)
 	accessor.SetFinalizers(finalizers.List())
-	_, err = c.client.FederatedTypeConfigs(tc.Namespace).Update(tc)
+	_, err = c.client.FederatedTypeConfigs(tc.Namespace).UpdateStatus(tc)
```
marun: ditto

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -160,6 +169,7 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 		if statusRunning {
 			c.stopController(statusKey, statusStopChan)
 		}
+		typeConfig.Status.ObservedGeneration = typeConfig.Generation
```
marun: I think the appropriate place to update the observed generation is just before returning ok.  afaik `observedGeneration` is an indication that a controller has _successfully_ processed the generation in question.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -168,6 +178,8 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 		return util.StatusAllOK
 	}
 
+	typeConfig.Status.ObservedGeneration = typeConfig.Generation
```
marun: ditto

pkg/controller/federatedtypeconfig/controller.go
sohankunkerkar: Ack. I will make the necessary changes.

pmorie:
Just to be explicit about this - observedGeneration is meant to contextualize other status fields. It's not terribly useful on its own, and my intention is that we will add (at least one) additional field in this PR that makes observedGeneration useful.

/hold

sohankunkerkar:
Referrencing #314 to get more clarity as to what we're planning to achieve.

pkg/controller/federatedtypeconfig/controller.go
sohankunkerkar: Yeah, makes sense, I will remove that initialization from that place.

pkg/controller/federatedtypeconfig/controller.go
sohankunkerkar: same comments as above.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -196,6 +211,18 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 		c.stopController(statusKey, statusStopChan)
 	}
 
+	typeConfig.Status.ObservedGeneration = typeConfig.Generation
+
+	typeConfig.Status.PropagationController = corev1a1.ControllerStatusRunning
```
pmorie: These two fields are both set to running - which may not be the correct value. You should set these fields based on the actual state of these controllers.

sohankunkerkar: Yeah, I made changes in the code to accommodate this requirement.

pkg/controller/federatedtypeconfig/controller.go
font: @sohankunkerkar were you able to confirm if this works? Otherwise, per @pmorie we may have a bug with CRD.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -213,9 +213,23 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 
 	typeConfig.Status.ObservedGeneration = typeConfig.Generation
 
-	typeConfig.Status.PropagationController = corev1a1.ControllerStatusRunning
+	if syncRunning {
 
-	typeConfig.Status.StatusController = corev1a1.ControllerStatusRunning
+		typeConfig.Status.PropagationController = corev1a1.ControllerStatusRunning
+	} else {
+
```
font: Let's remove these extra blank lines.

pmorie:
/approve

@font you have LGTM

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -95,11 +95,26 @@ type OverridePath struct {
...
+
 // FederatedTypeConfigStatus defines the observed state of FederatedTypeConfig
 type FederatedTypeConfigStatus struct {
 	// ObservedGeneration is the generation as observed by the controller consuming the FederatedTypeConfig.
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// PropagationController tracks the status of the sync controller.
+	// +optional
+	PropagationController ControllerStatus `json:"propagationController,omitempty"`
+	// +optional
```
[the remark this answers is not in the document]
sohankunkerkar: Done!

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -148,10 +148,14 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
...
 
+		typeConfig.Status.PropagationController = corev1a1.ControllerStatusNotRunning
+
+		typeConfig.Status.StatusController = corev1a1.ControllerStatusNotRunning
+
 		_, err = c.client.FederatedTypeConfigs(typeConfig.Namespace).UpdateStatus(typeConfig)
 		if err != nil {
 
-			runtime.HandleError(fmt.Errorf("Could not update ObservedGeneration of the CRD: %q: %v", key, err))
+			runtime.HandleError(fmt.Errorf("Could not update required fields of the CRD: %q: %v", key, err))
```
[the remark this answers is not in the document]
sohankunkerkar: Yeah, made necessary changes.

pkg/controller/federatedtypeconfig/controller.go
sohankunkerkar: Done!

pkg/controller/federatedtypeconfig/controller.go
sohankunkerkar: `UpdatesStatus()` method was used twice, one in the `ensureFinalizer` and other in `removeFinalizer`. so basically it was updating the status twice in the same iteration loop.  I commented out `UpdatesStatus()` in the `ensureFinalizer` method and it's working fine without any issues.

sohankunkerkar:
@font Done!

PR #404, opened 2018-11-08
Switch statefulset to deployment for federation-v2 controller manager

fisherxu:
With #394 add the leader elect, we can deploy the manager in multi instance with deployment.

/cc @font @irfanurrehman @marun @pmorie @shashidharatd

### DIFF
--- a/charts/federation-v2/values.yaml
+++ b/charts/federation-v2/values.yaml
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-## Configuration values for federation v2 controllermanager statefulset.
+## Configuration values for federation v2 controllermanager deployment.
 ## TODO(gyliu513): Add more parameters here like `feature-gates` etc.
 ##
 controllermanager:
--- a/docs/development.md
+++ b/docs/development.md
@@ -218,18 +218,18 @@ executable to aid in debugging, we need to kill the existing
 `federation-controller-manager` pod so that they will not step on each other. Follow
 these steps:
 
-1. Reduce the `federation-controller-manager` statefulset replicas to 0. This way
+1. Reduce the `federation-controller-manager` deployment replicas to 0. This way
    we can launch the necessary federation-v2 controllers ourselves via the test
    binary.
     ```bash
-    kubectl -n federation-system patch statefulsets.apps \
+    kubectl -n federation-system patch deployment.apps \
         federation-controller-manager -p '{"spec":{"replicas": 0}}'
     ```
 
     Once you've reduced the replicas to 0, you should see the
-    `federation-controller-manager` statefulset update to show 0 pods running:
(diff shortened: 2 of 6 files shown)


marun:
What motivated this PR?  For someone that doesn't want to enable leader election, statefulsets are preferable to deployments so I don't think it's reasonable to make deployments the default.

quinton-hoole-2:
@marun I'm not sure I understand your concern.  Could you elaborate?  Superficially, controller-manager seems stateless?  What was the motivation for deploying it as a StatefulSet rather than a Deployment?

marun:
@quinton-hoole The default controller manager is created by kubebuilder as a statefulset, and when I asked why they made the point that statefulset ensures that only one instance is running at a time.  A deployment does not have this guarantee.

quinton-hoole-2:
OK, gotcha.  So unless Federation has some fundamentally and compelling different requirement then Kubernetes, I agree that we should just stick with that default.

marun:
@fisherxu Do you intend to follow up on this PR?  It's been dormant for over a month.

shashidharatd:
@fisherxu, could you please rebase this pr. Thanks !

shashidharatd:
@marun, now that leader election is merged. Shall we take this pr forward?

This PR replaces the deployment with statefulset for controller-manager and is the right direction.
Technically, statefulset could also be used with multiple replicas but the controller-manager runs in reduced state, when a instance is down because of node going down or sth like that. Another instance could not be brought back soon because of the identity associated with the instance. So deployment seems to be the natural fit for multiple replicas of controller manager.
Providing both the options of using either deployment (leader election enabled) or statefulset (leader election disabled) is an cognitive overhead to the user (we need to take care of in documentation too). Given that in production, it is recommended to run with leader election enabled. It would be better for us to just support the leader election enabled by default and use deployment. Do you concur?

marun:
@shashidharatd Deployment is only a viable option if leader election is always enabled (not just enabled by default, but only enabled).  Otherwise there is a risk of more than one controller manager running.

irfanurrehman:
> @shashidharatd Deployment is only a viable option if leader election is always enabled (not just enabled by default, but only enabled). Otherwise there is a risk of more than one controller manager running.

@shashidharatd @marun I too think that we keep the option for leader election not exposed to the user at all and use deployment as the only option for the controller to run. Leader election to be inherently enabled in this mode. There wont be any need to specify anything about leader election in documentation either.

marun:
@irfanurrehman I think there will still be a need for leader election documentation - both for the arguments that configure it and the implications of having 1 vs many replicas.

irfanurrehman:
> @irfanurrehman I think there will still be a need for leader election documentation - both for the arguments that configure it and the implications of having 1 vs many replicas.

@marun I believe this is quite acceptable.

shashidharatd:
The commits are included in #657, hence closing this.
/close

PR #406, opened 2018-11-08
Update vendoring

marun:
This is an experiment checking to see the effects of `dep ensure -update`.


marun:
@irfanurrehman @shashidharatd @font @pmorie @gyliu513 

I was trying to add a dependency to `k8s.io/kubectl` and it changed other vendored code so I figured it was time to update deps.  The line count on this PR looks extreme, though.  Am I doing something dumb?

gyliu513:
@marun when I was working on https://github.com/kubernetes-sigs/federation-v2/pull/148#issuecomment-405803396 , I also found that many dependency files need some update, I think it is ok to update those files.

Also do we need to update versions for other dependencies in `Gopkg.toml` as well? like update kubernetes to 1.12 or so?

shashidharatd:
This `LGTM`, its high time we update the vendor code. the huge code size is understandable.

marun:
@gyliu513 We weren't pinning so kubebuilder was updated to the latest (see `Gopkg.lock`).

It still seems crazy to me that we have so much vendoring.  Our pkg path takes ~2.2MB.  vendor takes 351MB.

gyliu513:
@marun not sure if all vendors are used, we may need a cleanup about this. I can help on this after KubeCon if no one work for this. 

But for now, I'd prefer that we merge your PR first to make sure the `dep ensure` works well with current `Gopkg.toml`.

Also not sure if we can follow the way as https://github.com/kubernetes/cluster-registry , currently, for this project, they are not committing `vendor` https://github.com/kubernetes/cluster-registry/blob/master/docs/development.md#run-all-tests

marun:
For cluster registry not committing the vendored tree is essential to enabling vendoring by projects like us.  I'm less clear on the value of federation not committing vendored code, since I think the priority is simplifying development rather than vendoring by 3rd parties.

gyliu513:
Ah, I see, thanks @marun , then I think we need a task to do a clean up for the unused vendors similar as what we did for Istio https://github.com/istio/istio/pull/9050 , how about open an issue to trace this?

PR #408, opened 2018-11-09
Updated helm chart based on latest code.

gyliu513:
/cc @marun @shashidharatd @pmorie

I will update document as well for how to update the chart with https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/generate-install-crds-yaml.sh


shashidharatd:
/lgtm
Thanks @gyliu513!

PR #409, opened 2018-11-09
Updates k8s version in gke docs

danehans:
Although the [release notes](https://cloud.google.com/kubernetes-engine/release-notes#version-updates) indicate `1.11.2-gke.9` is GA, in testing I found `1.11.2-gke.15` is required for creating non-alpha 1.11 clusters.


docs/environments/gke.md
```diff
@@ -2,7 +2,7 @@
 
 Federation v2 can be deployed to and manage [GKE](https://cloud.google.com/kubernetes-engine/) clusters running
 Kubernetes v1.11 or greater. The following example deploys two GKE clusters named `cluster1` and `cluster2` using
-Kubernetes version `1.11.2-gke.9`.
+Kubernetes version `1.11.2-gke.15`.
 
 ```bash
 export ZONE=$(gcloud config get-value compute/zone)
```
gyliu513: How about update the script a bit by adding a version here?

```
export CLUSTER_VERSION=1.11.2-gke.15
```

Then we can use this variable in the following two `gcloud` command.

Issue #413, opened 2018-11-12
Cleanup vendor path


marun:
As per a recent vendor tree update (#406), the number of vendored dependencies seems excessive and it may be worth attempting to discovery and remove unused dependencies.

shashidharatd:
Addressed by #415

PR #415, opened 2018-11-13
Prune unused vendor packages

shashidharatd:
Addresses https://github.com/kubernetes-sigs/federation-v2/issues/413

/cc @marun @kubernetes-sigs/federation-v2-maintainers


Gopkg.toml
```diff
@@ -9,7 +9,7 @@ source = "gopkg.in/fsnotify/fsnotify.v1"
 
 [prune]
   go-tests = true
-  #unused-packages = true
+  unused-packages = true
```
gyliu513: Thanks @shashidharatd , one question, how about go through Gopkg.toml and check if we can remove all the unused `[[override]]`?

shashidharatd: AFAIK, these override directive will not cause inclusion of unused packages. they are used to override the package version or source etc...

shashidharatd: Anyways, i will update the packages based on what is generated by kubebuilde v1.0.4

marun:
I don't think we can prune non-go, as per the ci failure.

shashidharatd:
> I don't think we can prune non-go, as per the ci failure.

i excluded `github.com/kubernetes/repo-infra` from pruning now. should fix the ci issue.

shashidharatd:
CI build is successful now. @marun, @gyliu513 PTAL

marun:
I'm having trouble vendoring k8s.io/kubectl after this merge.  I'm going to see if reverting locally helps.

PR #417, opened 2018-11-13
Add informational output to federate subcommands

marun:
This is intended to be in keeping with the output provided when creating and deleting resources with kubectl.

### DIFF
--- a/pkg/kubefed2/federate/enable.go
+++ b/pkg/kubefed2/federate/enable.go
@@ -38,6 +38,7 @@ import (
 	apicommon "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
 	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+	ctlutil "github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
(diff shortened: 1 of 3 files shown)


pkg/kubefed2/federate/enable.go
```diff
@@ -247,6 +255,7 @@ func CreateResources(config *rest.Config, resources *typeResources) error {
 	if err != nil {
 		return fmt.Errorf("Error creating FederatedTypeConfig %q: %v", concreteTypeConfig.Name, err)
 	}
+	write(fmt.Sprintf("federatedtypeconfig.core.federation.k8s.io/%s created\n", concreteTypeConfig.Name))
```
pmorie: It might be helpful to have the namespace in this message, since you may not have supplied it as an arg.

pkg/kubefed2/federate/enable.go
```diff
@@ -304,11 +313,14 @@ func typeConfigForTarget(apiResource metav1.APIResource, namespace,
 func primitiveCRDs(typeConfig typeconfig.Interface, accessor schemaAccessor) ([]*apiextv1b1.CustomResourceDefinition, error) {
 	crds := []*apiextv1b1.CustomResourceDefinition{}
 
-	templateSchema, err := templateValidationSchema(accessor)
-	if err != nil {
-		return nil, err
+	// Namespaces do not require a template
+	if typeConfig.GetTarget().Kind != ctlutil.NamespaceKind {
```
pmorie: :+1:

PR #419, opened 2018-11-15
Update the comment of the function

xichengliudui:
Update the comment of the function

### DIFF
--- a/pkg/schedulingtypes/replicascheduler.go
+++ b/pkg/schedulingtypes/replicascheduler.go
@@ -343,7 +343,7 @@ func schedule(planner *planner.Planner, key string, clusterNames []string, curre
 	return result
 }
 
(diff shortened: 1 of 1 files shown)


sohankunkerkar:
@xichengliudui can you also update the command mentioned here:
https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/federating_crds.md#enable-propagation-of-your-federated-crd
you just need to append apply command before -f option.
```sh
$ kubectl *apply* -f ./example/crd/federatedBar.yaml --context=cluster1
federatedtypeconfig.core.federation.k8s.io/bars.example.io created
```

pmorie:
> @xichengliudui can you also update the command mentioned here:

Let's not expand the scope of this PR, @sohankunkerkar would you go ahead and just fix that in a PR since you know that it's an issue?

PR #423, opened 2018-11-16
Reorder primitive creation by crudtester to minimize test runtime

marun:


test/common/crudtester.go
```diff
@@ -102,20 +102,25 @@ func (c *FederatedTypeCrudTester) CheckLifecycle(desiredTemplate, desiredPlaceme
 }
 
 func (c *FederatedTypeCrudTester) Create(desiredTemplate, desiredPlacement, desiredOverride *unstructured.Unstructured) (*unstructured.Unstructured, *unstructured.Unstructured, *unstructured.Unstructured) {
-	template := c.createFedResource(c.typeConfig.GetTemplate(), desiredTemplate)
+	// Create placement and overrides first to ensure the sync
```
pmorie: This is a fine thing to do for now, but it points at deeper things like:

- controlling exactly when propagation happens in the presence of out of order creation of resources
- controlling how larger updates are rolled out

I have no good answers here, only starting to figure out the right questions :)

PR #425, opened 2018-11-16
Add support for 'federate enable' to be configured from a file

marun:
This is intended to simplify the encoding of configuration for the types whose propagation will be enabled be default.

### DIFF
--- a/pkg/kubefed2/federate/directive.go
+++ b/pkg/kubefed2/federate/directive.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package federate
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
+	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+)
+
+// FederateDirectiveSpec defines the desired state of FederateDirective.
+type FederateDirectiveSpec struct {
+	// The API version of the target type.
+	// +optional
+	TargetVersion string `json:"targetVersion,omitempty"`
+
+	// Which field of the target type determines whether federation
+	// considers two resources to be equal.
+	ComparisonField common.VersionComparisonField `json:"comparisonField"`
+
+	// The name of the API group to use for generated federation primitives.
+	// +optional
+	PrimitiveGroup string `json:"primitiveGroup,omitempty"`
+
+	// The API version to use for generated federation primitives.
+	// +optional
+	PrimitiveVersion string `json:"primitiveVersion,omitempty"`
+
+	// The list of name:path pairs of the fields to override in target template.
+	// +optional
+	OverridePaths []fedv1a1.OverridePath `json:"overridePaths,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/kubefed2/federate/enable.go
```diff
@@ -124,32 +131,56 @@ func NewCmdFederateEnable(cmdOut io.Writer, config util.FedConfig) *cobra.Comman
 
 // Complete ensures that options are valid and marshals them if necessary.
 func (j *enableType) Complete(args []string) error {
+	j.federateDirective = &FederateDirective{}
```
irfanurrehman: I think you wrote `NewFederateDirective()` to be used here, but missed using it.

pkg/kubefed2/federate/enable.go
```diff
@@ -124,32 +131,56 @@ func NewCmdFederateEnable(cmdOut io.Writer, config util.FedConfig) *cobra.Comman
...
+	ft := j.federateDirective
```
irfanurrehman: Not nitpicking on naming, but `fd` seems more apt here. I saw `ft` in later section of code and wondered what might it be.

pkg/kubefed2/federate/enable.go
```diff
@@ -124,32 +131,56 @@ func NewCmdFederateEnable(cmdOut io.Writer, config util.FedConfig) *cobra.Comman
...
+	ft.Name = args[0]
 
-	if j.rawComparisonField == string(apicommon.ResourceVersionField) ||
-		j.rawComparisonField == string(apicommon.GenerationField) {
-		j.comparisonField = apicommon.VersionComparisonField(j.rawComparisonField)
-	} else {
-		return fmt.Errorf("--comparison-field must be %q or %q",
-			apicommon.ResourceVersionField, apicommon.GenerationField,
-		)
+	if len(j.rawComparisonField) > 0 {
```
irfanurrehman: Why is this check necessary?

pkg/kubefed2/federate/util.go
```diff
@@ -18,15 +18,33 @@ package federate
...
 )
 
+func DecodeYAMLFromFile(filename string, obj interface{}) error {
+	f, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	return DecodeYAML(f, obj)
```
irfanurrehman: In case of error, does it need to be supplemented with a wrapper with message more relevant. I am not sure what strings `decode()` returns.

pkg/kubefed2/federate/enable.go
marun: Good catch.  I had originally named the type `FederateType`.

pkg/kubefed2/federate/enable.go
marun: Good question.  Removed.

pkg/kubefed2/federate/util.go
marun: I think the caller should be responsible for wrapping the error.  I've updated the call in `enable.go` accordingly.  The only other usage is in tests and I think that can evolve as needed.

marun:
@irfanurrehman I've added a fixup commit.  Please don't merge before squash.  Given the lack of movement on reviewable I think the fixup workflow is the best option to make life easier on reviewers.

pkg/kubefed2/federate/enable.go
marun: I recall now that an earlier iteration of this PR allowed arguments to override what was provided in a file.  Since this argument is defaulted, it would override what appeared in the file by default.  I have since switched to not allowing arguments to override a directive supplied by file.

pkg/kubefed2/federate/directive.go
```diff
@@ -0,0 +1,69 @@
...
+package federate
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
+	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+)
+
+// FederateDirectiveSpec defines the desired state of FederateDirective.
```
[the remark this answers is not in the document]
marun: I wanted to wait until we had a need for it in the api.  It should be cheap to move if/when we require.

PR #426, opened 2018-11-16
Deploy and test with generated primitives

marun:
This is the final PR in a series that implements generated primitives.  

 - Code-based primitives are removed from the tree and generated primitives are deployed in their place.           
 - Configuration for the generation of primitives is defined in `config/federatetypes`
 - Test fixtures have been updated to remove explicit version and kind to allow those details to be sourced from type configs


irfanurrehman:
Awsome job on this one @marun 
/lgtm

PR #427, opened 2018-11-16
Ensure rsp plugins are only started once by the scheduling manager

marun:
#366 added a manager for the scheduling controllers, but was untested in CI due:

 - managed e2e running 2 clusters but starting controllers directly, so the new manager was not involved
 - unmanaged e2e running the new manager but having only a single cluster, so the testing was skipped

When I ran unmanaged e2e locally with 2 clusters, the RSP tests would time out, and I traced the problem to the manager starting the plugins on every reconcile loop.  Adding the control logic that appears in this PR resolved the issue by ensuring a plugin is started only once.

#426 will ensure that the manager can be used for managed e2e, since the necessary type config will be created in a managed fixture.  @font has reported promising results with [kind](https://github.com/kubernetes-sigs/kind) will allow more than one cluster to be deployed in ci for unmanaged e2e.

### DIFF
--- a/pkg/controller/schedulingmanager/controller.go
+++ b/pkg/controller/schedulingmanager/controller.go
@@ -30,6 +30,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


marun:
cc: @xunpan

irfanurrehman:
@marun thanks a lot for doing this.
/lgtm
/approve

PR #432, opened 2018-11-19
Added script to sync up CRD resources in helm chart.

gyliu513:
/cc @marun @shashidharatd


shashidharatd:
@gyliu513, why don't we just remove one copy and replace all the references to it to another copy. Maintaining multiple copies of same files is anyway not beneficial.

gyliu513:
@shashidharatd Done by removed the manual section in user guide.

gyliu513:
@shashidharatd the reason that I want to keep both is because I want to give end user choice that they can install federation v2 via either helm chart or the deployment script, so here I was using a script to sync up the YAML templates to helm charts.

I was also planning to add some pre-check in `pre-commit.sh`, if there are difference between the YAML templates and helm charts, the pre-check will be failed, comments?

shashidharatd:
> @shashidharatd the reason that I want to keep both is because I want to give end user choice that they can install federation v2 via either helm chart or the deployment script, so here I was using a script to sync up the YAML templates to helm charts.

@gyliu513, what i meant is not to remove one of the method. I know we cannot completely rely on helm method as of now. We could remove the `config/federatedtypes` and wherever they are refered, use the `charts/federation-v2/templates` instead. like wise for hack/install.yaml too.

marun:
I think it might be worth skipping this step and targeting the results of #426.  That PR removes type configs and primitive crds from the tree in favor of generating them with `kubefed federate enable`.  Either the helm install should do the same (possible now that `kubefed2` will be shipped in the image thanks to @shashidharatd ), or there will need to be a script to generate the necessary yaml from the federate directives that are committed to the tree.

gyliu513:
Thanks @marun and @shashidharatd 

For `config/federatedtypes`, we can delete those files and use files under `charts/federation-v2/templates`.

But for `charts/federation-v2/charts/controllermanager/templates/crds.yaml`, I think we still need to keep this file as the `hack-install-latest.yaml` include all resources in this file while `charts/federation-v2/charts/controllermanager/templates/crds.yaml` only include the CRDs for federation.

So how about the following proposal:
1) Delete  `config/federatedtypes` and use files under  `charts/federation-v2/templates`.
2) Still keep this script but this script only sync up CRDs for federation.

marun:
@gyliu513 Your 2 points sound good to me.  And as you say, the non-primitive crds (core, mc and scheduling) will still need to be generated so crd.yaml still needs to be used.

How would you like to see helm chart delivery of primitives evolve in response to #426?  Would it be possible to use `kubefed2 federate enable`, or will results from the command need to be committed to the chart?

gyliu513:
@marun after we enable `kubefed2 federate enable`, we can leverage helm `pre-install` hook in helm at https://github.com/helm/helm/blob/master/docs/charts_hooks.md#hooks , and the hook will leverage a job to call `kubefed2 federate enable` to enable all CRD resources, I can update the chart once `kubefed2 federate enable` is ready.

So for this PR, I will use my proposal for now:
1) Create another PR to delete  `config/federatedtypes` and use files under  `charts/federation-v2/templates`.
2) Update this PR to and use script to sync up CRDs for federation.

marun:
@gyliu513 I think this PR is in good shape.  As per a comment on #427 I think using the type configs from the chart should not be done.  #426 should merge and the new directive files should be used by the chart to configure `kubefed2 federate enable` e.g.

```bash
kubefed2 federate enable -f ./config/federatedirectives/configmaps.yaml`
```

What do you think about adding a ci check based on the script added in this PR to ensure that the chart yaml stays updated?

gyliu513:
@marun that's exactly what I'm planning to do.

1) For this PR, it was mainly doing sync up for CRD resources in federation congtroller.
2) I can create another PR to update CI by adding some check, if the helm chart and the latest install CRDs are not synced, then CI will be failed. The CI will prompt a message telling developers to run `scripts/sync-up-helm-chart.sh` and re-commit.

marun:
@gyliu513 I think it would make sense to revisit this PR now that #426 has merged, since all primitives are now generated.

gyliu513:
@marun Done, this PR is focusing on sync up CRD for the helm chart.

marun:
@gyliu513 Good work!  Please add a ci check using your script that ensures the crd yaml is kept updated.  The existing check for install yaml may be a useful example in accomplishing this goal.

gyliu513:
@marun CI check is enabled.

marun:
Thanks!

/approve
/lgtm

PR #437, opened 2018-11-21
Use federated type from helm chart.

gyliu513:
This is a follow up PR for https://github.com/kubernetes-sigs/federation-v2/pull/432#issuecomment-440569216

/cc @marun @shashidharatd


marun:
I appreciate your initiative, but I don't think it's a good idea for this to merge ahead of #426.  #426 switches to using `kubefed2 federate enable` to generate type configs, and I'd prefer that it merge before switching helm over to generated type configs.  Otherwise #426 will either break the chart or have to take on responsibility for fixing it, and the PR is already gigantic.

gyliu513:
@marun Yes, this is not needed any more, let me close this one.

PR #438, opened 2018-11-21
Remove excess code

xichengliudui:
Remove excess code


pkg/controller/sync/version/adapter.go
```diff
@@ -45,7 +45,6 @@ type VersionAdapter interface {
 func NewVersionAdapter(client fedclientset.Interface, namespaced bool) VersionAdapter {
 	if namespaced {
 		return newNamespacedVersionAdapter(client)
-	} else {
-		return newClusterVersionAdapter(client)
 	}
+	return newNamespacedVersionAdapter(client)
```
poothia: i think it should be "return newClusterVersionAdapter(client)"

xichengliudui:
@poothia  Thank you for reminding me. I see the wrong

PR #442, opened 2018-11-21
Add validation of type config names in the relevant controllers

### LINKED ISSUE #433: Validate the names of FederationTypeConfigs
The naming of `FederatedTypeConfig` instances needs to conform to `<target plural name>.<target group name>` to ensure that federation will be enabled at most once for a given target type.  The simplest way to accomplish this is to add validation on any controllers that respond to `FederatedTypeConfig` (sync and rsp at the time of this writing).  Ideally a validation webhook would be used, but in the near term validation is complicated by the need to provision TLS certs.

marun:
Fixes #433.


pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -139,6 +140,13 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	// TODO(marun) Perform this defaulting in a webhook
 	corev1a1.SetFederatedTypeConfigDefaults(typeConfig)
```
gyliu513: Since the `typeConfig` was set by `corev1a1.SetFederatedTypeConfigDefaults`, just curious which case will trigger the validation `typeconfig.CheckTypeConfigName` failed?

marun: If the target plural name and group are not set, then that function will ensure a valid type config.  The check is to ensure that if the target plural name and group are set, that the name matches.

gyliu513: > If the target plural name and group are not set, 

When will this happen? Do you mean some mistake configuration of federated resource can cause this?

marun: There's nothing stopping someone from creating a type config with an arbitrary target that doesn't correspond to the name of the type config.  As per the linked issue, maintaining the correspondence between name and target configuration is essential to ensuring that a given federation control plane runs controllers for a given target type at most once.

marun:
rebased

PR #443, opened 2018-11-21
Cleanup operation logic for namespaces in the host cluster

### LINKED ISSUE #396: Control plane terminating a Federated Namespace when FederatedNamespacePlacement doesn't have the Host Cluster defined within clusters spec
Hello,

I just found out some strange behaviour (at least something that I wouldn't expect to happen)

Let's say we have 3 different clusters: cluster1, cluster2, cluster3, the Federation v2 control plane is running on cluster1, so we create all the CRDs related to federation in that cluster.

With the yaml definition below I would expect `foo` namespace being available on cluster2 and cluster3, but what is happening is that the namespace is being terminated on `cluster1` deleting the FederatedNamespacePlacement as well.

I also tried to create the FederatedNamespacePlacement in a different namespace, the behaviour is not the same but the namespace is still not being replicated because (I suppose) the `foo` namespace definition is being deleted.

Do you know if this is the expected behaviour? In that case, How should I create federated namespaces that I don't want to be present on Host Cluster?

Thanks in advanced,

```yaml
apiVersion: v1
kind: List
items:
- apiVersion: v1
  kind: Namespace
  metadata:
    name: foo
- apiVersion: core.federation.k8s.io/v1alpha1
  kind: FederatedNamespacePlacement
  metadata:
    name: foo
    namespace: foo
  spec:
    clusterNames:
    - cluster2
    - cluster3
```

marun:
There previously existed logic to avoid performing updates to namespaces in the host cluster.  This PR extends that logic to cover deletions and simplifies the comments for clarity.

Fixes #396

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -574,17 +574,17 @@ func (s *FederationSyncController) clusterOperations(selectedClusters, unselecte
 		if found {
 			clusterObj := clusterObj.(*unstructured.Unstructured)
 
-			// If we're a namespace kind and this is an object for the primary
-			// cluster, then skip the version comparison check as we do not
-			// track the cluster version for namespaces in the primary cluster.
-			// This avoids unnecessary updates that triggers an infinite loop
-			// of continually adding finalizers and then removing finalizers,
-			// causing PropagatedVersion to not keep up with the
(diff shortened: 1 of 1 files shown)


shashidharatd:
/lgtm
Thanks for the fix!

PR #446, opened 2018-11-22
Add scripts for creating and deleting kind clusters

font:
The following changes are implemented here to enable users to create and delete kind clusters for dev testing.

- This moves the functions that create kind clusters out of the CI script and into its own script to allow running locally for a dev workflow. Additional checks were added to prevent the script from clobbering user's existing kube config and warn about multiple `insecure-registry` configurations.
- Modify contexts to use `cluster[0-9]` to be consistent with the docs.
- Add delete clusters script.

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -118,7 +118,7 @@ function fixup-cluster() {
   export KUBECONFIG="${KUBECONFIG:-}:${kubeconfig_path}"
 
   # Simplify context name
-  kubectl config rename-context kubernetes-admin@kind-${i} kind-${i}
+  kubectl config rename-context kubernetes-admin@kind-${i} cluster${i}
 
   # TODO(font): Need to set container IP address in order for clusters to reach
   # kube API servers in other clusters until
@@ -134,7 +134,7 @@ function fixup-cluster() {
 
 function check-clusters-ready() {
   for i in $(seq ${1}); do
-    util::wait-for-condition 'ok' "kubectl --context kind-${i} get --raw=/healthz &> /dev/null" 120
+    util::wait-for-condition 'ok' "kubectl --context cluster${i} get --raw=/healthz &> /dev/null" 120
   done
 }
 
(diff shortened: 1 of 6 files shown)


scripts/create-clusters.sh
```diff
@@ -0,0 +1,129 @@
...
+
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+CONFIGURE_INSECURE_REGISTRY="${CONFIGURE_INSECURE_REGISTRY:-}"
+CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
+NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
+docker_daemon_config="/etc/docker/daemon.json"
+
+function create-and-configure-insecure-registry() {
+  # Run insecure registry as container
+  docker run -d -p 5000:5000 --restart=always --name registry registry:2
```
shashidharatd: (no action required) This could conflict with user environment, as the user might already be running a registry. would it be good if we make this configurable and optional?

font: I think that's a good suggestion that can be implemented in a follow-up.

scripts/util.sh
```diff
@@ -37,7 +37,7 @@ function util::wait-for-condition() {
   local error_msg="[ERROR] Timeout waiting for ${msg}"
 
   local counter=0
-  while ! ${condition}; do
+  while ! eval ${condition}; do
```
pmorie: just curious, how did you find that this needed to be changed?

font: I discovered this when the `condition` argument being passed to this function included a redirection such as:
``kubectl --context cluster${i} get --raw=/healthz &> /dev/null``

What happened is the redirection argument `&> /dev/null` was being passed to `kubectl` and it didn't know what to do with it. Using `eval` allows the `condition` args to be properly constructed into a single command and passed to the shell resulting in the proper execution.

PR #447, opened 2018-11-22
Add and update docs to use kind clusters

font:
**NOTE**: This PR is not ready for review until #446 has merged.

This adds a doc for supporting kind clusters. Subsequently, changes were made to the userguide doc to parameterize cluster names and simplify the example tutorial.


marun:
Off topic, it seems like a lot of work to maintain examples for every single type.  Would it make sense to limit the examples to core workload types?

font:
> Off topic, it seems like a lot of work to maintain examples for every single type. Would it make sense to limit the examples to core workload types?

What would you consider core types? Right now the example only includes the below resources:

``RESOURCE_TYPES="configmap deployment job secret service serviceaccount"``

Even though there are `clusterrole` and `clusterrolebinding` example yaml configs, they aren't used in the tutorial right now because there is no `federatedtypeconfig` yaml for enabling them. Therefore, I opted to avoid adding them due to the new `kubefed2 federate enable` command superseding those configs.

In any case I'm okay with keeping to core types, although the current ones in the tutorial seem rather core to me, with the exception of `configmap` and `secret` perhaps being a little redundant.

marun:
I would be ok with picking one of `secret` or `configmap` rather than including examples for both.  I don't think there's much point in having examples for `serviceaccount`.  #426 adds type config for `clusterrole` to allow testing of a cluster-scoped type, but I agree that adding examples for it isn't that useful.

font:
@marun sgtm. I'd like to address it in a new PR. I've updated this PR to just add the `kind` docs and I will open a new PR to only include yaml configs and tutorial examples for:

``RESOURCE_TYPES="deployment job secret service"``

We can continue the discussion there if we want to add/remove any examples and see what others think.

docs/environments/kind.md
```diff
@@ -0,0 +1,73 @@
...
+script if you'd like to change the default.
+
+### Configure Insecure Container Registry
+
+Please answer the following questions to determine if you need to set up an
+insecure container registry on your host:
+
+1. Is this the first time you're running the `create-clusters.sh` script?
+2. Are you planning on creating container images locally without pushing to a
+public container registry such as quay.io?
```
gyliu513: s/quay.io/`quay.io`

docs/environments/kind.md
```diff
@@ -0,0 +1,73 @@
...
+CONFIGURE_INSECURE_REGISTRY=y ./scripts/create-clusters.sh
+```
+
+This will automatically create the necessary dockerd daemon config and reload
+the docker daemon for you. Keep in mind that it will **not** do this for you
+if a config already exists, or your docker daemon is already configured with an
+`--insecure-registry` command line option.
+
+If you would like to manually make the changes to your docker daemon instead,
+add `172.17.0.1:5000` as an insecure registry host and reload or restart your
```
gyliu513: How about adding the link https://docs.docker.com/registry/insecure/ here to help end user understand how to add an insecure registry host.

docs/environments/kind.md
```diff
@@ -0,0 +1,73 @@
...
+### Configure Insecure Container Registry
+
+Please answer the following questions to determine if you need to set up an
+insecure container registry on your host:
+
+1. Is this the first time you're running the `create-clusters.sh` script?
+2. Are you planning on creating container images locally without pushing to a
+public container registry such as quay.io?
+
+If you answered yes to both of these questions, then you will need to
```
gyliu513: How about other cases?

- 1 No 2 Yes
- 1 No 2 No
- 1 Yes 2 No

docs/environments/kind.md
```diff
@@ -0,0 +1,73 @@
...
+docker daemon.
+
+### Run Script
+
+Run the following command to create `2` `kind` clusters:
+
+```bash
+./scripts/create-clusters.sh
+```
+
```
gyliu513: How about add the following:

If you want to create a cluster which has more than 2 clusters, run the following command:

```bash
NUM_CLUSTERS=<your number> ./scripts/create-clusters.sh
```

docs/environments/kind.md
```diff
@@ -0,0 +1,73 @@
...
+## Delete Clusters
+
+Run the following command to delete your `2` `kind` clusters. If you've
+specified a different `NUM_CLUSTERS` than the default, make sure to set it
+again before invoking the script.
+
+```bash
+./scripts/delete-clusters.sh
+```
```
gyliu513: How about add the following:

If you want to delete a cluster which has more than 2 clusters, run the following command:

```bash
NUM_CLUSTERS=<your number> ./scripts/delete-clusters.sh
```

docs/development.md
```diff
@@ -238,7 +238,7 @@ to build an image and a deployment config.
 If you just want to have this automated, then run the following command
 specifying your own image. This assumes you've used the steps [documented
 above](#setup-clusters-deploy-the-cluster-registry-and-federation-v2-control-plane) to
-set up two minikube clusters (`cluster1` and `cluster2`):
+set up two `kind` clusters (`cluster1` and `cluster2`):
```
pmorie: Do we really want to prescribe kind over minikube? Why not say "kind or minikube" ?

docs/environments/kind.md
```diff
@@ -0,0 +1,73 @@
...
+## Download and Install kind
+
+If you don't yet have `kind` installed, you can run the following script to
+download and install a known working version.
+
+```bash
+./scripts/download-e2e-binaries.sh
+```
+
+Make sure that your `GOBIN` directory is in your path as that is where `kind`
```
pmorie: Might want to clarify that path is `PATH` here

docs/environments/kind.md
pmorie: Is this really accurate? The teardown script also stops the local registry, iirc

docs/environments/kind.md
pmorie: That's addressed elsewhere in the doc but it couldn't hurt to reflect that in this sample, I guess.

marun:
(as much commentary on the scripts as the docs)

Given the size of the PRs in question, I think it makes sense to collapse #452 and evolve the docs along with the code.

Since the registry only needs to be configured and started once, consider separating out the functionality into a separate script.  I think that would make it easier to document (and maintain).  I think it would also make sense to include an example of using the registry (e.g. a `deploy-federation.sh` invocation as appears in `scripts/pre-commit.sh`) at the end of the section documenting the configuration of the insecure registry so the link is clear. 

I don't access external clusters so overwriting kubeconfig was fine for me, but I suspect this will be annoying for some.  Consider using `kubectl config` commands to add the new clusters (with a unique name e.g. fed-kind-clusterx) to an existing kubeconfig.

Issue #370, opened 2018-10-27
Support NodePort for Type MultiClusterServiceDNSRecord


danehans:
Currently, services must be of type `LoadBalancer` to populate the `MultiClusterServiceDNSRecord` status.dns.loadbalancer field. Exposing a Kubernetes service of type `NodePort` is common and should be supported by type `MultiClusterServiceDNSRecord`.

gyliu513:
@danehans I think one of the concerns for NodePort is security, any solutions to handle the security issues when using NodePort?

shashidharatd:
NodePort poses some restriction to the DNS based discovery mechanism.
1. The NodePort allocated by the service-controller would not be the same across all member clusters for the service. (workaround is to user specify NodePort in spec, but this is too restrictive and there is potential for port clash and is better avoided).
2. The DNS mechanism we currently use does not take care of `port` and so we don't have mechanism to discover the port.

Supporting `MultiClusterServiceDNSRecord` for NodePort is therefore not a production worthy feature. It can at the most serve for some demo purposes.

danehans:
@gyliu513 I am not following your question. NoePort is used by gke and many other k8s implementations to externally expose services.

danehans:
@shashidharatd thanks for your feeback. I should clarify my question. I do not care about the TCP/UDP port used by type `NodePort` as it relates to the `MultiClusterServiceDNSRecord`. With type `NodePort`, I care about the IP address used for populating the status ingress IP of `MultiClusterServiceDNSRecord`. As we do with type `LoadBalancer`, we should be able to obtain the node(es) IP to use for populating status.ingress of `MultiClusterServiceDNSRecord`. This will help with test/dev (i.e. minikube) and for on-prem k8s deployments where `NodePort` is more common. `MultiClusterServiceDNSRecord` could potentially support `NodePort` by:

1. Getting the label selector from the service.
2. Create a list of IPs for the Service by getting all pods with a label matching the label selector from 1 and filtering `.items[0].status.hostIP` for each pod.
3. Use the IP's from 2 to populate `MultiClusterServiceDNSRecord` status.ingress.ip.

Thoughts?

gyliu513:
@danehans The reason that I say NodePort is not security is because if you are using NodePort, then the service can be accessed from any node as follows:

1) Create a service and use NodePort.
2) After the service created, normal user will get the host IP address and one NodePort which in 31000 - 32000
[...]

PR #453, opened 2018-11-27
Updates service dns doc to refer to issue 370 and gke clusters

danehans:
Adds details about required cluster type (i.e. GKE) and sample service type due to #370.

### DIFF
--- a/docs/servicedns-with-externaldns.md
+++ b/docs/servicedns-with-externaldns.md
@@ -42,7 +42,8 @@ providers.
 Setting-up MCSDNS can be accomplished by referencing the following documentation:
 
 - The Federation-v2 [User Guide](userguide.md) to setup one or more Kubernetes clusters and the Federation
-control-plane. If running in GKE, the cluster hosting the ExternalDNS controller must have scope
+control-plane. Due to [Issue #370](https://github.com/kubernetes-sigs/federation-v2/issues/370), the environment running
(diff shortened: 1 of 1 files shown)


docs/servicedns-with-externaldns.md
```diff
@@ -42,7 +42,8 @@ providers.
 Setting-up MCSDNS can be accomplished by referencing the following documentation:
 
 - The Federation-v2 [User Guide](userguide.md) to setup one or more Kubernetes clusters and the Federation
-control-plane. If running in GKE, the cluster hosting the ExternalDNS controller must have scope
+control-plane. Due to [Issue #370](https://github.com/kubernetes-sigs/federation-v2/issues/370) you can not use minikube
```
gyliu513: I think we can still use minikube here as long as we can enable the minikube support service type `LoadBalancer`, how about make the limitation to `NodePort` but not `minikube`?

gyliu513: But after I install minikube, I can still deploy a helm chart like keepalived here https://github.com/gyliu513/work/tree/master/k8s/charts/keepalived and enable the minikube support `LoadBalancer` type, am I missing anything?

shashidharatd: one could also use [metallb](https://metallb.universe.tf/) to provision LoadBalancers in minikube clusters.

Issue #457, opened 2018-11-28
kubefed2 federate resource pluralization


danehans:
The command does not properly pluralize recouces that end in "y":
```
$ kubectl get federatedserviceentries
error: the server doesn't have a resource type "federatedserviceentries"
$ kubectl get serviceentries
NAME          AGE
httpbin-ext   8m
$ kubectl get federatedserviceentrys
NAME          AGE
httpbin-ext   8m
```

In the above example, you see `federatedserviceentrys` should be `federatedserviceentries`.

marun:
The command doesn't do any pluralization, it only prefixes `federated`.  If you check the plural name of the target and I'll bet you'll see that it is `serviceentrys`.

marun:
(Please close if my perspective is correct)

danehans:
@marun The target is `serviceentries` and not `serviceentrys`. If the target was `serviceentrys`, that would be a problem b/c the proper and expected pluralization of `ServiceEntry` is `serviceentries`. Istio users are expecting `ServiceEntry` to be pluralized as `serviceentries`. See:
```
$ kubectl get serviceentrys
error: the server doesn't have a resource type "serviceentrys"
$ kubectl get serviceentries
NAME          AGE
httpbin-ext   50m
$ kubectl get serviceentries --context cluster2
NAME          AGE
httpbin-ext   50m
```

The federated equivilent:
```
$ kubectl get federatedserviceentrys
NAME          AGE
httpbin-ext   53m
$ kubectl get federatedserviceentries
error: the server doesn't have a resource type "federatedserviceentries"
```

You can see the incorrect pluralization in the api resources as well:
```
$ kubectl api-resources | grep ServiceEntry
serviceentries                                                 networking.istio.io                 true         ServiceEntry
federatedserviceentryplacements                                primitives.federation.k8s.io        true         FederatedServiceEntryPlacement
federatedserviceentrys                                         primitives.federation.k8s.io        true         FederatedServiceEntry
```

I am submitting a PR to fix this issue.

sohankunkerkar:
I had similar kind of issues while trying to rename `servicedns` API type. It was doing `servicednss` instead of `servicednses`. I've created a PR in the controllers-tool repository to fix this issue.
https://github.com/kubernetes-sigs/controller-tools/pull/106/files
With this in place, we can define the resource name with the help of `+kubebuilder:resource:path=" servicednses"` annotation.
@marun @danehans

PR #458, opened 2018-11-28
Properly pluralizes type configs for resources ending in y

danehans:
Fixes https://github.com/kubernetes-sigs/federation-v2/issues/457


marun:
Looks like a gofmt fix is required.  I'm happy to merge once that's done.  I've reconsidered not doing pluralization at all. I think the current pluralization should remain in place (which this PR addresses), and a separate PR can change generated primitives to compose their plural names.

marun:
@danehans I'm confused, why is `federate.go` changing?

danehans:
@marun Not sure how the change to `federate.go` happened. Maybe when I ran the gofmt script. I apologize, I should've caught that. I resubmitted the PR and I think all looks good now.

PR #459, opened 2018-11-29
Minor fix for kind document.

gyliu513:
/cc @font

### DIFF
--- a/docs/environments/kind.md
+++ b/docs/environments/kind.md
@@ -25,7 +25,7 @@ Please answer the following question to determine if you need to set up an
 insecure container registry on your host:
 
 1. Are you planning on creating container images locally without pushing to a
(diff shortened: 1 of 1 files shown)


docs/environments/kind.md
```diff
@@ -25,7 +25,7 @@ Please answer the following question to determine if you need to set up an
 insecure container registry on your host:
 
 1. Are you planning on creating container images locally without pushing to a
-   public container registry such as `quay.io`? For example, you can build your
+   public container registry such as `quay.io`. For example, you can build your
```
font: This is a question so we shouldn't remove the question mark.

gyliu513: Thanks @font, I will keep this in mind and fix it when there are some other issues in the userguide.

PR #462, opened 2018-11-30
IngressDNSRecord should be in test-namespace in ingressdns-with-externaldns.md

xunpan:
IngressDNSRecord should be in the same namespace with FederatedIngress to make it work.

Currently, the doc uses default for IngressDNSRecord and uses test-namespace for FedertedIngress. It should not work.

@danehans


gyliu513:
/ok-to-test
/approve

@danehans can you help review?

danehans:
@xunpan thanks for catching this. /lgtm

PR #467, opened 2018-11-30
Sets type config template pluralname

danehans:
Sets the template pluralName field of a type config by using the name of the API resource.

### DIFF
--- a/pkg/kubefed2/federate/enable.go
+++ b/pkg/kubefed2/federate/enable.go
@@ -289,6 +289,7 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
 func typeConfigForTarget(apiResource metav1.APIResource, federateDirective *FederateDirective) typeconfig.Interface {
 	spec := federateDirective.Spec
 	kind := apiResource.Kind
+	pluralName := apiResource.Name
 	typeConfig := &fedv1a1.FederatedTypeConfig{
 		// Explicitly including TypeMeta will ensure it will be
 		// serialized properly to yaml.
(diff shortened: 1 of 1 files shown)


marun:
This is a good start.  Please do the same for the override and placement resources.

marun:
@danehans Arg, I'm not sure what I was thinking.  Overrides and placement can be automatically pluralized since 'override' and 'placement' is what is being pluralized rather than an arbitrary type name.

danehans:
No worries. Let me update the PR .

danehans:
I pass the unmanaged e2e tests locally, but fails in CI:
```
Ran 34 of 35 Specs in 135.331 seconds
SUCCESS! -- 34 Passed | 0 Failed | 0 Pending | 1 Skipped
--- PASS: TestE2E (135.33s)
PASS
ok  	github.com/kubernetes-sigs/federation-v2/test/e2e	135.389s
```

Any suggestions @marun?

pkg/kubefed2/federate/enable.go
```diff
@@ -308,7 +309,8 @@ func typeConfigForTarget(apiResource metav1.APIResource, federateDirective *Fede
 			ComparisonField:    spec.ComparisonField,
 			PropagationEnabled: true,
 			Placement: fedv1a1.APIResource{
-				Kind: fmt.Sprintf("Federated%sPlacement", kind),
+				Kind:       fmt.Sprintf("Federated%sPlacement", kind),
+				PluralName: fmt.Sprintf("federated%splacement", pluralName),
```
marun: The template should reuse the target's plural name in composing it's plural name, but placement doesn't need to.

marun:
@danehans The namespaced suite is failing, likely a transitory failure.  Once you push an update that pluralizes only the template the job will restart.

pkg/kubefed2/federate/enable.go
danehans: My bad, I removed the wrong line in the last commit. I just pushed a new commit that pluralizes only `typeConfig.Spec.Template.PluralName` using the API resource name.

marun:
Thank you!

/lgtm
/approve

Issue #354, opened 2018-10-23
Target Cluster Service Account Does Not Include nonResourceURLs


danehans:
While testing https://github.com/kubernetes-sigs/federation-v2/pull/332, the following `FederatedClusterRole` was not being propagated to target clusters:
```
apiVersion: generated.federation.k8s.io/v1alpha1
kind: FederatedClusterRole
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"generated.federation.k8s.io/v1alpha1","kind":"FederatedClusterRole","metadata":{"annotations":{},"name":"prometheus-istio-system","namespace":""},"spec":{"template":{"rules":[{"apiGroups":[""],"resources":["nodes","services","endpoints","pods","nodes/proxy"],"verbs":["get","list","watch"]},{"apiGroups":[""],"resources":["configmaps"],"verbs":["get"]},{"nonResourceURLs":["/metrics"],"verbs":["get"]}]}}}
  creationTimestamp: 2018-10-23T16:14:01Z
  finalizers:
  - federation.kubernetes.io/delete-from-underlying-clusters
  - orphan
  generation: 1
  name: prometheus-istio-system
  resourceVersion: "101608"
  selfLink: /apis/generated.federation.k8s.io/v1alpha1/federatedclusterroles/prometheus-istio-system
  uid: a7672882-d6de-11e8-b72c-42010a8a0002
spec:
  template:
    rules:
    - apiGroups:
      - ""
      resources:
      - nodes
      - services
      - endpoints
      - pods
      - nodes/proxy
      verbs:
      - get
      - list
      - watch
    - apiGroups:
      - ""
      resources:
      - configmaps
      verbs:
      - get
    - nonResourceURLs:
      - /metrics
      verbs:
      - get
```

I observed the following log messages on the controller-manager:
```
[...]

PR #468, opened 2018-11-30
Updates clusterrole rules to include all NonResourceURLs

danehans:
Rules update is specific to cluster roles and binding and not namespace roles and binding.

Fixes issue #354

### DIFF
--- a/pkg/kubefed2/join.go
+++ b/pkg/kubefed2/join.go
@@ -62,11 +62,18 @@ var (
 
 	// Policy rules allowing full access to resources in the cluster
 	// or namespace.
-	adminPolicyRules = []rbacv1.PolicyRule{
+	namespacedPolicyRules = []rbacv1.PolicyRule{
 		{
 			Verbs:     []string{rbacv1.VerbAll},
 			APIGroups: []string{rbacv1.APIGroupAll},
 			Resources: []string{rbacv1.ResourceAll},
+		},
+	}
+	clusterPolicyRules = []rbacv1.PolicyRule{
+		namespacedPolicyRules[0],
+		{
+			NonResourceURLs: []string{rbacv1.NonResourceAll},
+			Verbs:           []string{"get"},
 		},
 	}
 )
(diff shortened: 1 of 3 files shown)


danehans:
@marun I would like to get this merged before kubecon so I no longer have to implement the workaround for federated istio. Let me know your thoughts on the latest commit.

pkg/kubefed2/util/util.go
```diff
@@ -120,9 +120,9 @@ func RoleName(serviceAccountName string) string {
 	return fmt.Sprintf("federation-controller-manager:%s", serviceAccountName)
 }
 
-// CommonClusterRoleName returns the name of a ClusterRole and its
+// ClusterRoleName returns the name of a ClusterRole and its
```
marun: nit: `s/ClusterRoleName/HealthCheckRoleName/`

pkg/kubefed2/unjoin.go
```diff
@@ -345,7 +345,7 @@ func deleteClusterRoleAndBinding(clusterClientset client.Interface, saName, name
 	}
 
 	roleName := util.RoleName(saName)
-	commonClusterRoleName := util.CommonClusterRoleName(saName)
+	commonClusterRoleName := util.ClusterRoleName(saName)
```
marun: nit: `s/commonClusterRoleName/healthCheckRoleName/`

pkg/kubefed2/join.go
```diff
@@ -440,6 +447,19 @@ func createRBACSecret(hostClusterClientset, joiningClusterClientset client.Inter
 		glog.V(2).Infof("Created role and binding for service account: %s in joining cluster: %s",
 			saName, joiningClusterName)
 
+		glog.V(2).Infof("Creating health check cluster role and binding for service account: %s in joining cluster: %s", saName, joiningClusterName)
+
+		err = createHealthCheckClusterRoleAndBinding(joiningClusterClientset, saName, namespace, joiningClusterName,
+			dryRun)
```
marun: (No action required) Seems strange to break here given that the line above would be longer.

pkg/kubefed2/unjoin.go
danehans: ack

pkg/kubefed2/unjoin.go
```diff
@@ -345,7 +345,7 @@ func deleteClusterRoleAndBinding(clusterClientset client.Interface, saName, name
 	}
 
 	roleName := util.RoleName(saName)
-	commonClusterRoleName := util.CommonClusterRoleName(saName)
+	commonClusterRoleName := util.HealthCheckRoleName(saName)
```
marun: sorry, one more nit: `s/commonClusterRoleName/healthCheckRoleName/`

danehans: My bad, I should have caught this. I'll change.

PR #478, opened 2018-12-03
Use prettier to format the md file.

gyliu513:
Produced via: `cd docs; prettier --write $(find . -name "*.md")`


marun:
@gyliu513 Should this be a one-time change, or would it make sense to add a ci check to ensure that subsequent changes maintain this formatting?

danehans:
+1 for adding check to ci.

.travis.yml
```diff
@@ -13,6 +13,12 @@ sudo: required
 services:
   - docker
 
+before_install:
+  - sudo apt-get update
```
marun: This works, but I don't think using distro packages is the right way to install a tool that contributors will need to run.  That's we have scripts that download dependencies in a distro-agnostic way.  Will it be possible to download this functionality in a self-contained way?  If not, I think an alternative needs to be found.  Requiring that contributors install npm and node for non-functional formatting cleanup does not seem worth it to me. 

Will it be possible to avoid requiring npm and node and the prettifier  to be installed I'm concerned that this particular prettifier may at the prospect of adding a dependency on npm and node given it may not be , is there another mechanism that

gyliu513: @marun I also get some comments from knative document, they are now using a bot to generate a PR for the whole doc, please refer to https://github.com/knative/docs/issues/624 for detail.

Not sure for federation v2, we can also use a bot to do this?

marun: Who is maintaining the bot?  If it's not trivial, It's not clear to me that this level of effort is justified for a non-functional feature of alpha software.  I think we have many higher priority things to drive us to beta and GA.

gyliu513: @marun please refer to https://github.com/knative/docs/issues/624#issuecomment-444336763 for how Knative maintain the bot. Currently, they are using Knative BuildTeamplate as the bot. But for federation v2, I think we can investigate this after we enter beta and GA.

As I append above, for current document, how about merge the formatted md document first? I can delete the `enable CI` commit.

gyliu513: @marun thanks, done.

PR #484, opened 2018-12-05
Allow placement to be set by label selector

### LINKED ISSUE #401: Option to specify propagate resource to all cluster
Currently, the placement objects are designed such that if you wish to propagate the federated resource to a member cluster, then one need to explicitly specify the cluster name in corresponding placement object. This could be hard as the cluster members could added/removed over time and user needs to keep updating the placement objects of federated resource. There is a clear desire to be able to specify the intent that the federated resource should be propagated to member clusters without additional steps.

Background:
For `cross-cluster service discovery` feature, there is a need to update federation map in Cluster DNS. We could achieve this by creating a `FederatedConfigMap` in federation and specifying a placement object which would take care of propagating the configmap to all members clusters. it should also take care of propagating to any newly joined cluster.
Note: Currently, if the cluster name is not present in the placement object the resource is removed during propagation, which could have some impact on the service discovery within the cluster.

To facilitate such a behaviour, i do not have a clear solution, but have some thoughts.
1. Introduce some special keyword (like `*` or `all`) which when specified as `clusterNames` the resource will be propagated to all member clusters.
2. Introduce a new field in placement (like `propogateToAllClusters`) which when set to true, the controller will take care of federating to all member clusters.

/cc @irfanurrehman @marun @kubernetes-sigs/federation-wg 

p.s: This is partly brought out in https://github.com/kubernetes-sigs/federation-v2/issues/87, but i created a separate issue to disambiguate between the little different but related requirement.

marun:
Add support for determining placement via label selection.  This PR adds a new field named `clusterSelector` of type `LabelSelector` to placement resources.  The `clusterNames` field is retained to ensure that scheduling will still work, and `clusterSelector` will be ignored if `clusterNames` is provided to ensure that scheduling can take priority over manual placement.

Unit testing is included to validate the determination of selected cluster names between the 2 fields. 

Fixes #401


marun:
The latest push removes the ci script commit.

marun:
Rebased

PR #488, opened 2018-12-05
Fix validation generation for crd targets that lack validation

marun:
It was not a reasonable assumption that all crd targets would have a validation schema.


danehans:
Worked in my tests /lgtm

Issue #490, opened 2018-12-06
Added document for clusterSelector.


gyliu513:
With PR https://github.com/kubernetes-sigs/federation-v2/pull/484 , fed v2 now support placement policy with `clusterSelector`. But when joining a cluster to the federation controller plane, we cannot set labels for the joining cluster, it is better to enable `kubefed2 join` support adding labels.

My proposal here is adding a new option like `label` to `kubefed2` like `kubectl label`, and also enable `kubefed2 join` support `--label` option.

Setting label when joining a cluster:
```
./bin/kubefed2 join cluster1 --cluster-context cluster1 --host-cluster-context cluster1 --add-to-registry --v=2 --label foo=bar,foo1=bar1
```

Adding label to an existing federatedclusters:
```
kubefed2 label federatedclusters cluster1 foo=bar
```

Deleting label from an existing federatedclusters:
```
kubefed2 label federatedclusters cluster1 foo-
```

Currently, we can use `kubectl label federatedclusters` to set labels for federated clusters, so not sure if we still need `kubefed2 label`, but I think we need enable `kubefed2 join` support adding a federated cluster with some labels.

/cc @marun @kubernetes-sigs/federation-wg

marun:
I don't think this should be the responsibility of `kubefed2`.  Labels are a part of every single kube resource, and there is a wealth of existing tooling and commands that allow them to be edited.  The best code is the code we don't have to write.

gyliu513:
@marun so for now, if an end user want to add labels for the federated clusters, then two steps are needed: 

- Join the cluster
- Label the cluster via `kubectl label federatedclusters`

Shall we update the document a bit by clarifying this?

marun:
Maybe add a pointer to the kube docs on labeling?

Also, just because label selection is supported doesn't mean that clusters need to be labeled.  I think the base use case is distributing resources to all clusters, and an empty selector will ensure that regardless of whether clusters are labeled or not.

PR #492, opened 2018-12-06
Added document for clusterSelector.

gyliu513:
Fixed https://github.com/kubernetes-sigs/federation-v2/issues/490

/cc @marun

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -293,6 +293,74 @@ done
 If you were able to verify the resources removed and added back then you have
 successfully verified a working federation-v2 deployment.
 
+#### Using Cluster Selector
+
+In addition to specifying an explicit list of clusters that a resource should be
+propagated to via the `spec.clusterNames` field of a placement resource, it is possible
+to use the `spec.clusterSelector` field to provide a label selector that determines
+that list of clusters at runtime.
+
+If the goal is to select a subset of member clusters, make sure that the `FederatedCluster`
+resources that are intended to be selected have the appropriate labels applied.
+
+The following command is an example to label a `FederatedCluster`:
+
+```bash
+kubectl label federatedclusters -n federation-system cluster1 foo=bar
+```
+
+Please refer to [Kubernetes label command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#label)
+to get more detail for how `kubectl label` works.
+
+The following sections detail how `spec.clusterNames` and `spec.clusterSelector` are
+used in determining the clusters that a federated resource should be propagated to.
+
+##### Neither `spec.clusterNames` nor `spec.clusterSelector` is provided
+
+```yaml
+spec: {}
+```
+
+In this case, you can either set `spec: {}` as above or remove `spec` field from your
+placement policy. The resource will not be propagated to member clusters.
+
+##### Both `spec.clusterNames` and `spec.clusterSelector` are provided
+
+```yaml
+spec:
+  clusterNames:
+    - cluster2
+    - cluster1
+  clusterSelector:
+    matchLabels:
+      foo: bar
+```
+
+For this case, `spec.clusterSelector` will be ignored as `spec.clusterNames` is provided. This
+ensures that the results of runtime scheduling have priority over manual definition of a
+cluster selector.
+
+##### `spec.clusterNames` is not provided, `spec.clusterSelector` is provided but empty
+
+```yaml
+spec:
+  clusterSelector: {}
+```
+
+In this case, the resource will be propagated to all member clusters.
+
+##### `spec.clusterNames` is not provided, `spec.clusterSelector` is provided and not empty
+
+```yaml
+spec:
+  clusterSelector:
+    matchLabels:
+      foo: bar
+```
+
+In this case, the resource will only be propagated to member clusters that are labeled
+with `foo: bar`.
+
 ### Example Cleanup
 
 To cleanup the example simply delete the namespace:


docs/userguide.md
```diff
@@ -377,6 +377,68 @@ constructed using the mechanics provided by the core API types (`template`,
...
+    - cluster1
+  clusterSelector:
+    matchLabels:
+      foo: bar
+```
+
+For this case, `spec.clusterSelector` will be ignored as `spec.clusterNames` is
+provided, this can ensure that scheduling can take priority over manual placement.
+
+#### Define an empty `spec.clusterSelector`
```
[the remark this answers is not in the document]
gyliu513: Done

docs/userguide.md
```diff
@@ -377,6 +377,68 @@ constructed using the mechanics provided by the core API types (`template`,
...
+
+```yaml
+spec:
+  clusterSelector: {}
+```
+
+For this case, all clusters will be selected by the federate placement policy.
```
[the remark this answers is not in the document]
gyliu513: Done

docs/userguide.md
```diff
@@ -377,6 +377,68 @@ constructed using the mechanics provided by the core API types (`template`,
...
+#### Define a `spec.clusterSelector` with `matchLabels`
+
+```yaml
+spec:
+  clusterSelector:
+    matchLabels:
+      foo: bar
+```
+
+For this case, only clusters have the label of `foo: bar` will be selected by
```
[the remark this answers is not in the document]
gyliu513: Done

docs/userguide.md
```diff
@@ -247,6 +247,72 @@ done
...
+The following sections detail how `spec.clusterNames` and `spec.clusterSelector` are
+used in determining the clusters that a federated resource should be propagated to.
+
+#### Neither `spec.clusterNames` nor `spec.clusterSelector` is provided
+
+```yaml
+spec: {}
+```
+
+In this case, you can either set `sepc: {}` as above or remove `spec` section from your
```
tghartland: sepc -> spec

docs/userguide.md
```diff
@@ -247,6 +247,72 @@ done
...
+    - cluster2
+    - cluster1
+  clusterSelector:
+    matchLabels:
+      foo: bar
+```
+
+For this case, `spec.clusterSelector` will be ignored as `spec.clusterNames` is provided. This
+ensures that the results of dynamic scheduling have priority over manual definition of a
+cluster selector.
```
tghartland: This sentence is a bit confusing. The dynamic scheduling is the label based one, which *does not* have priority. And "manual" definition should refer to listing each cluster name, not selectors, right?

marun: 'dynamic scheduling' in this instance is something like RSP or job scheduling - a mechanism that actively  to determines which clusters to propagate to and with what configuration.  Label selection is relatively static by comparison - it only considers the labels on a local cluster resources rather than the runtime state of the underlying cluster.

Would you like to suggest alternate terminology?

docs/userguide.md
```diff
@@ -247,6 +247,72 @@ done
 If you were able to verify the resources removed and added back then you have
 successfully verified a working federation-v2 deployment.
 
+In addition to specifying an explicit list of clusters that a resource should be
```
marun: I think a section header is suggested here to differentiate from the example above.

docs/userguide.md
```diff
@@ -247,6 +247,72 @@ done
...
+propagated to via the `spec.clusterNames` field of a placement resource, it is possible
+to use the `spec.clusterSelector` field to provide a label selector that determines
+that list of clusters dynamically.
```
marun: Ah, I think I see the reason for @tghartland's comment below.  There needs to be terminology to clearly differentiate the results of 'scheduling' from the results of 'label selection'.  I think the following could work:

`s/dynamically/at runtime/`

tghartland: Yes, this is where I was getting the idea that dynamic = label selection. Using "at runtime" here would clear up some of the confusion.

docs/userguide.md
tghartland: It's clear to me now that I know the intention, and I think the terminology is fine for the intended meaning. The problem then is that I think it's quite easy to miss the change of context from the first sentence to the second.

Now that I think about it the reason for the label selector being ignored is that I assume RSP doesn't support labels yet. I think if that were mentioned here it would be much easier to read the second sentence as intended.

docs/userguide.md
gyliu513: Done

docs/userguide.md
gyliu513: Done

docs/userguide.md
gyliu513: Done

docs/userguide.md
gyliu513: Done by s/dynamically/runtime

gyliu513:
/cc @pmorie @irfanurrehman 

Can you help lgtm if no further comments as @marun is off till Dec 16?

docs/userguide.md
```diff
@@ -291,6 +291,74 @@ done
...
+The following sections detail how `spec.clusterNames` and `spec.clusterSelector` are
+used in determining the clusters that a federated resource should be propagated to.
+
+##### Neither `spec.clusterNames` nor `spec.clusterSelector` is provided
+
+```yaml
+spec: {}
+```
+
+In this case, you can either set `spec: {}` as above or remove `spec` section from your
```
marun: s/`spec` section/the `spec` field/

docs/userguide.md
```diff
@@ -291,6 +291,74 @@ done
...
+federate placement policy. The resource will not be propagated to member clusters.
```
marun: s/federate//

docs/userguide.md
gyliu513: Done

docs/userguide.md
gyliu513: Done

gyliu513:
@marun done.

marun:
Thank you!

/lgtm

Issue #496, opened 2018-12-06
Precommit script fails locally


danehans:
Whenever I run `./scripts/pre-commit.sh` locally, I get the following error:
```
Downloading test dependencies
Checking initial state of working tree
Verifying Gofmt
Please run hack/update-gofmt.sh to fix the following files:
./pkg/kubefed2/federate/federate.go
```

I plan to submit a PR that updates `federate.go` to be gofmt'd.

PR #497, opened 2018-12-06
Gofmt's federate.go

danehans:
Fixes issue #496


danehans:
This is an issue with my dev environment. I was using go v1.10, which appears to have a gofmt bug. Gofmt is trying to remove the spaces [here](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/kubefed2/federate/federate.go#L43-L46). I no longer experienced this issue after upgrading to go v1.11.

PR #498, opened 2018-12-06
Disable helm chart check until the crds can be updated reliably

### LINKED ISSUE #496: Precommit script fails locally

marun:
Fixes #496

@gyliu513 I'm disabling the ci check for now.  Some merges landed (without re-running their travis jobs) after the check was added and that broke master.  When I tried to fix the problem I ran into the same problems you did with apiserver version skew.  I will file an issue to capture what needs to happen for this check to be re-enabled.

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -159,11 +159,18 @@ join-cluster-list > /dev/null
 echo "Deploying federation-v2"
 ./scripts/deploy-federation.sh ${CONTAINER_REGISTRY_HOST}/federation-v2:e2e $(join-cluster-list)
 
-echo "Checking sync up status of helm chart"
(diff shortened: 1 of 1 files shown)


gyliu513:
> I will file an issue to capture what needs to happen for this check to be re-enabled.

@marun please link your issue here after it is opened ;-)

marun:
@gyliu513 Here it is: #505

PR #503, opened 2018-12-07
Update docs to reflect `kubefed federate`

marun:


docs/userguide.md
```diff
@@ -125,6 +125,50 @@ Status:
     Type:                  Ready
 ```
 
+## Enabling federation of an API type
+
+It is possible to enable federation of any Kubernetes API type (including CRDs) using the
+`kubefed2` command:
+
+```bash
+kubefed2 federate enable <target API type>
```
gyliu513: Do you want to add more detail for this command? Like enable federation with file name directly

```shell
kubefed2 federate enable -f "${filename}" 
```

And more options, like they can use the option `--dry-run -oyaml` to do some test before apply the resource?

docs/userguide.md
```diff
@@ -125,6 +125,50 @@ Status:
...
+which will prompt the sync controller for the type to be stopped.
+
+If the goal is to permanently disable federation of a type, passing the `--delete-from-api`
+flag will remove the `FederatedTypeConfig` and the primitive CRDS created by `enable`:
+
+```bash
+kubefed2 federate disable <FederatedTypeConfig name> --delete-from-api
+```
+
+**WARNING: All primitive custom resources for the type will be removed by this command**
```
gyliu513: Need a period in the end?

docs/userguide.md
marun: I avoided including more detail since my intention is to reduce the options going forward.  I'm not sure the `-f` option will be needed, for example, and `--comparison-field` is also likely to go away.  

I agree that the yaml output would be useful to document, I'll post something today.  And likely -o yaml should force dry run for simplification.

PR #508, opened 2018-12-10
Update federatednamespace-template.yaml

klingeleersbenjamin:
The sample namespace used for the user guide wasn't federated yet.

### DIFF
--- a/example/sample1/federatednamespace-template.yaml
+++ b/example/sample1/federatednamespace-template.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
-kind: Namespace
+apiVersion: primitives.federation.k8s.io/v1alpha1
+kind: FederatedNamespace
 metadata:
   name: test-namespace


klingeleersbenjamin:
I singed it

klingeleersbenjamin:
@gyliu513 

When following the userguide, the namespace is only made in the first cluster.

for r in configmaps secrets service deployment serviceaccount job; do
    for c in cluster1 cluster2; do
        echo; echo ------------ ${c} resource: ${r} ------------; echo
        kubectl --context=${c} -n test-namespace get ${r}
        echo; echo
    done
done

This command should show resources but because there is no namespace test-namespace in the second cluster none of the resources are made in the second cluster.



------------ cluster1 resource: configmaps ------------

NAME             DATA   AGE
test-configmap   1      40s



------------ cluster2 resource: configmaps ------------

No resources found.

gyliu513:
Namespaces working fine after @klingeleersbenjamin using master build.

PR #515, opened 2018-12-13
WIP Cluster taints and placement tolerations

tghartland:
I am currently only acting on NoSchedule taints, as these can be easily included in the scheduling. I am not sure how to proceed with PreferNoSchedule and NoExecute, as these are far more complex. I have not yet managed to find how PreferNoSchedule is done in kubernetes, and for NoExecute we will probably require a controller to watch for taint changes or something similar.

A `taint` command is added under `kubefed2`, which is used mostly the same as `kubectl taint`. Most of the utils are just copied from the `kubectl taint` utils. At the moment a major missing feature is being able to taint multiple clusters in a single command or taint based on label selections.

I will not have much more time to work on this until January but I am making the WIP PR for discussion about how best to implement the missing features (and what can be improved in what I already have).


marun:
@tghartland I appreciate your interest in implementing this feature.  Would it possible for you to document your use case in more detail in an issue?  I'd like to be able to be educated enough to review effectively, and I'm interested to understand why it is necessary for taints and tolerations to be supported directly in placement policy vs via a higher-order mechanism like scheduling.

tghartland:
@marun I've made an issue #524 with more information.

tghartland:
@marun I have not done any more work on this than what is already here. I wasn't sure how (or even if) the more complex PreferNoSchedule and NoExecute taints should be implemented, and how closely they should match the node taint behaviours.

Getting something working with annotations to just do just the NoSchedule effect is probably a better way to proceed for now. Will that require its own controller to make it able to react to changes after the initial placement calculation, or is there some way to just re-trigger the placement on notification of an annotation change?

marun:
@tghartland I think the existing controller should be respond to the annotation in the same way you've proposing to respond to a change in placement specification in this PR.  The reason to use an annotation instead of explicit API support would be to avoid including something in the API whose future is not yet agreed upon.

Issue #506, opened 2018-12-07
Enable CI testing of helm chart


marun:
Currently the helm chart is not tested in CI, so ensuring it is current with changes in master is a manual process.  As a starting point, a new test run could be added that tears down the most recent script-deployed control plane, deploys with helm, and runs the e2e tests.  Eventually the helm chart could replace the script-based deployment entirely when it proves reliable.

marun:
This was fixed by #525

PR #525, opened 2018-12-19
enable helm chart installation test in CI

xunpan:
#506

use helm chart to deploy federation. It is targeted to replace script
installation for federation. Currently, add it in CI test to make sure
it is stable enough before replace the script installation method

### DIFF
--- a/test/e2e/scheduling.go
+++ b/test/e2e/scheduling.go
@@ -175,6 +175,11 @@ var _ = Describe("ReplicaSchedulingPreferences", func() {
 					if err != nil {
 						tl.Fatalf("Failed waiting for matching overrides: %v", err)
 					}
+
+					err = deleteTestObj(typeConfig, kubeConfig, name, namespace)
+					if err != nil {
+						tl.Fatalf("Deletion of test object failed in fedeartion: %v", err)
+					}
 				})
 			}
 		})
@@ -241,6 +246,21 @@ func createTestObjs(tl common.TestLogger, fedClient clientset.Interface, typeCon
 	}
 
 	return name, nil
+}
+
+func deleteTestObj(typeConfig typeconfig.Interface, kubeConfig *restclient.Config, name, namespace string) error {
+	templateAPIResource := typeConfig.GetTemplate()
+	templateClient, err := util.NewResourceClient(kubeConfig, &templateAPIResource)
+	if err != nil {
+		return err
+	}
+
+	err = templateClient.Resources(namespace).Delete(name, &metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func waitForMatchingPlacement(tl common.TestLogger, typeConfig typeconfig.Interface, kubeConfig *restclient.Config, name, namespace string, expected map[string]int32) error {


scripts/deploy-federation-with-helm-chart.sh
```diff
@@ -0,0 +1,130 @@
...
+  docker build ${dockerfile_dir} -t "${IMAGE_NAME}"
+  ${DOCKER_PUSH_CMD}
+fi
+
+if ! kubectl get ns "${PUBLIC_NS}" > /dev/null 2>&1; then
+    kubectl create ns "${PUBLIC_NS}"
+fi
+
+go build -o bin/kubefed2 cmd/kubefed2/kubefed2.go
+if [[ "${SYNC_CHART}" ]]; then
```
gyliu513: When SYNC_CHART will be true?

xunpan: It is a function to update files before helm install.

scripts/delete-federation-with-helm-chart.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+
+# Unjoin clusters by removing objects added by kubefed2.
+HOST_CLUSTER="$(kubectl config current-context)"
+JOINED_CLUSTERS="$(kubectl -n "${NS}" get federatedclusters -o=jsonpath='{range .items[*]}{.metadata.name}{" "}{end}')"
+for c in ${JOINED_CLUSTERS}; do
+  ./bin/kubefed2 unjoin "${c}" --host-cluster-context "${HOST_CLUSTER}" --remove-from-registry --v=2 ${KF_NS_ARG}
+done
+
+${KCD} -n "${NS}" FederatedTypeConfig --all
```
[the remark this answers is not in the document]
xunpan: Currently we have no function to check federated resources before delete federation. I think it should be another function requirement if it is a case.

scripts/delete-federation-with-helm-chart.sh
```diff
@@ -0,0 +1,66 @@
...
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+
+KCD="kubectl --ignore-not-found=true delete"
+NS="${FEDERATION_NAMESPACE:-federation-system}"
+PUBLIC_NS=kube-multicluster-public
+NAMESPACED="${NAMESPACED:-}"
+
+KF_NS_ARG="--federation-namespace=${NS} "
+if [[ "${NAMESPACED}" ]]; then
+  KF_NS_ARG+="--registry-namespace=${NS}"
```
gyliu513: Why did `NAMESPACED` request `--registry-namespace`? Please check here https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/userguide.md#namespaced-federation for what does `NAMESPACED` mean.

scripts/delete-federation-with-helm-chart.sh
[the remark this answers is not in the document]
gyliu513: This is the use case, suppose we use the `helm install` install the federation, and then joined some clusters, also created some federated resources. 

Then if I run this script, it will be failed, as the deletion of delete those federated resources will be failed.

How about add some checking before `unjoin` in Line36, if there some federated resources, just report some warning message and return, do not unjoin and delete resources for such cases.

@xunpan @marun

scripts/deploy-federation-with-helm-chart.sh
```diff
@@ -0,0 +1,162 @@
...
+
+LATEST_IMAGE_NAME=quay.io/kubernetes-multicluster/federation-v2:latest
+USE_LATEST=
+if [[ "${IMAGE_NAME}" == "$LATEST_IMAGE_NAME" ]]; then
+  USE_LATEST=y
+fi
+
+KF_NS_ARGS="--federation-namespace=${NS} "
+if [[ "${NAMESPACED}" ]]; then
+  KF_NS_ARGS+="--registry-namespace=${NS} --limited-scope=true"
```
gyliu513: As above, it is not request `registry-namespace` for `NAMESPACED` federation.

scripts/delete-federation-with-helm-chart.sh
xunpan: I did not touch this part. It follows the rule in delete-federation.sh. 

I can fix it if it is an issue.

scripts/deploy-federation-with-helm-chart.sh
xunpan: as above

scripts/delete-federation-with-helm-chart.sh
xunpan: Yes. I checked related feature. It is used to test namespaced cluster registry.

scripts/pre-commit.sh
```diff
@@ -165,14 +165,26 @@ PATH="${PATH}:${base_dir}/bin" ./scripts/sync-up-helm-chart.sh
...
 
-echo "Deleting federation-v2"
-./scripts/delete-federation.sh
+echo "Deleting federation-v2 with helm chart"
+USE_CHART=true ./scripts/delete-federation.sh
 
 echo "Deploying namespaced federation-v2"
 FEDERATION_NAMESPACE=foo NAMESPACED=y ./scripts/deploy-federation.sh ${CONTAINER_REGISTRY_HOST}/federation-v2:e2e $(join-cluster-list)
 
+echo "Deleting namespaced federation-v2"
```
marun: Why is the scripted namespaced deployment not tested before being deleted?

xunpan: I did not test cases for both namesapced and global cluster which deployed by old scripts. 

Old script deployment way is tested for namespaced and global federation deployment to make sure related script works correctly. 

Test functions on helm chart depoyment to use the new way to test functions.

In such case, if helm chart is broken, we can switch back to script based testing as candidate. However, helm chart seems work fine based on my test. So, I think testing it on old script based depoloyment seems not necessary.

This is my thought. If we still need test on both script based deployments, I can add it back. 
Any comments?

scripts/delete-federation.sh
```diff
@@ -23,6 +23,46 @@ set -o pipefail
 
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 
+function delete-with-kubebuilder() {
```
gyliu513: I'd like to rename this function as `delete-with-kubectl`

scripts/deploy-federation.sh
```diff
@@ -47,6 +47,71 @@ set -o pipefail
 
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 
+function deploy-with-kubebuilder() {
```
gyliu513: s/deploy-with-kubebuilder/deploy-with-kubectl

scripts/deploy-federation.sh
```diff
@@ -28,7 +28,7 @@
 #
 #   $ minikube start
 #
-# This script depends on kubectl and kubebuilder being installed in
+# This script depends on kubectl and kubebuilder/helm being installed in
```
gyliu513:
```
# This script depends on kubectl and kubebuilder being installed in
# the path.  If you want to install Federation via helm chart, you may
# also need to install helm in the path. These and other test binaries
# can be installed via the download-binaries.sh script, which downloads
# to ./bin:
```

scripts/delete-federation.sh
marun: How about delete-script-deployment?

scripts/pre-commit.sh
marun: Both script based deployments need to remain tested in ci so long as they remain in the tree and a documented way of deploying fedv2.  Without ci testing, breakage will be inevitable.  As per my review comment, the way likely forward will be either be removing the script-based deployment if helm is deemed a suitable replacement, or minimizing the testing to a representative subset so that the ci cost isn't doubled.

xunpan: add tests for script based deployment

marun:
It appears that CI job is failing due to the deletion of the namespaced deployment not being completed successfully.

xunpan:
This should be a bug in testing scheduling.go. 
It creates test resources but not delete after test done. When `foo` namespace is deleted, related federateddeployment and federatedreplicaset resources are not deleted because the federation finalizer.
So, I just delete the resource after test is done.

I think it needs log enhancement beause related case logs nothing.

xunpan:
there is no code change for related issue in failed log.
And local test passed.

/retest

marun:
Thank you for your patient efforts!

/approve
/lgtm

Issue #83, opened 2018-05-23
Replace dynamic client usage with controller-runtime's generic client


font:
The current dynamic client in use has now been deprecated. We need to update the client-go vendored dependency and update the dynamic client to use the new version. See https://github.com/kubernetes/kubernetes/pull/63446 for more details.

marun:
This will likely depend on upgrading all dependencies to match kube 1.11, not just client-go.

marun:
The dynamic client has been superceded by the generic client defined by the `controller-runtime`:

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/example_test.go

PR #530, opened 2018-12-20
Upgrade client go to kubernetes-1.12.4

fisherxu:
See [comments here](https://github.com/kubernetes-sigs/federation-v2/pull/394#issuecomment-449002333) .

/hold


shashidharatd:
This also partially addresses https://github.com/kubernetes-sigs/federation-v2/issues/83 and still uses dynamic client. This LGTM as an intermediate step before migrating new kubebuilder 1.0.6. Would request @font / @marun to approve.

/lgtm

test/e2e/crd.go
```diff
@@ -23,21 +23,20 @@ import (
 	"github.com/pborman/uuid"
 
 	apicommon "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
+	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
```
marun: It is an (unfortunately) undocumented goal to format imports as follows:

```
[system imports]

[3rd party imports (e.g. uuid, glog)]

[kube and community imports]

[federation imports]

[testing imports]
```

test/e2e/crd.go
```diff
@@ -103,14 +102,13 @@ func validateCrdCrud(f framework.FederationFramework, targetCrdKind string, name
 	userAgent := fmt.Sprintf("test-%s-crud", strings.ToLower(targetCrdKind))
 
 	// Create the target crd in all clusters
-	var pools []dynamic.ClientPool
-	var hostPool dynamic.ClientPool
+	var configs []*rest.Config
+	var config *rest.Config
```
marun: s/config/hostConfig/ for clarity.

marun:
Thank you for doing this!  I have a minor nit about import ordering inline, but otherwise lgtm.

Gopkg.toml
```diff
@@ -24,11 +24,24 @@ required = ["github.com/kubernetes/repo-infra/verify/boilerplate/test"]
...
   name="k8s.io/apimachinery"
-  version="kubernetes-1.10.1"
+  version="kubernetes-1.12.4"
 
 [[override]]
   name="k8s.io/client-go"
-  version="kubernetes-1.10.1"
+  version="kubernetes-1.12.4"
+
+[[override]]
```
gyliu513: A question here: Why did you add another three overrides here?

fisherxu: If we just upgrade the client-go, this three older repo will not work, they should keep the same version :)

gyliu513: Got it, thanks @fisherxu :-)

test/e2e/crd.go
marun: Oh, that's because they are the same value.  Please remove the `config` variable and ensure instances of `hostPool` are changed to `hostConfig`.

fisherxu: have done :-)

marun:
When I try to run `dep ensure` locally against this PR, I see the following error:

```
grouped write of manifest, lock and vendor: failed to export k8s.io/client-go: fatal: failed to unpack tree object 5e6a3d4e34f694e895b13ae728111e726a5b69df
: exit status 128
```

I don't have this problem against master, which suggests that this PR may be leaving the tree in a problematic state for future vendoring.  I'm putting a hold on the PR until I can rule out that possibility.

/hold

marun:
False alarm.

/hold cancel
/approve
/lgtm

PR #533, opened 2018-12-21
stop scheduler and plugins if resource is deleted

### LINKED ISSUE #386: stop scheduling controller if federatedtypeconfig resource is deleted from system
In PR [366 ](https://github.com/kubernetes-sigs/federation-v2/pull/366), we implement basic functions to reconcile federatedtypeconfig for scheduling controller when the resource is created.

This ticket is used to trace related resource deletion case to make sure schedulingmanager works gracefully for all cases.

xunpan:
Fix  #386

stop scheduling controller if federatedtypeconfig resource is deleted
from system

I also make reconcile federatedtypeconfig log V(3) to avoid messing up
default messages.

@marun

### DIFF
--- a/pkg/controller/schedulingmanager/controller.go
+++ b/pkg/controller/schedulingmanager/controller.go
@@ -50,7 +50,8 @@ type SchedulerController struct {
 	config *util.ControllerConfig
 
 	runningPlugins sets.String
-	templateKinds  map[string]string
+	// mapping qualifiedname to template kind for managing plugins in scheduler
+	templateKindMap map[string]string
 }
 
 func StartSchedulerController(config *util.ControllerConfig, stopChan <-chan struct{}) {
@@ -68,10 +69,10 @@ func StartSchedulerController(config *util.ControllerConfig, stopChan <-chan str
 
 func newController(config *util.ControllerConfig, client corev1alpha1client.CoreV1alpha1Interface) *SchedulerController {
 	c := &SchedulerController{
-		config:         config,
-		scheduler:      make(map[string]schedulingtypes.Scheduler),
-		runningPlugins: sets.String{},
-		templateKinds:  make(map[string]string),
+		config:          config,
+		scheduler:       make(map[string]schedulingtypes.Scheduler),
+		runningPlugins:  sets.String{},
+		templateKindMap: make(map[string]string),
 	}
 
 	fedNamespace := config.FederationNamespace
@@ -176,7 +177,7 @@ func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.R
 		return util.StatusError
 	}
 	c.runningPlugins.Insert(qualifiedName.Name)
-	c.templateKinds[qualifiedName.Name] = templateKind
+	c.templateKindMap[qualifiedName.Name] = templateKind
 
 	return util.StatusAllOK
 }
@@ -191,11 +192,11 @@ func (c *SchedulerController) stopScheduler(schedulingKind string, qualifiedName
 		return
 	}
 
-	glog.Infof("Stopping plugin for %q with kind %q", qualifiedName.Name, c.templateKinds[qualifiedName.Name])
+	glog.Infof("Stopping plugin for %q with kind %q", qualifiedName.Name, c.templateKindMap[qualifiedName.Name])
 
-	scheduler.StopPlugin(c.templateKinds[qualifiedName.Name])
+	scheduler.StopPlugin(c.templateKindMap[qualifiedName.Name])
 	c.runningPlugins.Delete(qualifiedName.Name)
-	delete(c.templateKinds, qualifiedName.Name)
+	delete(c.templateKindMap, qualifiedName.Name)
 
 	// if all resources registered to same scheduler are deleted, the scheduler should be stopped
 	resources := schedulingtypes.GetSchedulingKinds(qualifiedName.Name)
--- a/pkg/controller/schedulingmanager/controller.go
+++ b/pkg/controller/schedulingmanager/controller.go
@@ -45,8 +45,7 @@ type SchedulerController struct {
 
 	worker util.ReconcileWorker
 
-	scheduler    map[string]schedulingtypes.Scheduler
-	stopChannels map[string]chan struct{}
+	scheduler map[string]schedulingtypes.Scheduler
 
 	config *util.ControllerConfig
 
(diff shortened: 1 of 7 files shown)


pkg/controller/schedulingmanager/controller.go
```diff
@@ -156,22 +169,57 @@ func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.R
...
 		c.scheduler[schedulingKind] = scheduler
+		c.stopChannels[schedulingKind] = stopChan
 	}
 
 	templateKind := typeConfig.GetTemplate().Kind
-	glog.Infof("Start plugin with kind %s for scheduling type %s", templateKind, schedulingKind)
-	err = scheduler.StartPlugin(typeConfig, c.stopChan)
+	glog.Infof("Start plugin kind %s for scheduling type %s", templateKind, schedulingKind)
+
+	stopChan := make(chan struct{})
```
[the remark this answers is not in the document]
xunpan: I like this comments. Personally, I recall and learn something from this.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -156,22 +169,57 @@ func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.R
...
 		runtime.HandleError(fmt.Errorf("Error starting plugin for %q : %v", templateKind, err))
 		return util.StatusError
 	}
 	c.runningPlugins.Insert(qualifiedName.Name)
+	c.templateKinds[qualifiedName.Name] = templateKind
 
 	return util.StatusAllOK
 }
+
+func (c *SchedulerController) stopScheduler(schedulingKind string, qualifiedName util.QualifiedName) {
```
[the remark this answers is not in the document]
xunpan: I'd like to keep the name bacause it tries to stop scheduler after all related plugins stopped.

xunpan:
I addressed related comments. 
Now, I start working for how to make related case tested with coverage.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -60,18 +60,18 @@ func StartSchedulerController(config *util.ControllerConfig, stopChan <-chan str
...
 	controller.Run(stopChan)
 }
 
-func newController(config *util.ControllerConfig, client corev1alpha1client.CoreV1alpha1Interface, stopChan <-chan struct{}) *SchedulerController {
+func newController(config *util.ControllerConfig, client corev1alpha1client.CoreV1alpha1Interface) *SchedulerController {
 	c := &SchedulerController{
-		scheduler:      make(map[string]schedulingtypes.Scheduler),
 		config:         config,
-		stopChan:       stopChan,
+		scheduler:      make(map[string]schedulingtypes.Scheduler),
```
marun: (No action required) Please avoid mixing non-functional changes like ordering with functional changes to simplify both review and future maintenance.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -113,31 +113,33 @@ func (c *SchedulerController) Run(stopChan <-chan struct{}) {
...
 	typeConfig := cachedObj.(*corev1a1.FederatedTypeConfig)
+	if typeConfig.Spec.PropagationEnabled == false || typeConfig.DeletionTimestamp != nil {
+		c.stopScheduler(schedulingType.Kind, qualifiedName)
+		return util.StatusAllOK
+	}
 
-	deleted := typeConfig.DeletionTimestamp != nil
-	if deleted {
-		// Do not support deletion yet
+	if c.runningPlugins.Has(qualifiedName.Name) {
```
marun: (No action required) I'm not sure why management of running plugins needs to be performed outside of the scheduling preference controller.   Or that the scheduling preference controller needs to be started separately from the plugin.  I think this suggests the possibility of simplification in future iterations of this area of functionality.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -47,10 +47,10 @@ type SchedulerController struct {
 
 	scheduler map[string]schedulingtypes.Scheduler
 
-	config   *util.ControllerConfig
-	stopChan <-chan struct{}
+	config *util.ControllerConfig
 
 	runningPlugins sets.String
+	templateKinds  map[string]string
```
marun: This map appears to associate a template kind with a scheduling kind.  I don't think this is obvious from the name `templateKinds` and encourage you to either choose a more descriptive name or add a comment to document your intent.

(No action required) As per a subsequent comment, I'm not sure why the scheduling preference controller shouldn't be tracking the association between scheduling and template kinds.

pkg/schedulingtypes/plugin.go
```diff
@@ -60,6 +60,8 @@ type Plugin struct {
 	placementClient util.ResourceClient
 
 	typeConfig typeconfig.Interface
+
+	stopChannel chan struct{}
```
marun: Nice cleanup!  

As per a previous question, is there a reason you switched the stop channels to be bidirectional?  Most stop channels that I've seen in kube are one way, and I assume that's because the signaling is intended to only flow from caller to callee.

pkg/controller/schedulingmanager/controller.go
xunpan: Got it. 
The reason here is that old logic is more clear to make `scheduler` near to old `stopChannels`. After I update code based on your comments to delete `stopChannels`, this is not recovered.

pkg/controller/schedulingmanager/controller.go
xunpan: Agree. This is also what I'm thinking when change the stop management code to the related module.

pkg/schedulingtypes/plugin.go
xunpan: we cannot close a receive-only channel. But I need to close the channel to stop related controllers.
```
$ CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/hyperfed cmd/hyperfed/main.go 
\# github.com/kubernetes-sigs/federation-v2/pkg/controller/schedulingpreference
pkg/controller/schedulingpreference/controller.go:180:7: invalid operation: close(s.stopChannel) (cannot close receive-only channel)
```

marun: Ah, I see.  It's necessary to create the stop channel as writeable so it can be closed, and then it can be passed as read-only.  Previously the channel was only being passed in so it was always read-only.

In future, please avoid resolving conversations in which you answer a question.  That way, whoever asked the question can determine for themselves whether the response satisfied their query.

marun:
Thank you!

/approve
/lgtm

marun:
ci seems to be having trouble - it's failing even before it gets to deployment or testing.

PR #539, opened 2019-01-03
Updated development guide for how to update toc.

gyliu513:


docs/development.md
```diff
@@ -333,3 +334,8 @@ In order to test the latest stable released version (tagged as `latest`) on
 your kubernetes cluster, follow the
 [automated](userguide.md#automated-deployment) or
 [manual](userguide.md#manual-deployment) instructions from the user guide.
+
+## Updating Document
+
+If you are going to add some new sections for the document, make sure the
```
marun: s/make sure the toc was also updated by.../make sure to update the toc.  This can be done manually or with [doctoc].../

Given that we're avoiding adding a dependency for developers on npm (at least for now), I think it would make sense to mention doctoc as an option but not a requirement.

gyliu513: Done, thanks @marun

docs/development.md
```diff
@@ -333,3 +334,8 @@ In order to test the latest stable released version (tagged as `latest`) on
 your kubernetes cluster, follow the
 [automated](userguide.md#automated-deployment) or
 [manual](userguide.md#manual-deployment) instructions from the user guide.
+
+## Updating Document
+
+If you are going to add some new sections for the document, make sure to update the toc.
```
marun: Hmm, sorry I missed this earlier but I think 'Table of Contents' would be preferable to 'toc' so there's no confusion.

gyliu513: @marun Done, but I was using 'table of contents' instead, hope it is OK.

marun:
Thank you!

/lgtm

PR #541, opened 2019-01-03
tiny fix with gosimple

carmark:


marun:
I appreciate your initiative in wanting to perform cleanup on the codebase.  I'm not sure I'd suggest cleanup of this repo at this point, though, since the codebase is alpha and any given piece of code has an uncertain future.  It certainly makes sense to cleanup code that is intended to be maintained over the long term (i.e. beta and beyond), we're just not quite there yet.

When performing maintenance, I appreciate the separation of different kinds of changes into separate commit to ensure a clean history (e.g. import ordering in one commit, boolean comparison in another).

If you're interested in implementing functional changes to the repo, please reach out on the #sig-multicluster slack channel and I or other contributors will be happy to assist.

carmark:
Hello @marun , many thanks for advice.  I did split the commit and remove the import order fix, since it is unnecessary for now.

By the way, I have joined the slack channel, but could you please share the roadmap or todo list? or should I pick them from the issue list?

Thanks,

pkg/controller/ingressdns/controller.go
```diff
@@ -214,7 +214,7 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 
 	glog.V(2).Infof("Starting to reconcile IngressDNS resource: %v", key)
 	startTime := time.Now()
-	defer glog.V(2).Infof("Finished reconciling IngressDNS resource %v (duration: %v)", key, time.Now().Sub(startTime))
```
marun: (No action required) The duplication across controllers suggests the need for a common controller framework.

pkg/controller/dnsendpoint/ingress_test.go
```diff
@@ -182,9 +182,9 @@ func TestGetEndpointsForIngressDNSObject(t *testing.T) {
 	for testName, tc := range testCases {
 		t.Run(testName, func(t *testing.T) {
 			endpoints, err := getIngressDNSEndpoints(&tc.dnsObject)
-			if tc.expectError == false && err != nil {
+			if !tc.expectError && err != nil {
 				t.Fatalf("Unexpected error: %v", err)
-			} else if tc.expectError == true && err == nil {
```
marun: (No action required) I'm definitely in favor of avoiding comparing booleans with a value.

pkg/kubefed2/federate/disable.go
```diff
@@ -121,12 +121,7 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		Name:      j.targetName,
 	}
 
-	err = DisableFederation(cmdOut, hostConfig, typeConfigName, j.delete, j.DryRun)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return DisableFederation(cmdOut, hostConfig, typeConfigName, j.delete, j.DryRun)
```
marun: (No action required) I think this kind of change is of questionable value.  It might look cleaner, but if there are more than one return statement in a function - as is the case in most of these changes in this PR - I think having a default return statement may well be preferable:
 
 - new code can be added after the last error-checked statement without modifying the last statement
 - the last error-checked statement can be removed without further modification (i.e. adding a return statement) 

I'm not suggesting code should be one way or the other, though, and I wouldn't reject code submitted in either style.  Similarly, I don't think there is a basis for 'cleanup' of either form.  Whoever submits a given piece of functionality gets to decide what they think is best, and maintainers making functional changes can modify that initial decision if it makes sense to do so.

marun:
@carmark Fedv2 is a small project and the roadmap and backlog are not quite formalized.  Also, as with any community effort, the best way to get involved is to identify a mentor and work with them to familiarize yourself with the project and its goals and how your interests can be aligned to the current effort.  I'm happy to help, just reach out to me on slack, and I would also encourage you to join the next federation working group meeting (see the slack channel topic for details).

marun:
Thank you for your initiative!

/approve
/lgtm

marun:
Ugh, I need to decide if the owner reference test is fixable.  It's flaking really hard lately.

PR #544, opened 2019-01-07
Skip check of owner reference addition for namespaced resources 

marun:
Coverage is already achieved by testing cluster-scoped resources since the codepath for both is the same, and the garbage collection of namespaced resources takes an arbitrary amount of time that is not suitable for inclusion in an e2e test.

cc: @xunpan


marun:
rebased

PR #548, opened 2019-01-15
Factor host cluster interaction out of the sync controller

marun:
This change factors the logic of federated type composition out of the sync controller and into an accessor type that creates a logical federated resource to encapsulates one or more underlying kubernetes resources.  This separation is intended to simplify future maintenance.

### DIFF
--- a/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
+++ b/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
@@ -201,12 +201,12 @@ func (f *FederatedTypeConfig) GetTemplate() metav1.APIResource {
 }
 
 func (f *FederatedTypeConfig) GetPlacement() metav1.APIResource {
-	namespaced := f.isPrimitiveNamespaced()
+	namespaced := f.GetFederatedNamespaced()
 	return apiResourceToMeta(f.Spec.Placement, namespaced)
 }
 
 func (f *FederatedTypeConfig) GetOverride() metav1.APIResource {
-	namespaced := f.isPrimitiveNamespaced()
+	namespaced := f.GetFederatedNamespaced()
 	return apiResourceToMeta(f.Spec.Override, namespaced)
 }
 
@@ -222,7 +222,7 @@ func (f *FederatedTypeConfig) GetEnableStatus() bool {
 	return f.Spec.EnableStatus
 }
 
-func (f *FederatedTypeConfig) isPrimitiveNamespaced() bool {
+func (f *FederatedTypeConfig) GetFederatedNamespaced() bool {
 	// Special-case the scope of namespace primitives since it will
 	// hopefully be the only instance of the scope of a federation
 	// primitive differing from the scope of its target.
@@ -234,6 +234,16 @@ func (f *FederatedTypeConfig) isPrimitiveNamespaced() bool {
 		return true
 	}
 	return f.Spec.Namespaced
+}
+
+func (f *FederatedTypeConfig) GetFederatedKind() string {
+	// TODO(marun) Use the constant in pkg/controller/util
+	if f.Name == "namespaces" {
+		// The template type is 'Namespace', so return
+		// 'FederatedNamespace' for consistency with other types.
+		return "FederatedNamespace"
+	}
+	return f.GetTemplate().Kind
 }
 
 func apiResourceToMeta(apiResource APIResource, namespaced bool) metav1.APIResource {
--- a/pkg/controller/sync/accessor.go
+++ b/pkg/controller/sync/accessor.go
@@ -1,0 +1,288 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sync
+
+import (
+	"github.com/golang/glog"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
+	fedclientset "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/sync/placement"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/sync/version"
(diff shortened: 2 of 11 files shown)


pkg/controller/sync/placement/resource.go
```diff
@@ -32,21 +32,19 @@ type ResourcePlacementPlugin struct {
...
 }
 
-func NewResourcePlacementPlugin(client util.ResourceClient, targetNamespace string, triggerFunc func(pkgruntime.Object), defaultAll bool) PlacementPlugin {
-	return newResourcePlacementPluginWithOk(client, targetNamespace, triggerFunc, defaultAll)
+func NewResourcePlacementPlugin(client util.ResourceClient, targetNamespace string, triggerFunc func(pkgruntime.Object)) PlacementPlugin {
+	return newResourcePlacementPluginWithOk(client, targetNamespace, triggerFunc)
 }
 
-func newResourcePlacementPluginWithOk(client util.ResourceClient, targetNamespace string, triggerFunc func(pkgruntime.Object), defaultAll bool) *ResourcePlacementPlugin {
+func newResourcePlacementPluginWithOk(client util.ResourceClient, targetNamespace string, triggerFunc func(pkgruntime.Object)) *ResourcePlacementPlugin {
```
xunpan: "WithOK" seems tell nothing. What about remove it or use Impl as postfix? Same as computePlacementWithOK()

pkg/controller/sync/accessor.go
```diff
@@ -0,0 +1,275 @@
...
+			// Ensure the federated name is namespace qualified.
+			federatedName.Namespace = federatedName.Name
+		} else {
+			// Ensure the target name is not namespace qualified.
+			targetName.Namespace = ""
+		}
+
+		// A namespace is only federated if it has a corresponding
+		// placement resource.
+		var err error
```
xunpan: := seems more concise for err definition

pkg/controller/sync/accessor.go
```diff
@@ -0,0 +1,275 @@
...
+	if !a.placementPlugin.HasSynced() {
+		glog.V(2).Infof("Placements not synced")
+		return false
+	}
+	return true
+}
+
+func (a *resourceAccessor) FederatedResource(eventSource util.QualifiedName) (FederatedResource, error) {
+	if a.targetIsNamespace && a.isSystemNamespace(eventSource.Name) {
+		glog.V(7).Infof("Ignoring system namespace %q", eventSource.Name)
```
xunpan: question: any reasaon to choose level 7 for this message. Is there any rule in using level of glog.V?

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -236,6 +236,16 @@ func (f *FederatedTypeConfig) isPrimitiveNamespaced() bool {
 	return f.Spec.Namespaced
 }
 
+func (f *FederatedTypeConfig) GetFederatedKind() string {
+	// TODO(marun) Use the constant in pkg/controller/util
+	if f.Name == "namespaces" {
+		// The template type is 'Namespace', so return
+		// 'FederatedNamespace' for consistency with other types.
+		return "FederatedNamespace"
```
xunpan: It is used for messages. Make a "FederatedNamespace" seems a little confused. Because no one can get this as a resource in the cluster. "federated Namespace" seems better to me.

pkg/controller/sync/placement/resource.go
marun: Please limit yourself to commenting on changes rather than existing code.  `withOK` is existing code.

If you examine the definition of `ResourcePlacementPlugin` you will notice that it has a method `computePlacementWithOk` that returns a boolean indication of whether placement could be successfully retrieved.  If you examine the definition of `NamespacedPlacementPlugin`, you will find that this method is used to determine whether namespace placement was retrieved, and if so, the intersection of namespace and resource placement is used.  If namespace placement was not retrieved, the resource placement is used by itself.

Also, please avoid reviewing PRs that clearly state 'this PR is not ready for review' (and/or WIP in the title).  If a feature requires more than 1 PR to implement, I will frequently post them in a chain with subsequent PRs marked WIP until their predecessor has merged.  This avoids reviewing follow-on PRs until all review-related changes from previous PRs have been incorporated.

pkg/controller/sync/accessor.go
marun: It's not possible to use `:=` when one of the targets is already defined (in this case, as a struct variable.

pkg/controller/sync/accessor.go
marun: The value of logging when a resource is ignored is extremely low, hence the low log level.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
marun: As per the comment, `FederatedNamespace` is for consistency with the kind of every other type (e.g. FederatedDeployment, FederatedConfigMap, etc).  I don't think it makes sense to diverge from the kind form for namespaces.

pkg/controller/sync/placement/resource.go
xunpan: Got it. I will review it after it is ready for review.

pkg/controller/sync/accessor.go
```diff
@@ -0,0 +1,275 @@
...
+	for _, obj := range a.templateStore.List() {
+		visitFunc(obj)
+	}
+}
+
+func (a *resourceAccessor) isSystemNamespace(namespace string) bool {
+	// TODO(font): Need a configurable or discoverable list of namespaces
+	// to not propagate beyond just the default system namespaces e.g.
+	switch namespace {
+	case "kube-system", "kube-public", "default", a.fedNamespace:
```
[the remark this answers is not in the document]
marun: 'system namespace' in this context means 'namespace which is expected to exist in member clusters and therefore should not be created or deleted by federation'.  That is true of the namespaces listed here, but is not true of the multicluster public namespace.  I should add comment that clarifies this.

marun:
@shashidharatd Thank you for your review!  #545 already forces the explicit specification of namespace placement for namespace propagation.  The checks in the code remain in case someone adds placement for a system namespace that instructs removal of the namespace.

The reunification PR will transition to having an explicit `FederatedNamespace` that contains overrides and placement (the template will continue to be the containing namespace).  Using the containing namespace as the template ensures uniform handling of propagation.  Otherwise it would be possible to specify a template and overrides that specified a different namespace in the host cluster than was actually present, which I think is more confusing.

shashidharatd:
Thanks for the clarification @marun. @Xunpan, I guess you are reviewing this pr too. Would wait for your inputs before merging. Thanks!

pkg/controller/util/deletionhelper/deletion_helper.go
```diff
@@ -27,6 +27,7 @@ import (
 	finalizersutil "github.com/kubernetes-sigs/federation-v2/pkg/controller/util/finalizers"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
```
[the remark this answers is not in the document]
marun: Done.  This is legacy code sourced from fedv1, and it has bad habits like not aliasing apimachinery runtime to avoid potential conflict with golang's runtime package.

marun:
Fixed duplicate import.  prow is once again working so this PR should be mergeable.

marun:
@shashidharatd Rebased.

PR #554, opened 2019-01-22
Kindly remind the user to replace the parameter of domain

shlao:
This PR fixes the potential problem that users maybe configure the wrong domain 'example.org', because we use the 'example.com' in this article.

Here is the extract from section:[Creating service resources](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/ingress-service-dns-with-coredns.md#creating-service-resources) 
$ cat <<EOF | kubectl create -f -
apiVersion: multiclusterdns.federation.k8s.io/v1alpha1
kind: Domain
... ...
domain: example.com
... ...

### DIFF
--- a/docs/ingress-service-dns-with-coredns.md
+++ b/docs/ingress-service-dns-with-coredns.md
@@ -31,8 +31,7 @@ Install Federation-v2 with minikube in [User Guide](https://github.com/kubernete
 
 ## Installing ExternalDNS
 
-Install ExternalDNS with CoreDNS as backend in your host cluster. You can follow the [tutorial](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md).  
-**Note**: You should replace parameter of `parameters: example.org` with `parameters: example.com` when [Installing CoreDNS](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md#installing-coredns)
+Install ExternalDNS with CoreDNS as backend in your host cluster. You can follow the [tutorial](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md).
 
 To make it work for federation resources, you need to use below ExternalDNS deployment instead of the one in the tutorial.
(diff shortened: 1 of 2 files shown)


docs/ingress-service-dns-with-coredns.md
```diff
@@ -34,6 +34,8 @@ Install Federation-v2 with minikube in [User Guide](https://github.com/kubernete
 Install ExternalDNS with CoreDNS as backend in your host cluster. You can follow the [tutorial](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md).
 
 To make it work for federation resources, you need to use below ExternalDNS deployment instead of the one in the tutorial.
+**Note**: You should replace parameter of `parameters: example.org` with `parameters: example.com` when [Installing CoreDNS](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md#installing-coredns)
```
gyliu513: Can you please move this before Line 36?

/cc @xunpan

shlao: I have tried that: delete the empty line-35, but the output looks ugly.

gyliu513: I do not mean delete line 35, but update the document as follows:

Install ExternalDNS with CoreDNS as backend in your host cluster. You can follow the [tutorial](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md).

**Note**: You should replace parameter of `parameters: example.org` with `parameters: example.com` when [Installing CoreDNS](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md#installing-coredns)

To make it work for federation resources, you need to use below ExternalDNS deployment instead of the one in the tutorial.

**Note**: You should replace value of `ETCD_URLS` with your own etcd client service IP address.

shlao: Copy, do it now

shlao:
I am sorry that I rebase with the remote branch
I close this PR and submit a new one.

PR #555, opened 2019-01-22
Allow join command to run in idempotent mode

csrwng:
Introduces a flag to the JoinCluster function that specifies whether to run in idempotent mode. If true, creates that fail because the resource already exists will not fail the entire operation.

### DIFF
--- a/pkg/kubefed2/join.go
+++ b/pkg/kubefed2/join.go
@@ -19,6 +19,7 @@ package kubefed2
 import (
 	goerrors "errors"
 	"io"
+	"reflect"
 	"strings"
 	"time"
 
@@ -91,6 +92,7 @@ type joinFederationOptions struct {
 	addToRegistry   bool
 	limitedScope    bool
 	hostClusterName string
+	errorOnExisting bool
 }
 
 // Bind adds the join specific arguments to the flagset passed in as an
@@ -106,6 +108,8 @@ func (o *joinFederationOptions) Bind(flags *pflag.FlagSet) {
 		"Whether the federation namespace (configurable via --federation-namespace) will be the only target for federation.  If true, join will add a service account with access only to the federation namespace in the target cluster.")
 	flags.StringVar(&o.hostClusterName, "host-cluster-name", "",
 		"If set, overrides the use of host-cluster-context name in resource names created in the target cluster. This option must be used when the context name has characters invalid for kubernetes resources like \"/\" and \":\".")
+	flags.BoolVar(&o.errorOnExisting, "error-on-existing", false,
+		"Whether the join operation will throw an error if it encounters existing artifacts with the same name as those it's trying to create. If false, the join operation will update existing artifacts to match its own specification.")
 }
 
 // NewCmdJoin defines the `join` command that joins a cluster to a
@@ -187,13 +191,13 @@ func (j *joinFederation) Run(cmdOut io.Writer, config util.FedConfig) error {
 	}
 
 	return JoinCluster(hostConfig, clusterConfig, j.FederationNamespace, j.ClusterNamespace,
-		hostClusterName, j.ClusterName, j.secretName, j.addToRegistry, j.limitedScope, j.DryRun)
+		hostClusterName, j.ClusterName, j.secretName, j.addToRegistry, j.limitedScope, j.DryRun, j.errorOnExisting)
 }
 
 // JoinCluster performs all the necessary steps to join a cluster to the
 // federation provided the required set of parameters are passed in.
 func JoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, clusterNamespace,
-	hostClusterName, joiningClusterName, secretName string, addToRegistry, limitedScope, dryRun bool) error {
+	hostClusterName, joiningClusterName, secretName string, addToRegistry, limitedScope, dryRun, errorOnExisting bool) error {
 	hostClientset, err := util.HostClientset(hostConfig)
 	if err != nil {
 		glog.V(2).Infof("Failed to get host cluster clientset: %v", err)
@@ -213,13 +217,13 @@ func JoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, cl
 	}
 
 	glog.V(2).Infof("Performing preflight checks.")
-	err = performPreflightChecks(clusterClientset, joiningClusterName, hostClusterName, federationNamespace)
+	err = performPreflightChecks(clusterClientset, joiningClusterName, hostClusterName, federationNamespace, errorOnExisting)
 	if err != nil {
 		return err
 	}
 
 	if addToRegistry {
-		err = addToClusterRegistry(hostConfig, clusterNamespace, clusterConfig.Host, joiningClusterName, dryRun)
+		err = addToClusterRegistry(hostConfig, clusterNamespace, clusterConfig.Host, joiningClusterName, dryRun, errorOnExisting)
 		if err != nil {
 			return err
 		}
@@ -248,7 +252,7 @@ func JoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, cl
 
 	secret, err := createRBACSecret(hostClientset, clusterClientset,
 		federationNamespace, joiningClusterName, hostClusterName,
-		secretName, limitedScope, dryRun)
+		secretName, limitedScope, dryRun, errorOnExisting)
 	if err != nil {
 		glog.V(2).Infof("Could not create cluster credentials secret: %v", err)
 		return err
@@ -259,7 +263,7 @@ func JoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, cl
 	glog.V(2).Info("Creating federated cluster resource")
 
 	_, err = createFederatedCluster(fedClientset, joiningClusterName, secret.Name,
-		federationNamespace, dryRun)
+		federationNamespace, dryRun, errorOnExisting)
 	if err != nil {
 		glog.V(2).Infof("Failed to create federated cluster resource: %v", err)
 		return err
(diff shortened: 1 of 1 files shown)


pkg/kubefed2/join.go
```diff
@@ -374,13 +378,17 @@ func createFederatedCluster(fedClientset *fedclient.Clientset, joiningClusterNam
 		return fedCluster, nil
 	}
 
-	return fedClientset.CoreV1alpha1().FederatedClusters(federationNamespace).Create(fedCluster)
+	fedCluster, err := fedClientset.CoreV1alpha1().FederatedClusters(federationNamespace).Create(fedCluster)
+	if idempotent && errors.IsAlreadyExists(err) {
```
gyliu513: How about create a function named as `IsIdempotentMode` to replace `idempotent && errors.IsAlreadyExists(err)`?

/cc @xunpan

xunpan:
@csrwng 
Thanks for your PR. Could you please explain more about your use case?

For federation, kubefed2 join operation creates related resources automatically to make joined cluster work. When reverse the join operation, we unjoin the cluster. The related resource should be deleted automatically.

So my question is, what is the case that the related resources is not cleaned up but still need to installation forcely?

marun:
@csrwng It appears that you intend to vendor fedv2 and call the `Join` function.  I would caution you that federation is alpha software, and the join command is neither extensively tested nor provides any guarantee of interface stability.  You may want to consider exposing the idempotent flag in the `kubefed2` binary instead (and I would be supportive of that addition), since that is likely to be a better supported interface.

Also, is there a reason the`join` command shouldn't just default to idempotency?

csrwng:
Hi @xunpan and @marun thanks for taking a look.
Yes, as @marun implied, we aim to invoke the join function from a controller that is part of [OpenShift Hive](https://github.com/openshift/hive). The use case is that when you create a new cluster through Hive, we automatically join that cluster to the Hive host. If any step in the join operation fails, we need to be able to retry the operation until it succeeds. We didn't want to re-write the same join operation inside our controller and thought it'd be better to vendor the federation code to do so.

@marun to your question as to why join isn't idempotent by default. I think it should be. However, I didn't know whether changing the way it works would be acceptable to existing users of the federation CLI. Please let me know if it's ok to make it the default and I'll rework this PR.

As for the reverse function (unjoin), we need to be able to invoke it, but only the part that affects the host cluster. When a Hive cluster is deleted, we would like to remove the local FederatedCluster, secret, and the cluster registry Cluster. However, we really don't care to do anything in the target cluster as it may not be accessible anymore and all we want to do is to delete it. Not sure if the host part can be broken out from the target part inside kubefed2.

xunpan:
1. I think default behavior should be current one instead of `idempotent` one. 
    - if system exists some same name resources but not created by federation, it is easy to get error from command line.
    - our `unjoin` should clean all things and `join` should report error if anything is not cleaned.

2. I agree with marun. If we add this flag, we'd better export it in command line. 
However, if we enhance it, we need to make log message meaningful. E.g.
If a resource exists, `Infof` message should tell that the resource is not created by current command line operation but exists in the system already.

pkg/kubefed2/join.go
csrwng: Something like
```go
if ignoreIfIdempotent(idempotent, err) {
...
}
```
?
imho, I would think that it would obscure the fact that we're only ignoring the error when the item already exists. But I can make the change if you feel strongly about it.

csrwng:
Thank you for your review so far. I've added a commit to expose the flag in the CLI and log when items already exist.

marun:
@csrwng FYI the ci job is showing a go vet failure.

csrwng:
Thanks! fix pushed

marun:
@xunpan I think fedv2 should follow the example of `kubectl apply`.  `join` could create resources if they don't exist, and not complain if resources already exist in the desired form.  An error should only be reported if creation was not successful or an existing resource was not of the desired form (this latter characteristic would require implementation).

xunpan:
I think it is fine if we checking `an existing resource was not of the desired`. We should make sure all related resources in desired status. Or else, it is not good to reuse any resources that is unknown from `kubefed2`

csrwng:
@xunpan @marun should the join then behave like an apply and not only skip, but also update existing resources with their desired state?

marun:
@csrwng I think apply-like behavior should be the default.  I think it would make sense to retain the existing behavior with an optional flag that returns an error if any of the resources already exist.

csrwng:
Thank you @marun. I’ll submit an update for that.

csrwng:
@marun @xunpan I have updated the join operation to behave like apply, added a flag for errorOnExisting, and unit tests for the modified functions.

marun:
@csrwng I appreciate your initiative, but I'm concerned at the amount of faking going on in your unit tests.  fedv2 has a strategy of using real fixture in preference to maintaining fake fixture, and I think I'd prefer to have that strategy maintained in this instance.

csrwng:
@marun ack, will switch the tests
Is the implementation what you were thinking?

pkg/kubefed2/join.go
```diff
@@ -357,18 +356,27 @@ func registerCluster(crClientset *crclient.Clientset, clusterNamespace, host, jo
...
-	if idempotent && errors.IsAlreadyExists(err) {
-		glog.V(2).Infof("Cluster %s already exists in the cluster registry", joiningClusterName)
-		return cluster, nil
+	existingCluster, err := crClientset.ClusterregistryV1alpha1().Clusters(clusterNamespace).Get(cluster.Name, metav1.GetOptions{})
+	switch {
+	case err != nil && !errors.IsNotFound(err):
+		glog.V(2).Infof("Cannot retrieve cluster registry cluster %s due to %v", cluster.Name, err)
+		return nil, err
+	case err == nil && errorOnExisting:
+		return nil, fmt.Errorf("cluster registry cluster %s already exists", cluster.Name)
```
marun: We've recently switched to `github.com/pkg/errors`, and there is a ci check that will fail if `fmt.Errorf` is used:

https://github.com/kubernetes-sigs/federation-v2/pull/563

(No action required) A refactor of join/unjoin should likely replace the info logging with error wrapping.

cc: @gyliu513 (since we've discussed this before)

pkg/kubefed2/join.go
```diff
@@ -757,18 +867,31 @@ func populateSecretInHostCluster(clusterClientset, hostClientset client.Interfac
 		v1Secret.Name = secretName
 	}
 
-	v1SecretResult, err := hostClientset.CoreV1().Secrets(namespace).Create(&v1Secret)
-	if err != nil {
-		if idempotent && errors.IsAlreadyExists(err) {
-			glog.V(2).Infof("Secret %s already exists in host cluster", secretName)
-		} else {
-			glog.V(2).Infof("Could not create secret in host cluster: %v", err)
+	if secretName != "" {
```
marun: (No action required) Consider documenting this conditional e.g

```golang
secretMightExist := secretName != ""
if secretMightExist {
 ...
}
```

pkg/kubefed2/join.go
```diff
@@ -757,18 +867,31 @@ func populateSecretInHostCluster(clusterClientset, hostClientset client.Interfac
...
+		existingSecret, err := hostClientset.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
+		switch {
+		case err != nil && !errors.IsNotFound(err):
+			glog.V(2).Infof("Could not get secret %s in host cluster: %v", secretName, err)
 			return nil, err
+		case err == nil && errorOnExisting:
+			return nil, fmt.Errorf("secret %s already exists in host cluster", secretName)
+		case err == nil:
+			existingSecret.Data = v1Secret.Data
```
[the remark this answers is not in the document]
csrwng: @marun sorry I'm just getting back to this PR today. I still think that overwriting the secret is the right thing to do since the service account on the target cluster could have been (re-)created and you have a different token. To mitigate risk of overwriting stuff we could merge the Data map, but not sure that is a better result.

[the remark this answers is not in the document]
marun: To be clear, I'm worried because join can be configured with an arbitrary secret name, and a secret is a generic type rather than federation-specific.  It would be entirely possible, then, for join to unintentionally override an arbitrary secret.  Maybe an existing secret should result in an error unless the name is the default (i.e. not an arbitrary name provided by the user)?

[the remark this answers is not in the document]
csrwng: That sounds fine. I'll change it. I should hopefully submit an updated (hopefully cleaner) PR today.

csrwng:
@marun finally made the changes to the tests and squashed, ptal

csrwng:
@marun bump

pkg/kubefed2/kubefed2_suite_test.go
```diff
@@ -0,0 +1,91 @@
...
+	RunSpecsWithDefaultAndCustomReporters(t, "v1 Suite", []Reporter{test.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func() {
+	crds := inject.Injector.CRDs
+	crds = append(crds, clusterRegistryCRD())
+	testenv = &test.TestEnvironment{CRDs: crds}
+
+	var err error
+	config, err = testenv.Start()
```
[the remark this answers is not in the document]
csrwng: @marun if ok with you I’ll just remove the tests. They test non-public functions in this package and don’t make sense to move to an e2e type test. They are really meant as unit tests which is why using fake clients made the most sense to me initially. In e2e maybe we can increase coverage for the join operation in a different PR. Sound ok?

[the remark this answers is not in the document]
marun: I'm ok with that.  

I don't think a test that requires a ton of faking or a kube api server makes for a good unit test.  One is expensive to write and maintain, the other to run.  That's why we've purposely blurred the lines between integration and e2e in our e2e package, to reduce the cost of both maintaining and running tests that require api interaction.  If you wanted to expose those non-public functions to enable testing, I'd be fine with that too.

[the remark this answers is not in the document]
csrwng: Thanks, done.

pkg/kubefed2/join.go
```diff
@@ -392,7 +412,29 @@ func createFederatedCluster(fedClientset *fedclient.Clientset, joiningClusterNam
...
+		fedCluster, err = fedClientset.CoreV1alpha1().FederatedClusters(federationNamespace).Update(existingFedCluster)
+		if err != nil {
+			glog.V(2).Infof("Could not update federated cluster %s due to %v", fedCluster.Name, err)
+			return nil, err
+		}
+		return fedCluster, nil
+	default:
+		fedCluster, err = fedClientset.CoreV1alpha1().FederatedClusters(federationNamespace).Create(fedCluster)
+		if err != nil {
+			glog.V(2).Infof("Could not created federated cluster %s due to %v", fedCluster.Name, err)
```
marun: nit: s/created/create/

marun:
@xunpan @gyliu513 This lgtm.  Would one of you be able to review and merge once you're satisfied?

pkg/kubefed2/join.go
csrwng: Thanks, fixed.

marun:
@shashidharatd PTAL when you have a chance.

shashidharatd:
This LGTM has already undergone elaborate review and is as expected by other reviewers. Thanks @csrwng for doing this !
/lgtm

pkg/kubefed2/join.go
```diff
@@ -561,20 +626,54 @@ func createClusterRoleAndBinding(clientset client.Interface, saName, namespace,
...
+			}
+		}
+	default:
+		_, err = clientset.RbacV1().ClusterRoleBindings().Create(binding)
+		if err != nil {
+			glog.V(2).Infof("Could not create cluster role binding for service account: %s in joining cluster: %s due to: %v",
+				saName, clusterName, err)
+			return err
+		}
 	}
```
xunpan: To me, this can be simplified with:
```go
case err == nil:
    err = clientset.RbacV1().ClusterRoleBindings().Delete(existingBinding.Name, &metav1.DeleteOptions{})
    ...
    fallthrough
default:
    _, err = clientset.RbacV1().ClusterRoleBindings().Create(binding)
    ...
```

Reasons:
1. more branches in testing
2. `join` is not frequent operation. update vs. delete&update is not critical.

@marun

marun: @csrwng Is there a reason you wanted to update an existing binding if possible instead of always deleting?

csrwng: @marun my thinking was that the most common use case is that if there was an existing rolebinding that it would point to the same roleref, which means an update would be just ok. The delete/recreate in my mind is just to cover an edge case.

csrwng:
@marun btw, I think I need your approval again

PR #556, opened 2019-01-23
Kindly remind the user to replace the parameter of domain

shlao:
This PR fixes the potential problem that users maybe configure the wrong domain 'example.org', because we use the 'example.com' in this article.

Here is the extract from section: [Creating service resources](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/ingress-service-dns-with-coredns.md#creating-service-resources) 
$ cat <<EOF | kubectl create -f -
apiVersion: multiclusterdns.federation.k8s.io/v1alpha1
kind: Domain
... ...
domain: example.com
... ...

I add two spaces after [link](https://stackoverflow.com/questions/26626256/how-to-insert-a-line-break-br-in-markdown)


docs/ingress-service-dns-with-coredns.md
```diff
@@ -31,7 +31,8 @@ Install Federation-v2 with minikube in [User Guide](https://github.com/kubernete
 
 ## Installing ExternalDNS
 
-Install ExternalDNS with CoreDNS as backend in your host cluster. You can follow the [tutorial](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md).
+Install ExternalDNS with CoreDNS as backend in your host cluster. You can follow the [tutorial](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md).  
+**Note**: You should replace parameter of `parameters: example.org` with `parameters: example.com` when [Installing CoreDNS](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/coredns.md#installing-coredns)
```
marun: s/parameter of//

marun:
@gyliu513 `/ok-to-test` is a no-op for travis.  CI will run regardless of the message prow reports.

gyliu513:
@marun yes, but I found that without `ok-to-test`, the `needs-ok-to-test` label will block the merge of the PR.

marun:
@gyliu513 If that's true we should make sure that gets fixed in prow.  Prow shouldn't be blocking merges that are tested in travis.  At present, though, prow seems to be inactive on our repo, and I've reported this to infra.

docs/ingress-service-dns-with-coredns.md
shlao: Ye! brevity!

gyliu513:
@shlao can you squash the two commits to one?

shlao:
Ok, Let me try it.

shlao:
@gyliu513
1. I consult my colleague, he says that：The owner of this repo can squash committed logs or use the first log.
To confirm that, I create a test repo, new PR with another id. 

2. If I rebase with following steps, then the github will erases many commits
  1) git reset --soft HEAD~2
  2) git commit -m "Kindly remind the user to replace the parameter of domain"
  3) git status
     ... Your branch and 'origin/Rebase' have diverged
  4) git push -u origin  xxx:xxx -f

So it is better, I think, that the owner will do the right operation. Or can you tell me a new method?

gyliu513:
@shlao CI will merge this PR, so the PR owner need to squash the commit, please refer to https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit for how to squash.

shlao:
@gyliu513  If I have to force the push to my KindlyRemind. I think some commits will be lost.

gyliu513:
@shlao Yes, you can force push and we only need one commit for such PR.

shlao:
@gyliu513  Thanks

PR #560, opened 2019-01-23
Added additionalprintercolumn "Ready"  for the federatedcluster

sohankunkerkar:
I'm adding an additional column called `READY` in the federatedcluster type to print the readiness status for a cluster when you run `kubectl get federatedclusters -n federation-system` 
```
$ kubectl get federatedclusters -n federation-system
NAME         AGE           READY
cluster1     20m           true
cluster2     20m           true
```
We need to bump up the version for `kubebuilder` (probably v1.0.6/1.0.7) in order to support the`kubebuilder:printcolumn` annotation.

### DIFF
--- a/charts/federation-v2/charts/controllermanager/templates/crds.yaml
+++ b/charts/federation-v2/charts/controllermanager/templates/crds.yaml
@@ -74,6 +74,10 @@ metadata:
     controller-tools.k8s.io: "1.0"
   name: federatedclusters.core.federation.k8s.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.conditions[?(@.type=='Ready')].status
+    name: ready
+    type: string
   group: core.federation.k8s.io
   names:
     kind: FederatedCluster
--- a/pkg/apis/core/v1alpha1/federatedcluster_types.go
+++ b/pkg/apis/core/v1alpha1/federatedcluster_types.go
@@ -70,6 +70,7 @@ type FederatedClusterStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:resource:path=federatedclusters
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name=ready,type=string,JSONPath=.status.conditions[?(@.type=='Ready')].status
 type FederatedCluster struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`


marun:
afaik upgrading kubebuilder will not be a trivial endeavor.  It's evolution has proven a poor fit for fedv2, and it will probably be necessary to figure out how to use the code generation bits (now driven by makefile rather than the kubebuilder binary) separate from the prescriptively defined controller logic.

cc: @pmorie

marun:
@shashidharatd What are your thoughts on the effort required to update to newer versions of kubebuilder to support this PR?

marun:
@sohankunkerkar Would you be able to revisit this PR after #716 merges?

sohankunkerkar:
@marun sure thing!  I will try to complete this PR by tomorrow.

marun:
@sohankunkerkar Is this ready for review or is further work required?  I'm assuming the directive needs to be followed by a code generation step.

sohankunkerkar:
@marun ideally I need to add a single line of code to meet the requirements for this PR. Something like this:
`// +kubebuilder:printcolumn:name="ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status"`
However, I found a bug in the `additionalPrinterColumn` feature in kubebuilder. Let me explain you the exact scenario.
With the above annotation, it genrates the following CRD:
```yaml
  additionalPrinterColumns:
  - JSONPath: .status.conditions[?(@.type==\"Ready\")].status
    description: descr1
    format: date
    name: toy
    priority: 3
    type: string
```
and if you do `kubectl get federatedclusters -n federation-system`, it will return
```
NAME       
cluster1     
cluster2   
```
After debugging this issue for some time, I realize that the kubebuilder annotation should create a`JSONPath` in double quotes (which the present code is not handling it appropriately) like
`".status.conditions[?(@.type==\"Ready\")].status"` in order to print what we need at this moment:
```
$ kubectl get federatedclusters -n federation-system
NAME         AGE           READY
cluster1     20m           true
cluster2     20m           true
```
I'm still figuring out the best way to fix this issue.

marun:
@sohankunkerkar Ah, ok.  It sounds like a further update to kubebuilder is required then?  It should be easy to vendor an updated version when it becomes available, hopefully before beta.

sohankunkerkar:
@marun ready for review. There's no need to escape double quotes for `Ready` field so that was a catch here. You can't review generated CRD's since they are ignored due to .gitignore file.

marun:
Thank you your patience and diligent coordination on the kubebuilder side of things!

/approve

I'll leave it to @xunpan or @shashidharatd to merge.

shashidharatd:
/lgtm
Thanks @sohankunkerkar

PR #562, opened 2019-01-25
fix federation use default value to cover member cluster service node port when node port is auto-allocate by controller

Dingshujie:


marun:
/approved
/lgtm

Whoops.  We really need to have code like this unit tested.

cc: @shashidharatd

Issue #381, opened 2018-11-01
Switch to github.com/pkg/errors


poothia:
There is a general consensus on moving towards `github.com/pkg/errors`... and it's being said that `fmt.errorf` is confusing and has an overhead to it. 

reference:
https://github.com/kubernetes/kubeadm/issues/1183#issue-372862070
https://github.com/kubernetes/kubernetes/pull/70271#issue-225967737 

Other references:
https://github.com/golang/lint/issues/284#issue-218626145
https://hyperledger-fabric.readthedocs.io/en/v1.1.0-alpha/error-handling.html
https://github.com/hashicorp/packer/issues/4097#issue-186668507

poothia:
if there is a consensus on this, i would like to work on it.

marun:
I would be happy to see us transition to the `errors` package.  I think the following article makes a more compelling case: 

https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

There are ~100 instances of embedding errors as strings in the codebase today, I encourage you to propose a PR that adds the dependency on the `errors` package and updates the code to use the `errors.Wrap` method.  Please submit a switch for creation of non-formatted errors to `errors.New` in a separate PR, as I see that work being orthogonal to wrapping errors.

poothia:
@marun  thanks for the reply and the pointers, will keep those in mind.

marun:
@poothia Were you going to follow up on this?

PR #558, opened 2019-01-23
Switch to github.com/pkg/errors

poothia:
closes #381

PR #563, opened 2019-01-25
vendoring change for github.com/pkg/errors

poothia:
part of #381


marun:
Please include this in #558 as its own commit, not a separate PR.

PR #527, opened 2018-12-20
Use a hash of the template field as the template version

marun:
Previously the resourceVersion of the template resource was used, but resourceVersion changes when status changes.  Switching to a hash paves the way to store propagated version in the status of a template resource.

Part 1 of #479

PR #564, opened 2019-01-25
Unify template, placement and overrides for a federated type into a single resource

marun:
This PR proposes that the components of a logical federated type (template, placement, overrides) be included in a single resource rather than separately.


marun:
Tests are now at parity with master.  Docs and examples are next.

marun:
I've updated the docs and example yaml in the latest commit.

I'm thinking the group name should change from `primitives.federation.k8s.io` since a federated type now defines fields for what used to be separate resources.  Maybe go back to `generated.federation.k8s.io`?

marun:
I've removed the `approved` label to avoid accidental merge of this non-trivial PR.

pkg/controller/sync/placement.go
```diff
@@ -0,0 +1,113 @@
...
+// placement of the federated namespace containing the resource.
+//
+// If federation is limited to a single namespace, placement is
+// determined as the intersection of resource and namespace placement
+// if namespace placement exists.  If namespace placement does not
+// exist, resource placement will be used verbatim.  This is possible
+// because the single namespace by definition must exist on member
+// clusters, so namespace placement becomes a mechanism for limiting
+// rather than allowing propagation.
+func computeNamespacedPlacement(resource, fedNamespace *unstructured.Unstructured, clusters []*fedv1a1.FederatedCluster, limitedScope bool) (selectedClusters, unselectedClusters []string, err error) {
```
shashidharatd: (suggestion) It would be good to uniformly name the parameters across the functions. Like the `resource` has become `obj` in this function and henceforth.

pkg/controller/util/placement.go
```diff
@@ -24,11 +24,17 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
-type GenericPlacementSpec struct {
+type GenericPlacementFields struct {
 	ClusterNames    []string              `json:"clusterNames,omitempty"`
 	ClusterSelector *metav1.LabelSelector `json:"clusterSelector"`
```
shashidharatd: (suggestion) Shouldn't we use `omitempty` json tag for this field as well?

pkg/schedulingtypes/plugin.go
```diff
@@ -161,59 +126,45 @@ func (p *Plugin) TemplateExists(key string) bool {
...
-}
+	isDirty := false
 
-func (p *Plugin) ReconcileOverride(qualifiedName util.QualifiedName, result map[string]int64) error {
-	override, err := p.overrideClient.Resources(qualifiedName.Namespace).Get(qualifiedName.Name, metav1.GetOptions{})
+	newClusterNames := []string{}
+	for name := range result {
+		newClusterNames = append(newClusterNames, name)
+	}
+	clusterNames, err := util.GetClusterNames(fedObject)
```
shashidharatd: `util.GetClusterNames` seems only work if placement contains slice of clusterNames and does not work with label based clusterSelector. Better to add a TODO, if we could not solve it in this pr, which could be taken up later.

test/e2e/scheduling.go
```diff
@@ -107,7 +107,7 @@ var _ = Describe("ReplicaSchedulingPreferences", func() {
 		cluster2      int32
 		noPreferences bool
 	}{
-		"replicas spread equally in clusters with no explicit per cluster preferences": {
+		"maru replicas spread equally in clusters with no explicit per cluster preferences": {
```
shashidharatd: Unintentional change :)

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -31,31 +31,27 @@ type FederatedTypeConfigSpec struct {
...
-	// member clusters the resource should be federated to. If not provided, the
-	// group and version will default to those provided for the template
-	// resource.
-	Placement APIResource `json:"placement"`
-	// Configuration for the override type that holds information about how the
-	// resource should be changed from the template when in certain member
-	// clusters. If not provided, the group and version will default to those
-	// provided for the template resource.
-	Override APIResource `json:"override"`
+	// Configuration for the federated type that defines how the target
```
irfanurrehman: [Action not necessary]: This came across a little confusing in the first read. To me, it read as `Configuration .... that defines how target type should be propagated...`. Consider rephrasing.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -31,31 +31,27 @@ type FederatedTypeConfigSpec struct {
...
+	// type should be propagated to multiple clusters.
+	FederatedType APIResource `json:"federatedType"`
```
irfanurrehman: Is `Federated` prefix necessary here? This is already inside `FederatedTypeConfigSpec`.
My suggestion would be to try and reduce our usage of `Federated` prefix where ever logically possible. The code base already has an overdose of the same :).

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -47,9 +47,11 @@ type Controller struct {
 	// Arguments to use when starting new controllers
 	controllerConfig *util.ControllerConfig
 
-	// The namespace placement resource which will be needed to start
-	// sync controllers when running namespaced.
-	namespacePlacement *metav1.APIResource
+	// The federated namespace api resource will be needed to start
+	// sync controllers for namespaced federated types.  Federated
+	// namespace placement is used in determining placement of
```
irfanurrehman: [Action not necessary] I believe this means the `placement` field in the actual API resource. Consider rephrasing.

pkg/controller/sync/accessor.go
```diff
@@ -216,62 +219,71 @@ func (a *resourceAccessor) FederatedResource(eventSource util.QualifiedName) (Fe
...
+
+	resource, err := util.ObjFromCache(a.federatedStore, kind, key)
+	if err != nil {
+		return nil, err
+	}
+	if resource == nil {
+		// The event source may be a federated resource that was
+		// deleted, but is more likely a non-federated resource in the
+		// target cluster.
+		glog.V(7).Infof("%s %q was not found which indicates that the %s is not federated",
```
irfanurrehman: [Query]: I doubt that this would happen. The `federatedController` is started only when this type is federated eventually resulting in objects in `federatedStore`. The resulting action if this type in no longer federated is stopping the sync controller, which means that this `reconcile` loop using this `accessor` is already stopped. 
Even if the sync controller `reconcile` loop waits to clear all the residual events, the typeconfig controller wouldn't have completed the  `disable/delete` (or `unfederate`) operation.
It does not harm to keep this log info any which ways; it would be great if you can correct/clear my understanding. Is this meant only for namespaces?

marun: For a given federated type, reconciliation is triggered on changes for all target resources in member clusters.  It is very likely that some of those target resources are local to a member cluster and have no equivalent federated resource in the host cluster (e.g. the deployments managing system components in the `kube-system` namespace).

pkg/schedulingtypes/plugin.go
marun: `GetClusterNames` considers both `clusterNames` and `clusterSelector`

pkg/controller/sync/placement.go
marun: Done

pkg/controller/util/placement.go
marun: Done

test/e2e/scheduling.go
marun: Done

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
marun: Updated, PTAL

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
marun: I agree that 'federated' is an overused prefix, but I don't think `Type` is sufficiently descriptive.  I'm happy to consider an alternate name, `FederatedType` is just a sane default.

pkg/controller/federatedtypeconfig/controller.go
marun: Updated, PTAL

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
irfanurrehman: Lets continue with `FederatedType`. Thanks.

pkg/controller/sync/accessor.go
irfanurrehman: Thanks for the clarification.

xunpan:
The code looks good to me.

pkg/schedulingtypes/plugin.go
irfanurrehman: @shashidharatd, if an RSP is present, it will always override the `placement.clusternames` to the names which result from scheduling the per cluster user preferences specified in the RSP. It will write new `placement.clusternames` if not already present on the object. `GetClusterNames()` is used here to get and determine if a write() on the object is needed or not, and avoid write() if the object already has the same `clusternames`.

pkg/controller/sync/resource.go
```diff
@@ -80,27 +77,21 @@ func (r *federatedResource) TargetName() util.QualifiedName {
...
+	return r.federatedResource
 }
 
 func (r *federatedResource) TemplateVersion() (string, error) {
-	return GetTemplateHash(r.template)
+	obj := r.federatedResource
+	if r.targetIsNamespace {
+		obj = r.namespace
+	}
+	return GetTemplateHash(obj.Object, r.targetIsNamespace)
```
irfanurrehman: [Query]: Apologies for asking as this question is not from the change in this PR. I probably missed it when version manager was introduced. Why is a hash calculated and used in version manager for federated resources? Do CRD resources not get their own ResourceVersion` values auto updated on changes? Or this is to be able to determine part of the whole object (say `spec.overrides` now)?

pkg/controller/sync/placement.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	clusterNames := getClusterNames(clusters)
+	clusterSet := sets.NewString(clusterNames...)
+	selectedSet := sets.NewString(selectedNames...)
+	return clusterSet.Intersection(selectedSet).List(), clusterSet.Difference(selectedSet).List(), nil
+}
+
+func selectedClusterNames(rawPlacement *unstructured.Unstructured, clusters []*fedv1a1.FederatedCluster) ([]string, error) {
```
irfanurrehman: s/rawPlacement/rawResource/

pkg/controller/util/placement.go
```diff
@@ -51,21 +57,21 @@ func GetPlacementDirective(rawPlacement *unstructured.Unstructured) (*PlacementD
 	if err != nil {
 		return nil, err
 	}
-	selector, err := metav1.LabelSelectorAsSelector(placement.Spec.ClusterSelector)
+	selector, err := metav1.LabelSelectorAsSelector(placement.Spec.Placement.ClusterSelector)
```
irfanurrehman: s/placement.spec/resource.spec/  as it becomes confusing percolating from comment on `line 90` in `placement.go`

docs/userguide.md
```diff
@@ -592,8 +593,8 @@ behaviour is unacceptable.
 
 The RSP can be considered as more user friendly mechanism to distribute the
 replicas, where the inputs needed from the user at federated control plane are
-reduced. The user only needs to create the RSP resource and the mapping template
-resource, to distribute the replicas. It can also be considered as a more
+reduced. The user only needs to create the RSP resource and associated federated
+resource to distribute the replicas. It can also be considered as a more
```
irfanurrehman: s/create the RSP resource and associated federated resource to distribute the replicas/
create the RSP resource and associated federated resource with only `spec.template` field filled in to distribute the replicas./

irfanurrehman:
Thanks @marun for this piece of work. As usual, this work is quite neat and only few nits found.

pkg/controller/sync/resource.go
marun: Resource version changes when anything in the resource changes, so it's not appropriate for determining the identity of a field (e.g. template, overrides).  The switch to using a hash for the template was made in https://github.com/kubernetes-sigs/federation-v2/pull/527, and this PR switches to using a hash for overrides too since it is no longer in its own resource.

pkg/controller/sync/placement.go
marun: Done

pkg/controller/util/placement.go
marun: I've changed `rawResource` to `resource` in this method, but I think `placement` is the appropriate name here since it is an instance of `GenericPlacement`.  

Also, I envision a future iteration that unmarshals an unstructured federated type to an internal api struct rather than marshaling separately for overrides and placement, which should simplify things.

pkg/controller/sync/placement.go
marun: I've changed `rawPlacement` to `resource` for consistency with the other functions in this file.

docs/userguide.md
marun: Done

marun:
@shashidharatd @irfanurrehman Thoughts on a new group name to replace `primitives.federation.k8s.io`?

pkg/controller/sync/resource.go
irfanurrehman: Got it, thanks!

pkg/controller/util/placement.go
irfanurrehman: Ok

pkg/controller/sync/placement.go
irfanurrehman: > @shashidharatd @irfanurrehman Thoughts on a new group name to replace `primitives.federation.k8s.io`?

Reasons for thinking on this change? `primitives` still sounds quite ok to me. `core`, `base`, `primary` are the other options that come to my mind.

irfanurrehman:
> @shashidharatd @irfanurrehman Thoughts on a new group name to replace `primitives.federation.k8s.io`?

Reasons for thinking of this change? `primitives` seems fine to me for the unified API also. `core`, `base` and `primary` are some names which come to my mind, if the change is necessary.

marun:
@irfanurrehman Given that 'primitive' is the term we have used up till now to describe 'template', 'placement', and 'override' resources, and those elements are just part of a federated type, I would like to de-emphasize the term.  Retaining it in the group name would seem to invite confusion, then.

I'm going to change it to 'generated', since I think that is more in keeping with what the group is used for - generated API types - than any of `core` (already used for non-generated types) or `base` and `primary`.

marun:
@shashidharatd @irfanurrehman Do you have further comments/concerns, or is this ready for merge?

marun:
Rebased

pmorie:
Overall LGTM.

pkg/controller/sync/placement_test.go
```diff
@@ -80,10 +80,10 @@ func TestSelectedClusterNames(t *testing.T) {
 				},
 			}
 			if testCase.clusterNames != nil {
```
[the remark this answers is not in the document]
marun: Correct

pkg/controller/status/controller.go
```diff
@@ -196,8 +196,8 @@ func (s *FederationStatusController) isSynced() bool {
 		glog.V(2).Infof("Cluster list not synced")
 		return false
 	}
-	if !s.templateController.HasSynced() {
-		glog.V(2).Infof("Template not synced")
+	if !s.federatedController.HasSynced() {
+		glog.V(2).Infof("Federated type not synced")
```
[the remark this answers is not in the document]
marun: Agree.  That's already the case for the sync controller.

docs/userguide.md
```diff
@@ -368,10 +365,11 @@ kubectl label federatedclusters -n federation-system cluster1 foo=bar
...
-used in determining the clusters that a federated resource should be propagated to.
+The following sections detail how `spec.placement.clusterNames` and
+`spec.placement.clusterSelector` are used in determining the clusters that a federated
+resource should be propagated to.
 
-##### Neither `spec.clusterNames` nor `spec.clusterSelector` is provided
+##### Neither `spec.placement.clusterNames` nor `spec.placement.clusterSelector` is provided
 
 ```yaml
 spec: {}
```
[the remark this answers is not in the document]
marun: That is implicit, but I agree that it makes sense to be explicit here.  Done.

docs/userguide.md
```diff
@@ -256,11 +253,11 @@ kubectl apply -R -f example/sample1
 **NOTE:** If you get the following error while creating a test resource i.e.
 
 ```
-unable to recognize "example/sample1/Federated<type>-placement.yaml": no matches for kind "Federated<type>Placement" in version "primitives.federation.k8s.io/v1alpha1",
+unable to recognize "example/sample1/federated<type>.yaml": no matches for kind "Federated<type>" in version "primitives.federation.k8s.io/v1alpha1",
```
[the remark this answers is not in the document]
marun: I think this comment is probably more appropriate as a general troubleshooting tip, but I think that change can be made as part of docs refactoring leading up to beta.

pkg/schedulingtypes/plugin.go
shashidharatd: ok. SGTM for now

shashidharatd:
LGTM too. Thanks @marun

irfanurrehman:
> @irfanurrehman Given that 'primitive' is the term we have used up till now to describe 'template', 'placement', and 'override' resources, and those elements are just part of a federated type, I would like to de-emphasize the term. Retaining it in the group name would seem to invite confusion, then.
> 
> I'm going to change it to 'generated', since I think that is more in keeping with what the group is used for - generated API types - than any of `core` (already used for non-generated types) or `base` and `primary`.
> 
> Edit: I'm going to submit this change in a follow-up PR. I don't think it should block merge of this one.

`generated` seems quite fine to me.



> @shashidharatd @irfanurrehman Do you have further comments/concerns, or is this ready for merge?

This is LGTM for me; @pmorie if your review is complete, please merge.

gyliu513:
@marun can you please share some motivation for this PR? What is the advantage of using a single file for all of the resources?

marun:
@gyliu513 Having a single resource represent a federated type is a huge win when it comes to comprehending the federation API.  Most of the 'gotchas' of the API were related to the split into separate resources.  With unification:

- It will no longer be necessary to access multiple resources to determine the configuration for a given federated type.  
 - It won't be necessary to consider the order in which placement, override and template resources are created to ensure the desired resource form appears in member clusters.
 - Resource cleanup is simplified - only a single resource to delete

This also paves the way for ensuring a federated resource has enough status to aid in effective troubleshooting, a key target of the first beta release.

gyliu513:
Thanks @marun for the detailed explanation, +1 this is really a huge win.

FYI @xunpan

shashidharatd:
/lgtm
Thanks @marun

PR #568, opened 2019-01-30
Wait for test namespace deletion for unmanaged deployments

### LINKED ISSUE #567: Teardown timeout causes CI failures
#545 switched finalization for federated namespaces to the placement resources instead of the namespaces.  Since it merged, CI jobs have started timing out when attempting to teardown a test federation.  The likely cause is a race between namespace deletion and controller shutdown, such that if controllers are shut down before federated resources are marked for deletion and finalized, the containing namespaces will be stuck in `Terminating` state indefinitely and block teardown.  The fix will be to wait for test namespace removal in `AfterEach` to maximize the chances of finalization before controllers are shut down.

/kind bug

CC

marun:
Waiting for test namespace deletion before controller shutdown ensures that finalizers will be removed so that deletion of resources (and the containing namespace) is possible.  This increases test execution time considerably, though, so the (potentially unsafe) flag `--wait-for-finalization` has been added and can be set to false for disposeable clusters.

Fixes #567


marun:
I seem to have broken the tests with the reversal of fixture teardown ordering.

marun:
I've backed out the attempted optimization of removal of federated namespaces for now.

@shashidharatd I've based the PR chain for reunification on this to avoid flakes.

PR #571, opened 2019-02-05
Update docs and comments for 'federate enable/disable'

irfanurrehman:
As a follow up for #534 

cc @marun @gyliu513 @xunpan

PR #572, opened 2019-02-05
s/FederateDirective/EnableTypeDirective

irfanurrehman:
As a precursor to use keyword `federate` for resources and namespaces only.
Also `TypeDirective` sounds more suitable in use for `types`.

Based on #571. Only the last commit is reviewable.

cc @marun @gyliu513 @xunpan

### DIFF
--- a/config/enabletypedirectives/clusterroles.rbac.authorization.k8s.io.yaml
+++ b/config/enabletypedirectives/clusterroles.rbac.authorization.k8s.io.yaml
@@ -1,4 +1,4 @@
 apiVersion: core.federation.k8s.io/v1alpha1
-kind: FederateDirective
+kind: EnableTypeDirective
 metadata:
   name: clusterroles.rbac.authorization.k8s.io
--- a/config/enabletypedirectives/configmaps.yaml
+++ b/config/enabletypedirectives/configmaps.yaml
@@ -1,4 +1,4 @@
 apiVersion: core.federation.k8s.io/v1alpha1
-kind: FederateDirective
+kind: EnableTypeDirective
 metadata:
   name: configmaps
--- a/config/enabletypedirectives/deployments.apps.yaml
+++ b/config/enabletypedirectives/deployments.apps.yaml
@@ -1,6 +1,6 @@
 apiVersion: core.federation.k8s.io/v1alpha1
-kind: FederateDirective
+kind: EnableTypeDirective
 metadata:
-  name: serviceaccounts
+  name: deployments.apps
 spec:
   comparisonField: Generation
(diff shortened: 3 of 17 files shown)


marun:
@irfanurrehman Please rebase and I'll review!

irfanurrehman:
@marun rebased!

marun:
I'm not sure the name `TypeDirective` is descriptive enough.  Perhaps `EnableTypeDirective` or `EnableDirective`, in keeping with the new name of the command?

irfanurrehman:
I think the name `TypeDirective` is quite apt. The `Directive` [instruction(s)] which encompasses the intent and accompanying information needed to enable/disable (not just enable) a `Type`. Its a little deviant from this that we do not need the same for disable as this information persists in `FederatedTypeConfig` API. Perhaps something to revisit when `TypeDirective` itself becomes an API.
Having said that, I am fine with changing to a better name, if there is any; `EnableTypeDirective` or `EnableDirective` doesn't sound right somehow.. !

marun:
@irfanurrehman I don't think the directives in question are good for more than enabling a type.  The ftc is what is required to disable, and the fact that a directive may enable the ftc to be discovered is incidental.

irfanurrehman:
@marun updated to `EnableTypeDirective`. PTAL, thanks!

marun:
Broad strokes of the PR look fine to me, but CI is legitimately failing.

Also, is `pkg/kubfed2/federate` going to be renamed at some point?

poothia:
@marun looks like outage at travis itself... like this https://github.com/travis-ci/travis-ci/issues/9327 one.

marun:
@poothia Not sure what you mean, the build failure looks legit.

poothia:
@marun sorry, my bad... had some issues with my firewall... it was blocking my travis-CI.

gyliu513:
> Also, is `pkg/kubfed2/federate` going to be renamed at some point?

@marun do you mean `pkg/kubfed2/federate/directive.go`?

marun:
@gyliu513 I meant the 'federate' part of the path, since the contents no longer reflect that name.

irfanurrehman:
> Also, is `pkg/kubfed2/federate` going to be renamed at some point?

@marun, I would be segregating and replacing the `federate` pkg as `type` and `resource` pkgs in the next PR. Another thought that did come across to me was keeping all files into the base `kubefed2` pkg. Let me know if you have better suggestions?

marun:
@irfanurrehman I think the files for enable/disable belong in their own package.

marun:
Thank you!

/lgtm

PR #576, opened 2019-02-07
Fix help text for kubefed2 enable

pmorie:
Mistakenly still said "federate enable" in the help text rather than "enable"


marun:
I think this can be closed, as the change was already made by #571.

Issue #577, opened 2019-02-07
kubefed2 join does not work correctly with some kubectl context names for the hosting cluster


pmorie:
**What happened**:

I used `kubefed2 join` with a cluster whose kubectl context contained a slash (`default/pmorie-dev-api-devcluster-openshift-com:6443/kube:admin`), which is a perfectly valid kubectl context name. However, the name of the kubectl context for the host cluster is used to construct the names of resources, which will yield invalid resource names in cases like this where the name of the kubectl context is a valid context name but not a valid resource name. We should construct the names of these resources using values that will always yield a valid resource name.

Looking at possibilities for which value to use, I thought about using the name of the cluster in the kubeconfig (instead of the name of the context), but it appears there are basically no validation requirements for kubeconfig cluster names: https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/validation.go#L167

Tactically, perhaps we could consider using `-federation` when we construct resource names.

**What you expected to happen**:

I would expect the join operator to proceed even if my context name does not construct a valid k8s resource name.

**How to reproduce it (as minimally and precisely as possible)**:

Use `kubefed2 join` with a host cluster context that contains a slash.

/kind bug

shashidharatd:
@pmorie, can this be closed?

marun:
Fixed by #578

PR #578, opened 2019-02-07
Add --host-cluster-name to kubefed2 join to override use of kubeconfig context name for resources

pmorie:
Addresses #577

### DIFF
--- a/pkg/kubefed2/join.go
+++ b/pkg/kubefed2/join.go
@@ -17,7 +17,9 @@ limitations under the License.
 package kubefed2
 
 import (
+	goerrors "errors"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -84,10 +86,11 @@ type joinFederation struct {
 }
 
 type joinFederationOptions struct {
-	clusterContext string
-	secretName     string
-	addToRegistry  bool
-	limitedScope   bool
+	clusterContext  string
+	secretName      string
+	addToRegistry   bool
+	limitedScope    bool
+	hostClusterName string
 }
 
 // Bind adds the join specific arguments to the flagset passed in as an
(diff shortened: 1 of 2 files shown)


marun:
This LGTM

@xunpan Can you please take a look?

pkg/kubefed2/unjoin.go
```diff
@@ -110,6 +114,10 @@ func (j *unjoinFederation) Complete(args []string) error {
 		j.clusterContext = j.ClusterName
 	}
 
+	if j.hostClusterName == "" && strings.ContainsAny(j.HostClusterContext, ":/") {
+		glog.Fatal("host-cluster-name must be set if the name of the host cluster context contains one of \":\" or \"/\"")
```
xunpan: It is preferred to return an error instead of directly `glog.Fatal()`

pmorie: Valid comments, will incorporate.

pmorie:
@xunpan comments addressed, PTAL

pkg/kubefed2/join.go
```diff
@@ -145,6 +149,10 @@ func (j *joinFederation) Complete(args []string) error {
 		j.clusterContext = j.ClusterName
 	}
 
+	if j.hostClusterName == "" && strings.ContainsAny(j.HostClusterContext, ":/") {
+		glog.Fatal("host-cluster-name must be set if the name of the host cluster context contains one of \":\" or \"/\"")
+	}
+
```
xunpan: `join` should have similar logic as `unjoin` for messages and `hostClusterName` checking logic.
```
	if j.hostClusterName == "" && strings.ContainsAny(j.HostClusterContext, ":/") {
		return goerrors.New("host-cluster-name must be set if the name of the host cluster context contains one of \":\" or \"/\"")
	}

	if j.hostClusterName != "" && strings.ContainsAny(j.hostClusterName, ":/") {
		return goerrors.New("host-cluster-name may not contain \"/\" or \":\"")
	}
```

pmorie:
@xunpan addressed, PTAL

PR #586, opened 2019-02-11
Determine the comparison field at runtime rather than configuring

### LINKED ISSUE #480: Remove the need to specify ComparisonField for federated types
A `FederatedTypeConfig` requires that a value for `comparisonField` be provided.  This value is used to determine what field of a target resource determines equality when deciding whether an update is required.  The 2 valid values are:

 - `ResourceVersion`
    - Used when a resource does not have a controller that records status on it, since `resourceVersion` is incremented when a change is made to any fields of the resource
 - `Generation`
    - Used when a resource has a controller managing it and writing status.  `generation` is only incremented when a change to the spec is observed.

For resources that are not managed by a controller, `generation` may never be populated (this includes `namespace`, `secret`, `configmap`, and likely others), so it is not possible to choose one or the other for all types. 

The recent implementation of `kubefed2 federate` has revealed that requiring the user to choose a comparison field is a burden.  It would be preferable that a user could federate any type without having to know how comparison should be performed.  This could be achieved by switching from the use of a version sourced from a field populated by kube to a version computed by hashing the contents of the spec of a target resource.   Similarly to how generation-based comparison is implemented, equality could be determined by comparing the metadata and content-based version of a target resource with the metadata of the spec embedded in the template and the stored version of the target resource.

marun:
As per a suggestion from @irfanurrehman, the field used in deriving the version of a cluster object is now determined at runtime rather than having to be configured at the type level.  This allows use of the `generation` field where an object is expected to be modified regularly by a controller, and fallback to the `resourceVersion` field if the generation field is not being set.  Removing the requirement to configure `comparisonField` when enabling federation of a type will remove the possibility of a user inadvertently choosing the wrong field and the potential for variance in type configuration across deployments of federation. 

Fixes #480

### DIFF
--- a/pkg/controller/util/propagatedversion.go
+++ b/pkg/controller/util/propagatedversion.go
@@ -17,64 +17,45 @@ limitations under the License.
 package util
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
-	"strconv"
+	"strings"
 
-	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type ComparisonHelper interface {
-	GetVersion(objectMeta metav1.Object) string
-	Equivalent(objMeta1, objectMeta2 metav1.Object) bool
-}
+const (
+	generationPrefix      = "gen:"
+	resourceVersionPrefix = "rv:"
+)
 
-// NewComparisonHelper instantiates and returns a Resource or Generation Helper
-// struct that implements the ComparisonHelper interface based on the version
-// comparison type passed in.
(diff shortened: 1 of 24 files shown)


marun:
@gyliu513 I'm running `hack/sync-up-helm-chart.sh` locally and no changes are detected, but this PR should result in changes and CI is detecting this.  Any idea why the script wouldn't work?

marun:
Holding for review

pkg/controller/util/propagatedversion.go
```diff
@@ -17,64 +17,45 @@ limitations under the License.
...
-func NewComparisonHelper(comparisonField common.VersionComparisonField) (ComparisonHelper, error) {
-	switch comparisonField {
-	case common.ResourceVersionField:
-		return &ResourceHelper{}, nil
-	case common.GenerationField:
-		return &GenerationHelper{}, nil
-	default:
-		return nil, errors.Errorf("Unrecognized version comparison field %v", comparisonField)
+// ObjectVersion retrieves the field type-prefixed value used for
+// determining currency of the given cluster object.
```
shashidharatd: `determining currency`?, is this some term. or is it currency == version?

marun: 'determining whether the object is current, up-to-date'

PR #587, opened 2019-02-11
Enable configurable replicas retention by the sync controller

### LINKED ISSUE #570: A way to prevent controller from overwriting certain fields
I would like to have a way to define a list of fields for resource that if they're updated on some cluster won't trigger federation controller to sync and overwrite them.

**Why is this needed**:
I want to specify, which clusters contain my deployment and I want local versions of HPA to scale them based on local usage.

I have federated Deployment an federated HPA (enabled) defined
https://gist.github.com/lyxsus/bca5bb7ab95ebd014b60186fa3882357

I expected deployment to be scaled up to 2, instead federation controller and HPA started to race for "replicas" field, oscillating number of pods between 1 and ~2:
https://gist.github.com/lyxsus/7d705ef4ba726897277b6429f13fc822

Initially, I thought there must be some property in FederatedDeployment like `minReplicas` and `maxReplicas` for this use case, but that isn't the case and talks on Slack suggested that it's more general problem and a case where Service NodePorts and ExternalIP fields were undesirably overwritten in a same way.

/kind feature

marun:
`kubefed2 enable` is updated to add a `retainReplicas` field for target types that expose a `spec.replicas` field.  If set on a federated resource, the sync controller will attempt to retain a replicas field from the target resource when performing updates to the target.  This is intended to avoid thrashing the scheduler when in-cluster HPA is targeting a scalable resource.

The granularity of this retention is all clusters or no-clusters, with the expectation that where HPA is intended to be applied selectively, a user will define 2 resources (one with `retainReplicas: true` and one with `retainReplicas: false`) along with partitioning placement.

Fixes #570 

TODO

- [x] Add testing
- [x] Add documentation

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -33,6 +33,8 @@
     - [Automated Deployment](#automated-deployment-1)
     - [Joining Clusters](#joining-clusters)
     - [Deployment Cleanup](#deployment-cleanup-1)
+  - [Local Value Retention](#local-value-retention)
+    - [Scalable Resources](#scalable-resources)
   - [Higher order behaviour](#higher-order-behaviour)
     - [Multi-Cluster Ingress DNS](#multi-cluster-ingress-dns)
     - [Multi-Cluster Service DNS](#multi-cluster-service-dns)
@@ -527,6 +529,44 @@ employed by deployment:
 ```bash
 NAMESPACED=y FEDERATION_NAMESPACE=<namespace> ./scripts/delete-federation.sh
 ```
+
+## Local Value Retention
+
+In most cases, the federation sync controller will overwrite any
+changes made to resources it manages in member clusters.  The
+exceptions appear in the following table.  Where retention is
+conditional, an explanation will be provided in a subsequent section.
+
+| Resource Type  | Fields                    | Retention   | Requirement                                                                  |
+|----------------|---------------------------|--------------------------------------------------------------------------------------------|
+| All            | metadata.resourceVersion  | Always      | Updates require the most recent resourceVersion for concurrency control.     |
+| Scalable       | spec.replicas             | Conditional | The HPA controller may be managing the replica count of a scalable resource. |
+| Service        | spec.clusterIP,spec.ports | Always      | A controller may be managing these fields.                                   |
+| ServiceAccount | secrets                   | Conditional | A controller may be managing this field.                                     |
+
+### Scalable
+
+For scalable resources (those that have a scale subtype
+e.g. `ReplicaSet` and `Deployment`), retention of the `spec.replicas`
+field is controlled by the `retainReplicas` boolean field of the
+federated resource.  `retainReplicas` defaults to `false`, and should
+be set to `true` only if the resource will be managed by HPA in member
+clusters.
+
+Retention of the replicas field is possible for all
+clusters or no clusters.  If a resource will be managed by HPA in some
+clusters but not others, it will be necessary to create a separate
+federated resource for each retention strategy (i.e. one with
+`retainReplicas: true` and one with `retainReplicas: false`).
+
+### ServiceAccount
+
+A populated `secrets` field of a `ServiceAccount` resource managed by
+federation will be retained if the managing federated resource does
+not specify a value for the field.  This avoids the possibility of the
+sync controller attempting to repeatedly clear the field while a local
+serviceaccounts controller attempts to repeatedly set it to a
+generated value.
 
 ## Higher order behaviour
 
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -375,7 +375,7 @@ func (s *FederationSyncController) clusterOperations(selectedClusters, unselecte
 				continue
 			}
 
-			desiredObj, err = s.objectForUpdateOp(desiredObj, clusterObj)
+			desiredObj, err = s.objectForUpdateOp(desiredObj, clusterObj, fedResource.Object())
 			if err != nil {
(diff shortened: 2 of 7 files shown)


marun:
Rebased

marun:
Holding for review

pkg/controller/sync/controller.go
```diff
@@ -459,7 +459,7 @@ func (s *FederationSyncController) objectForUpdateOp(desiredObj, clusterObj *uns
 	if targetKind == util.ServiceAccountKind {
 		return serviceAccountForUpdateOp(desiredObj, clusterObj)
 	}
-	return desiredObj, nil
+	return retainReplicas(desiredObj, clusterObj, fedObj)
```
irfanurrehman: Logically, the check on line 561 should happen here.
```
if retainReplicas {
    return retainReplicas(desiredObj, clusterObj, fedObj)
}

return desiredObj, nil
```
or change name to something like `retainReplicasIfNeeded` which actually sounds crude.

irfanurrehman:
I had trouble understanding this 
```
The granularity of this retention is all clusters or no-clusters, with
the expectation that where HPA is intended to be applied selectively,
a user will need to define 2 resources (one with `retainReplicas: true`
and one with `retainReplicas: false`) and define appropriate placement.
```
I believe you meant 2 different `hpas` (different names) targeting 2 different `resources`? In which case only this comment should be sufficient -> `The granularity of this retention is all clusters or no-clusters.`

pkg/controller/sync/controller.go
marun: Why is it problematic for the function to conditionally modify the resource?  Given that the test for retention is non-trivial, I would prefer not to cluster objectForUpdateOp with it.

marun:
@irfanurrehman The point of that comment was to document that it's not possible for a given resource to retain replicas in some clusters but not others.  If it was desirable for a given deployment to be managed by HPA in some clusters but not others, the scheme proposed by this PR would require that a user create one deployment for clusters where it would be targeted by HPA and set `retainReplicas: true` on that resource, and create another deployment for the clusters where it would not be targeted by HPA and ensure `retainReplicas: false` (the default).

pkg/controller/util/constants.go
```diff
@@ -40,6 +40,10 @@ const (
 	// ServiceAccount fields
 	SecretsField = "secrets"
 
+	// Scale types
```
xunpan: `Scale types` seems not clear enough. What about ` HPA impact fields`?

xunpan:
The code LGTM.

For the design, I'm thinking if we can make it more general feature.
`HPA` + `Deployment` is a way for `policy` + `resource`. It should be possible for other CRDs to design in similar way. A field in a resource spec is updated by related controller with a policy. If that's ture, current design for retain cannot work in configurable way. It only take effect for fixed `replicas` item.

@marun

marun:
@xunpan I can imagine there being a general-purpose solution at some point, but I don't think we have enough examples to derive that solution just yet.  I'm assuming we'll collect examples as user adoption increases.  The problem is that kubernetes does not dictate how controllers interact with API resources, so all the ways in which federation can end up fighting with those controllers is largely unknown.

pkg/controller/sync/controller.go
irfanurrehman: Readability; however, I don't see any harm in it being the way it is except the impression to the code reader here that all resources can have `replicas`.

irfanurrehman:
> @irfanurrehman The point of that comment was to document that it's not possible for a given resource to retain replicas in some clusters but not others. If it was desirable for a given deployment to be managed by HPA in some clusters but not others, the scheme proposed by this PR would require that a user create one deployment for clusters where it would be targeted by HPA and set `retainReplicas: true` on that resource, and create another deployment for the clusters where it would not be targeted by HPA and ensure `retainReplicas: false` (the default).

Understand the intent now.

irfanurrehman:
@marun This is LGTM for me. Are we expecting review from anybody else?

marun:
@irfanurrehman No, but I need to add testing/docs.

marun:
I've rebased and added added docs to the user guide.  Testing is next.

marun:
Added unit test

marun:
@irfanurrehman I think this is ready for review.

docs/userguide.md
```diff
@@ -528,6 +530,35 @@ employed by deployment:
...
+| Scalable        | spec.replicas             | The HPA controller may be managing the replica count of a scalable resource. |
+| Services        | spec.clusterIP,spec.ports | The services controller may be managing these fields.                        |
+| ServiceAccounts | secret                    | The serviceaccounts controller may be managing this field                    |
+
+
+### Scalable Resources
+
+For scalable resources (those that have a scale subtype
+e.g. `ReplicaSet` and `Deployment`), retention of the `spec.replicas`
+field is controlled by the `retainReplicas` boolean field of the
```
irfanurrehman: It does not mention anywhere that except `spec.replicas` no other fields listed in the table in the previous section can be controlled by user. To avoid confusion, it might make sense to add a note.

irfanurrehman:
@marun thanks for doing this.
Looks fine to me except a suggestion in documentation.

docs/userguide.md
marun: Done

marun:
Docs have been updated as requested.  PTAL

irfanurrehman:
waiting for the fix on CI failure from @shashidharatd.

marun:
Rebased

irfanurrehman:
Thanks for doing this.
/lgtm

PR #594, opened 2019-02-13
build kubefed2 on demands with parameter control

xunpan:
Fix#589

build kubefed2 on demands with parameter control: when the script is used independently, kubefed2 should be build. If it is used in pre-commit.sh, the build process should be ignored.

@gyliu513 @marun

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -163,7 +163,7 @@ echo "Deploying federation-v2"
 ./scripts/deploy-federation.sh ${CONTAINER_REGISTRY_HOST}/federation-v2:e2e $(join-cluster-list)
 
 echo "Checking sync up status of helm chart"
-PATH="${PATH}:${base_dir}/bin" ./scripts/sync-up-helm-chart.sh
+BUILD_KUBEFED="false" PATH="${PATH}:${base_dir}/bin" ./scripts/sync-up-helm-chart.sh
 
 echo "Checking helm chart state of working tree"
 check-git-state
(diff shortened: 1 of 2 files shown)


marun:
Thank you for doing this!   We don't have a formal bash style guide, but I'd like to see quoting of substitutions consistent with the script content e.g.

```bash
BUILD="${BUILD}" vs BUILD=${BUILD}

**Comment by xunpan** [comment]
Related code are quoted.

**Comment by gyliu513** [comment]
@marun @xunpan 

I found some examples in script not using quote,  like the following:

- https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/deploy-federation.sh#L162
- https://github.com/kubernetes-sigs/federation-v2/blob/master/scripts/delete-federation.sh#L92

What is the advantage of adding quote?

**Comment by xunpan** [comment]
bash will treat message in "" as a whole. E.g.
```bash
$ mkdir "hi job"
$ MD="hi job"
$ echo $MD
hi job
$ cd $MD
bash: cd: hi: No such file or directory
$ cd "$MD"
$ pwd
/tmp/hi job
```
For this case, I agreed to use "" based on marun's comments because 
1. it handles `${ROOT_DIR}` correctly even with blank in the path
2. it is consistent style in the context for bash code.

@marun any comments?

marun:
@xunpan I think that's correct, yes.  Thank you!

At some point we'll probably want a bash style guide that codifies the intentional elements of style in the existing scripts, and a corresponding effort to fix the inconsistent elements.  I don't think it's a high priority yet, though.

(No action required) I think it's preferable to use `y` for boolean true and empty string as boolean false, such that test for empty (`-z`) and its negation can be used in conditional statements.  I don't feel strongly enough to ask for it to change in this PR though.

/approve
/lgtm

PR #595, opened 2019-02-13
Fix unjoin to use namespace identity instead of context name as criteria for removal

### LINKED ISSUE #593: Kubefed2 unjoin of cluster hosting control plane deletes federation control plane
**What happened**:
I was attempting to restore a previously installed control plane on a new cluster (minishift) for disaster recovery testing. Certificates did not match, so I attempted to unjoin/rejoin the host cluster to reestablish control, but the unjoin removed the federation-system namespace, thus deleting the federation-controller-manager-0 deployment config and all cluster definitions.

Kubefed2 unjoin should not delete federation controller's namespace from the host cluster.

Federate a cluster, then unjoin it. Observe that federation namespace is deleted, removing entire control plane as well as the connection.

**Environment**:

Minishift with current OCP:

Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:08:12Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2019-02-08T23:07:29Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}

- Federation version

git d474b39dd49

hack/latest

namespaced

- Others
kubefed2 version: version.Info{Version:"v0.0.4", GitCommit:"2cdf5d37240d9b8b33e2715deb75fbb7f9e003ad", GitTreeState:"clean", BuildDate:"2018-12-10T23:03:18Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}

/kind bug

marun:
Previously unjoin was comparing the context names of the host and unjoining cluster, and removing the federation namespace in the unjoining cluster if the names didn't match.   If the host context name was not specified, though, a value of empty string (`""`) would be used and the federation namespace would always be removed even for the host cluster.  This PR switches to comparing the identities of the federation namespace in the host and unjoining clusters and only removing the namespace if the 2 don't match.

I've cleaned up the logging in this function, which now deviates from the rest of join/unjoin in its formatting.  I'll leave it up to the reviewers whether this PR should be updated for consistency or the rest of the code should be updated in a subsequent PR.

Fixes #593

### DIFF
--- a/pkg/kubefed2/unjoin.go
+++ b/pkg/kubefed2/unjoin.go
@@ -17,13 +17,15 @@ import (
 	"io"
 
 	"github.com/golang/glog"
+	"github.com/pkg/errors"
 
 	fedclient "github.com/kubernetes-sigs/federation-v2/pkg/client/clientset/versioned"
+	controllerutil "github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/options"
 	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/util"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	client "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


xunpan:
/lgtm
The UID should be UUID set by kubernetes. It is unique even cross clusters.

PR #596, opened 2019-02-14
forcely delete a member cluster information in the host cluster 

xunpan:
Fix#592

forcely delete federation information in the host cluster if a member cluster
is not available

### DIFF
--- a/pkg/kubefed2/unjoin.go
+++ b/pkg/kubefed2/unjoin.go
@@ -144,7 +144,13 @@ func (j *unjoinFederation) Run(cmdOut io.Writer, config util.FedConfig) error {
 	clusterConfig, err := config.ClusterConfig(j.clusterContext, j.Kubeconfig)
 	if err != nil {
 		glog.V(2).Infof("Failed to get unjoining cluster config: %v", err)
-		return err
+
+		if !j.forceDeletion {
+			return err
+		}
+		// If configuration for the member cluster cannot be successfully loaded,
+		// forceDeletion indicates that resources associated with the member cluster
+		// should still be removed from the host cluster.
 	}
 
 	hostClusterName := j.HostClusterContext
@@ -167,10 +173,15 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, 
 		return err
 	}
 
-	clusterClientset, err := util.ClusterClientset(clusterConfig)
-	if err != nil {
-		glog.V(2).Infof("Failed to get unjoining cluster clientset: %v", err)
-		return err
+	var clusterClientset *client.Clientset
+	if clusterConfig != nil {
+		clusterClientset, err = util.ClusterClientset(clusterConfig)
+		if err != nil {
+			glog.V(2).Infof("Failed to get unjoining cluster clientset: %v", err)
+			if !forceDeletion {
+				return err
+			}
+		}
 	}
 
 	fedClientset, err := util.FedClientset(hostConfig)
@@ -180,15 +191,18 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace, 
 	}
 
 	if removeFromRegistry {
-		removeFromClusterRegistry(hostConfig, clusterNamespace, clusterConfig.Host, unjoiningClusterName, dryRun)
+		removeFromClusterRegistry(hostConfig, clusterNamespace, unjoiningClusterName, dryRun)
 	}
 
-	deletionSucceeded := deleteRBACResources(clusterClientset, federationNamespace, unjoiningClusterName, hostClusterName, dryRun)
+	var deletionSucceeded bool
+	if clusterClientset != nil {
+		deletionSucceeded = deleteRBACResources(clusterClientset, federationNamespace, unjoiningClusterName, hostClusterName, dryRun)
 
-	err = deleteFedNSFromUnjoinCluster(hostClientset, clusterClientset, federationNamespace, unjoiningClusterName, dryRun)
-	if err != nil {
-		glog.Errorf("Error deleting federation namespace from unjoin cluster: %v", err)
-		deletionSucceeded = false
+		err = deleteFedNSFromUnjoinCluster(hostClientset, clusterClientset, federationNamespace, unjoiningClusterName, dryRun)
+		if err != nil {
+			glog.Errorf("Error deleting federation namespace from unjoin cluster: %v", err)
+			deletionSucceeded = false
+		}
 	}
 
 	// deletionSucceeded when all operations in deleteRBACResources and deleteFedNSFromUnjoinCluster succeed.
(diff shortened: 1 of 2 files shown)


pkg/kubefed2/unjoin.go
```diff
@@ -183,12 +193,15 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
 		removeFromClusterRegistry(hostConfig, clusterNamespace, clusterConfig.Host, unjoiningClusterName, dryRun)
```
yamt: is this `clusterConfig.Host` safe?

yamt: this `host` argument seems unused. i guess you can just remove it.

xunpan: Good catch. Thanks for your comments. It is resolved.

pkg/kubefed2/unjoin.go
```diff
@@ -167,10 +172,15 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
 		return err
 	}
 
-	clusterClientset, err := util.ClusterClientset(clusterConfig)
-	if err != nil {
-		glog.V(2).Infof("Failed to get unjoining cluster clientset: %v", err)
-		return err
+	var clusterClientset *client.Clientset = nil
```
marun: The default value of pointers is nil, so assigning nil is unnecessary.

pkg/kubefed2/unjoin.go
```diff
@@ -144,7 +144,12 @@ func (j *unjoinFederation) Run(cmdOut io.Writer, config util.FedConfig) error {
 	clusterConfig, err := config.ClusterConfig(j.clusterContext, j.Kubeconfig)
 	if err != nil {
 		glog.V(2).Infof("Failed to get unjoining cluster config: %v", err)
-		return err
+
+		if !j.forceDeletion {
+			return err
+		}
+		// if member cluster is not available with forceDeletion,
+		// deleting federation information in the host cluster for the unjoining cluster
```
marun: -> If configuration for the member cluster cannot be successfully loaded, forceDeletion indicates that resources associated with the member cluster should still be removed from the host cluster.

pkg/kubefed2/unjoin.go
```diff
@@ -180,15 +190,18 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
 	}
 
 	if removeFromRegistry {
-		removeFromClusterRegistry(hostConfig, clusterNamespace, clusterConfig.Host, unjoiningClusterName, dryRun)
+		removeFromClusterRegistry(hostConfig, clusterNamespace, unjoiningClusterName, dryRun)
 	}
 
-	deletionSucceeded := deleteRBACResources(clusterClientset, federationNamespace, unjoiningClusterName, hostClusterName, dryRun)
+	var deletionSucceeded bool = false
```
marun: The default value for bool is false, so assigning false is unnecessary.  Both of the following forms are preferred:

```golang
// shorthand-style, useful for initializing non-pointer types  
deletionSucceeded := false
// var-style, with default to zero-value for type (false for bool)
var deletionSucceeded bool
```

pkg/kubefed2/unjoin.go
```diff
@@ -167,10 +172,15 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
 		return err
 	}
 
-	clusterClientset, err := util.ClusterClientset(clusterConfig)
-	if err != nil {
-		glog.V(2).Infof("Failed to get unjoining cluster clientset: %v", err)
-		return err
+	var clusterClientset *client.Clientset = nil
+	if clusterConfig != nil {
+		clusterClientset, err = util.ClusterClientset(clusterConfig)
```
marun: Please check the definition of `ClusterClientSet`.  Since it calls `client.NewForConfigOrDie`, it will never return an error and will instead panic which is not what the code here expects.  I'm not sure why an error is being returned, I'm guessing this suboptimal code was inherited from fedv1.

pkg/kubefed2/unjoin.go
```diff
@@ -180,15 +190,18 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
 	}
 
 	if removeFromRegistry {
-		removeFromClusterRegistry(hostConfig, clusterNamespace, clusterConfig.Host, unjoiningClusterName, dryRun)
+		removeFromClusterRegistry(hostConfig, clusterNamespace, unjoiningClusterName, dryRun)
```
marun: (No action required) In future PRs, please ensure that unrelated non-functional changes are separated into their own commits to ease future maintenance.

xunpan:
Thanks for your detailed comments. Addressed.

pkg/kubefed2/unjoin.go
```diff
@@ -167,10 +173,15 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
...
 	}
 
-	clusterClientset, err := util.ClusterClientset(clusterConfig)
-	if err != nil {
-		glog.V(2).Infof("Failed to get unjoining cluster clientset: %v", err)
-		return err
+	var clusterClientset *client.Clientset
+	if clusterConfig != nil {
+		clusterClientset, err = util.ClusterClientset(clusterConfig, forceDeletion)
+		if err != nil {
```
marun: This error handling is redundant and can be removed, as per a previous comment about `util.ClusterClientset`.

pkg/kubefed2/util/util.go
```diff
@@ -91,8 +91,12 @@ func HostClientset(config *rest.Config) (*client.Clientset, error) {
 
 // ClusterClientset provides a kubernetes API compliant clientset to
 // communicate with the joining cluster's kubernetes API server.
-func ClusterClientset(config *rest.Config) (*client.Clientset, error) {
-	return client.NewForConfigOrDie(config), nil
+func ClusterClientset(config *rest.Config, returnErr bool) (*client.Clientset, error) {
```
marun: I think changing how this client helper works should be tackled separately rather than differentiating just `ClusterClientset` to ensure consistency.

xunpan:
I make other related functions return error instead of panic in library function.
It is in another commit to make commit history clear.

marun:
Functionally this looks good, thank you!  Please ensure a good commit message for `Fix#592` (e.g. `Ensure 'unjoin --force' always attempts removal of resources from host cluster`) and I'll merge.

marun:
@xunpan Thank you!

I realize now why I've been missing changes that you've been making.  I don't get notified on commit changes, only on PR comments.  When you've responded to feedback and pushed changes, please remember to add a comment (even as simple as PTAL) so that I know to re-review. 

/approve
/lgtm

PR #599, opened 2019-02-15
Replace references to 'federation primitives' with 'federated type'

marun:
This change is in response to the unification of federation primitives into federated types.  The term 'primitives' is no longer relevant, and continuing to include it in documentation or code is likely to cause confusion.  As part of this change, the api group that federated types are generated into has been changed from `primitives.federation.k8s.io` to `types.federation.k8s.io`.  I'm not fixed on the name, but there does need to be differentiation from the `core` group, as everything in the core group is essential to propagation but the contents of the generated group are likely to depend on the propagation requirements of a federation deployment.

### DIFF
--- a/charts/federation-v2/templates/clusterroles.rbac.authorization.k8s.io.yaml
+++ b/charts/federation-v2/templates/clusterroles.rbac.authorization.k8s.io.yaml
@@ -5,7 +5,7 @@ metadata:
   name: clusterroles.rbac.authorization.k8s.io
 spec:
   federatedType:
-    group: primitives.federation.k8s.io
+    group: types.federation.k8s.io
     kind: FederatedClusterRole
     pluralName: federatedclusterroles
     version: v1alpha1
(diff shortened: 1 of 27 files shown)


marun:
Rebased

marun:
@xunpan Please take care to avoid lgtm'ing a PR if nobody else has had a chance to review. Unfortunately we don't have a way to disable auto-approve for approvers, so if I don't remember to remove the approved label a careless lgtm can merge a PR before other reviewers get to see it.

xunpan:
Got it. 

My initial thought was that it is a simple rename work. The changed code is fine. If it is CI passed, it should be OK. Nevertheless, it is fine to double reviewed for code quality :-)

marun:
@irfanurrehman @shashidharatd Are you ok with this?

marun:
Rebased

shashidharatd:
> @irfanurrehman @shashidharatd Are you ok with this?
yes, this LGTM

PR #600, opened 2019-02-18
Log improvements

yamt:


pkg/controller/federatedcluster/clusterclient.go
```diff
@@ -147,8 +146,7 @@ func getZoneNameForNode(node corev1.Node) (string, error) {
 			return value, nil
 		}
 	}
-	return "", errors.Errorf("Zone name for node %s not found. No label with key %s",
-		node.Name, LabelZoneFailureDomain)
+	return "", nil
```
marun: Why is it desirable to remove detail from these messages?

cc: @shashidharatd

yamt: because this message is shown on every iterations and it's too verbose IMO.

marun: I think we need to find a way to reduce the incidence of logging, then, rather than removing it entirely.  My understanding is that this will be useful information for anyone trying to troubleshoot federated dns or ingress.

yamt: ok. how about having a "already-warned" flag in ClusterClient and make it warn only the first time?

shashidharatd: Do agree with @marun, we need to reduce the incidence of logging. But at the same time, when one is not using the federated dns feature, this error should not be overwhelming the logs. I don't have an elegant solution for this but maybe we could keep a variable in controller per cluster to say, error is already reported.

yamt: i've pushed new patches to do that.

shashidharatd:
/lgtm
Thanks @yamt. @marun, you have any comments?

marun:
@shashidharatd Should there be a flag (defaulting to false) that disables the check?  It seems strange to be focusing on a symptom (the frequency of an error message) rather than fixing the cause (a check that should only be performed if necessary).

shashidharatd:
> @shashidharatd Should there be a flag (defaulting to false) that disables the check? It seems strange to be focusing on a symptom (the frequency of an error message) rather than fixing the cause (a check that should only be performed if necessary).

oh yes, we already have a feature gate [CrossClusterServiceDiscovery](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/features/features.go#L47) to which this zone & region label were checked and status updated. we could use this feature gate flag to update the `FederatedCluster` Status. So disabling the feature, we would not see such errors.

@yamt, Apologies for my earlier suggestion. Could you please consider the new suggestion. @marun, do you concur.
Thanks !

yamt:
> > @shashidharatd Should there be a flag (defaulting to false) that disables the check? It seems strange to be focusing on a symptom (the frequency of an error message) rather than fixing the cause (a check that should only be performed if necessary).
> 
> oh yes, we already have a feature gate [CrossClusterServiceDiscovery](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/features/features.go#L47) to which this zone & region label were checked and status updated. we could use this feature gate flag to update the `FederatedCluster` Status. So disabling the feature, we would not see such errors.
> 
> @yamt, Apologies for my earlier suggestion. Could you please consider the new suggestion. @marun, do you concur.
> Thanks !

isn't a feature gate something to be removed sometime after the feature got GA?
i'm not sure if it's appropriate to use it for this purpose.

shashidharatd:
> isn't a feature gate something to be removed sometime after the feature got GA?
i'm not sure if it's appropriate to use it for this purpose.

I think, that's a wrong perception that feature gates need to be removed after GA You could find some features which are GA but still managed through feature gates https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates

marun:
@shashidharatd According to the page you linked to:

```
Feature gates are a set of key=value pairs that describe alpha or experimental features.
...
A GA feature is also referred to as a stable feature. It means:

- The corresponding feature gate is no longer needed.
- Stable versions of features will appear in released software for many subsequent versions.
```

In light of that, I think it necessary to re-evaluate the use of 'feature gates' as the primary way to configure the controller manager.

cc: @pmorie

shashidharatd:
Even though features which are GA are defaulted to true, there could be some kubernetes cluster deployments who wants to selectively choose a feature. Say for e.g. `SupportIPVSProxyMode` feature gate. It is already GA, but it is an alternative to `IP table proxy`. I am not sure whether those feature gates would be removed ever in future.

yamt:
i made a bit research.

some feature gates have been removed after being stable for a while.
eg. ExternalTrafficLocalOnly https://github.com/kubernetes/kubernetes/pull/56948
discussion: https://github.com/kubernetes/kubernetes/issues/46404

otoh, it seems that some of feature gates have been introduced purely as a controlling knob.
eg. VolumeSubpath https://github.com/kubernetes/kubernetes/pull/61044

marun:
In the context of federation, the features being exposed only have one implementation and are primary to the project (propagation, scheduling, dns programming).  Does kube use feature gates to control primary features?

yamt:
kube controller manager has several enable/disable style command line options like --enable-garbage-collector.
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/
to me, that way seems more straightforward than feature gates.
but anyway i'm not familiar with k8s enough to judge either ways.

marun:
I've added a new issue (#611) to track the possibility of moving away from feature gates.  For now, as per @shashidharatd's comment, the checks for zone/region label should only can be enabled if CrossClusterServiceDiscovery is enabled (and maybe it should be disabled by default?).

marun:
Thank you!

/approve

@shashidharatd I've leave it to you to lgtm.  

@shashidharatd  What are your thoughts on disabling the feature gate by default?  I'm not sure how many people using federation will be configuring the necessary external dependencies or running in a compatible environment such that it would make sense to keep it enabled by default.

shashidharatd:
@yamt, This LGTM. Please resolve the conflict. Thanks!

yamt:
the ci failure doesn't seem related to this pr but looks legit.
i filed a separate pr for that. https://github.com/kubernetes-sigs/federation-v2/pull/620

PR #601, opened 2019-02-18
Fix service account sync

marun:
Previously service account propagation had been configured to use the generation field for versioning, which was masking the need to retain the secrets field when it was populated by the in-cluster service
account controller.  Switching to the resource version field for versioning resulted in thrashing between the sync and service account controllers.  The sync controller would continually clear the 'secrets' field, and the service account controller would continually generate a new token secret and set it in the 'secrets' field.  This change ensures that the sync controller retains the 'secrets' field of service accounts to avoid this problem.

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -452,8 +452,12 @@ func (s *FederationSyncController) objectForUpdateOp(desiredObj, clusterObj *uns
 	// Pass the same ResourceVersion as in the cluster object for update operation, otherwise operation will fail.
 	desiredObj.SetResourceVersion(clusterObj.GetResourceVersion())
 
-	if s.typeConfig.GetTarget().Kind == util.ServiceKind {
+	targetKind := s.typeConfig.GetTarget().Kind
+	if targetKind == util.ServiceKind {
 		return serviceForUpdateOp(desiredObj, clusterObj)
+	}
+	if targetKind == util.ServiceAccountKind {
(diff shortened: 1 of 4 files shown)


pkg/controller/sync/controller.go
```diff
@@ -452,9 +452,13 @@ func (s *FederationSyncController) objectForUpdateOp(desiredObj, clusterObj *uns
 	// Pass the same ResourceVersion as in the cluster object for update operation, otherwise operation will fail.
 	desiredObj.SetResourceVersion(clusterObj.GetResourceVersion())
 
-	if s.typeConfig.GetTarget().Kind == util.ServiceKind {
+	targetKind := s.typeConfig.GetTarget().Kind
+	if targetKind == util.ServiceKind {
```
shashidharatd: (no action required) maybe switch is better than multiple if's

marun:
@shashidharatd I discovered the issue while implementing #586.  The crud test for service accounts started failing because resource version was the field used automatically.

pkg/controller/sync/controller.go
marun: I'm thinking we'll want a configurable way to do this before too long, so I'm not too worried about optimizing it.

xunpan:
It LGTM.
This is also reviewed by shashidharatd and the left issue (manual setting secret deletion case) is commented in TODO, I think this can be merged.

/lgtm

PR #602, opened 2019-02-19
Implement a generic informer

marun:
The generated clients need to be removed as a precondition to implementing HA (due to a client-go dependency), but the generic client intended to replace the generated clients does not support the watch call necessary for informers.  This commit hacks together a generic informer to work around that omission so that the generic client is a viable replacement for the generated clients.

### DIFF
--- a/pkg/controller/util/genericinformer.go
+++ b/pkg/controller/util/genericinformer.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"github.com/pkg/errors"
+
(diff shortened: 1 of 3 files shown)


marun:
@shashidharatd Hopefully this PR can unblock your efforts to remove the generated clients.

pkg/controller/util/genericinformer.go
```diff
@@ -0,0 +1,79 @@
...
+			},
+			WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
+				// Watch needs to be set to true separately
+				opts.Watch = true
+				isNamespaceScoped := namespace != "" && mapping.Scope.Name() != meta.RESTScopeNameRoot
+				return client.Get().NamespaceIfScoped(namespace, isNamespaceScoped).Resource(mapping.Resource.Resource).VersionedParams(&opts, scheme.ParameterCodec).Watch()
+			},
+		},
+		obj,
+		NoResyncPeriod,
```
shashidharatd: (suggestion) Better to configure this `ResyncPeriod` by passing parameter to `NewGenericInformer`

marun: I was mimicking the behavior of `NewResourceInformer`, which similarly uses `NoResyncPeriod` because no callers needed to vary the value.  If/when we find a caller that needs to set the resync period, I'm entirely fine with the code being updated to meet that requirement. Until then I invoke YAGNI.

shashidharatd: SGTM

shashidharatd:
This LGTM, would wait for other reviewers, if they have any comments.

marun:
Rebased

marun:
Holding for review

PR #606, opened 2019-02-21
Test orphanDependents=true as well

yamt:


test/common/crudtester.go
```diff
@@ -79,17 +79,15 @@ func NewFederatedTypeCrudTester(testLogger TestLogger, typeConfig typeconfig.Int
 	}, nil
 }
 
-func (c *FederatedTypeCrudTester) CheckLifecycle(desiredFedObject *unstructured.Unstructured) {
+func (c *FederatedTypeCrudTester) CheckLifecycle(desiredFedObject *unstructured.Unstructured, orphanDependents *bool) {
```
marun: `CheckLifecycle` is an expensive test, which is why it only validates the golden path. That doesn't suggest adding `orphanDependents` parameter to validate the non-golden path.

test/common/crudtester.go
```diff
@@ -238,10 +236,16 @@ func (c *FederatedTypeCrudTester) CheckDelete(fedObject *unstructured.Unstructur
 	name := qualifiedName.Name
 	namespace := qualifiedName.Namespace
 
+	clusterNames, err := util.GetClusterNames(fedObject)
```
marun: This is unnecessary.  `testClusters` is guaranteed to include all of clusters a resource could be propagated to.

test/e2e/crud.go
```diff
@@ -77,7 +77,10 @@ var _ = Describe("Federated", func() {
 				testObjectFunc := func(namespace string, clusterNames []string) (*unstructured.Unstructured, error) {
 					return common.NewTestObject(typeConfig, namespace, clusterNames, fixture)
 				}
-				validateCrud(f, tl, typeConfig, testObjectFunc)
+				orphanDependents := false
+				validateCrud(f, tl, typeConfig, testObjectFunc, &orphanDependents)
+				orphanDependents = true
+				validateCrud(f, tl, typeConfig, testObjectFunc, &orphanDependents)
```
marun: It is not necessary to validate non-golden path deletion for every single resource.  To do so increases execution cost with negligible increase in test coverage.  It should be sufficient to test orphaning for a single resource type, and to write a test that specifically creates and deletes a resource.   Then the test needs to at least attempt cleanup of the now orphaned resources in member clusters, since the resources will otherwise not be removed when testing against a namespaced deployment.

test/common/crudtester.go
yamt: This is necessary because clusters included in testClusters but not in ClusterNames will not have orphaned resources.

yamt:
@marun i think it's still better to test orphan case in e2e rather than unit tests because it can collide with garbage collector. how do you think?

marun:
@yamt What makes you think it would collide with the garbage collector?

yamt:
> @yamt What makes you think it would collide with the garbage collector?

The use of the same finalizer, FinalizerOrphanDependents.

The following is a copy-and-paste from slack.

yamamoto   [8 days ago]
doesn't our controller race with the gc removing "orphan" finalizer?
yamamoto   [8 days ago]
when our controller runs first, i guess we should not remove the "orphan" finalizer. otherwise the gc might be confused.
yamamoto   [8 days ago]
when the gc runs first, it removes the "orphan" finalizer and our controller misses it.

marun:
@yamt I realize you are correct that a race condition exists (see #622).  I think it would be preferable to fix that issue (the race condition).  Would you be interested in tackling it?

marun:
@yamt I missed the import of our conversation on slack.  There is no harm in the federation controller removing the orphan finalizer before the gc sees it, because there shouldn't be any dependent resources of a federated resource that the gc needs to remove owner references for.  The reverse is not true.  As per #622, if the gc removes the finalizer before the sync controller sees it, orphaning will not take place.

yamt:
> There is no harm in the federation controller removing the orphan finalizer before the gc sees it, because there shouldn't be any dependent resources of a federated resource that the gc needs to remove owner references for.

i guess nothing prevents another controller unknown to us from creating dependent resources. i agree the opposite case is the bigger concern though.

marun:
Closing in favor of #669

PR #610, opened 2019-02-25
Short name support in kubefed2 disable and reject case if multiple resource name matched

### LINKED ISSUE #608: kubefed2 disable should support the use of kind and short name

`kubefed2 disable` only supports identifying the type config to enable by name (group-qualified plural name e.g. deployment.apps).  It should be possible to specify kind (Deployment) or short name (deploy) as well, as is already possible with `kubefed2 enable`.  Only if an api resource is not available (e.g. mycrd.mygroup was deleted before the associated federated type config) should the name of the type config be mandatory.  

**Why is this needed**:

enable/disable should be consistent in the type of identifiers they support.

**Note**: 

The search for an api resource in `enable` currently stops at the first match it finds.  It should be checking that the provided identity is not ambiguous and matches only one api resource.  In the case of deployments, both `deployment.apps` (desired) and `deployments.extensions`(legacy) will be matched by the kind `Deployment` and the short name `deploy`.   If the search finds more than one api resource, it should report an error indicating that a group-qualified plural name must be provided.

/kind feature

xunpan:
- `kubefed2 disable` is supported to use kind and short name in command line 

- if multiple resource names are matched by kind/short name, reject the operation and print error message to let end user specify a group-qualified name.

Fixes #608

### DIFF
--- a/pkg/kubefed2/federate/disable.go
+++ b/pkg/kubefed2/federate/disable.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
@@ -121,9 +122,21 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return errors.Wrap(err, "Failed to get host cluster config")
 	}
 
+	// If . is specified, the target name is assumed as a group qualified name.
+	// In such case, ignore the lookup to make sure deletion of a federatedtypeconfig
+	// for which the corresponding target has been removed.
+	name := j.targetName
+	if !strings.Contains(j.targetName, ".") {
+		apiResource, err := LookupAPIResource(hostConfig, j.targetName, "")
+		if err != nil {
+			return err
+		}
+		name = typeconfig.GroupQualifiedName(*apiResource)
+	}
+
 	typeConfigName := ctlutil.QualifiedName{
 		Namespace: j.FederationNamespace,
-		Name:      j.targetName,
+		Name:      name,
 	}
 
 	return DisableFederation(cmdOut, hostConfig, typeConfigName, j.delete, j.DryRun)
--- a/pkg/kubefed2/federate/util.go
+++ b/pkg/kubefed2/federate/util.go
@@ -91,6 +91,17 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
 	// TODO(marun) Consider using a caching scheme ala kubectl
 	lowerKey := strings.ToLower(key)
 	var targetResource *metav1.APIResource
+	var matchedResources []string
+	var matchResource = func(resource metav1.APIResource, gv schema.GroupVersion) {
+		if targetResource == nil {
+			targetResource = resource.DeepCopy()
+			targetResource.Group = gv.Group
+			targetResource.Version = gv.Version
+		}
+
+		matchedResources = append(matchedResources, groupQualifiedName(resource.Name, gv.Group))
+	}
+
 	for _, resourceList := range resourceLists {
 		// The list holds the GroupVersion for its list of APIResources
 		gv, err := schema.ParseGroupVersion(resourceList.GroupVersion)
@@ -106,24 +117,20 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
 				lowerKey == strings.ToLower(resource.Kind) ||
 				lowerKey == fmt.Sprintf("%s.%s", resource.Name, gv.Group) {
 
-				targetResource = &resource
-				break
+				matchResource(resource, gv)
+				continue
 			}
 			for _, shortName := range resource.ShortNames {
 				if lowerKey == strings.ToLower(shortName) {
-					targetResource = &resource
+					matchResource(resource, gv)
 					break
 				}
 			}
-			if targetResource != nil {
-				break
-			}
 		}
-		if targetResource != nil {
-			targetResource.Group = gv.Group
-			targetResource.Version = gv.Version
-			break
-		}
+
+	}
+	if len(matchedResources) > 1 {
+		return nil, errors.Errorf("Multiple resources are matched by %q: %s. A group-qualified plural name must be provided.", key, strings.Join(matchedResources, ", "))
 	}
 
 	if targetResource != nil {
@@ -146,4 +153,13 @@ func resourceKey(apiResource metav1.APIResource) string {
 		version = apiResource.Version
 	}
 	return fmt.Sprintf("%s.%s/%s", apiResource.Name, group, version)
+}
+
+func groupQualifiedName(name, group string) string {
+	apiResource := metav1.APIResource{
+		Name:  name,
+		Group: group,
+	}
+
+	return typeconfig.GroupQualifiedName(apiResource)
 }


pkg/kubefed2/federate/util.go
```diff
@@ -106,24 +108,36 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
...
-				break
+				if targetResource != nil {
+					multiMatched = true
+				} else {
+					targetResource = resource.DeepCopy()
+					targetResource.Group = gv.Group
+					targetResource.Version = gv.Version
+				}
+
+				message += fmt.Sprintf(", %s.%s", resource.Name, gv.Group)
```
marun: Consider collecting matched resource names in a slice and using `strings.Join` to compose a message from the slice.

marun:
@xunpan I have a minor comment inline but functionally this looks good.

pkg/kubefed2/federate/util.go
yamt: why don't you use GroupQualifiedName?

xunpan:
@yamt 

Here is an example
```
$ kubefed2 enable deployment                                                                                                                                          
F0226 02:44:10.241239    3416 enable.go:109] error: Multiple resources are matched: deployments.extensions, deployments.apps. A group-qualified plural name must be provided.
```

pkg/kubefed2/federate/util.go
xunpan: You can refer to the context comments
```
// The list holds the GroupVersion for its list of APIResources
gv, err := schema.ParseGroupVersion(resourceList.GroupVersion)
```

yamt:
> @yamt
> 
> Here is an example
> 
> ```
> $ kubefed2 enable deployment                                                                                                                                          
> F0226 02:44:10.241239    3416 enable.go:109] error: Multiple resources are matched: deployments.extensions, deployments.apps. A group-qualified plural name must be provided.
> ```

i got it. thank you!

yamt:
at least for "kubefed2 enable", i think it's more user friendly to pick one of them as kubectl does rather than complaining about the ambiguity. (i don't know how kubectl decides which to use though.)
how do you think?

xunpan:
In my opinion, if we accept instead of reject, we may blindly picked one to federate as there is no preference of matched resources.
However, if end user federate a resource (picked one) which is not what they want.
1. They need to delete the federated resource and try to federate what they want again. 
2. In such case, they may be puzzled because there is no message tell what happens.

Rejection can tell what is correct and avoid misleading.

marun:
@yamt Try running `kubectl -n kube-system get deploy -v 6`, and note that the api called is for `apis/extensions/v1beta1`.  It will return deployments with `apiVersion: extensions/beta1`.  Then try running `kubectl -n kube-system get deployment.apps -v 6`, and note that the api called is now `apis/apps/v1`, and the same deployments with `apiVersion: apps/v1`.  This is fine for an existing resource, since there is automatic conversion between the versions.

For our use-case, though, a user needs to target a specific group/version since we can't support more than one in a federated type.  In cases where there are multiple results for a kind/name/shortname, we need to warn the user and ensure that they provide a qualified name that returns only a single result.  If we don't do this, and someone tries to federate `deploy`, they will end up federating the beta `deployment.extensions` instead of the GA `deployment.apps`.

pkg/kubefed2/federate/disable.go
```diff
@@ -119,9 +119,14 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return errors.Wrap(err, "Failed to get host cluster config")
 	}
 
+	apiResource, err := LookupAPIResource(hostConfig, j.targetName, "")
```
marun: If the target name is qualified (i.e. includes a `.`), I think it can be assumed that the user is providing the group-qualified name likely to correspond to the ftc name.

pkg/kubefed2/federate/util.go
```diff
@@ -106,24 +108,36 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
...
-				break
+				if targetResource != nil {
+					multiMatched = true
+				} else {
+					targetResource = resource.DeepCopy()
+					targetResource.Group = gv.Group
+					targetResource.Version = gv.Version
+				}
+
+				matchedResources = append(matchedResources, fmt.Sprintf("%s.%s", resource.Name, gv.Group))
```
marun: As per a previous comment, please use `pkg/apis/core/typeconfig:GroupQualifiedName` to compose the name to ensure that core resources (lacking a group) don't get an unnecessary suffix.

pkg/kubefed2/federate/util.go
```diff
@@ -106,24 +108,36 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
 				lowerKey == strings.ToLower(resource.Kind) ||
 				lowerKey == fmt.Sprintf("%s.%s", resource.Name, gv.Group) {
 
-				targetResource = &resource
-				break
+				if targetResource != nil {
```
marun: (No action required) Consider creating an anonymous function that can be reused across both match points.

pkg/kubefed2/federate/disable.go
xunpan: make sense. Resolved

pkg/kubefed2/federate/util.go
xunpan: Resolved. I just make a private function because currently it is only used in this file.

pkg/kubefed2/federate/util.go
xunpan: addressed

pkg/kubefed2/federate/util.go
```diff
@@ -101,29 +103,37 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
 			continue
 		}
 		for _, resource := range resourceList.APIResources {
+			var matchResource = func() {
```
marun: Nice cleanup!  In the interest of minimizing indentation and simplifying loop construction, consider moving the function definition outside of the loop and accept parameters for the loop-local variables, e.g.

```golang
var matchResource = func(resource metav1.APIResource, gv schema.GroupVersion) {
...
}
```

pkg/kubefed2/federate/util.go
```diff
@@ -101,29 +103,37 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
...
-			}
-		}
-		if targetResource != nil {
-			targetResource.Group = gv.Group
-			targetResource.Version = gv.Version
-			break
 		}
+
+	}
+	if multiMatched {
```
marun: Consider checking `len(matchedResources) > 1` to remove the need for the multiMatched variable.

pkg/kubefed2/federate/util.go
```diff
@@ -101,29 +103,37 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
...
-				targetResource = &resource
-				break
+				matchResource()
+				continue
 			}
 			for _, shortName := range resource.ShortNames {
 				if lowerKey == strings.ToLower(shortName) {
-					targetResource = &resource
+					matchResource()
 					break
```
marun: This should be `continue` to ensure that duplicate short names are also detected (e.g. `deploy` amiguously refers to both `deployment.apps` and `deployment.extensions`).

pkg/kubefed2/federate/util.go
```diff
@@ -101,29 +103,37 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
...
-		}
-		if targetResource != nil {
-			targetResource.Group = gv.Group
-			targetResource.Version = gv.Version
-			break
 		}
+
+	}
+	if multiMatched {
+		return nil, errors.Errorf("Multiple resources are matched: %s. A group-qualified plural name must be provided.", strings.Join(matchedResources, ", "))
```
marun: Consider indicating the name that was supplied in the error message e.g. `Multiple resources are matched by "<supplied name>":`

yamt:
@marun thank you for explanation.

i was assuming that there was a machine-readable "preferred order for create" somewhere.
but it seems it isn't.
- `kubectl get deploy` just uses the server-preferred order. that is, `extensions/v1beta1` for compatibility reasons.
- `kubectl create deploy` is nothing generic. it's just a hardcoded cobra.Command alias.

pkg/kubefed2/federate/disable.go
```diff
@@ -121,9 +122,21 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return errors.Wrap(err, "Failed to get host cluster config")
 	}
 
+	// If . is specified, the target name is assumed as a group qualified name.
+	// In such case, ignore the lookup to make sure deletion of a federatedtypeconfig
+	// for which the corresponding target has been removed.
```
yamt: just curious: in what case is this extra check necessary?

pkg/kubefed2/federate/util.go
yamt: @marun this `break` just escapes the `resource.ShortNames` loop, doesn't it?

xunpan: For a resource, I guess it has at most one short name matched. 
deploy is a short name for both `deployment.apps` and `deployment.extensions`

pkg/kubefed2/federate/disable.go
marun: This command can't assume that the target crd exists at the time it is invoked (in which case lookup would fail), and I think it's nicer to delete/disable the ftc rather than failing if that occurs.

pkg/kubefed2/federate/util.go
xunpan: @yamt, yes, it breaks `shortNames` loop.

marun:
Nice work, thank you!

/approve
/lgtm

PR #615, opened 2019-02-27
Add shortnames prefixed with f when enabling a type

andersjanmyr:


andersjanmyr:
I signed it

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -67,7 +67,8 @@ type APIResource struct {
 	// Lower-cased plural name of the resource (e.g. configmaps).  If
 	// not provided, it will be computed by lower-casing the kind and
 	// suffixing an 's'.
-	PluralName string `json:"pluralName,omitempty"`
+	PluralName string   `json:"pluralName,omitempty"`
+	ShortNames []string `json:"shortNames,omitempty"`
```
marun: This struct is only used in FederatedTypeConfig type, and only includes the fields required by a controller to access the api of a kubernetes type.  A short name is a UX nicety that controllers won't use, so specifying short names in this context does not make sense.

pkg/kubefed2/federate/enable.go
```diff
@@ -253,6 +253,10 @@ func typeConfigForTarget(apiResource metav1.APIResource, enableTypeDirective *En
 	spec := enableTypeDirective.Spec
 	kind := apiResource.Kind
 	pluralName := apiResource.Name
+	shortNames := []string{}
```
marun: This function is only responsible for creating a type config, and short names don't belong in the type config.  Please relocate this logic to the `GetResources` function and pass the computed short names in to `federateTypeCRD`.

pkg/kubefed2/federate/enable.go
```diff
@@ -253,6 +253,10 @@ func typeConfigForTarget(apiResource metav1.APIResource, enableTypeDirective *En
...
+	for _, shortName := range apiResource.ShortNames {
+		shortNames = append(shortNames, fmt.Sprintf("f%s", shortName))
```
marun: Is `f<short name>` sufficiently descriptive, or would a lengthier prefix like 'fed' be desirable?

(No action required) Not required for this PR, but it may be desirable in future to support configuring either the prefix or the short name for an enable invocation.

marun:
@andersjanmyr The CI failure is due to making a type change without updating the generated code.  Updating the code to avoid a type change (as suggested inline) will fix that problem.

test/e2e/crd.go
```diff
@@ -98,8 +98,9 @@ func validateCrdCrud(f framework.FederationFramework, targetCrdKind string, name
 			},
 		},
 	})
+	shortNames := []string{"tapir"}
 
-	targetCrd := federate.CrdForAPIResource(targetAPIResource, validationSchema)
+	targetCrd := federate.CrdForAPIResource(targetAPIResource, validationSchema, shortNames)
```
marun: I think it's preferable to pass nil here, since short names don't figure in this test.

andersjanmyr: I was planning to update the validation once I figured out where it was validated. But, the test error didn't help me out.

andersjanmyr:
I intermittently get the error `concurrent map read and map write`.
https://travis-ci.org/kubernetes-sigs/federation-v2/builds/499542063
```
VersionManager ClusterVersionManager 
  should refresh and update after out-of-band creation
  /home/travis/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/version.go:550
fatal error: concurrent map read and map write
```

scripts/pre-commit.sh
```diff
@@ -107,6 +107,7 @@ function check-git-state() {
   for line in "${output}"; do
     echo "${line}"
   done
+  git diff
```
marun: Is this intentional?

andersjanmyr: Yes, it's very helpful to figure out what has changed in the files.

andersjanmyr: Example: https://travis-ci.org/kubernetes-sigs/federation-v2/builds/499900877#L1443

xunpan:
@andersjanmyr 
Thanks for your effort working on this. 

For files under `charts/federation-v2/templates/`, it is generated code. You can follow [Test Your Changes](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md#test-your-changes). 

After execute:
```
```
You can get updated files under `charts/federation-v2/templates/` which helps to fix your error in CI

andersjanmyr:
@xunpan I got passed it and the build is working now. I think it can be merged.

marun:
@andersjanmyr Please squash down to a single commit.  Otherwise lgtm.

andersjanmyr:
@marun I have done as you asked. The problem is that the build fails even though I'm pushing the exact same code.

marun:
@andersjanmyr I've filed #623 to track the issues related to race detection.  CI should stabilize once #620 and #624 have merged.  The fix for scheduling is less about CI instability as ensuring we can  enable race detection in CI and prevent the introduction of further races.

marun:
Thank you for your patience!

/approve
/lgtm

PR #616, opened 2019-02-27
Remove typed client usage

shashidharatd:
This is in continuation of https://github.com/kubernetes-sigs/federation-v2/pull/604. wherever possible, used the generic client in place cluster-registry client and kube clients. Also removed generated unit tests for api's which use generated federation client. they seems redundant.

/cc @marun


pkg/controller/federatedcluster/controller.go
```diff
@@ -88,13 +80,11 @@ func StartClusterController(config *util.ControllerConfig, stopChan <-chan struc
 
 // newClusterController returns a new cluster controller
 func newClusterController(config *util.ControllerConfig, clusterMonitorPeriod time.Duration) (*ClusterController, error) {
-	client, kubeClient, crClient := config.AllClients("cluster-controller")
+	client := genericclient.NewForConfigOrDieWithUserAgent(config.KubeConfig, "cluster-controller")
```
marun: (No action required) Shades of java-land naming?  ;)

pkg/kubefed2/util/util.go
```diff
@@ -96,8 +96,8 @@ func ClusterClientset(config *rest.Config) (*kubeclient.Clientset, error) {
 
 // ClusterRegistryClientset provides a cluster registry API compliant
 // clientset to communicate with the cluster registry.
-func ClusterRegistryClientset(config *rest.Config) (*crclient.Clientset, error) {
-	return crclient.NewForConfig(config)
+func ClusterRegistryClientset(config *rest.Config) (generic.Client, error) {
```
marun: (No action required) What's the advantage of maintaining this method vs just replacing it with calls to `generic.New`?

shashidharatd: No advantage, just that code change was big. It can be taken up as next step in simplifying the code.

shashidharatd:
> As an aside, are you intending to implement my suggestion of removing the requirement to pass a context to methods in the client interface?

Yes @marun, would consider it in next code optimisation.

PR #620, opened 2019-03-04
Build a Scheme for generic client at init time

yamt:
apis.Scheme is a global object.  AddToScheme to that is not
a concurrency safe operation and occasionally causes
"fatal error: concurrent map read and map write" panics
during e2e tests.

### DIFF
--- a/pkg/client/generic/genericclient.go
+++ b/pkg/client/generic/genericclient.go
@@ -21,12 +21,10 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
-	k8sscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
-	crscheme "k8s.io/cluster-registry/pkg/client/clientset/versioned/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/kubernetes-sigs/federation-v2/pkg/apis"
+	"github.com/kubernetes-sigs/federation-v2/pkg/client/generic/scheme"
 )
 
 type Client interface {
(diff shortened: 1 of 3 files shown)


yamt:
this is intended to fix this failure https://travis-ci.org/kubernetes-sigs/federation-v2/builds/499642962?utm_source=github_status&utm_medium=notification

pkg/client/generic/genericclient.go
```diff
@@ -43,7 +43,8 @@ type genericClient struct {
 }
 
 func New(config *rest.Config) (Client, error) {
-	scheme := apis.Scheme
+	scheme := runtime.NewScheme()
+	apis.AddToScheme(scheme)
```
shashidharatd: (no action required) we could alternatively do call AddToScheme of k8s and cluster-registry in init() of federation api package (like the sigs.k8s.io/controller-runtime does it). Calling `runtime.NewScheme` should also be fine as it is called once per new client.

marun: Given that the kube client appears to use a global, I think that would be a preferable strategy.

marun:
I've added #623 to track all the races that need to be fixed before we can enable race detection in CI, and this PR includes one of the required fixes.

pkg/client/generic/genericclient.go
yamt: does fed v2 types depend on clusterregistry types? otherwise, it seems a bit awkward to me to AddToScheme there.

@shashidharatd where does controller-runtime do it? i couldn't find it by a simple grep.

shashidharatd: @yamt, please fix this. Thanks!

shashidharatd: @yamt, sorry, i meant, in the new kubebuilder(>=v1.0.5). you can find that when a new type is created. there is also an auto generated file to add the Scheme to the global variable. like below
```
func init() {
	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
	AddToSchemes = append(AddToSchemes, v1alpha1.SchemeBuilder.AddToScheme)
}
```

yamt: i pushed a new patch after a private chat with @shashidharatd

shashidharatd:
/lgtm
Thanks @yamt

Issue #623, opened 2019-03-05
Enable race detection in CI


marun:
The recent changes to replace the generated clients introduced CI instability.  Anders Janmyr was first to notice this, and thought to enable the race detector (`-race').  Races were detected in the following code:

 - [x] Generic client construction (new race, fixed by #620)
 - [x] Version manager (new race, fixed by #624)
 - [x] Passing a shared config to generic client construction (new, fixed by #624)
 - [x] Scheduling (old, fixed by #626)
 - [x] Enable race detection in managed test run (since controllers are run in-process) (fixed by #626)

To prevent the introduction of new races, race detection needs to be turned on in CI for the managed fixture (which runs the controllers in-process).  CI won't pass until the above races are fixed, though.

PR #624, opened 2019-03-05
Fix races in version manager and generic client construction

marun:
- The switch to the generic client voilated some of the assumptions behind version manager client usage.  This PR updates the version manager to be fully compatible with the generic client.
- The generic client will trigger race detection if a client is constructed with a shared kubeconfig object that is changed after client construction.  The most common change in the codebase is the addition of a user agent.  This PR ensures a request for a new client with a user agent is always constructed with a copied kubeconfig.

Partially addresses #623.

### DIFF
--- a/pkg/controller/sync/version/cluster.go
+++ b/pkg/controller/sync/version/cluster.go
@@ -17,35 +17,28 @@ limitations under the License.
 package version
 
 import (
-	"context"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
-	genericclient "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 )
 
-type clusterVersionAdapter struct {
-	client genericclient.Client
-}
+type clusterVersionAdapter struct{}
 
-func newClusterVersionAdapter(client genericclient.Client) VersionAdapter {
-	return &clusterVersionAdapter{client}
-}
-
-func (a *clusterVersionAdapter) TypeName() string {
+func (*clusterVersionAdapter) TypeName() string {
 	return "ClusterPropagatedVersion"
 }
 
-func (a *clusterVersionAdapter) List(namespace string) (pkgruntime.Object, error) {
-	clusterPropagatedVersionList := &fedv1a1.ClusterPropagatedVersionList{}
-	err := a.client.List(context.TODO(), clusterPropagatedVersionList, namespace)
-	return clusterPropagatedVersionList, err
+func (*clusterVersionAdapter) NewListObject() pkgruntime.Object {
+	return &fedv1a1.ClusterPropagatedVersionList{}
 }
 
-func (a *clusterVersionAdapter) NewVersion(qualifiedName util.QualifiedName, ownerReference metav1.OwnerReference, status *fedv1a1.PropagatedVersionStatus) pkgruntime.Object {
+func (*clusterVersionAdapter) NewObject() pkgruntime.Object {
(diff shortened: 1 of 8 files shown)


test/e2e/framework/unmanaged.go
```diff
@@ -187,8 +187,7 @@ func (f *UnmanagedFramework) KubeClient(userAgent string) kubeclientset.Interfac
 }
 
 func (f *UnmanagedFramework) Client(userAgent string) genericclient.Client {
-	restclient.AddUserAgent(f.Config, userAgent)
-	return genericclient.NewForConfigOrDie(f.Config)
+	return genericclient.NewForConfigOrDieWithUserAgent(f.Config)
```
yamt: you forgot to pass userAgent?

yamt:
lgtm

marun:
Rebased

pkg/controller/sync/version/manager.go
```diff
@@ -301,7 +304,9 @@ func (m *VersionManager) writeVersion(qualifiedName util.QualifiedName) util.Rec
 	creationNeeded := len(metaAccessor.GetResourceVersion()) == 0
 
 	if creationNeeded {
-		createdObj, err := m.adapter.Create(obj)
+		// Make a copy to avoid mutating the map outside of a lock.
+		createdObj := obj.DeepCopyObject()
```
irfanurrehman: I understand the change on line 226 as `List()` is used inside a poll loop and it could effectively create a new `List Object` on each call to `adapter.List()` in earlier code, which could be a lot.
However I do not understand the change from earlier `adapter.Create()` and `adapter.Update()`. A copy of the object could be made inside those earlier methods, rather then exposing the client outside.
Is this done to keep `client` only at one place (as in `VersonManager` only now), because the fix was needed for `List()` anyways ?

pkg/controller/sync/version/cluster.go
```diff
@@ -17,35 +17,28 @@ limitations under the License.
...
 
-type clusterVersionAdapter struct {
-	client genericclient.Client
-}
+type clusterVersionAdapter struct{}
 
-func newClusterVersionAdapter(client genericclient.Client) VersionAdapter {
-	return &clusterVersionAdapter{client}
+func (*clusterVersionAdapter) TypeName() string {
+	return "ClusterPropagatedVersion"
```
irfanurrehman: Does it make sense to use `reflect.TypeOf(&fedv1a1.ClusterPropagatedVersion{}).Name()` rather then hard coding the string here.

irfanurrehman:
@marun thanks for this.
Looks ok to me. Only few queries.

pkg/controller/sync/version/cluster.go
marun: Your suggestion is functionally equivalent, but I think it's out of scope here.  I'm not actually changing this particular code, though the diff may appear otherwise due to all the deletions.  I'm also not inclined to optimize the version stuff since there's a good chance it will be changing in the near future (e.g. to support selector-based overrides).

pkg/controller/sync/version/manager.go
marun: The version adapters previously internalized the client because they were calling entirely different methods internally.  For example, the adapter for `PropagatedVersion` was calling `PropagatedVersion.Get` method which would only accept the specific object type.  I don't think there's a good reason to internalize the client now that the only thing that differs about the client is the object type (the client just accepts `pkgruntime.Object`).  Given that there is only a single caller for both methods, removing the encapsulation actually requires less code (since only one implementation is required rather than 2) and exposes the copy requirement such that its purpose can be documented in relation to the map storage.

irfanurrehman: Thanks for the explanation.

pkg/controller/sync/version/cluster.go
irfanurrehman: Ok.

irfanurrehman:
This looks ok to me; If no more review is expected I will go ahead and lgtm this.

PR #626, opened 2019-03-05
Fix races in replica scheduling code

irfanurrehman:
I have chosen to use fine grained locking, which seems more suitable given the small number of values stored in the maps/sets in question.

cc @marun @shashidharatd

### DIFF
--- a/pkg/controller/util/safe_map.go
+++ b/pkg/controller/util/safe_map.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"sync"
+)
+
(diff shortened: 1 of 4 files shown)


irfanurrehman:
> I think you've covered the case where multiple threads try to use maps in these controllers. Is there a guarantee that reconcile will never be called at the same time for the same key (for any operation), such that only locking the maps (and sets) is sufficient to prevent leaking resources (i.e. multiple threads starting the same plugin but only the last plugin to be started is recorded)?

Yes, you have a very valid point. I have updated the code to avoid such a situation. PTAL!

pkg/controller/util/safe_map.go
```diff
@@ -0,0 +1,61 @@
...
+limitations under the License.
+*/
+
+package util
+
+import (
+	"sync"
+)
+
+type SafeMap struct {
```
shashidharatd: Is there any particular reason we could not use thread-safe concurrent map type from golang library https://golang.org/pkg/sync/#Map

irfanurrehman: Because of the recommendations there.. :-). I had a look at the implementation too and found it doing lot of additional stuff then that was needed here, especially the `Range()` method (which I could accomplish for now implementing `GetAll()` here).

marun:
@irfanurrehman Please enable race detection for the managed e2e suite as part of this PR.

irfanurrehman:
> @irfanurrehman Please enable race detection for the managed e2e suite as part of this PR.

@marun Not sure if you got a notification. I did rebase and push the commit to enable checking race in CI today morning.

marun:
@irfanurrehman  I'm a bit behind in reviews but will make time tomorrow.

marun:
Much appreciated, thank you!

/lgtm

PR #628, opened 2019-03-06
support federated resource CRD sharing for namespaced installation

xunpan:
Keep the default command behavior same as before for cluster federation deployment.

For namespaced deployment,
when enable a federated resource, if the federaed* CRD exists, reuse it. If the CRD does not exist, create it. "--error-on-existing=false" need to be used.
when disable a federated resource,  CRD should not be deleted because other deployment may use it. "--delete-crd=false" need to be used.

### DIFF
--- a/pkg/kubefed2/enable/enable.go
+++ b/pkg/kubefed2/enable/enable.go
@@ -30,6 +30,7 @@ import (
 
 	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apiextv1b1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


marun:
@xunpan What use case are you trying to solve with `--delete-crd`?  And how does it relate to namespaced deployment?

xunpan:
@marun cluster-scoped fedeartion deployment need to delete CRDs when it is deleted.

marun:
As per a conversation on slack, I think deletion of a namespaced federation without intending to delete any CRDs (because other deployments may exist in the cluster) is already possible by deleting the federation namespace (which will also remove the FTCs for the deployment).  I don't think additional support in kubefed, as proposed in this PR, is suggested.

xunpan:
agreed to not maintain `disable`. 
As discussed, we need to make further enhance in the `disable` operation to avoid using it in namespaced federation. Because it will delete CRDs and break other namespaced federation deployment.

marun:
I think you're on the right track - there needs to be a change to the `enable` command to enable federation of a type that has previously been enabled for a different deployment in the same cluster.  I'm not sure it makes sense to reuse the `--error-on-existing` flag from the `join` command, though.

I think an `enable` command updated to be work with existing federated type crds would have the following characteristics:

 - attempt to retrieve a crd for the federated type
 - if a crd for the federated type does not exist, create it
  - if a crd exists, check that its contents against what would be generated
    - if the contents match, use it without error
    - if the contents do not match
       - if `--update-existing=true` (the default), update the crd and use it
       - if `--update-existing=false`, throw an error

Alternately an existing crd could always be updated, avoiding the requirement for a `--use-existing` flag.  Given the guarantee of backwards compatibility provided by kubernetes (and the backwards compatibility guarantee that federation will need to provide from beta forward), this may be entirely reasonable.

xunpan:
I'd like to make it update by default without option control. We do not need to make it more flexbile unless the option is useful.

marun:
@xunpan We may want to revisit this decision in the future, but I think it's fine for now.

pkg/kubefed2/enable/enable.go
```diff
@@ -228,11 +229,27 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
...
+		_, err = crdClient.CustomResourceDefinitions().Update(existingCRD)
+		if err == nil {
+			write(fmt.Sprintf("customresourcedefinition.apiextensions.k8s.io/%s updated\n", resources.CRD.Name))
+		} else {
+			return errors.Wrapf(err, "Error creating CRD %q", resources.CRD.Name)
+		}
+
+	} else if apierrors.IsNotFound(err) {
+		_, err = crdClient.CustomResourceDefinitions().Create(resources.CRD)
+		if err == nil {
```
marun: Minimize indentation.

xunpan:
CI fails more than once in 
```
Mar 12 02:28:53.629: INFO: Timed out waiting to become leader, err: timed out waiting for the condition
panic: 
Your test failed.
```

I'm checking

pkg/kubefed2/enable/enable.go
```diff
@@ -234,11 +235,24 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
...
+	existingCRD, err := crdClient.CustomResourceDefinitions().Get(resources.CRD.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		_, err = crdClient.CustomResourceDefinitions().Create(resources.CRD)
+		if err != nil {
+			return errors.Wrapf(err, "Error creating CRD %q", resources.CRD.Name)
+		}
+		write(fmt.Sprintf("customresourcedefinition.apiextensions.k8s.io/%s created\n", resources.CRD.Name))
+	} else if err != nil {
+		return errors.Wrapf(err, "Error getting CRD %q", resources.CRD.Name)
+	} else {
```
marun: Please avoid using `else` here and the unnecessary indentation it requires.

xunpan: I think it needed. Or else, we will I think it needed. Or else, after creating CRD, we will update again. It is not good to me from logic.

marun: Ah, my mistake.

marun:
This lgtm, but #645 needs to be merged and rebased on before merge.

marun:
Please rebase

xunpan:
done

xunpan:
as there is no code change after rebase, just mark it
/lgtm

marun:
Thank you!

/lgtm
/approve

PR #617, opened 2019-02-28
WIP Multiple limited-scope federations should be deployed

xunpan:
Install multiple namespaced fedeartion is failed. It need to be fixed to make feature 

- [x] multiple namespaced federation deployment
  - [x] kubefedv2 should work correctly with global CRD Federated* and ClusterRole/ClusterRoleBindings
  - [x] scripts should work for deploying and deleting limited-scope federations
- [x] helm chart support to install multiple namespaced federation deployment
- [x] controller log issues
- [ ] more test

Fixes #557

PR #629, opened 2019-03-06
support cluster RBAC for namespaced installation

xunpan:
In namespaced federation, make RBAC for clusterrolebinding and clusterrole name contain namespace
So that each namespaced federation has its owned RBAC.

E.g. after installation of namespaced federation in `foo` namespace.
```
$ kubectl get clusterrole  | grep fed 
federation-controller-manager:foo:healthcheck-cluster1-cluster1        15s
$kubectl get clusterrolebindings  | grep fed
federation-controller-manager:foo:healthcheck-cluster1-cluster1   23s
```

This is better than sharing clusterrole and clusterolebindings for all namespaced federation as I implemented in https://github.com/kubernetes-sigs/federation-v2/pull/617 after considering marun's advise.

This method is simple to implement and easy to maintain.

### DIFF
--- a/pkg/kubefed2/join.go
+++ b/pkg/kubefed2/join.go
@@ -778,7 +778,7 @@ func createHealthCheckClusterRoleAndBinding(clientset kubeclient.Interface, saNa
 		return nil
 	}
 
-	roleName := util.HealthCheckRoleName(saName)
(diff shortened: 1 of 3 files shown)


marun:
Thank you!

/approve
/lgtm

PR #630, opened 2019-03-06
support multiple namespaced federation with script

xunpan:
- cluster scope resource CRDs are created on demand and not deleted in delete-deployment operation to avoid breaking other namespaced federations
 - an option to clean all cluster scope resources.

### DIFF
--- a/scripts/delete-federation.sh
+++ b/scripts/delete-federation.sh
@@ -24,8 +24,10 @@ set -o pipefail
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 
 function delete-script-deployment() {
-  # Remove cluster registry CRD
-  ${KCD} -f vendor/k8s.io/cluster-registry/cluster-registry-crd.yaml
+  if [[ ! "${NAMESPACED}" || "${DELETE_CLUSTER_RESOURCE}" ]]; then
+      # Remove cluster registry CRD
+      ${KCD} -f vendor/k8s.io/cluster-registry/cluster-registry-crd.yaml
+  fi
 
   # Remove public namespace
   if [[ ! "${NAMESPACED}" ]]; then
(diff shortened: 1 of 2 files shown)


scripts/delete-federation.sh
```diff
@@ -34,13 +45,20 @@ function delete-script-deployment() {
...
   done
 
   # Remove federation CRDs, namespace, RBAC and deployment resources.
   if [[ ! "${USE_LATEST}" ]]; then
     if [[ "${NAMESPACED}" ]]; then
-      ${KCD} -n "${NS}" -f hack/install-namespaced.yaml
+      if [[ "${DELETE_CLUSTER_RESOURCE}" ]]; then
+        ${KCD} -n "${NS}" -f hack/install-namespaced.yaml
+      else
+        ${KCD} -n "${NS}" role federation-role
```
marun: (No action required) This seems like it could become a maintenance hassle.  I guess that's a vote in favor of deprecating script deployment.

scripts/delete-federation.sh
```diff
@@ -24,23 +24,38 @@ set -o pipefail
...
+        ${KCD} -n "${NS}" rolebinding federation-rolebinding
+        ${KCD} -n "${NS}" service federation-controller-manager-service
+        ${KCD} -n "${NS}" statefulset federation-controller-manager
```
marun: This is now a deployment thanks to #657

scripts/delete-federation.sh
```diff
@@ -24,23 +24,38 @@ set -o pipefail
...
+  fi
 
   # Remove federation CRDs, namespace, RBAC and deployment resources.
   if [[ ! "${USE_LATEST}" ]]; then
     if [[ "${NAMESPACED}" ]]; then
-      ${KCD} -n "${NS}" -f hack/install-namespaced.yaml
+      if [[ "${DELETE_CLUSTER_RESOURCE}" ]]; then
+        ${KCD} -n "${NS}" -f hack/install-namespaced.yaml
+      else
+        ${KCD} -n "${NS}" role federation-role
```
marun: Is it necessary to delete these resources explicitly vs deleting the containing namespace?

xunpan:
When a namespace is deleted, resources in the namespace are deleted automatically. So:
1. yes. the explicitly deletion is not needed if the namespace is deleted.
2. I also fix a nother unnecessary deletion of namespaced resource with a new commit.

@marun

scripts/delete-federation.sh
```diff
@@ -24,23 +24,33 @@ set -o pipefail
...
 
   # Disable federation of all types
-  for ftc in $(kubectl get federatedtypeconfig -n "${NS}" -o=jsonpath={.items..metadata.name}); do
-    ./bin/kubefed2 disable "${ftc}" --delete-from-api --federation-namespace="${NS}"
-  done
+  if [[ ! "${NAMESPACED}" ]]; then
+    for ftc in $(kubectl get federatedtypeconfig -n "${NS}" -o=jsonpath={.items..metadata.name}); do
+      ./bin/kubefed2 disable "${ftc}" --delete-from-api --federation-namespace="${NS}"
+    done
+  else
```
marun: (No action required) else -> elif (and save some indentation)

marun:
Thank you!

/lgtm
/approve

PR #635, opened 2019-03-07
Handle events with DeletedFinalStateUnknown

yamt:
The following handlers need the same change but it seems that
they are currently unused:
	NewTriggerOnMetaAndSpecChanges
	NewTriggerOnMetaAndFieldChanges

### DIFF
--- a/pkg/controller/util/handlers.go
+++ b/pkg/controller/util/handlers.go
@@ -28,6 +28,13 @@ import (
 func NewTriggerOnAllChanges(triggerFunc func(pkgruntime.Object)) *cache.ResourceEventHandlerFuncs {
 	return &cache.ResourceEventHandlerFuncs{
 		DeleteFunc: func(old interface{}) {
+			if deleted, ok := old.(cache.DeletedFinalStateUnknown); ok {
+				// This object might be stale but ok for our current usage.
+				old = deleted.Obj
+				if old == nil {
+					return
+				}
+			}
 			oldObj := old.(pkgruntime.Object)
 			triggerFunc(oldObj)
 		},
--- a/pkg/controller/util/handlers.go
+++ b/pkg/controller/util/handlers.go
@@ -19,9 +19,6 @@ package util
 import (
 	"reflect"
 
-	"github.com/pkg/errors"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
 )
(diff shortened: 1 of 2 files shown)


pkg/controller/util/handlers.go
```diff
@@ -31,6 +31,10 @@ import (
 func NewTriggerOnAllChanges(triggerFunc func(pkgruntime.Object)) *cache.ResourceEventHandlerFuncs {
 	return &cache.ResourceEventHandlerFuncs{
 		DeleteFunc: func(old interface{}) {
+			if deleted, ok := old.(cache.DeletedFinalStateUnknown); ok {
+				// This object might be stale but ok for our current usage
+				old = deleted.Obj
+			}
 			oldObj := old.(pkgruntime.Object)
```
marun: Ugh, this code is terrible.  Would you be up for ensuring that all handlers only call `triggerFunc` for a non-nil (successfully cast) `pkgruntime.Object` in a subsequent PR?  All of the handlers passed as triggerFunc are assuming this.

marun:
Please remove NewTriggerOnMetaAndSpecChanges and NewTriggerOnMetaAndFieldChanges to avoid accidental use of broken behavior.

pkg/controller/util/handlers.go
yamt: can you elaborate what you mean?
i know the assumption and it's why i made this change.

marun: The deletion handler should check that the object provided is non-nil before calling triggerFunc:

```golang
if oldObj != nil {
   triggerFunc(oldObj)
}
```

yamt: i don't think it can be nil.

marun: Please review the delta fifo code responsible for generating events of type `DeletedFinalStateUnknown`:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L474

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L502

In both cases, the deleted object may not be present and may be passed as nil in the event.

yamt: in both cases, i don't see how it happens without bugs.
but i guess it's better to be safe than sorry. i'll add a nil check.

pkg/controller/util/handlers.go
```diff
@@ -100,10 +44,7 @@ func NewTriggerOnMetaAndFieldChanges(field string, triggerFunc func(pkgruntime.O
 		},
 		UpdateFunc: func(old, cur interface{}) {
 			curObj := cur.(pkgruntime.Object)
-			oldMeta := getFieldOrPanic(old, "ObjectMeta").(metav1.ObjectMeta)
```
marun: As per a previous comment, please remove `NewTriggerOnMetaAndSpecChanges`
 and `NewTriggerOnMetaAndFieldChanges` rather than updating them.  There are no non-test clients, so continuing to maintain the code has zero value.

yamt: i have removed them.
i guess it's just `diff -p` being confused by the removal.

marun: Ah, my apologies.  As much as I'd like to blame the tool, I think I'm the one who's confused.

marun:
Trying again...

/approve

PR #637, opened 2019-03-08
Create the registry namespace from the clusterregistry chart

andersjanmyr:
When the cluster registry is enabled, it should also create the namespace.
In order to have access to the `registryNamespace` property in both charts it needs to be `.global`.


charts/federation-v2/charts/clusterregistry/templates/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Values.global.registryNamespace }}
```
xunpan: For namespaced deployment, there is no need to create this namespace because cluster registry are in the same namespace with federation

marun:
Please rebase to pick up the CI fix.

marun:
@xunpan Are changes still required before this merges?

andersjanmyr:
@marun The comment above is valid. I haven't found a solution that will cover both use cases. Helm is not very nice to use when we need conditional logic. I guess we could use a pre-install hook, but it's kind of ugly. Maybe, I should just close the ticket?

marun:
@andersjanmyr I'm not a helm expert, I'll defer to @xunpan.

@xunpan Can you see a way to achieve the goals of this PR without requiring a pre-install hook?

marun:
@xunpan Should this be closed?

xunpan:
I agree to close this ticket. @andersjanmyr 's comments makes sense.

Registry namespace is expected to be shared among several deployment in some use cases. In such case, it cannot easily managed by helm. pre-hook makes it a little heavy.

FYI: we plan to make registry namespace into federation by default. So, this creation task is not used frequently.

PR #639, opened 2019-03-11
Refactor makefile and reuse docker image build steps from imagebuild.sh

shashidharatd:
This PR does following things. Although none are critical and is done for easier maintenance.

- Makefile was using different steps to build image. It is unified with `./scripts/imagebuild.sh` now.
- ~~Building binary image inside docker was unnecessary overhead as the binaries are built from CI in fresh environment every-time. In local development environment it takes additional time to build. Hence removed.~~
- Fix few go vet issues in e2e test code.
- Simplified Makefile targets for vet, fmt & test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ ifeq ($(DIFF), 1)
 endif
 
 ifneq ($(VERBOSE),)
-VERBOSE_FLAG = -v
+    VERBOSE_FLAG = -v
 endif
 BUILDMNT = /go/src/$(GOTARGET)
 BUILD_IMAGE ?= golang:1.11.2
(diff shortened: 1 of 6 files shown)


marun:
@shashidharatd Isn't building in docker desirable for someone on osx?

shashidharatd:
> @shashidharatd Isn't building in docker desirable for someone on osx?

oops, i became over enthusiastic to reduce the time building binaries in my local environment and forgot the Non-Linux users. I will rollback the changes.

Makefile
```diff
@@ -104,15 +94,8 @@ kubefed2: $(BIN_DIR)
...
-	$(DOCKER) push $(REGISTRY)/$(TARGET):$(GIT_VERSION)
-	if git describe --tags --exact-match >/dev/null 2>&1; \
-	then \
-		$(DOCKER) tag $(REGISTRY)/$(TARGET):$(GIT_VERSION) $(REGISTRY)/$(TARGET):$(GIT_TAG); \
-		$(DOCKER) push $(REGISTRY)/$(TARGET):$(GIT_TAG); \
-		$(DOCKER) tag $(REGISTRY)/$(TARGET):$(GIT_VERSION) $(REGISTRY)/$(TARGET):latest; \
-		$(DOCKER) push $(REGISTRY)/$(TARGET):latest; \
-	fi
+push: container
+	bash ./scripts/imagebuild.sh push
```
irfanurrehman: I think it will be useful to be able to give your own image name [ IMAGE=xxxx ]. Most of the people invoking this will not push to the default (official) image location.

scripts/imagebuild.sh
```diff
@@ -21,43 +21,56 @@ set -o pipefail
...
 dockerfile_dir="${base_dir}/images/federation-v2"
 
+SUBCMD=
+if [[ $# == 1 ]]; then
+    SUBCMD="$1"
+fi
+
+REGISTRY=${REGISTRY:-quay.io}
+REPO=${REPO:-kubernetes-multicluster}
+IMAGE=${REGISTRY}/${REPO}/federation-v2
```
irfanurrehman: If this is what going to be triggered from make and/or available choice for user. this should be overrideable by the user.

irfanurrehman:
@shashidharatd my take on the Makefile is for it to be the mechanism of choice for users and then make based commands to be adopted in CI including building and pushing the image (explicitly done by imagebuild.sh before this PR). 

I would have rather kept the mechanism of building and pushing as it is in Makefile and call `make build` and/or `make push` from `imagebuild.sh`. `imagebuild.sh` would remain as the CI build script only then.

Makefile
shashidharatd: @irfanurrehman, it is possible, like below:
```
$> REGISTRY=quay.io REPO=<user-account> make push
```

scripts/imagebuild.sh
shashidharatd: yes they are overridable, please check above comment.

shashidharatd:
Probably, lets hold this pr until, @yamt, works on supporting OSX in pr https://github.com/kubernetes-sigs/federation-v2/pull/665

Makefile
irfanurrehman: Probably I wasn't clear the first time. I meant a canonical name which can be documented also and `IMAGE=<repo>/<username>/<image-name>:<tag>` as one override option seems to fit that better. Say the user now wants to override tag and actual image name not just `REGISTRY` and `REPO`, it wont be possible. This was missing from earlier Makefile too and its probably better to have it.

scripts/imagebuild.sh
irfanurrehman: As above.

irfanurrehman:
@shashidharatd Check #680 also please.

Makefile
pmorie: Agree with @irfanurrehman. Coincidentally, I just went through implementing a script that changed repo and registry, and then needed to change the image name.

shashidharatd:
No longer required, hence closing.

Issue #642, opened 2019-03-12
CI failed in Leader Elector


xunpan:
**Which test(s) are failing**:
`Leader Elector`

**Since when has it been failing**:
Since today

**Reason for failure**:
```
Mar 12 02:28:53.629: INFO: Timed out waiting to become leader, err: timed out waiting for the condition
panic: 
Your test failed.
```
**Anything else we need to know**:
The case is not synchronized, so it is not to be executed independently.

/kind failing-test

shashidharatd:
@xunpan, Can you give more context.
Which type of test are you executing. unit/e2e test? managed/unmanaged?
what is the command you used to execute test?

Note: It has been passing in CI.

shashidharatd:
@xunpan, never mind. here is one more instance of failed CI log https://travis-ci.org/kubernetes-sigs/federation-v2/builds/505079399?utm_source=github_status&utm_medium=notification

PR #645, opened 2019-03-12
Fix leader election e2e test

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/federation-v2/issues/642

It is observed that leader election e2e test is flaky due to multiple go routines involved in the test, which results in timeout issue. This pr introduces the synchronisation between go routine methods in the test.

It is also observed that the leader election library from client-go is not concurrent safe which results in error like below:
```
WARNING: DATA RACE
Read at 0x00c000836148 by goroutine 204:
  github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).IsLeader()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:195 +0x7a
  github.com/kubernetes-sigs/federation-v2/test/e2e.waitToBecomeLeader.func1()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/leaderelection.go:109 +0x41
  github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/apimachinery/pkg/util/wait.pollImmediateInternal()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:245 +0x38
  github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/apimachinery/pkg/util/wait.PollImmediate()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:241 +0x5e
  github.com/kubernetes-sigs/federation-v2/test/e2e.waitToBecomeLeader()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/leaderelection.go:108 +0xb1
  github.com/kubernetes-sigs/federation-v2/test/e2e.glob..func8.2.4()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/test/e2e/leaderelection.go:92 +0x89

Previous write at 0x00c000836148 by goroutine 24:
  github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).tryAcquireOrRenew()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:284 +0xb7b
  github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/client-go/tools/leaderelection.(*LeaderElector).renew.func1.1.1()
      /home/s70553/gopath/src/github.com/kubernetes-sigs/federation-v2/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go:231 +0x62
```
So leader election test is run with race detector disabled. The concurrency issue, only affects the tests and not the actual functionality of leader election. Also appreciate, any alternate suggestions to resolve this issue.

/cc @xunpan @marun

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -24,7 +24,6 @@ import (
 	"os"
 
 	"github.com/golang/glog"
-	configlib "github.com/kubernetes-sigs/kubebuilder/pkg/config"
 	"github.com/kubernetes-sigs/kubebuilder/pkg/install"
 	"github.com/kubernetes-sigs/kubebuilder/pkg/signals"
 	"github.com/spf13/cobra"
@@ -33,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/apiserver/pkg/util/logs"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/kubernetes-sigs/federation-v2/cmd/controller-manager/app/leaderelection"
 	"github.com/kubernetes-sigs/federation-v2/cmd/controller-manager/app/options"
@@ -45,6 +45,10 @@ import (
 	"github.com/kubernetes-sigs/federation-v2/pkg/features"
 	"github.com/kubernetes-sigs/federation-v2/pkg/inject"
 	"github.com/kubernetes-sigs/federation-v2/pkg/version"
+)
+
+var (
+	kubeconfig, masterURL string
 )
 
 // NewControllerManagerCommand creates a *cobra.Command object with default parameters
@@ -76,6 +80,8 @@ member clusters and does the necessary reconciliation`,
 
 	opts.AddFlags(cmd.Flags())
 	cmd.Flags().BoolVar(&verFlag, "version", false, "Prints the Version info of controller-manager")
+	cmd.Flags().StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	cmd.Flags().StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


marun:
@shashidharatd Can you link to a gist of the full error?  My recent experience suggests that we should not be ignoring races.  The library may not be concurrency-safe, but our tests need to be.

shashidharatd:
@marun, here is the [log](https://gist.github.com/shashidharatd/43eccfe5b1f97c9034d2032c22b0c9da), when the test case is run with race detector.

marun:
@shashidharatd I think the fix may be to run one or both controller managers out-of-process for the purposes of the test.  As you say, this is a test-only problem, but I would rather not have flakey tests and the race detector is a good indication of a problem rather than something to ignore.

shashidharatd:
> @shashidharatd I think the fix may be to run one or both controller managers out-of-process for the purposes of the test. As you say, this is a test-only problem, but I would rather not have flakey tests and the race detector is a good indication of a problem rather than something to ignore.

even if we run the controller manager out-of-process, it is difficult to detect which is master. we might have to rely on logs to identify the master. I am inclined to skip this test in CI. Is that fine?

marun:
If leader election is a feature worth having, it should be tested.  Start an out-of-process controller manager and watch its logs until it has started as the leader.  Then start an in-process manager and kill the first.

xunpan:
I guess a simple way to handle this race issue is to wait a second after secondary leader run.
```
    secondaryLeaderElector.Run(SecondaryContext)
+ // wait second leader elected
+ time.Sleep(1 * time.Second)
    waitToBecomeLeader(tl, secondaryLeaderElector) 
```
`time.Sleep` sleep enough time to make sure leader is elected. After the time, if leader is not elected, it has issues. So, keep `-race` should be fine.

shashidharatd:
@marun, ptal. Thanks !

marun:
Awesome, thank you for fixing the test! :)

/lgtm

PR #646, opened 2019-03-12
Update kube-openapi to include compatibility fixes required by openshift

marun:
[k8s.io/kube-openapi](https://github.com/kubernetes/kube-openapi) has seen recent fixes to ensure compliance with the openapi standard.  Recent builds of openshift return openapi that reflects these fixes but that break our older vendored version of kube-openapi.  This PR updates the vendored version of kube-openapi to ensure compatibility with both openshift and future versions of kubernetes.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1003,11 +1003,12 @@
   version = "kubernetes-1.12.4"
 
 [[projects]]
-  digest = "1:e0d6dcb28c42a53c7243bb6380badd17f92fbd8488a075a07e984f91a07c0d23"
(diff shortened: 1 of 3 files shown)


pmorie:
This is for any 1.13 kube, not just openshift afaik.

pmorie:
This change is necessary to work against any kube that has https://github.com/kubernetes/kubernetes/pull/71192 , so k/k master/1.14

marun:
CI is broken pending #645

marun:
Rebased

PR #648, opened 2019-03-14
Configmap configures federation controller

### LINKED ISSUE #114: Component configuration for controller
Follow-on to #101; this is to add (where it makes sense) additional configuration parameters beyond basic feature gating. For example: the push reconciler is enabled/disabled by feature gate, but the resync duration for the push reconciler would be configured by a distinct component configuration API field mapped to a CLI flag.

xunpan:
A `federation-v2` configmap is used to configure federation controller. It will override command line options if exists.

After this is merged, I will:
- [ ] make installer script uses configmap for configuration
- [ ] make helm chart uses configmap for configuration
- [ ] make `kubectl disable` reject in namespaced federation.

Fixes #114

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,0 +1,22 @@
+kind: ConfigMap 
+apiVersion: v1 
+metadata:
+  name: federation-v2
+data:
+  install-crds: "false"
+  federation-namespace: federation-system
+  registry-namespace: kube-multicluster-public
+  cluster-available-delay: "20s"
+  cluster-unavailable-delay: "60s"
+  limited-scope: "false"
+  uster-monitor-period: "40s"
+  leader-elect: "true"
+  leader-elect-lease-duration: "15s"
+  leader-elect-renew-deadline: "10s"
+  leader-elect-retry-period: "5s"
+  leader-elect-resource-lock: configmaps
+  feature-gates: |
+    PushReconciler=true
+    SchedulerPreferences=true
+    CrossClusterServiceDiscovery=true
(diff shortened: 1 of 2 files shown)


shashidharatd:
@Xunpan, It is better we follow k8s community best practices for component configuration. Here is a relevant doc which discusses the issue and solution for component configuration https://docs.google.com/document/d/1arP4T9Qkp2SovlJZ_y790sBeiWXDO6SG10pZ_UUU-Lc/edit

xunpan:
@shashidharatd 
I had a simple sync-up with marun before. This configmap implementation is a first step for component configuration and get feedback further. @marun

Yes. I will check the k8s best practises for future enhancement.

marun:
@shashidharatd I suggested a starting point of configmap, but as per the linked document there seem to be many things we need to consider before moving forward.

config/config.yaml
```diff
@@ -0,0 +1,22 @@
+kind: ConfigMap
```
shashidharatd: Shall we define a new API in federation for Controller-Manager configuration? validation and defaulting would be difficult in the current form (as a configmap).

xunpan: I prefer to have new API for configuration management. I add comments in issue ticket for discussion: https://github.com/kubernetes-sigs/federation-v2/issues/114#issuecomment-475083315

xunpan:
close this ticket because we've decided to use FederatedConfig for configuration https://github.com/kubernetes-sigs/federation-v2/pull/685

PR #649, opened 2019-03-14
User guide doesn't cover `unjoin`

### LINKED ISSUE #638: User guide doesn't cover `unjoin`
Pretty much what it sounds like, the userguide doesn't cover `kubefed2 unjoin` and should.

poothia:
User guide doesn't cover `unjoin`
fixes #638

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -13,6 +13,7 @@
   - [Operations](#operations)
     - [Join Clusters](#join-clusters)
     - [Check Status of Joined Clusters](#check-status-of-joined-clusters)
+    - [Unjoin Clusters](#unjoin-clusters)
   - [Enabling federation of an API type](#enabling-federation-of-an-api-type)
   - [Disabling federation of an API type](#disabling-federation-of-an-api-type)
   - [Example](#example)
@@ -174,6 +175,14 @@ Status:
     Status:                True
     Type:                  Ready
 ```
+### Unjoin Clusters
+
+If required, federation allows you to unjoin clusters using `kubefed2` tool.
+
+```bash
+./bin/kubefed2 unjoin cluster2 --cluster-context cluster2 --host-cluster-context cluster1 --remove-from-registry --v=2
+```
+You can repeat these steps to unjoin any additional clusters.
 
 ## Enabling federation of an API type
 


docs/userguide.md
```diff
@@ -174,6 +175,13 @@ Status:
     Status:                True
     Type:                  Ready
 ```
+### Unjoin Clusters
+
+If required, federation allows you to unjoin clusters using `kubefed2` tool.
+
+```bash
+kubefed unjoin gondor --host-cluster-context=rivendell
```
xunpan: could you use the same cluster name in `join` case above? It makes consistency to users for reading.

docs/userguide.md
```diff
@@ -174,6 +175,13 @@ Status:
     Status:                True
     Type:                  Ready
 ```
+### Unjoin Clusters
+
+If required, federation allows you to unjoin clusters using `kubefed2` tool.
```
gyliu513: How about clarify the what the command can help like `The following command un-joins cluster2 from federation controller`?

docs/userguide.md
```diff
@@ -174,6 +175,13 @@ Status:
...
+
+```bash
+kubefed2 unjoin cluster2 --host-cluster-context cluster1
```
gyliu513: How about add some comments as `Join Clusters`?

```
You can repeat above step to un-join any additional clusters.
```

gyliu513: I'd like you use a full command here as join including `cluster-context`

gyliu513: How about adding a NOTE here as `join`?

```
NOTE: cluster-context will default to use the un-joining cluster name if not specified.
```

docs/userguide.md
poothia: Can you please elaborate.
I am trying to match the format of the whole document.
In the above command it is mentioned `--host-cluster-context`, so I feel this is understood.

Thanks

docs/userguide.md
poothia: Sure I will add it

docs/userguide.md
gyliu513: Alright, how about use the following command instead?

```
./bin/kubefed2 unjoin cluster2 --cluster-context cluster2 --host-cluster-context cluster1 --remove-from-registry --v=2
```

poothia: Ya, I was about to comment the same.
Thank you for your inputs.

Will update it with a new patchset

poothia: Would love to get an opinion, whether it is required to add `--remove-from-registry` in the command

gyliu513: Since the `join` is including `add-to-registry` https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/userguide.md#join-clusters , so I'd prefer adding `remove-from-registry` for unjoin

poothia: Please review @gyliu513

docs/userguide.md
```diff
@@ -174,6 +175,13 @@ Status:
     Status:                True
     Type:                  Ready
 ```
+### Unjoin Clusters
+
+If required, federation allows you to unjoin clusters using `kubefed2` tool.
+
+```bash
+./bin/kubefed2 unjoin cluster2 --cluster-context cluster2 --host-cluster-context cluster1 --add-to-registry --v=2
```
gyliu513: s/--add-to-registry/--remove-from-registry

docs/userguide.md
```diff
@@ -174,6 +175,13 @@ Status:
...
+```
```
gyliu513: How about adding a NOTE here as `join`?

```
NOTE: cluster-context will default to use the un-joining cluster name if not specified.
```

docs/userguide.md
poothia: Sorry for that. Updated !

PR #650, opened 2019-03-14
Some OSX support

yamt:
This allows e2e run on OSX.


marun:
I'm ok with this merging, with the caveat that I don't think it makes sense for fedv2 to commit to supporting osx compatibility for the controller manager.

/approved

PR #651, opened 2019-03-14
Update diagrams after API unification

pmorie:
And make some minor changes to README text


marun:
Lgtm.  Will leave it to a non-rher to do the honors.

xunpan:
good diagram. It LGTM.
One question: where is the link to docs/images/propagation.png from the doc? I think we need such kind of description and link.

shashidharatd:
LGTM. Thanks @pmorie. You might have missed linking `docs/images/propagation.png`. is that the case?

pmorie:
> LGTM. Thanks @pmorie. You might have missed linking docs/images/propagation.png. is that the case?

@shashidharatd yep, I'm just adding it for now. I'll make a small doc PR that contextualizes it as a follow up.

shashidharatd:
/lgtm
Thanks @pmorie

Issue #565, opened 2019-01-29
User guide shoud state implicit GOPATH dependency


llarsson:
**What happened**:

Trying to run through the User Guide from some other directory than `${GOPATH}/src/github.com/kubernetes-sigs/federation-v2` fails while downloading binaries and cannot build federation using scripts.

Making sure to clone to the `${GOPATH}/src/github.com/kubernetes-sigs/federation-v2` directory works as intended.

**What you expected to happen**:

Since nothing was stated about which directory is OK to clone the repo to, I expected the User Guide to work.

**How to reproduce it (as minimally and precisely as possible)**:

 1. Check out the repo to a path under e.g. `/tmp`.
 2. Deploy some clusters.
 3. Try to run `./scripts/download-binaries.sh` and watch `kubebuilder` complain about not running in the right place.
 4. Try to run `./scripts/deploy-federation-latest.sh cluster2` and watch it fail while building Go code.

/kind bug

matti:
well, I also stumbled in to this - I wouldn't clone the repo, but helm install failed with

```
"federatedclusters.core.federation.k8s.io" already exists
```

marun:
@matti That error message indicates that installation had already been attemped on the target cluster.  Attempting re-installation will fail because the crds have already been installed.

matti:
@marun yeah I figured that out later when I actually understood what was happening.

aviweit:
I also suggest to denote GOPATH setting in userguide, as @llarsson mentions (even just as an aside comment). I encountered same issue yesterday when attempting to install federation v2.

Thanks.

marun:
I don't think the userguide should require mention GOPATH, so I'll take it as an action item to limit it to documenting deployment with helm.  Script-based deployment is likely to go away, and while it remains it is very much intended for developers only.

llarsson:
Good! As long as a user does not fail with their experiments because they need implicit knowledge of what language the software was written in and therefore know what steps to take to set up a development environment, I am satisfied. 😀 Because as a user, I do not typically expect that.

On March 14, 2019 6:30:08 PM GMT+01:00, Maru Newby <notifications@github.com> wrote:
>I don't think the userguide should require mention GOPATH, so I'll take
>it as an action item to limit it to documenting deployment with helm. 
>Script-based deployment is likely to go away, and while it remains it
>is very much intended for developers only.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
[...]

PR #655, opened 2019-03-15
User guide shoud state implicit GOPATH dependency

poothia:
#565 
Added some clarity in the userguide doc.


marun:
@poothia Did you read the comments on the issue you're proposing to resolve?  I think it would be preferable to limit the userguide to deployment with helm and use of released binaries.  Details of building from source would be limited to the devguide.

poothia:
Closing as per @marun comments.

PR #657, opened 2019-03-18
Replace statefulset with deployment for controller-manager.

shashidharatd:
This pr supersedes https://github.com/kubernetes-sigs/federation-v2/pull/404. uses the same commits and adds another commit to remove enabling/disabling leader-election and so leader election is always enabled.

/cc @marun @irfanurrehman


marun:
Thank you!

/lgtm

PR #660, opened 2019-03-18
Implement -o yaml for federate resource

irfanurrehman:
This PR is on top of #647 
#647 commits are not reviewable here.

To make it easier for review, the enhancements are implemented in a PR chain:
 - [x] Implements -o yaml option [this PR]
 - [x] Accept group qualified type name [#661]
 - [x] Usable without federation API, it needs a k8s cluster (if only yaml is needed)  [#662]
 - [x] Optional --enable-type to enable type also while federating resource  [#663]

cc @marun @xunpan

### DIFF
--- a/pkg/kubefed2/enable/enable.go
+++ b/pkg/kubefed2/enable/enable.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
@@ -346,25 +345,7 @@ func writeObjectToYAML(obj pkgruntime.Object, w io.Writer) error {
 		return err
 	}
 
-	// Convert to unstructured to filter status out of the output.  If
-	// status is included in the yaml, attempting to create it in a
-	// kube API will cause an error.
 	unstructuredObj := &unstructured.Unstructured{}
 	unstructured.UnstructuredJSONScheme.Decode(json, nil, unstructuredObj)
-	delete(unstructuredObj.Object, "status")
-	// Also remove unnecessary field
-	metadataMap := unstructuredObj.Object["metadata"].(map[string]interface{})
-	delete(metadataMap, "creationTimestamp")
-
-	updatedJSON, err := unstructuredObj.MarshalJSON()
-	if err != nil {
-		return err
-	}
-
-	data, err := yaml.JSONToYAML(updatedJSON)
-	if err != nil {
-		return err
-	}
-	_, err = w.Write(data)
-	return err
+	return util.WriteUnstructuredToYaml(unstructuredObj, w)
 }
--- a/pkg/kubefed2/federate/federate.go
+++ b/pkg/kubefed2/federate/federate.go
@@ -62,10 +62,13 @@ type federateResource struct {
 	typeName          string
 	resourceName      string
 	resourceNamespace string
+	output            string
+	outputYAML        bool
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.resourceNamespace, "namespace", "n", "default", "The namespace of the resource to federate.")
+	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 }
 
 // Complete ensures that options are valid.
@@ -84,6 +87,12 @@ func (j *federateResource) Complete(args []string) error {
 		// TODO: irfanurrehman: Can a target namespace be federated into another namespace?
 		glog.Infof("Resource to federate is a namespace. Given namespace will itself be the container for the federated namespace")
 		j.resourceNamespace = ""
+	}
+
+	if j.output == "yaml" {
+		j.outputYAML = true
+	} else if len(j.output) > 0 {
+		return errors.Errorf("Invalid value for --output: %s", j.output)
 	}
 
 	return nil
@@ -136,11 +145,28 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 		Name:      j.resourceName,
 	}
 
-	_, err = FederateResource(hostConfig, qualifiedTypeName, qualifiedResourceName, j.DryRun)
-	return err
+	resources, err := GetFedResources(hostConfig, qualifiedTypeName, qualifiedResourceName)
+	if err != nil {
+		return err
+	}
+
+	if j.outputYAML {
+		err := util.WriteUnstructuredToYaml(resources.FedResource, cmdOut)
+		if err != nil {
+			return errors.Wrap(err, "Failed to write federated resource to YAML")
+		}
+		return nil
+	}
+
+	return CreateFedResource(hostConfig, resources, j.DryRun)
 }
 
-func FederateResource(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName, dryrun bool) (*unstructured.Unstructured, error) {
+type fedResources struct {
+	typeConfig  typeconfig.Interface
+	FedResource *unstructured.Unstructured
+}
+
+func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName) (*fedResources, error) {
 	typeConfig, err := lookupTypeDetails(hostConfig, qualifiedTypeName)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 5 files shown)


marun:
@irfanurrehman Would it be possible to break this PR into component parts?  Some of them seem logically separable and smaller PRs are much easier to review.

irfanurrehman:
> @irfanurrehman Would it be possible to break this PR into component parts? Some of them seem logically separable and smaller PRs are much easier to review.

@marun I can do that. To make it easy for the review I did separated the components into individual commits anyways.

irfanurrehman:
@marun updated. PTAL!

marun:
@irfanurrehman Thank you for breaking up the changes.  I've taken the liberty of adding WIP to the title of this and subsequent PRs to ensure reviews won't accidentally merge the PRs out-of-order.

irfanurrehman:
> @irfanurrehman Thank you for breaking up the changes. I've taken the liberty of adding WIP to the title of this and subsequent PRs to ensure reviews won't accidentally merge the PRs out-of-order.

ACK.

pkg/kubefed2/enable/enable.go
```diff
@@ -346,25 +345,8 @@ func writeObjectToYAML(obj pkgruntime.Object, w io.Writer) error {
 		return err
 	}
 
-	// Convert to unstructured to filter status out of the output.  If
-	// status is included in the yaml, attempting to create it in a
-	// kube API will cause an error.
+	// Use unstructured to filter unwanted fields out of the output.
```
marun: nit: this comment doesn't seem relevant given that removal of unwanted fields is delegated to yaml conversion.

pkg/kubefed2/util/yaml_writer.go
```diff
@@ -0,0 +1,45 @@
...
+package util
+
+import (
+	"github.com/ghodss/yaml"
+	"io"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func WriteUnstructuredToYaml(unstructuredObj *unstructured.Unstructured, w io.Writer) error {
```
marun: I think it's potentially surprising that a reusable utility method intended to output yaml modifies its input.  That was fine when the code was internal to `enable`, but now that it's available for reuse I think either documentation or a DeepCopy of the input before modification is suggested.

pkg/kubefed2/federate/federate.go
```diff
@@ -156,6 +165,14 @@ func FederateResource(hostConfig *rest.Config, qualifiedTypeName, qualifiedName
 		return nil, errors.Wrapf(err, "Error getting %s from %s %q", typeConfig.GetFederatedType().Kind, typeConfig.GetTarget().Kind, qualifiedName)
 	}
 
+	if outputYAML {
+		err := util.WriteUnstructuredToYaml(fedResource, cmdOut)
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to write federated resource to YAML")
+		}
+		return fedResource, nil
```
marun: Why does it make sense to return the resource if it has been output to yaml?  Note how `enable` separated generation of the resource from its creation in the `Run` method.  That also avoids the problem of cluttering the method signature of `FederateResource` with 2 variables with a very narrow use.

pkg/kubefed2/federate/util.go
```diff
@@ -25,6 +25,7 @@ import (
 var systemMetadataFields = []string{"selfLink", "uid", "resourceVersion", "generation", "creationTimestamp", "deletionTimestamp", "deletionGracePeriodSeconds"}
 
 func RemoveUnwantedFields(resource *unstructured.Unstructured) {
+	unstructured.RemoveNestedField(resource.Object, "status")
```
marun: This looks like a bug discovered during implementation of yaml output, and accordingly seems deserving of its own commit.

pkg/kubefed2/util/yaml_writer.go
```diff
@@ -0,0 +1,45 @@
...
+	// If status is included in the yaml, attempting to create it in a
+	// kube API will cause an error.
+	delete(unstructuredObj.Object, "status")
```
marun: What's the difference between `unstructured.RemoveNestedField` and `delete`?

pkg/kubefed2/util/yaml_writer.go
```diff
@@ -0,0 +1,45 @@
...
+	// Also remove unnecessary field
```
marun: I know this is copy-pasta, but this comment seems redundant.

pkg/kubefed2/federate/federate.go
```diff
@@ -176,7 +193,7 @@ func lookupTypeDetails(config *rest.Config, qualifiedTypeName ctlutil.QualifiedN
 		return nil, errors.Wrapf(err, "Error retrieving API resource for FederatedTypeConfig %q", qualifiedTypeName)
 	}
 
-	glog.Infof("FederatedTypeConfig: %q found", qualifiedTypeName)
+	glog.V(2).Infof("FederatedTypeConfig: %q found", qualifiedTypeName)
```
marun: Please move these unrelated non-functional log changes to a separate commit.

pkg/kubefed2/enable/enable.go
irfanurrehman: Updated!

pkg/kubefed2/util/yaml_writer.go
irfanurrehman: Updated!

pkg/kubefed2/federate/federate.go
irfanurrehman: The return param enables simplify (reading the same resource again in) the federate e2e test. Agree that the actual resource need not be returned for the yaml case here, updated to nil.
I am not sure if this breaks separation between functional and test code. If it does, let me know and I will update accordingly. 

To answer your suggestion about separating the generate and create, `FederateResource()` is actually doing that, however I did choose to use the bigger unit to write test on rather then using multiple smaller ones.

pkg/kubefed2/federate/util.go
irfanurrehman: Put into another commit!

pkg/kubefed2/util/yaml_writer.go
irfanurrehman: The difference is suggested by the name `NestedField`. However if I got your point right, it makes sense to keep uniformity in usage across the code. Have updated to `RemoveNestedField`.

pkg/kubefed2/util/yaml_writer.go
irfanurrehman: Updated!

pkg/kubefed2/federate/federate.go
irfanurrehman: I think these changed the functionality, such that the yaml output in case of -o yaml becomes clean in the default verbosity case. I have anyways separated them into another commit.

irfanurrehman:
@marun Updated, PTAL!

pkg/kubefed2/federate/federate.go
```diff
@@ -136,11 +145,11 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 		Name:      j.resourceName,
 	}
 
-	_, err = FederateResource(hostConfig, qualifiedTypeName, qualifiedResourceName, j.DryRun)
+	_, err = FederateResource(cmdOut, hostConfig, qualifiedTypeName, qualifiedResourceName, j.outputYAML, j.DryRun)
 	return err
 }
 
-func FederateResource(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName, dryrun bool) (*unstructured.Unstructured, error) {
+func FederateResource(cmdOut io.Writer, hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName, outputYAML, dryrun bool) (*unstructured.Unstructured, error) {
```
marun: I still think this method should be decomposed for consistency with `enable`, such that generating the fed resource is one method call and creating it is another.  Fine to have a convenience method that calls generate and then enable, but I think generating yaml output should only be in the cli-invoked Run.

irfanurrehman: I was in the process of doing this change in `federate ns --content` already, because it works well for multiple resources. Nevertheless moved it here in the simpler form here now.

irfanurrehman:
@marun Updated, PTAL!

marun:
@irfanurrehman I tried this out locally and noticed the following:

 - federating a namespace would have the following output.  Would it make sense to change the log level to suppress non-error output if `-o=yaml`, rather than changing the log level by default?

I0329 10:42:07.462737    6687 federate.go:88] Resource to federate is a namespace. Given namespace will itself be the container for the federated namespace
 - The spec for a namespace includes a finalizer field, which I think is the result of controller action and should not be included.
 - Output of all types includes apiVersion and kind in its output. I recommend stripping those.
 - There was no addition to the changelog.
 - There is no doc change introducing the feature.

I apologize for being so sticky in my reviews, though.  I realize that I'm unnecessarily slowing things down given how many PRs are still to go.  My usual tendency is to fix things immediately, but I don't think that is required in a PR series where I can count on you to refine things.  I'm going to merge this one and you can fix the issues mentioned above in a subsequent PR.

/lgtm

PR #661, opened 2019-03-18
Accept group qualified type name for type in federate resource

irfanurrehman:
This implements:
 - [x] Accept group qualified type name in federate resource

cc @marun @xunpan

### DIFF
--- a/pkg/kubefed2/federate/federate.go
+++ b/pkg/kubefed2/federate/federate.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/rest"
@@ -52,8 +53,8 @@ var (
 		flag otherwise.`
 
 	federate_example = `
-		# Federate resource named "my-dep" in namespace "my-ns" of type identified by FederatedTypeConfig "deployment.apps"
-		kubefed2 federate deployment.apps my-dep -n "my-ns" --host-cluster-context=cluster1`
+		# Federate resource named "my-dep" in namespace "my-ns" of kubernetes type "deploy"
+		kubefed2 federate deploy my-dep -n "my-ns" --host-cluster-context=cluster1`
 	// TODO(irfanurrehman): implement —contents flag applicable to namespaces
 )
 
@@ -74,7 +75,7 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
 // Complete ensures that options are valid.
 func (j *federateResource) Complete(args []string) error {
 	if len(args) == 0 {
-		return errors.New("FEDERATED-TYPE-NAME is required")
+		return errors.New("TYPE-NAME is required")
 	}
 	j.typeName = args[0]
 
@@ -82,12 +83,6 @@ func (j *federateResource) Complete(args []string) error {
 		return errors.New("RESOURCE-NAME is required")
 	}
 	j.resourceName = args[1]
-
-	if j.typeName == ctlutil.NamespaceName {
-		// TODO: irfanurrehman: Can a target namespace be federated into another namespace?
-		glog.Infof("Resource to federate is a namespace. Given namespace will itself be the container for the federated namespace")
-		j.resourceNamespace = ""
-	}
 
 	if j.output == "yaml" {
 		j.outputYAML = true
@@ -104,7 +99,7 @@ func NewCmdFederateResource(cmdOut io.Writer, config util.FedConfig) *cobra.Comm
 	opts := &federateResource{}
 
 	cmd := &cobra.Command{
-		Use:     "federate FEDERATED-TYPE-NAME RESOURCE-NAME",
+		Use:     "federate TYPE-NAME RESOURCE-NAME",
 		Short:   "Federate creates a federated resource from a kubernetes resource",
 		Long:    federate_long,
 		Example: federate_example,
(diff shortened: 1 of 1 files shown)


marun:
Please ready this PR for review by rebasing, editing the command, and removing WIP.

irfanurrehman:
@marun rebased.
Did not get your meaning of `editing the top comment`?

marun:
> @marun rebased.
> Did not get your meaning of `editing the top comment`?

The first comment on this PR still says:

```
This is implemented on top of #660
Only the last commit is reviewable.
```

When you make a new patch in a series available, make sure to remove text like that.

pkg/kubefed2/federate/federate.go
```diff
@@ -188,24 +186,25 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", qualifiedTypeName)
+		return nil, errors.Wrap(err, "Failed to get federation client")
 	}
 
-	_, err = enable.LookupAPIResource(config, typeConfig.Name, typeConfig.APIVersion)
+	typeConfig := &fedv1a1.FederatedTypeConfig{}
+	err = client.Get(context.TODO(), typeConfig, typeName.Namespace, typeName.Name)
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error retrieving API resource for FederatedTypeConfig %q", qualifiedTypeName)
```
marun: Would it make sense to provide special handling for `IsNotFound` and give an indication of how to fix that?

pkg/kubefed2/federate/federate.go
```diff
@@ -166,12 +160,16 @@ type fedResources struct {
...
 }
 
-func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName) (*fedResources, error) {
+func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName, outputYAML bool) (*fedResources, error) {
 	typeConfig, err := lookupTypeDetails(hostConfig, qualifiedTypeName)
 	if err != nil {
 		return nil, err
 	}
 
+	if typeConfig.Spec.Target.Kind == ctlutil.NamespaceKind && !outputYAML {
```
marun: Would it make sense to move this to `CreateFedResource`?

pkg/kubefed2/federate/federate.go
irfanurrehman: Updated, I am not sure if the `recommendation` has the best language.
However I now intend to update this to `Try using --enable-type to also enable the type while federating the resource.` in #663.

pkg/kubefed2/federate/federate.go
irfanurrehman: Yes, it does. Updated!

irfanurrehman:
@marun appreciate you understanding that the feature is broken into multiple PRs mainly to aid review. Couple of improvements that you pointed out in the first PR were actually intended in subsequent PRs. 
Am listing some answers to the pointers you had [here](https://github.com/kubernetes-sigs/federation-v2/pull/660#issuecomment-478092485)

> @irfanurrehman I tried this out locally and noticed the following:
> 
> * federating a namespace would have the following output.  Would it make sense to change the log level to suppress non-error output if `-o=yaml`, rather than statically setting a lower log level?
> 
> ```
> I0329 10:42:07.462737    6687 federate.go:88] Resource to federate is a namespace. Given namespace will itself be the container for the federated namespace
> ```

Leaving other info logs to a suppressed default verbosity case was intentional.
I found the ns message which received [this comment](https://github.com/kubernetes-sigs/federation-v2/pull/661#discussion_r270565189) to be the only important msg to convey and kept its verbosity as is. Other logs seemed appropriate for higher verbosity.


> * The spec for a namespace includes a finalizer field, which I think is the result of controller action and should not be included.
> * The template generated for all types includes `apiVersion` and `kind` in its output. I recommend stripping those.

I will push a separate PR for the above 2. However small, they don't seem to fit in any of the PRs in the current chain.

> * There was no addition to the changelog.
> * There is no doc change introducing the feature.

I did intend to update these once the feature is complete, which logically will complete with `federate ns --content`. I feared repeated updates especially in the docs if updated prior to that. Let me know if thts ok.

> I apologize for being so sticky in my reviews, though. I realize that I'm unnecessarily slowing things down given how many PRs are still to go. My usual tendency is to fix things immediately, but I don't think that is required in a PR series where I can count on you to refine things. I'm going to merge this one and you can fix the issues mentioned above in a subsequent PR.

Appreciate, Thanks!

pkg/kubefed2/federate/federate.go
```diff
@@ -188,24 +183,28 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", qualifiedTypeName)
+		return nil, errors.Wrap(err, "Failed to get federation client")
 	}
 
-	_, err = enable.LookupAPIResource(config, typeConfig.Name, typeConfig.APIVersion)
+	typeConfig := &fedv1a1.FederatedTypeConfig{}
+	err = client.Get(context.TODO(), typeConfig, typeName.Namespace, typeName.Name)
+	if apierrors.IsNotFound(err) {
+		return nil, errors.Wrapf(err, "FederatedTypeConfig %q not found. Try 'kubefed2 enable type %s' before federating the resource", typeName, typeName.Name)
```
marun: (No action required) Wrapping the error makes the first part of the error message redundant.  I'll make sure this gets fixed by the end of the series but its non-functional for now.

pkg/kubefed2/federate/federate.go
```diff
@@ -52,8 +53,8 @@ var (
 		flag otherwise.`
 
 	federate_example = `
-		# Federate resource named "my-dep" in namespace "my-ns" of type identified by FederatedTypeConfig "deployment.apps"
-		kubefed2 federate deployment.apps my-dep -n "my-ns" --host-cluster-context=cluster1`
+		# Federate resource named "my-dep" in namespace "my-ns" of kubernetes type "deploy"
+		kubefed2 federate deploy my-dep -n "my-ns" --host-cluster-context=cluster1`
```
marun: (No action required) Using the `deploy` short-name will result in an error message because it is ambiguous (it can mean `deployments.extensions` - the beta version - or `deployment.apps` - the GA version).  Will need to to fix before this PR series is done.

marun:
Thank you!  A couple of minor issues that can be fixed in subsequent PRs.

/lgtm

PR #662, opened 2019-03-18
Make federate resource usable without federation API

irfanurrehman:
This implements:
  - [x] Federate resource usable without federation API, it needs a k8s cluster (if only yaml translation is needed). 

cc @marun @xunpan

### DIFF
--- a/pkg/kubefed2/federate/federate.go
+++ b/pkg/kubefed2/federate/federate.go
@@ -43,10 +43,11 @@ var (
 	federate_long = `
 		Federate creates a federated resource from a kubernetes resource.
 		The target resource must exist in the cluster hosting the federation
-		control plane. The control plane must have a FederatedTypeConfig
-		for the type of the kubernetes resource. The new federated resource
-		will be created with the same name and namespace (if namespaced) as
-		the kubernetes resource.
+		control plane. If the federated resource needs to be created in the
+		API, the control plane must have a FederatedTypeConfig for the type
+		of the kubernetes resource. If using with flag '-o yaml', it is not
+		necessary for the FederatedTypeConfig to exist (or even for the
+		federation API to be installed in the cluster).
 
 		Current context is assumed to be a Kubernetes cluster hosting
 		the federation control plane. Please use the --host-cluster-context
@@ -140,7 +141,7 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 		Name:      j.resourceName,
 	}
 
-	resources, err := GetFedResources(hostConfig, qualifiedTypeName, qualifiedResourceName)
+	resources, err := GetFedResources(hostConfig, qualifiedTypeName, qualifiedResourceName, j.outputYAML)
 	if err != nil {
 		return err
 	}
@@ -161,8 +162,8 @@ type fedResources struct {
 	FedResource *unstructured.Unstructured
 }
 
-func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName) (*fedResources, error) {
-	typeConfig, err := lookupTypeDetails(hostConfig, qualifiedTypeName)
+func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName ctlutil.QualifiedName, outputYAML bool) (*fedResources, error) {
+	typeConfig, err := getTypeConfig(hostConfig, qualifiedTypeName.Name, qualifiedTypeName.Namespace, outputYAML)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


marun:
@irfanurrehman Please update so I can review.

irfanurrehman:
@marun rebased. PTAL!

pkg/kubefed2/federate/federate.go
```diff
@@ -43,10 +43,11 @@ var (
...
 		Federate creates a federated resource from a kubernetes resource.
 		The target resource must exist in the cluster hosting the federation
-		control plane. The control plane must have a FederatedTypeConfig
-		for the type of the kubernetes resource. The new federated resource
-		will be created with the same name and namespace (if namespaced) as
-		the kubernetes resource.
+		control plane. If the federated resource needs to be created in the
+		API, the control plane must have a FederatedTypeConfig for the type
+		of the kubernetes resource. If using with flag '-o yaml', it is not
+		necessary for the FederatedTypeConfig to exist. Infact federate can
```
marun: s/exist. Infact.*/exist (or even for the federation API to be installed in the cluster)./

pkg/kubefed2/federate/federate.go
```diff
@@ -183,32 +185,50 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
 	}, nil
 }
 
-func lookupTypeDetails(hostConfig *rest.Config, typeName ctlutil.QualifiedName) (*fedv1a1.FederatedTypeConfig, error) {
-	apiResource, err := enable.LookupAPIResource(hostConfig, typeName.Name, "")
+func getTypeConfig(hostConfig *rest.Config, typeName, namespace string, outputYAML bool) (typeconfig.Interface, error) {
```
marun: I'm concerned that this function is doing too much.  What do you think of having one function that retrieves a type config and another that generates a type config?

In terms of the logic of choosing how to use the functions, would it make sense for the yaml path to attempt to retrieve a type config and if retrieval fails, warn that it is generating a type config?  And is there a reason to ever have the non-yaml path generate a type config?

xunpan:
it LGTM for the function

pkg/kubefed2/federate/federate.go

pkg/kubefed2/federate/federate.go
irfanurrehman: I think the logic and flow of the function is fine.
I did choose to lookup the federated type also (instead of retrieving the CRD) which needs the FederatedTypeName.
I can create a FederatedTypeName from the TypeName here using some strings magic, but using the typeconfig interfaces avoids duplication of code and possibility of confusion.
So we do generate typeConfig anywhich ways.

I have split the function into two and cleared out naming and comments to address your concern. The logical flow remains same.

irfanurrehman: PS: This is further cleaned up while rebasing #663. You can choose to review that too, if you have no further comments here, as its ready for review.

marun: Given that yaml output may be used to create resources in the future (and not just output to the screen for demonstration purposes), I think it is potentially problematic to prefer a generated typeconfig if one is available in the API.  There is no guarantee that a given federated deployment will use the default generated typeconfig - customization is possible (i.e. the name, kind, group, and version of the federated type).

irfanurrehman: Got your point; Updated accordingly. PTAL!

irfanurrehman:
@marun poke; In case you missed the notification from [my comment](https://github.com/kubernetes-sigs/federation-v2/pull/662#discussion_r272531320) after the last requested change!

irfanurrehman:
> I would like to see the strategies of generation vs lookup of a type config cleanly separated. I think the current iteration of this PR intermingles the two strategies and that is likely to complicate troubleshooting when something goes wrong. Please consider the following pseudocode:

@marun I understood your comment [here](https://github.com/kubernetes-sigs/federation-v2/pull/662#discussion_r271567534) as to what you were referring to, but resisted that change because of reasons listed in [reply](https://github.com/kubernetes-sigs/federation-v2/pull/662#discussion_r272209435). I thought and probably still think it to be better to source the strings from generated typeConfig rather then a function [as written now](https://github.com/kubernetes-sigs/federation-v2/pull/662/files#diff-824920fa5a6d5edf320e4a62dffa4ddeR230).

Anywhichways, I have updated as per your request. PTAL!
I do agree that the code looks simpler now; but its a pretty small piece of code that we are talking about.
On the other hand appreciate your diligence and patience.. :-)

pkg/kubefed2/federate/federate.go
```diff
@@ -183,32 +185,68 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
-	typeName.Name = typeconfig.GroupQualifiedName(*apiResource)
+	glog.V(2).Infof("API Resource for %s found", typeName)
 
-	client, err := genericclient.New(hostConfig)
+	typeConfig, err := attemptToLookupFederatedType(hostConfig, *apiResource, namespace)
+	if err == nil {
+		return typeConfig, nil
+	}
+	if apierrors.IsNotFound(err) && !outputYAML {
+		return nil, errors.Errorf("%v. Try 'kubefed2 enable type %s' before federating the resource", err, typeName)
```
marun: Please prefer Wrapf for instances like this:

errors.Wrapf(err, "Try kubefed2...", typeName)

pkg/kubefed2/federate/federate.go
```diff
@@ -183,32 +185,68 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
+	}
+	if outputYAML {
+		glog.V(1).Infof("Lookup of installed type failed: %v - Falling back on generated default values instead.", err)
```
marun: Please avoid embedding an error in the middle of a message, and prefer appending it to the end e.g.

```
Log("Falling back to a generated type config due to lookup failure: %v", err)
```

pkg/kubefed2/federate/federate.go
```diff
@@ -183,32 +185,68 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
+func attemptToLookupFederatedType(hostConfig *rest.Config, apiResource metav1.APIResource, namespace string) (typeconfig.Interface, error) {
+	resolvedTypeName := typeconfig.GroupQualifiedName(apiResource)
+	// Lookup federated API availability
+	federatedTypeName := federatedTypeName(apiResource)
+	_, err := enable.LookupAPIResource(hostConfig, federatedTypeName, "")
 	if err != nil {
-		return nil, errors.Wrap(err, "Failed to get federation client")
+		return nil, err
+	}
+	if err == nil {
```
marun: This conditional is redundant, since by definition anything after the `err != nil` block is the case where `err == nil`.

pkg/kubefed2/federate/federate.go
```diff
@@ -183,32 +185,68 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
+	}
+
+	return nil, err
+}
+
+func attemptToLookupFederatedType(hostConfig *rest.Config, apiResource metav1.APIResource, namespace string) (typeconfig.Interface, error) {
+	resolvedTypeName := typeconfig.GroupQualifiedName(apiResource)
+	// Lookup federated API availability
+	federatedTypeName := federatedTypeName(apiResource)
+	_, err := enable.LookupAPIResource(hostConfig, federatedTypeName, "")
```
marun: What is your intent in looking up the api resource of the federated type?  If the federated type were not present for whatever reason, wouldn't an appropriate error be returned when attempting to create an instance of the type?

If it is desirable to lookup the api resource in this function, is there a reason to search for a computed type name (via `federatedTypeName')?  There is no guarantee that the computed name is valid, and performing the check after retrieving the type config would allow for the configured type name to be used.

pkg/kubefed2/federate/federate.go
irfanurrehman: I did intentionally replace `Wrapf` with `Errorf`  because an error as below (actual output):
`error: federatedtypeconfigs.core.federation.k8s.io "deployments.apps" not found. Try 'kubefed2 enable type deployments.apps' before federating the resource`
reads better compared to
`error: Try 'kubefed2 enable type deployments.apps before federating the resource because federate failed with: resourcefederatedtypeconfigs.core.federation.k8s.io "deployments.apps" not found`

pkg/kubefed2/federate/federate.go
irfanurrehman: As in previous comment, I did embed the error in between because it did just read better to me.
Updating this anyways as this sequence also looks ok.

pkg/kubefed2/federate/federate.go
irfanurrehman: Silly miss, updated!

pkg/kubefed2/federate/federate.go
irfanurrehman: 2 reasons:
1. It provides clear indication to the user as to what is missing; its however debatable if the typeconfig could exist without the federated API CRD or vice versa, as the recommended way is to use `kubefed2 enable` or `helm install`. This is but a possibility.
2. A more pressing reason. I did observe that the create resource on a resource client for an API resource sometimes returns with error: `the server could not find the requested resource` even when the resource CRD is available; if its called immediately after the resource CRD is created. The lookup however always succeeds. Its probably has something to do with when is the resource reflected in the discovery API (for lookup) and against the calls of dynamic client in the API server.
This is further pronounced in #663 which implements `--enable-type`, where I have implemented a retry loop with a short timeout (in case the lookup succeeded but create returns `NotFound` error) in create to ensure this is handled correctly.

Having stated above I could further think of a possible scenario, where an user wants to use a precreated typeconfig (probably to override target `group/version` ), haven't used `kubefed2 enable` so the federated API resource is not available; creates typeconfig only to convert resources (-o yaml). This code would still default to a generated type if lookup is kept here.
I am not sure if this is an ideal scenario, but probably is a possibility.
For this possibility to work cleanly and still serve the reasons 1 and 2 above, moving The lookup  of federated type immediately before create resource. The retry loop is introduced in #663.

irfanurrehman: EDIT: On second observations, I was wrong on `Lookup()` always succeeding, as I saw a failure on `Lookup()` too sometimes when running with --enable-type. I m removing the `Lookup()` completely here and relying on failure reported from create resource alone.

irfanurrehman:
@marun in case you are satisfied with this and merge this, please go ahead to check the last commit in #663; I will update the WIP, PRs first comment and rebase tomorrow; but this will save us a day. I think this set of PRs is falling back a bit to some extent because of our timing difference.

pkg/kubefed2/federate/federate.go
marun: I take your point about the NotFound error being embeddable, but in cases where there is no specific error (as is the case here), the error message should almost always be placed at the end of a log message.  That ensures that a really long message doesn't result in suffixed content being confused with the error content.

marun:
Thank you for your patience in driving this PR to a mergeable state!

/lgtm

PR #663, opened 2019-03-18
Implement optional --enable-type in federate resource

irfanurrehman:
This implements:
  - [x] Optional --enable-type to enable type also while federating resource

cc @marun @xunpan

### DIFF
--- a/pkg/kubefed2/federate/federate.go
+++ b/pkg/kubefed2/federate/federate.go
@@ -19,6 +19,7 @@ package federate
 import (
 	"context"
 	"io"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
@@ -28,6 +29,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 
 	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
@@ -37,6 +39,11 @@ import (
 	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/enable"
 	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/options"
 	"github.com/kubernetes-sigs/federation-v2/pkg/kubefed2/util"
+)
+
+const (
+	createResourceRetryTimeout  = 10 * time.Second
+	createResourceRetryInterval = 1 * time.Second
 )
 
 var (
@@ -66,11 +73,13 @@ type federateResource struct {
 	resourceNamespace string
 	output            string
 	outputYAML        bool
+	enableType        bool
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.resourceNamespace, "namespace", "n", "default", "The namespace of the resource to federate.")
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
+	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 }
 
 // Complete ensures that options are valid.
@@ -89,6 +98,10 @@ func (j *federateResource) Complete(args []string) error {
 		j.outputYAML = true
 	} else if len(j.output) > 0 {
 		return errors.Errorf("Invalid value for --output: %s", j.output)
+	}
+
+	if j.enableType && j.outputYAML {
+		return errors.New("Flag '--enable-type' cannot be used with '--output [yaml]'")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/kubefed2/federate/federate.go
```diff
@@ -178,13 +192,20 @@ func GetFedResources(hostConfig *rest.Config, qualifiedTypeName, qualifiedName c
...
-	return &fedResources{
-		typeConfig:  typeConfig,
-		FedResource: fedResource,
-	}, nil
+	return fedResource, nil
 }
 
-func getTypeConfig(hostConfig *rest.Config, typeName, namespace string, outputYAML bool) (typeconfig.Interface, error) {
+func getTypeResources(hostConfig *rest.Config, typeName, namespace string, enableType, outputYAML bool) (*enable.TypeResources, error) {
+	if enableType {
```
marun: Why should `--enable-type` result in creating a federated resource with a generated type config even when a type config might already exist in the API?

pkg/kubefed2/federate/federate.go
```diff
@@ -124,50 +134,54 @@ func NewCmdFederateResource(cmdOut io.Writer, config util.FedConfig) *cobra.Comm
 	return cmd
 }
 
+type FedResources struct {
```
marun: Please rename `FedResources` to clearly indicate it's purpose.

pkg/kubefed2/federate/federate.go
```diff
@@ -308,12 +329,30 @@ func CreateFedResource(hostConfig *rest.Config, resources *fedResources, dryrun
...
+		err = wait.PollImmediate(createResourceRetryInterval, createResourceRetryTimeout, func() (bool, error) {
+			_, err := fedClient.Resources(resources.FedResource.GetNamespace()).Create(resources.FedResource, metav1.CreateOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			if err != nil {
+				return false, err
+			}
+			return true, nil
+		})
```
marun: Where is the check for an error condition?

pkg/kubefed2/federate/federate.go
```diff
@@ -124,50 +134,54 @@ func NewCmdFederateResource(cmdOut io.Writer, config util.FedConfig) *cobra.Comm
...
 
+	fedResources := &FedResources{
+		FedResource:   fedResource,
+		TypeResources: typeResources,
+	}
+
 	if j.outputYAML {
-		err := util.WriteUnstructuredToYaml(resources.FedResource, cmdOut)
+		objs := []interface{}{}
+		if j.enableType {
```
marun: Why should --enable-type be respected if yaml output is requested?

pkg/kubefed2/federate/federate.go
```diff
@@ -66,11 +73,14 @@ type federateResource struct {
 	resourceNamespace string
 	output            string
 	outputYAML        bool
+	enableType        bool
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.resourceNamespace, "namespace", "n", "default", "The namespace of the resource to federate.")
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
+	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If provided, the command will enable the provided kuberneted API type for propagation (with default values), before federating the resource. "+
```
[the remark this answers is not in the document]

pkg/kubefed2/federate/federate.go
```diff
@@ -124,50 +134,54 @@ func NewCmdFederateResource(cmdOut io.Writer, config util.FedConfig) *cobra.Comm
...
+	typeResources, err := getTypeResources(hostConfig, j.typeName, j.FederationNamespace, j.enableType, j.outputYAML)
+	if err != nil {
+		return err
+	}
+	fedResource, err := GetFedResource(hostConfig, typeResources.TypeConfig, qualifiedResourceName)
 	if err != nil {
 		return err
 	}
 
+	fedResources := &FedResources{
```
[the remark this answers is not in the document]
irfanurrehman: Done!

pkg/kubefed2/federate/federate.go
irfanurrehman: That was a mistake, updated!

pkg/kubefed2/federate/federate.go
irfanurrehman: Done!

pkg/kubefed2/federate/federate.go
irfanurrehman: Added!

pkg/kubefed2/federate/federate.go
irfanurrehman: I did initially think of an use case that a user might want to collect all yamls, before creating any of them. Revisiting the thought, this might become problematic (unless we want to make `enable type` idempotent) if many resources of the same type needs to be federated, in which case the type yamls will be repeated. The updated implementation considers `--enable-type` and `-o yaml` mutually exclusive.

irfanurrehman:
> I think it would be preferable for `--enable-type` to:
> 
> * detect whether type enablement was required
> * if so, enable type federation

@marun Updated accordingly, PTAL!

pkg/kubefed2/federate/federate.go
```diff
@@ -141,29 +154,40 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 		Name:      j.resourceName,
 	}
 
-	resources, err := GetFedResources(hostConfig, qualifiedTypeName, qualifiedResourceName, j.outputYAML)
+	artifacts, err := GetFederateArtifacts(hostConfig, qualifiedTypeName, qualifiedResourceName, j.enableType, j.outputYAML)
```
marun: I'm not sure I understand why it is necessary to collect 'artifacts' required by enable vs just ensuring that a type is enabled before attempting federation.  The strategy as proposed appears to avoid treating the enable functionality as a black box, which will have negative implications when inevitable changes need to be made to that functionality.  Is there a good reason not to do the following?

```golang
if enableType {
  // This should be a function on the enable side of things.  The lack of a returned error indicates that enable was successful.
  ensureEnabled(...)
}

// Whatever operations are necessary to federate the type, including looking up the type config.
...
```

irfanurrehman: 1. For the flow you have suggested - I have kept any modification to the API at the end which minimises partial updates to the API in case of other failures, which includes enabling the type. I think, its better to keep it this way. Let me know if you think otherwise.

2. The artifacts which are collected are not the ones computed solely for enable. These are anyways computed for federate functionality. The exception is `apiResource`, which is computed but need not be passed further if not for `enableType`. `apiResource` infact can be fetched again from `typeConfig`, rather then being passed as another parameter.

3. About treating the enable as a black box - I did choose not to repeat the operations which are common to both (`lookup api type and generate the typeconfig`), which would be wasteful especially in the next PR (enable ns with content). I think its better this way and the coupling between the two functionalities is really really small. If you think strongly otherwise, I would not mind changing this such that only the two main enable functions (`GetResources` and `CreateResources`) are utilised in federate.

marun: 1. Should `federate` commit to duplicating the error handling of `enable`?
2. Does the expense of an api call or two justify exposing `federate` to the internal details of `enable`?
3. As per my preliminary comments on #741, `enable` should be required at most once per type vs the proposed approach of once per resource.

I feel strongly that `enable` and `federate` are orthogonal concerns, and that `federate` should not be dependent on the internal details of `enable` until such time as performance is determined to be a concern.

irfanurrehman: > * Are you proposing to minimize partial updates in `federate` but not in `enable`?

Did not get you? Enable already follows this flow, resources are created in the end in enable, isn't it?
> * Does the expense of some api calls justify exposing `federate` to the internal details of `enable`?

If you talking about my point 2 above, I was talking about artifacts, they are not really for enable, but needed for federate anyways. If you talking about my point 3. I already mentioned that I am quite ok to update such that no changes are made to enable and federate does not use anything internal ( eg `typeResources` ) from enable.
> * As per my preliminary comments on #741, `enable` should be required at most once per type vs the proposed approach of once per resource.

Thats understood. I was waiting for this PR to be complete before updating #741.

I have updated this PR accordingly, PTAL!

pkg/kubefed2/federate/federate.go
```diff
@@ -300,28 +322,56 @@ func getNamespace(typeConfig typeconfig.Interface, qualifiedName ctlutil.Qualifi
...
-			return errors.Wrapf(err, "Error creating %s %q", fedKind, qualifiedFedName)
+			return err
 		}
 	}
 
 	glog.Infof("Successfully created %s %q from %s", fedKind, qualifiedFedName, typeConfig.GetTarget().Kind)
 	return nil
 }
+
+func CreateResources(cmdOut io.Writer, hostConfig *rest.Config, artifacts *federateArtifacts, namespace string, enableType, dryrun bool) error {
```
marun: nit: s/dryrun/dryRun/

PR #665, opened 2019-03-19
Revert "Some OSX support"

marun:
Reverts kubernetes-sigs/federation-v2#650

I regret having merged this.  I missed the consequences of `kubefed2` being generated as `kubefed2-linux`, and I think there needs to be greater thought put into supporting osx without breaking current expectations about binary names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,11 +37,9 @@ endif
 BUILDMNT = /go/src/$(GOTARGET)
 BUILD_IMAGE ?= golang:1.11.2
 
-HYPERFED_TARGET = bin/hyperfed-linux
-CONTROLLER_TARGET = bin/controller-manager-linux
-KUBEFED2_TARGET = bin/kubefed2-linux
-
-CONTROLLER_TARGET_NATIVE = bin/controller-manager
+HYPERFED_TARGET = bin/hyperfed
+CONTROLLER_TARGET = bin/controller-manager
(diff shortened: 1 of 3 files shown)


marun:
cc: @yamt

yamt:
@marun is it acceptable for you if they were `kubefed2` for linux and `kubefed2-native` for native?

marun:
@yamt I'm ok with the build differentiating by suffix (both -linux and -darwin) so long as invocation can remain without the suffix (e.g. via a symlink that varied by platform).

I'm also wondering if we want to start shipping a darwin build of kubefed2 - the catch being I'm not sure how we would test it.

yamt:
@marun well, currently on linux both of `kubefed2` and `kubefed2-linux` are linux binaries.
do you prefer a symlink?

i guess shipping `kubefed2` binaries for some popular platforms as it's done for `kubectl` makes some sense.
https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl-binary-using-curl

marun:
@yamt I guess I'd be fine with just sticking with an un-suffixed linux binary, I figured a symlink might allow both linux and darwin to avoid having to add a suffix.

marun:
@yamt I'll close this.  Would you be willing to post a new PR that removes the suffix from kubefed2?

yamt:
@marun i'll take a look tomorrow.

PR #666, opened 2019-03-19
Ensure all script boolean handling uses y/<empty> vs true/false

marun:


xunpan:
LGTM

xunpan:
I checked the code again and think it is ready to merge.
/lgtm

PR #667, opened 2019-03-20
Fix namespace deletion handling

### LINKED ISSUE #618: Deleting a FederatedNamespace does not delete the federated namespaces
**What happened**:
I have a federated namespace which is placed on 2 clusters.
```
# namespace.yaml
apiVersion: types.federation.k8s.io/v1alpha1
kind: FederatedNamespace
metadata:
  name: test-namespace
  namespace: test-namespace
spec:
  placement:
    clusterNames:
      - cluster1
      - cluster2
```

 When I delete the federated resource with `kubectl delete -f namespace.yaml`, the federated resources are not deleted. If I instead of deleting the `FederatedNamespace`, I update it by setting 
the clusterNames to an empty array, the clusters are deleted.
```
  placement:
    clusterNames: []
```

I expected the federated resources to be deleted.

Apply the `namespace.yaml`, then delete it again. 

**Environment**:
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-04T04:48:55Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T06:59:37Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
- Federation version (master build, or tag v0.0.6)
- Install method (helm)
- Install configuration (cluster-scoped)

/kind bug

marun:
This commit fixes a test bug and the deletion helper bug it was masking to correct previously broken namespace deletion handling.

- Symptom: Deleting a federated namespace would not result in the removal of the namespaces it was managing in member clusters.
- Test bug: When 2 clusters (one of them the host cluster) were targeted by the crudtester's CheckPropagationChange method, the namespace in the non-host cluster would be removed.  A subsequent call to CheckDelete would automatically succeed since the namespace no longer existed in non-host member clusters.
- Deletion helper bug: The key being used to attempt retrieval of namespaces in member clusters was for the federated namespace (i.e. `<namespace>/<namespace>`) rather than the namespace that contained it (i.e. `<namespace>`). A call to the informer for the bad key returned no results, which the deletion helper took to mean that managed resources had already been removed and that the federated resource was safe to delete.

Fixes #618

### DIFF
--- a/pkg/controller/sync/resource.go
+++ b/pkg/controller/sync/resource.go
@@ -190,6 +190,7 @@ func (r *federatedResource) EnsureDeletion() error {
 	r.DeleteVersions()
 	_, err := r.deletionHelper.HandleObjectInUnderlyingClusters(
 		r.federatedResource,
+		r.TargetName().String(),
 		func(clusterObj pkgruntime.Object) bool {
 			// Skip deletion of a namespace in the host cluster as it will be
 			// removed by the garbage collector once its contents are removed.
--- a/pkg/controller/util/deletionhelper/deletion_helper.go
+++ b/pkg/controller/util/deletionhelper/deletion_helper.go
@@ -103,7 +103,7 @@ func (dh *DeletionHelper) EnsureFinalizers(obj runtime.Object) (
 // when done.
 // Callers are expected to keep calling this (with appropriate backoff) until
 // it succeeds.
-func (dh *DeletionHelper) HandleObjectInUnderlyingClusters(obj runtime.Object, skipDelete func(runtime.Object) bool) (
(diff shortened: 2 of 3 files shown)


marun:
@shashidharatd At @xunpan's urging I stripped out the log changes from this one.  I'll submit separately to make sure this critical fix is as easy to review as possible.

shashidharatd:
LGTM. waiting for @xunpan

PR #669, opened 2019-03-20
Fix orphan deletion

### LINKED ISSUE #622: Orphaning deletion of federated resource is not reliable
The sync controller is intended to respect the `orphan` finalizer added to a federated resource when a deletion request specifies `propagationPolicy: Orphan` (via a raw or custom client request - `kubectl delete` doesn't support setting it).  However, as per a discussion intiated by @yamt on #606, a race exists between the sync controller and the garbage collector:

 - Desired behavior (sync controller wins)
   - The sync controller sees the `orphan` finalizer before the garbage collector removes it. 
   - Target resources in member clusters will not be deleted.
 - Undesired behavior (gc wins)
   - The garbage collector, having ensured the removal of any owner references to the federated resource, removes the `orphan` finalizer.  
   - The sync controller subsequently checks for the `orphan` finalizer and does not find it, and proceeds to delete target resources from member clusters.

There doesn't appear to be an obvious way to synchronize the sync controller with the garbage collector, which suggests that the sync controller cannot rely on the `orphan` finalizer as an indication that resources in member clusters should be orphaned.  ~Since it's not possible to set the `orphan` via kubectl anyway, it seems reasonable to require the addition of a federation-specific finalizer (e.g. `federationOrphan`) to indicate that the sync controller should orphan resources.  It may be desirable to implement something like `kubefed2 orphaning-delete` that would set the deletion timestamp and apply the federation-specific finalizer.~

Edit: The finalizer added by the sync controller is `federation.kubernetes.io/delete-from-underlying-clusters`, and in the absence of using the `orphan` finalizer,  a user could just remove the federation finalizer to orphan target resources.  To handle the case of the gc removing the federated resource before reconciliation by the sync controller, it would be necessary to call delete on the version manager if the federated resource could not be retrieved by the federated resource accessor to avoid tracking versions for deleted resources.  

/kind bug

marun:
fedv2 has been using the `orphan` finalizer in attempt to integrate with kube deletion policy.  This use of `orphan` proved problematic, though, since kube deletion policy is internal to a single cluster and the sync controller ends up racing with the gc (as per #622).  This PR switches to a new finalizer (`federation.k8s.io/sync-controller`) that ensures that deletion of federated resources will be visible to the sync controller and a new annotation (`federation.k8s.io/orphan`) in place of the `orphan` finalizer to indicate that managed resources should be retained on deletion of their managing federated resource.  

Documentation of the new deletion policy and use of finalizers is included, as is a test of orphan deletion (superceding that proposed by #606).

Due to the finalizer change, this PR will break deletion of federated resources reconciled by previous versions of the sync controller.  It may be desirable to highlight this fact along with a workaround (e.g. `kubectl patch` command to remove the previous finalizers) in something like a changelog. 

Fixes #622

cc: @yamt

### DIFF
--- a/test/e2e/crd.go
+++ b/test/e2e/crd.go
@@ -209,8 +209,23 @@ overrides:
 		return targetObj, overrides, nil
 	}
 
-	validateCrud(f, tl, concreteTypeConfig, testObjectsFunc)
+	crudTester, targetObject, overrides := initCrudTest(f, tl, typeConfig, testObjectsFunc)
+	// Make a copy for use in the orphan check.
+	deletionTargetObject := targetObject.DeepCopy()
+	crudTester.CheckLifecycle(targetObject, overrides)
 
+	if namespaced {
+		// This check should not fail so long as the main test loop
+		// skips for limited scope.  If it does fail, manual cleanup
+		// after CheckDelete may need to be implemented.
+		if framework.TestContext.LimitedScope {
+			tl.Fatalf("Test of orphaned deletion assumes deletion of the containing namespace")
+		}
+		// Perform a check of orphan deletion.
+		fedObject := crudTester.CheckCreate(deletionTargetObject, nil)
+		orphanDeletion := true
+		crudTester.CheckDelete(fedObject, orphanDeletion)
+	}
 }
 
 func waitForCrd(config *rest.Config, tl common.TestLogger, apiResource metav1.APIResource) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Unreleased
- - [#698](https://github.com/kubernetes-sigs/federation-v2/pull/698) -
+ - [#622](https://github.com/kubernetes-sigs/federation-v2/pull/622) -
+   Switched the sync controller to using a new finalizer
+   (`federation.k8s.io/sync-controller` instead of
+   `federation.kubernetes.io/delete-from-underlying-clusters`) and
+   replaced the use of the kube `orphan` finalizer in favor of an
+   annotation to avoid conflicting with the garbage collector.  This
+   change in finalizer usage represents a breaking change since
+   resources reconciled by previous versions of the sync controller
+   will have the old finalizer.  The old finalizer would need to be
+   manually removed from a resource for that resource to be garbage
+   collected after deletion.
+- [#698](https://github.com/kubernetes-sigs/federation-v2/pull/698) -
    Fix the generated CRD schema of scalable resources to define the
    `retainReplicas` of type `boolean` rather than the invalid type
    `bool`.
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -16,6 +16,7 @@
     - [Unjoin Clusters](#unjoin-clusters)
   - [Enabling federation of an API type](#enabling-federation-of-an-api-type)
   - [Disabling federation of an API type](#disabling-federation-of-an-api-type)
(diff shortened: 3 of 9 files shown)


marun:
I've also kicked off maintenance of a changelog for this PR (as discussed in this week's wg sync).

yamt:
looks good to me. thank you.

irfanurrehman:
@marun this looks good to me too.
I don't like the fact that we cannot use `--cascade=false` anymore; Does providing a delete via `kubefed2` with this option make sense ?

marun:
@irfanurrehman I don't think we should duplicate deletion functionality from kubectl.  I am supportive of adding a command that would set or remove the annotation (e.g. `kubefed2 enable-orphaning-deletion`).

irfanurrehman:
> @irfanurrehman I don't think we should duplicate deletion functionality from kubectl. I am supportive of adding a command that would set or remove the annotation (e.g. `kubefed2 enable-orphaning-deletion`).

@marun Created https://github.com/kubernetes-sigs/federation-v2/issues/690

test/e2e/crd.go
```diff
@@ -209,8 +209,23 @@ overrides:
...
+		// This check should not fail so long as the main test loop
+		// skips for limited scope.  If it does fail, manual cleanup
+		// after CheckDelete may need to be implemented.
+		if framework.TestContext.LimitedScope {
+			tl.Fatalf("Test of orphaned deletion assumes deletion of the containing namespace")
+		}
+		// Perform a check of orphan deletion.
+		fedObject := crudTester.CheckCreate(deletionTargetObject, nil)
+		orphanDeletion := true
+		crudTester.CheckDelete(fedObject, orphanDeletion)
```
xunpan: It is minor but I think there is no need for variable `orphanDeletion` because we deleted related part in this diff but keep it here.

xunpan:
Just 1 minor comments above. The logic LGTM.

test/e2e/crd.go
marun: 'true' doesn't provide any indication of my intent unless someone is familiar with the signature of `CheckDelete`, so for clarity I am declaring a variable whose name indicates my intent.  Greater programmers than me have suggested that code's primary purpose is to be read, and secondary purpose to be executed, and this is an example of emphasizing the former purpose.

xunpan: I agree on the point for code readability. Code should make itself clear.

My question is, did this work only for bool `true` but not for `false` because I found your deletion of
 - https://github.com/kubernetes-sigs/federation-v2/pull/669/files#diff-42695a0efcb8096fdb6a618bb1d79dd0R97
 - https://github.com/kubernetes-sigs/federation-v2/pull/669/files#diff-09cefc57af0991da6c524792e4183f3aR95

marun: The only reason `orphanDeletion` appeared everywhere before is because the parameter was being passed as a pointer to a bool.  It's not possible to take the pointer to a literal, so it was necessary to create a variable.  The only time I thought it important enough to retain the use of a variable was for the non-default 'orphan' case.  I think, though, that this is a nitty thing to worry about.

xunpan: We can close this discussion. It is not an issue. I tried to understand related consideration and learn from the code. Though the last comments does not convince me :-) for the change, it is fine for current code to me.

irfanurrehman:
@marun please rebase and I will merge, thanks.

marun:
Rebased

Issue #670, opened 2019-03-20
Helm chart rbac doesn't appear to be tested in CI


marun:
A number of people have complained that they are not able to get federation working by following the user guide.  One of the problems was diagnosed by @pmorie as an rbac rules issue (allowing access to the old `primitives.federation.k8s.io` group instead of the new `types.federation.k8s.io`).  While the fix for that is easy, I want to make sure CI will catch issues like this before they merge and affect users.

marun:
With the removal of script-based deployment, rbac for a deployment of federation is only provisioned by helm so it is always tested.

PR #671, opened 2019-03-20
Fix helm chart rules to allow types instead of primitives

marun:
This has been breaking new users attempting to try federation.  I had assumed that the CI testing of helm would catch this sort of thing, but as per #670 that doesn't appear to be the case.


xunpan:
LGTM. 

The CI failed but seems not related to the change. I will check the log and restart the CI to have a new try.

PR #673, opened 2019-03-21
Update crd test to use a permitted group/version

### LINKED ISSUE #670: Helm chart rbac doesn't appear to be tested in CI

marun:
Previously the crd e2e test was using a random group/version, but this was breaking a helm deploy that was granted only selective permissions.  Switching to the default group/version used by enable ensures the test will always have sufficient privileges to run.

This PR is now in support of #710.

### DIFF
--- a/pkg/kubefed2/enable/directive.go
+++ b/pkg/kubefed2/enable/directive.go
@@ -52,8 +52,8 @@ type EnableTypeDirective struct {
 }
 
 func (ft *EnableTypeDirective) SetDefaults() {
-	ft.Spec.FederationGroup = defaultFederationGroup
-	ft.Spec.FederationVersion = defaultFederationVersion
+	ft.Spec.FederationGroup = DefaultFederationGroup
+	ft.Spec.FederationVersion = DefaultFederationVersion
 }
 
 func NewEnableTypeDirective() *EnableTypeDirective {
--- a/pkg/kubefed2/enable/enable.go
+++ b/pkg/kubefed2/enable/enable.go
@@ -44,8 +44,8 @@ import (
 )
 
 const (
-	defaultFederationGroup   = "types.federation.k8s.io"
-	defaultFederationVersion = "v1alpha1"
+	DefaultFederationGroup   = "types.federation.k8s.io"
+	DefaultFederationVersion = "v1alpha1"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


marun:
This PR intentionally reverts #671 to check that CI will fail if cluster-admin is not granted to the federation namespace service account when deploying with helm.

marun:
I've rebased this on an experimental fix for the version manager race to avoid the false negatives I've been seeing on this PR.

marun:
The check of helm-deployed federation was failing with the old and invalid rbac roles.  The latest push fixes the roles, and CI should now be expected to pass.

marun:
@xunpan This test failure appears persistent.  It appears that a namespaced control plane deployed by helm is not granted sufficient permission to function in the absence of cluster-admin being granted on the namespace.

xunpan:
@marun
The 1st error I got is:
```
$ kubectl -n foo logs federation-controller-manager-6b55d87f87-666ct 
Federation v2 controller-manager version: version.Info{Version:"v0.0.1-alpha.0", GitCommit:"unknown", GitTreeState:"unknown", BuildDate:"unknown", GoVersion:"go1.11.4", Compiler:"gc", Platform:"linux/amd64"}
W0322 01:03:30.273664       1 client_config.go:548] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0322 01:03:30.276078       1 controller-manager.go:114] Federation will be limited to the "foo" namespace
I0322 01:03:30.283685       1 leaderelection.go:205] attempting to acquire leader lease  foo/federation-controller-manager...
E0322 01:03:30.390031       1 leaderelection.go:270] error retrieving resource lock foo/federation-controller-manager: configmaps "federation-controller-manager" is forbidden: User "system:serviceaccount:foo:default" cannot get configmaps in the namespace "foo
```

Current chart does not include enough permission in role binding after remove cluster-admin. I will add related part in and then you can merge it with this fix.

marun:
Replaced by #683 and #710.

xunpan:
LGTM

PR #675, opened 2019-03-21
Fix version manager race

marun:
Previously more than one thread could potentially modify VersionManager.versions.  This PR ensures that version writes are
synchronous to avoid that possibility.

### DIFF
--- a/pkg/controller/sync/version/manager.go
+++ b/pkg/controller/sync/version/manager.go
@@ -63,8 +63,6 @@ type VersionManager struct {
 
 	hasSynced bool
 
-	worker util.ReconcileWorker
-
 	versions map[string]pkgruntime.Object
 
 	client generic.Client
@@ -80,13 +78,6 @@ func NewVersionManager(client generic.Client, namespaced bool, federatedKind, ta
 		client:        client,
 	}
 
-	v.worker = util.NewReconcileWorker(v.writeVersion, util.WorkerTiming{
-		Interval:       time.Millisecond * 50,
-		RetryDelay:     time.Nanosecond * 1, // Effectively 0 delay
-		InitialBackoff: time.Second * 1,
-		MaxBackoff:     time.Second * 10,
-	})
-
 	return v
 }
 
@@ -101,8 +92,6 @@ func (m *VersionManager) Sync(stopChan <-chan struct{}) {
 	if !ok {
 		return
 	}
-
-	m.worker.Run(stopChan)
 }
 
 // HasSynced indicates whether the manager's in-memory state has been
@@ -197,9 +186,7 @@ func (m *VersionManager) Update(resource VersionedResource,
 
 	m.Unlock()
 
-	m.worker.Enqueue(qualifiedName)
-
-	return nil
+	return m.writeVersion(obj, qualifiedName)
 }
 
 // Delete removes the named propagated version from the manager.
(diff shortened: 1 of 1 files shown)


marun:
The hassle of having to use the meta accessor suggests a switch to unstructured.

marun:
Related: #658

marun:
After a conversation with @irfanurrehman, I've come to the conclusion that the current scheme to write versions to the api asynchronously is too complicated.  A switch to synchronous writes to the API is suggested.

marun:
@xunpan @irfanurrehman Updated to write synchronously.  This means that there should never be conflicting access to objects in the map, since only a single thread at a time will be interacting with the version manage for a given key.

marun:
Need to address the namespaced version manager before review.

pkg/controller/sync/version/manager.go
```diff
@@ -276,146 +263,137 @@ func (m *VersionManager) versionQualifiedName(qualifiedName util.QualifiedName)
...
+			// Clear the resource version and retry.
+			resourceVersion = ""
+			return false, nil
+		}
 
-	// Version has been updated or deleted since the last version
-	// manager write.  Attempt to refresh and retry.
+		// Update was successful. All returns should be true even in
+		// the event of an error since the next reconcile can refresh
+		// the resource version on the next reconcile.
```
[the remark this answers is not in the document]
marun: Done

yamt:
lgtm

marun:
Latest push added a check for errors other than `NotFound` and `Conflict` on update.

pkg/controller/sync/version/manager.go
```diff
@@ -276,146 +263,140 @@ func (m *VersionManager) versionQualifiedName(qualifiedName util.QualifiedName)
...
+
+			glog.V(4).Infof("Creating %s %q", adapterType, qualifiedName)
+			err = m.client.Create(context.TODO(), createdObj)
+			if apierrors.IsAlreadyExists(err) {
+				glog.V(4).Infof("%s %q was created by another process.  Will refresh the resource version and attempt to update.", adapterType, qualifiedName)
+				refreshVersion = true
+				return false, nil
+			}
+			if err != nil {
+				runtime.HandleError(errors.Wrapf(err, "Failed to create version %s %q", adapterType, key))
```
xunpan: s/version/resource version/

pkg/controller/sync/version/manager.go
```diff
@@ -276,146 +263,140 @@ func (m *VersionManager) versionQualifiedName(qualifiedName util.QualifiedName)
...
-func (m *VersionManager) refreshVersion(obj pkgruntime.Object) error {
-	// A read lock is not suggested due to the name and namespace of a
-	// resource being immutable.
-	qualifiedName := util.NewQualifiedName(obj)
-
-	glog.V(4).Infof("Refreshing %s version %q from the API", m.federatedKind, qualifiedName)
-	refreshedObj := m.adapter.NewObject()
-	err := m.client.Get(context.TODO(), refreshedObj, qualifiedName.Namespace, qualifiedName.Name)
+func (m *VersionManager) getResourceVersionFromAPI(qualifiedName util.QualifiedName) (string, error) {
+	glog.V(4).Infof("Retrieving resourceVersion for %s %q from the API", m.federatedKind, qualifiedName)
```
xunpan: `s/resourceVersion/resource verion/`

pkg/controller/sync/version/manager.go
```diff
@@ -276,146 +263,140 @@ func (m *VersionManager) versionQualifiedName(qualifiedName util.QualifiedName)
...
+				refreshVersion = true
+				return false, nil
+			}
+			if err != nil {
+				runtime.HandleError(errors.Wrapf(err, "Failed to create version %s %q", adapterType, key))
+				return false, nil
+			}
+
+			// Update the resource version that will be used for update.
+			resourceVersion, err = getResourceVersion(createdObj)
```
xunpan: For the created resource, it seems no need to update again.
Do you trying to reuse the logic at the end of block to update local cached obj?

marun: The version detail is stored in the status of the resource.  Status is not set on creation - it needs to be set explicitly through the status subresource after creation.

pkg/controller/sync/version/manager.go
marun: It's a field called resourceVersion.

pkg/controller/sync/version/manager.go
marun: 'version' in this case refers to the version resource - the formatted message would be `Failed to create version PropagatedVersion <myns>/<type>-<myname>`.  Reading that suggests that 'version' should just be removed.

pkg/controller/sync/version/manager.go
```diff
@@ -276,146 +263,140 @@ func (m *VersionManager) versionQualifiedName(qualifiedName util.QualifiedName)
...
+			err := setResourceVersion(createdObj, "")
+			if err != nil {
+				runtime.HandleError(errors.Wrapf(err, "Failed to clear the resource version for %s %q", adapterType, key))
+				return false, nil
+			}
+
+			glog.V(4).Infof("Creating %s %q", adapterType, qualifiedName)
+			err = m.client.Create(context.TODO(), createdObj)
+			if apierrors.IsAlreadyExists(err) {
+				glog.V(4).Infof("%s %q was created by another process.  Will refresh the resource version and attempt to update.", adapterType, qualifiedName)
```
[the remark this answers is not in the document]
marun: Done

marun:
@xunpan I've updated the log messages for consistency in a new commit - PTAL.

xunpan:
LGTM

PR #679, opened 2019-03-21
Fix binary name break in scripts/imagebuild.sh

pmorie:
In the 0.0.7 release, the binary name built in CI changed from `hyperfed` to `hyperfed-linux`. This broke the imagebuild.sh script that pushes to quay during a release; this PR makes a tactical fix to that script. Longer term, we should just run a make command to build and push the image to reduce the risk of name changes causing future breaks.

### DIFF
--- a/scripts/imagebuild.sh
+++ b/scripts/imagebuild.sh
@@ -21,7 +21,7 @@ set -o pipefail
 base_dir="$(cd "$(dirname "$0")/.." ; pwd)"
 dockerfile_dir="${base_dir}/images/federation-v2"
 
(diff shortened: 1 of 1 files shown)


pmorie:
Looks like #650 is where the break was introduced; it got canary builds and we didn't notice :(

Canary builds are broken after this build of master in travis: https://travis-ci.org/kubernetes-sigs/federation-v2/builds/508220319, ex: https://travis-ci.org/kubernetes-sigs/federation-v2/builds/508394573

I'm going to green-button this because it will fix canary builds.

PR #683, opened 2019-03-22
make federation-role has enough permission for namespaced deployment

xunpan:
cluster-admin is removed for the controller role binding in https://github.com/kubernetes-sigs/federation-v2/pull/673

However, namespaced permission is not set correctly to have permission to access `configmaps`, `secrets`, `events` for controller-manager functions.
I just make it has consist rule with clusterrole.

@marun

### DIFF
--- a/charts/federation-v2/charts/controllermanager/templates/role.yaml
+++ b/charts/federation-v2/charts/controllermanager/templates/role.yaml
@@ -14,29 +14,53 @@ rules:
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - update
 - apiGroups:
   - multiclusterdns.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - create
+  - update
 - apiGroups:
   - core.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - create
+  - update
 - apiGroups:
   - types.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - update
 - apiGroups:
   - clusterregistry.k8s.io
   resources:
-  - '*'
+  - clusters
   verbs:
-  - '*'
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - secrets
+  - events
+  verbs:
+  - get
+  - create
+  - update
 {{- end }}
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -190,17 +190,19 @@ if [[ ! "${NAMESPACED}" ]]; then
   fi
 fi
 
+USE_CHART="${USE_CHART:-}"
 # Create a permissive clusterrolebinding to allow federation controllers to run.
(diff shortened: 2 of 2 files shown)


marun:
If the aim is to grant cluster-admin privileges, I think it is preferable to just grant that role to the service account.  And while I think what you're proposing is a good stop-gap, I've filed #687 to ensure that more restrictive permissions are implemented in time for beta1.

xunpan:
Based on discussion in #687 and local test for unmanaged namespaced test for namespaced chart deployment. I update the fix for namespaced federation with chart.

After this is done, I will fix cluster scoped chart and all script part.

charts/federation-v2/charts/controllermanager/templates/role.yaml
```diff
@@ -12,31 +12,69 @@ rules:
 - apiGroups:
   - scheduling.federation.k8s.io
   resources:
-  - '*'
```
marun: For federation api groups, I think `*` is preferable.  Otherwise maintenance will be required, and I don't think there is any advantage to requiring that.

charts/federation-v2/charts/controllermanager/templates/role.yaml
```diff
@@ -12,31 +12,69 @@ rules:
...
+  - replicaschedulingpreferences
   verbs:
-  - '*'
```
marun: Is `*` not equivalent to `get/watch/list/update`?

xunpan: No. It is not equivalent, e.g. it has no `create`. Because this is created by client but not controller.

charts/federation-v2/charts/controllermanager/templates/role.yaml
xunpan: Actually, when I'm doing this, I'm thinking it is a balance between convenience and security. My preferred solution is to use * in `resources` and `verbs` for all api groups created by federation because it seems less peopole will reuse federation groups for their own purpose. So, it is convenient in extendibility for federation controller manages their own resource groups with full permission. 

However, I'm not full of knowledge about K8s RBAC. And strict checking is fine for most of case. So, I follow the rules metioned in #687 .

For this comments, did you mean for all api groups? If true, what is the purpose to metion each resource independently in #687, I'm a little confused.

marun: I meant that `*` should be used for resources for all federation groups.

xunpan: all federated resource groups with resource '*'. `clusterregistry.k8s.io` is an exception because we only need `cluster` in this independent maintained group.

xunpan: @marun

charts/federation-v2/charts/controllermanager/templates/role.yaml
```diff
@@ -14,29 +14,53 @@ rules:
...
+  - watch
+  - list
+  - update
 - apiGroups:
   - multiclusterdns.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
```
marun: Shouldn't this be *?

marun:
Thank you!  This PR ensures that the helm chart will have sufficient permissions to run, and we can refine it further in the run up to beta.

PR #685, opened 2019-03-22
add FederatedConfig resource

xunpan:
This is the first PR of fixing https://github.com/kubernetes-sigs/federation-v2/issues/114

Add new API `FederatedConfig` to record configurations of controller manager and be used by command line on demand.

It is intend to replace previous `ConfigMap` PR https://github.com/kubernetes-sigs/federation-v2/pull/648


marun:
@xunpan Please remove the generated controller.

@shashidharatd Is it necessary for the generated client to be committed for a new type?

xunpan:
@marun, I removed the controller because test related issue.

shashidharatd:
> @shashidharatd Is it necessary for the generated client to be committed for a new type?

@marun, when we do `kubebuilder generate`, kb is generating the generated client, and it looks like there is no way to switch off the auto generation from the current version of kubebuilder (maybe i am not aware of it). So as a temporary measure, we could keep these generated clients and remove totally in relevant pr, when we upgrade kubebuilder version.

pkg/apis/core/v1alpha1/federatedconfig_types_test.go
```diff
@@ -0,0 +1,83 @@
+/*
```
shashidharatd: please do not commit the api test files.

pkg/apis/core/v1alpha1/v1alpha1_suite_test.go
```diff
@@ -0,0 +1,52 @@
+/*
```
shashidharatd: Same as previous comment.

pkg/inject/zz_generated.kubebuilder.go
```diff
@@ -68,12 +72,18 @@ func init() {
 
 		// Add Kubernetes informers
 
+		if c, err := federatedconfig.ProvideController(arguments); err != nil {
+			return err
+		} else {
+			arguments.ControllerManager.AddController(c)
+		}
```
shashidharatd: This snippet needs to removed as we are not using the generated controller.

pkg/controller/federatedconfig/controller.go
```diff
@@ -0,0 +1,88 @@
+/*
```
shashidharatd: please remove the controller code

pkg/apis/core/v1alpha1/federatedconfig_types.go
```diff
@@ -29,12 +31,35 @@ import (
 type FederatedConfigSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "kubebuilder generate" to regenerate code after modifying this file
+
+	LimitedScope       bool                 `json:"limited-scope"`
+	RegistryNamespace  string               `json:"registry-namespace,omitempty"`
+	ControllerDuration DurationConfig       `json:"controller-duration"`
+	LeaderElect        LeaderElectConfig    `json:"leader-elect"`
+	FeatureGates       []FeatureGatesConfig `json:"feature-gates"`
```
shashidharatd: Many of the fields does not have omitempty json tag, although they are optional and can have default setting.

pkg/apis/core/v1alpha1/federatedconfig_types.go
```diff
@@ -29,12 +31,35 @@ import (
...
+type FeatureGatesConfig struct {
+	Name    string `json:"name"`
+	Enabled bool   `json:"enabled"`
 }
 
 // FederatedConfigStatus defines the observed state of FederatedConfig
 type FederatedConfigStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "kubebuilder generate" to regenerate code after modifying this file
+
```
shashidharatd: unnecessary.

hack/sample/federatedconfig.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: core.federation.k8s.io/v1alpha1
```
shashidharatd: We do not store sample for any other types. remove this please.

hack/install-latest.yaml
```diff
@@ -164,6 +164,40 @@ status:
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
+metadata:
```
shashidharatd: IIRC, `hack/install-latest.yaml` should be checked in by release build or while cutting the release manually.

pkg/apis/core/v1alpha1/federatedconfig_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// Important: Run "kubebuilder generate" to regenerate code after modifying this file
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// FederatedConfig
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:path=federatedconfigs
+type FederatedConfig struct {
```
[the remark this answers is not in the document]
xunpan: makes sense. I will handle it.

pkg/apis/core/v1alpha1/federatedconfig_types_test.go
xunpan: It was deleted. I will merge 2 commits into 1 to make it clear: the deletion files do not need to be committed at first time.

hack/install-latest.yaml
xunpan: This is changed to avoid CI fail. It seems not harmful to set a new CRD. 
Did you mean we need to change current CI not fail for potential case?

xunpan: after sync-up with @shashidharatd, we agree to keep this change. We will have discussion later to see if install-latest.yaml is only for release but not in code.

marun: `install-latest.yaml` has nothing to do with release and needs be maintained for any PR that would affect its contents.

marun:
Functionally this looks good.  I'm not sure about the name, though - `FederationConfig` seems a bit generic.  Maybe we can bikeshed in tomorrow's meeting?

marun:
I think this is a good starting point and in keeping with the discussion we had in the wg this morning.  We can always revise before beta (including moving it to another group) if needed.

/approve
/lgtm

Issue #676, opened 2019-03-21
The managed e2e suite is sometimes failing CI on shutdown due to a panic in the scheduling controller


marun:
Relevant portion of the CI log captured in a gist: https://gist.github.com/marun/df4bef5cd9cf5b5721f423144e219c81

I'm not able to reproduce the failure locally, and it doesn't always occur in CI.  When I run the managed e2e suite, though, I do see a ton of error log spew that appears to be from the scheduling controller.  I'm assuming that the scheduling controller is not being shutdown cleanly and under the right circumstances this results in a full-blown panic.

PR #689, opened 2019-03-24
Refactor scheduling manager to correct goroutine usage

irfanurrehman:
Fixes https://github.com/kubernetes-sigs/federation-v2/issues/676

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -145,8 +145,8 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
 	}
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulerPreferences) {
-		if _, err := schedulingmanager.StartSchedulerController(opts.Config, stopChan); err != nil {
-			glog.Fatalf("Error starting scheduler controller: %v", err)
+		if _, err := schedulingmanager.StartSchedulingManager(opts.Config, stopChan); err != nil {
+			glog.Fatalf("Error starting scheduling manager: %v", err)
 		}
 	}
 
--- a/pkg/controller/schedulingmanager/controller.go
+++ b/pkg/controller/schedulingmanager/controller.go
@@ -17,25 +17,21 @@ limitations under the License.
 package schedulingmanager
 
 import (
-	"sync"
-
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
+
+	"k8s.io/apimachinery/pkg/util/runtime"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
 	corev1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/schedulingpreference"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 	"github.com/kubernetes-sigs/federation-v2/pkg/schedulingtypes"
-
-	"k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
-	restclient "k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/cache"
 )
 
-type SchedulerController struct {
-	sync.RWMutex
+type SchedulingManager struct {
 	// Store for the FederatedTypeConfig objects
 	store cache.Store
 	// Informer for the FederatedTypeConfig objects
@@ -43,37 +39,56 @@ type SchedulerController struct {
 
 	worker util.ReconcileWorker
 
-	scheduler map[string]schedulingtypes.Scheduler
+	//schedulers map[string]*SchedulerWrapper
+	schedulers *util.SafeMap
 
 	config *util.ControllerConfig
-
-	runningPlugins sets.String
-	// mapping qualifiedname to template kind for managing plugins in scheduler
-	federatedKindMap map[string]string
 }
 
-func StartSchedulerController(config *util.ControllerConfig, stopChan <-chan struct{}) (*SchedulerController, error) {
+type SchedulerWrapper struct {
+	// To signal shutdown of scheduler and any associated routine.
+	stopChan chan struct{}
+	// Mapping qualifiedname to federated kind for managing plugins in scheduler.
+	// This is needed because typeconfig could be of any name and we run plugins
+	// by federated kinds (eg FederatedDeployment). This also avoids running multiple
+	// plugins in case multiple typeconfigs are created for same federated kind.
+	pluginMap *util.SafeMap
+	// Actual scheduler.
+	schedulingtypes.Scheduler
+}
 
-	userAgent := "SchedulerController"
+func (s *SchedulerWrapper) HasPlugin(typeConfigName string) bool {
+	_, ok := s.pluginMap.Get(typeConfigName)
+	return ok
+}
+
+func StartSchedulingManager(config *util.ControllerConfig, stopChan <-chan struct{}) (*SchedulingManager, error) {
+	userAgent := "SchedulingManager"
 	kubeConfig := config.KubeConfig
 	restclient.AddUserAgent(kubeConfig, userAgent)
 
-	controller, err := newController(config)
+	manager, err := newSchedulingManager(config)
 	if err != nil {
 		return nil, err
 	}
 
-	glog.Infof("Starting scheduler controller")
-	controller.Run(stopChan)
-	return controller, nil
+	glog.Infof("Starting scheduling manager")
+	manager.Run(stopChan)
+	return manager, nil
 }
 
-func newController(config *util.ControllerConfig) (*SchedulerController, error) {
-	c := &SchedulerController{
-		config:           config,
-		scheduler:        make(map[string]schedulingtypes.Scheduler),
-		runningPlugins:   make(sets.String),
-		federatedKindMap: make(map[string]string),
+func newSchedulerWrapper(schedulerInterface schedulingtypes.Scheduler, stopChan chan struct{}) *SchedulerWrapper {
+	return &SchedulerWrapper{
+		stopChan:  stopChan,
+		pluginMap: util.NewSafeMap(),
+		Scheduler: schedulerInterface,
+	}
+}
+
+func newSchedulingManager(config *util.ControllerConfig) (*SchedulingManager, error) {
+	c := &SchedulingManager{
+		config:     config,
+		schedulers: util.NewSafeMap(),
 	}
 
 	c.worker = util.NewReconcileWorker(c.reconcile, util.WorkerTiming{})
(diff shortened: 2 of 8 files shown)


pkg/controller/schedulingmanager/controller.go
```diff
@@ -43,37 +42,55 @@ type SchedulerController struct {
...
-	// mapping qualifiedname to template kind for managing plugins in scheduler
-	federatedKindMap map[string]string
+type SchedulerWrapper struct {
+	// To signal shutdown of scheduler and any associated routine.
+	stopChan chan struct{}
+	// Mapping qualifiedname to template kind for managing plugins in scheduler.
+	// This is needed because typeconfig could be of any name and we run plugins
+	// by federated kinds (eg FederatedDeployment). This also avoids running multiple
+	// plugins in case multiple typeconfigs are created for same federated kind.
+	pluginMap *util.SafeMap
```
xunpan: we're using 2 level locks here: a global `sync.RWMutex` for all (`schedulers`) and a local `util.SafeMap` lock for `pluginMap`.  It seems a little complex. Could you explain more about values to make this lock to `pluginMap`

xunpan:
Further, I did not meet this CI error before. Could you explain more about the case that why the issue happens?

cmd/controller-manager/app/controller-manager.go
```diff
@@ -145,7 +145,7 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
 	}
 
 	if utilfeature.DefaultFeatureGate.Enabled(features.SchedulerPreferences) {
-		if _, err := schedulingmanager.StartSchedulerController(opts.Config, stopChan); err != nil {
+		if _, err := schedulingmanager.StartSchedulingManager(opts.Config, stopChan); err != nil {
 			glog.Fatalf("Error starting scheduler controller: %v", err)
```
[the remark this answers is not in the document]
irfanurrehman: Ah, missed that.  Thanks for pointing it out, update!

pkg/controller/schedulingmanager/controller.go
```diff
@@ -43,37 +42,55 @@ type SchedulerController struct {
...
 	kubeConfig := config.KubeConfig
 	restclient.AddUserAgent(kubeConfig, userAgent)
 
-	controller, err := newController(config)
+	manager, err := newSchedulingManager(config)
 	if err != nil {
 		return nil, err
 	}
 
 	glog.Infof("Starting scheduler controller")
```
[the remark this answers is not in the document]
irfanurrehman: Update!

pkg/controller/schedulingmanager/controller.go
irfanurrehman: The general recommendation if using locks is to lock only for the scope needed and make that scope as small as possible. The lock in pluginMap is map specific and locks the reads and writes to it; The lock scope is minimal and thus locking overhead is minimal too. The safe map is recommended for usage in rest of the code too to avoid races.

The controller local mutex is mainly to protect rest of the data in reconcile from races (`SchedulingManager.schedulers` and `SchedulerWrapper.stopChan`). I could have used this same mutex to protect `pluginMap` too but chose otherwise as it makes one variable less in the overhead of maintaining all read/write accesses whenever we change code next.

irfanurrehman:
> Further, I did not meet this CI error before. Could you explain more about the case that why the issue happens?

The earlier code leaked running goroutines; Just run the e2e locally in managed mode. You do not need to run all tests you can simply run the managed test with `--ginkgo.focus='Scheduling'`. You will find that some informers continue running even after cleanup and report errors as listed in #676. Under right conditions it causes panic too as reported by the failed test in #675.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -152,76 +189,59 @@ func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.R
...
 	defer c.Unlock()
-	scheduler, ok := c.scheduler[schedulingKind]
+	scheduler, ok := c.schedulers[schedulingKind]
 	if !ok {
 		return
 	}
 
-	if !c.runningPlugins.Has(qualifiedName.Name) {
+	// This typeConfig might not have an associated plugin but another type could.
+	if !scheduler.HasPlugin(typeConfigName) && scheduler.pluginMap.Size() > 0 {
```
xunpan: It seems no need to add `scheduler.pluginMap.Size() > 0`

It seems try to cover the case that `!scheduler.HasPlugin(typeConfigName)` and `scheduler.pluginMap.Size() == 0`. If that's true, the later logic for handling plugin seems make no sense, right?

pkg/controller/schedulingmanager/controller.go
```diff
@@ -93,53 +110,73 @@ func newController(config *util.ControllerConfig) (*SchedulerController, error)
...
+	go func() {
+		<-stopChan
+		c.shutdown()
+	}()
+}
+
+func (c *SchedulingManager) shutdown() {
+	for _, scheduler := range c.schedulers {
+		// Indicate scheduler and associated goroutines to be stopped in schedulingpreference controller.
+		close(scheduler.stopChan)
```
xunpan: we need to stop plugins if shutdown.

xunpan:
@irfanurrehman 
Got it. It is clear to me know. Thanks for your explanation.

pkg/controller/schedulingmanager/controller.go
irfanurrehman: Nope, its covering the case `!scheduler.HasPlugin(typeConfigName)` and `scheduler.pluginMap.Size() > 0` exactly as in code. In other words this stop is not for the plugin associated with this typeConfig, but there is/are some other plugin(s) still associated with this scheduler.
Looking at your query I figured I made the condition check unnecessarily complex. A better way of expressing the same thing is the updated code. PTAL.

pkg/controller/schedulingmanager/controller.go
irfanurrehman: That was the leak in earlier code :-), which is now stopped via this stopchannel [here](https://github.com/kubernetes-sigs/federation-v2/blob/1804943a6eed0d349f0f930c39f439c094e71378/pkg/controller/schedulingpreference/controller.go#L175) and [here](https://github.com/kubernetes-sigs/federation-v2/blob/1804943a6eed0d349f0f930c39f439c094e71378/pkg/schedulingtypes/replicascheduler.go#L149).

pkg/controller/schedulingmanager/controller.go
xunpan: The result is what I expected. Thanks :-)

pkg/controller/schedulingmanager/controller.go
xunpan: got it. Thanks.

xunpan:
LGTM

I did not mark it to let others have a chance to review if needed.

marun:
Thank you for fixing this!  I'll make time to review tomorrow.

irfanurrehman:
> After digging into the code, I'm not clear on why so much locking is required. The `ReconcileWorker` only runs a single goroutine (and likely that should remain the case for the scheduling manager), so the scheduler code is essentially synchronous. The only case of another process needing state is the e2e code which needs to wait for plugins to start. Would it be enough to use a `SafeMap` to track schedulers and do away with the other locking?
> 
> I'll wait for you answer before doing any more review.

Well, the need of locking did arise only because the test code depends on the access to the info in maps (`pluginMap` and `schedulers`) and if not locked, the testing with `-race` cribs. If we are sure we won't change the worker to run on more threads atleast for now, I can change the usage of two locks to one. On the other hand if we are sure that the code would be single threaded, there is virtually no overhead of the locks (no waiting thread ever). Let me know your suggestions.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -43,37 +42,55 @@ type SchedulerController struct {
...
+}
 
-	runningPlugins sets.String
-	// mapping qualifiedname to template kind for managing plugins in scheduler
-	federatedKindMap map[string]string
+type SchedulerWrapper struct {
+	// To signal shutdown of scheduler and any associated routine.
+	stopChan chan struct{}
+	// Mapping qualifiedname to template kind for managing plugins in scheduler.
+	// This is needed because typeconfig could be of any name and we run plugins
```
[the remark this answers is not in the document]
irfanurrehman: I meant the user can create a FederatedTypeConfig with any name. Right now there is no stopping the user from doing that. We need to put in a prescription/recommendation somewhere in docs if the user creates FederatedTypeConfigs outside the installation method(s) we provide.

irfanurrehman:
> > After digging into the code, I'm not clear on why so much locking is required. The `ReconcileWorker` only runs a single goroutine (and likely that should remain the case for the scheduling manager), so the scheduler code is essentially synchronous. The only case of another process needing state is the e2e code which needs to wait for plugins to start. Would it be enough to use a `SafeMap` to track schedulers and do away with the other locking?
> > I'll wait for you answer before doing any more review.
> 
> Well, the need of locking did arise only because the test code depends on the access to the info in maps (`pluginMap` and `schedulers`) and if not locked, the testing with `-race` cribs. If we are sure we won't change the worker to run on more threads atleast for now, I can change the usage of two locks to one. On the other hand if we are sure that the code would be single threaded, there is virtually no overhead of the locks (no waiting thread ever). Let me know your suggestions.

EDIT: Additionally using `util.SafeMap` totally avoids complex readability by hiding the lock and keeps the lock scope minimal. What say about using the same for both maps and doing away with exposed locking local to the code.

marun:
Please do simplify to just using SafeMap.

pkg/controller/schedulingmanager/controller.go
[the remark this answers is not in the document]
marun: The type config name has to be <plural name>.<group>.  This is validated by `util.CheckTypeConfigName` today and in the future we'll use a validation webhook.

irfanurrehman:
> Please do simplify to just using SafeMap.

@marun Updated, PTAL!

pkg/controller/schedulingmanager/controller.go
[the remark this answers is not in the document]
marun: @irfanurrehman Are you attending to address my comment on this line?

pkg/controller/schedulingmanager/controller.go
```diff
@@ -152,76 +182,56 @@ func (c *SchedulerController) reconcile(qualifiedName util.QualifiedName) util.R
...
-	// if all resources registered to same scheduler are deleted, the scheduler should be stopped
-	resources := schedulingtypes.GetSchedulingKinds(qualifiedName.Name)
-	result := c.runningPlugins.Intersection(resources)
-	if result.Len() == 0 {
-		glog.Infof("Stopping scheduler schedulingpreference controller for %q", schedulingKind)
-		scheduler.Stop()
-
-		delete(c.scheduler, schedulingKind)
+	scheduler := abstractScheduler.(*SchedulerWrapper)
+	if scheduler.HasPlugin(typeConfigName) {
```
marun: (No action required) I think `HasPlugin` is redundant here:

```golang
kind, ok := scheduler.pluginMap.Get(typeConfigName)
if !ok {
  ...
}
```

(No action required) I also think kind should be cast once rather than in multiple places.

pkg/controller/schedulingmanager/controller.go
```diff
@@ -17,63 +17,78 @@ limitations under the License.
...
+	//schedulers map[string]*SchedulerWrapper
+	schedulers *util.SafeMap
 
 	config *util.ControllerConfig
+}
 
-	runningPlugins sets.String
-	// mapping qualifiedname to template kind for managing plugins in scheduler
-	federatedKindMap map[string]string
+type SchedulerWrapper struct {
```
marun: (No action necessary) I'm not sure I understand the point of SchedulerWrapper.  What is the advantage to including these fields here vs on `schedulingtypes.Scheduler` (along with the necessary methods to interact with them)?

marun:
@irfanurrehman Thanks for cleaning this up!  I have some comments that could be addressed at some point, but I'd rather merge this to fix CI than iterate until its perfect.  Feel free to submit a follow-on PR if you're inclined to address any of my comments.

/lgtm

PR #698, opened 2019-03-27
Fix retain replicas field type (was bool, should be boolean)

marun:


marun:
@shashidharatd This bug was discovered in testing against openshift based on kube 1.13, and reveals that our use of an old version of kubebuilder has also influenced the version of kube-apiserver that we are testing against in a managed test run (since we source it from the kubebuilder archive and the version we uses supplies 1.12).  This adds motivation to keeping pace with kubebuilder releases so that we're not †esting against an old version of kube.  I've also filed an issue (#699) to track updating CI to use a more recent release of kube (likely 1.13 until kind and kubebuilder update to 1.14).

shashidharatd:
Thanks @marun for the fix.
Agree with keeping pace with projects which we depend on.
/lgtm

Issue #612, opened 2019-02-26
Minimize the member cluster resources tracked by the control plane


marun:
The federated control plane can be scoped to all namespaces or a single namespace, and will then watch all resources in the selected namespace(s) for enabled types (in all member clusters) whether or not those resources are federated.  Since each tracked resource imposes a memory cost on the federation control plane, it would be desirable to minimize memory usage by not tracking resources that were not federated.

Possible solutions:

 - Where federated resources existed in a subset of available namespaces, support limiting the scope of the control plane to a subset of available namespaces (via a label selector) to limit caching to the selected namespaces.
   - Implementation note: Simple to implement, informers already support filtering by label.
 - Where federated resources potentially existed in all namespaces, support limiting the scope of the control plane to resources identified as being federated.  Resources could be labeled on propagation and the control plane would only track resources with the expected labeling. (this was suggested by **Anders Janmyr** on slack)
   - Implementation note: Propagation logic would have to be updated to handle the case where an un-labeled resource existed in a member cluster and needed to be 'adopted'.  Since the resource wouldn't exist in the informer, the federated updater would need to catch `AlreadyExists` on a call to create, retrieve the resource to retain fields, and then perform an update.

Edit: I'm implementing the second option.

TODO:

 - [x] - Refactor cluster field retention for reuse (#707)
 - [x] - Simplify placement computation (#708)
 - [x] - Refactor sync controller reconcile loop (#713)
 - [x] - Switch to having the FederatedNamespace define the template (#714)
 - [x] - Only watched managed resources (#715)
 - [x] - Add helpers to [ check for | add | remove ] the managed label (#788)
 - [x] - Refactor updater for extensibility (#785)
 - [x] - Factor unmanaged dispatcher out of the managed dispatcher (#789)
 - [x] - Add deletion handling with the new updater to the sync controller (#735)
 - [ ] - Ensure managed label is removed from unmanaged resources (#724)

/kind feature

quinton-hoole-2:
@marun Have we measured what that memory cost is?

PR #707, opened 2019-04-01
Refactor cluster field retention for reuse

marun:
Previously the methods to retain cluster fields were in the same file as the sync controller and returned the desired object even though the
original was being changed in-place.  This change isolates the functions in a separate file to simplify maintenance, no longer returns the
desired object, and adds a function doc explaining that the object is modified in-place.

1st in a series targeting #612


yamt:
lgtm

PR #708, opened 2019-04-01
Simplify placement computation

marun:
Placement computation is updated to return a set of the selected clusters rather than slices of both selected and unselected clusters.  This is in anticipation of collapsing the current sync loops for add/update and deletion into a single sync loop.

2nd in a series targeting #612


yamt:
lgtm

PR #710, opened 2019-04-01
make federation-role has enough permission for cluster-scope federation deployment

xunpan:
cluster-admin is removed for the controller role binding in #673

However,  permission is not set correctly to have permission to make controller work.

- [ ] document update to tell `CRD` federation needs RBAC

### DIFF
--- a/charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
+++ b/charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
@@ -13,35 +13,57 @@ rules:
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - update
 - apiGroups:
   - multiclusterdns.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - create
+  - update
+  - delete
 - apiGroups:
   - core.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - create
+  - update
 - apiGroups:
   - types.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - get
+  - watch
+  - list
+  - update
 - apiGroups:
   - clusterregistry.k8s.io
   resources:
-  - '*'
+  - clusters
   verbs:
-  - '*'
+  - get
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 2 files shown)


test/e2e/crd.go
```diff
@@ -43,9 +46,69 @@ import (
 	. "github.com/onsi/ginkgo"
 )
 
+const crdGroup string = "example.com"
```
marun: I would prefer that you follow the example I set in #673:

https://github.com/kubernetes-sigs/federation-v2/pull/673/files#diff-39161408e007fa958c893de39a5332d3R85

I don't think it is desirable for this crd testing to validate kube's rbac implementation, and changing the group and version to something the test will have permission for allows the test to run with minimal effort.

charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
```diff
@@ -13,35 +13,57 @@ rules:
...
+  - get
+  - watch
+  - list
+  - update
 - apiGroups:
   - multiclusterdns.federation.k8s.io
   resources:
   - '*'
   verbs:
-  - '*'
```
marun: nit: If all verbs are allowed, why not use `*`?

charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
```diff
@@ -13,35 +13,57 @@ rules:
...
+  - clusters
   verbs:
-  - '*'
+  - get
 - apiGroups:
   - ""
   resources:
-  - '*'
+  - configmaps
```
marun: What does federation use configmaps for?

I'm thinking that federation should only have access to secrets in the federation system namespace.  I'm assuming that would require a role rather than a cluster role?

xunpan: it is leader elector which uses `configmap`

charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
xunpan: a full supported kube verb list should be `get, list, watch, create, update, patch, delete, deletecollection, and proxy`. So it is different.

test/e2e/crd.go
xunpan: ok. I will rebase it.

marun:
Thank you!

/approve
/lgtm

PR #713, opened 2019-04-02
Refactor sync controller reconcile loop

marun:
The existing federated updater required that cluster operations be pre-computed.  This complicated the proposal to cleanly implement support for limiting the federation informers to managed resources and then adopt resources as necessary (which would involve converting an add to an update).

This PR refactors the sync controller's reconcile loop and implements a new updater to allow cluster operations to be dispatched individually.  It also ensures that reconciliation events are logged on federated resources where possible.

The current federated updater is left in place for now since the deletion helper is still using it.  Future PRs in this series are intended to rewrite the deletion helper to use the new updater and remove the old federated updater.

3rd in a series targeting #612

### DIFF
--- a/pkg/controller/sync/updater.go
+++ b/pkg/controller/sync/updater.go
@@ -1,0 +1,221 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sync
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+)
+
+type FederatedUpdater interface {
+	NoChanges() bool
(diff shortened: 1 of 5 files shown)


marun:
Rebased

marun:
Rebased

pkg/controller/sync/updater.go
```diff
@@ -0,0 +1,222 @@
...
+	u.operationCount += 1
+	const op = "create"
+	go u.clusterOperation(clusterName, op, func(client util.ResourceClient) (string, error) {
+		u.recordEvent(clusterName, op, "Creating")
+
+		obj, err := u.fedResource.ObjectForCluster(clusterName)
+		if err != nil {
+			return "", err
+		}
+		createdObj, err := client.Resources(obj.GetNamespace()).Create(obj, metav1.CreateOptions{})
```
shashidharatd: (query) Is there a possibility that an client operation gets stuck and never return? If that happens there is a possibility that the go routine would be leaked and many such incidences would keep increasing memory.

marun: The behavior here is the same as the old updater, and any request should eventually timeout (it just might not be within 30s).  As per a TODO in this file, I am intending to switch to the generic client to take advantage of context cancellation, which should ensure immediate cleanup rather than timeout.

marun:
Rebased

pkg/controller/sync/updater.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (u *federatedUpdaterImpl) Delete(clusterName string) {
+	u.operationCount += 1
+	const op = "delete"
+	go u.clusterOperation(clusterName, op, func(client util.ResourceClient) (string, error) {
+		u.recordEvent(clusterName, op, "Deleting")
+
+		qualifiedName := u.fedResource.TargetName()
+		orphanDependents := false
+		err := client.Resources(qualifiedName.Namespace).Delete(qualifiedName.Name, &metav1.DeleteOptions{OrphanDependents: &orphanDependents})
```
xunpan: `DeleteOptions` works but it is preferred to use `PropagationPolicy` instead.

marun: Good catch.  This really shouldn't be provided at all.  Removed.

marun:
Addressed comment, PTAL

pkg/controller/sync/updater.go
shashidharatd: Ack

PR #714, opened 2019-04-02
Switch to having the FederatedNamespace define the template

### LINKED ISSUE #652: Changes to the namespace containing a FederatedNamespace are not propagated
**What happened**:

- In a federation of more than one cluster, I created a FederatedNamespace.
- I checked that the namespace was propagate to the second cluster
- I added a label to the namespace containing the FederatedNamespace
- The namespace in the second cluster was not updated with the label

- The namespace in the second cluster should have been updated with the label

/kind bug

marun:
Previously the template for a FederatedNamespace namespace was sourced from the containing namespace.  This divergence from the norm of sourcing the template from a field in the federated resource was a source of user confusion.

Fixes #652 

4th in a series targeting #612

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -330,7 +330,7 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
 
 		// Resource should not exist in the named cluster
 		if !selectedClusterNames.Has(clusterName) {
-			if clusterObj != nil && !fedResource.SkipClusterChange(clusterObj) {
+			if clusterObj != nil && !fedResource.SkipClusterDeletion(clusterObj) {
 				updater.Delete(clusterName)
 			}
 			continue
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -328,13 +328,9 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
 			clusterObj = rawClusterObj.(*unstructured.Unstructured)
 		}
 
-		if clusterObj != nil && fedResource.SkipClusterChange(clusterObj) {
-			continue
-		}
-
 		// Resource should not exist in the named cluster
 		if !selectedClusterNames.Has(clusterName) {
-			if clusterObj != nil {
+			if clusterObj != nil && !fedResource.SkipClusterChange(clusterObj) {
 				updater.Delete(clusterName)
 			}
 			continue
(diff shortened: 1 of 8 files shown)


marun:
cc: @andersjanmyr

marun:
Rebased

marun:
Rebased

shashidharatd:
Overall LGTM. This is step in the right direction. This would probably simplify user understanding. Would leave it to other reviewers to comment and merge.

pkg/controller/sync/resource.go
```diff
@@ -158,45 +155,29 @@ func (r *federatedResource) SkipClusterChange(clusterObj pkgruntime.Object) bool
 
 // TODO(marun) Marshall the template once per reconcile, not per-cluster
 func (r *federatedResource) ObjectForCluster(clusterName string) (*unstructured.Unstructured, error) {
-	// Federation of namespaces uses Namespace resources as the
-	// template for resource creation in member clusters. All other
-	// federated types rely on a template type distinct from the
-	// target type.
-	//
-	// Namespace is the only type that can contain other resources,
```
irfanurrehman: Doesn't this comment still hold good?

pkg/controller/sync/controller.go
```diff
@@ -328,13 +328,9 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
 		}
 
-		if clusterObj != nil && fedResource.SkipClusterChange(clusterObj) {
-			continue
-		}
-
 		// Resource should not exist in the named cluster
 		if !selectedClusterNames.Has(clusterName) {
-			if clusterObj != nil {
+			if clusterObj != nil && !fedResource.SkipClusterChange(clusterObj) {
```
irfanurrehman: Github review does not allow to give comment on most of the portion of unchanged code.
The comments in the definition of this function might need a little change now.

```
        // The Namespace type is a special case because it is the only
	// container in the Kubernetes API.  Federation presumes a
	// separation between the template and target resources, but a
	// namespace in the host cluster is necessarily both template and
	// target.
```

irfanurrehman:
@marun few nits in comments; overall lgtm!

pkg/controller/sync/resource.go
marun: The comment still holds, but it does not belong in this function.  I've added a note to our technical writer to have mention of it added to the documentation.

pkg/controller/sync/controller.go
marun: Good catch.  I've changed the name of this function to `SkipClusterDelete` and updated the documentation accordingly (including the block deleted from `ObjectForCluster`).

marun:
Rebased and added commit to `Explain why sync controller never deletes namespaces in host cluster`.

xunpan:
It is clear now. LGTM

shashidharatd:
Taking the liberty to add LGTM :) as the comments raised by @irfanurrehman have been addressed.
/lgtm
Thanks @marun

PR #715, opened 2019-04-02
Only watched managed resources

### LINKED ISSUE #549: resource in joined cluster will be deleted if federatedresource is created with the same name
**What would you like to be added**: I create a deployment named nginx in joined cluster,then create a nginx federateddeployment in the kubernetes federation, the old nginx deployment will be deleted ,with nginx federateddeployment create new one.

**Why is this needed**:This will causing business disruption

/kind feature

### LINKED ISSUE #612: Minimize the member cluster resources tracked by the control plane

marun:
This PR labels managed resources with `federation.k8s.io/managed: true` and the federated informer will only watch resources so labeled.  Existing resources will be updated to include the label, and this is implicitly tested by the crud test for namespaces (since the namespace in the host cluster exists and will be adopted).

The label will be removed from namespaces in the host cluster in the event of a propagation change resulting in the host cluster not existing in the selected set.  The label is not yet removed from a namespace in the host cluster when a FederatedNamespace is deleted.  The label is also not yet removed from an orphaned resource.  Both of these changes will be made in a subsequent PR that rewrites the deletion helper.

Fixes #549

TODO:
 - [x] - Add test that federated resource without placement for a member cluster does not delete an unlabeled resource in that cluster. 

5th in a series targeting #612

### DIFF
--- a/pkg/controller/sync/updater.go
+++ b/pkg/controller/sync/updater.go
@@ -19,7 +19,7 @@ package sync
 import (
 	"fmt"
 	"strings"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
@@ -47,14 +47,12 @@ type updaterResult struct {
 }
 
 type federatedUpdaterImpl struct {
-	sync.RWMutex
-
 	fedView util.FederationView
 
 	fedResource FederatedResource
 
 	resultChan     chan updaterResult
-	operationCount int
+	operationCount int32
 
 	timeout time.Duration
 }
@@ -69,7 +67,7 @@ func NewFederatedUpdater(fedView util.FederationView, fedResource FederatedResou
 }
 
 func (u *federatedUpdaterImpl) NoChanges() bool {
-	return u.operationCount == 0
+	return atomic.LoadInt32(&u.operationCount) == 0
 }
 
 func (u *federatedUpdaterImpl) Wait() (map[string]string, bool) {
@@ -78,7 +76,7 @@ func (u *federatedUpdaterImpl) Wait() (map[string]string, bool) {
 
 	timedOut := false
 	start := time.Now()
-	for i := 0; i < u.operationCount; i++ {
+	for i := int32(0); i < atomic.LoadInt32(&u.operationCount); i++ {
 		now := time.Now()
 		if !now.Before(start.Add(u.timeout)) {
 			timedOut = true
@@ -107,7 +105,7 @@ func (u *federatedUpdaterImpl) Wait() (map[string]string, bool) {
 }
 
 func (u *federatedUpdaterImpl) Create(clusterName string) {
-	u.operationCount += 1
+	u.incrementOperationCount()
 	const op = "create"
 	go u.clusterOperation(clusterName, op, func(client util.ResourceClient) (string, error) {
 		u.recordEvent(clusterName, op, "Creating")
@@ -117,15 +115,29 @@ func (u *federatedUpdaterImpl) Create(clusterName string) {
 			return "", err
 		}
 		createdObj, err := client.Resources(obj.GetNamespace()).Create(obj, metav1.CreateOptions{})
-		if err != nil {
+		if err == nil {
+			return util.ObjectVersion(createdObj), nil
+		}
+		// TODO(marun) Figure out why attempting to create a namespace that
+		// already exists indicates ServerTimeout instead of AlreadyExists.
+		alreadyExists := apierrors.IsAlreadyExists(err) || u.fedResource.TargetKind() == util.NamespaceKind && apierrors.IsServerTimeout(err)
+		if !alreadyExists {
 			return "", err
 		}
-		return util.ObjectVersion(createdObj), err
+
+		// Attempt to update the existing resource to ensure that it
+		// is labeled as a managed resource.
+		clusterObj, err := client.Resources(obj.GetNamespace()).Get(obj.GetName(), metav1.GetOptions{})
+		if err != nil {
+			return "", errors.Wrapf(err, "Failed to retrieve object potentially requiring adoption for cluster %q", clusterName)
+		}
+		u.Update(clusterName, clusterObj)
+		return "", errors.Errorf("An update will be attempted instead of a creation due to an existing resource in cluster %q", clusterName)
 	})
 }
 
 func (u *federatedUpdaterImpl) Update(clusterName string, clusterObj *unstructured.Unstructured) {
-	u.operationCount += 1
+	u.incrementOperationCount()
 	const op = "update"
 	go u.clusterOperation(clusterName, op, func(client util.ResourceClient) (string, error) {
 		obj, err := u.fedResource.ObjectForCluster(clusterName)
@@ -160,7 +172,7 @@ func (u *federatedUpdaterImpl) Update(clusterName string, clusterObj *unstructur
 }
 
 func (u *federatedUpdaterImpl) Delete(clusterName string) {
-	u.operationCount += 1
+	u.incrementOperationCount()
 	const op = "delete"
 	go u.clusterOperation(clusterName, op, func(client util.ResourceClient) (string, error) {
 		u.recordEvent(clusterName, op, "Deleting")
@@ -172,6 +184,10 @@ func (u *federatedUpdaterImpl) Delete(clusterName string) {
 		}
 		return "", err
 	})
+}
+
+func (u *federatedUpdaterImpl) incrementOperationCount() {
+	atomic.AddInt32(&u.operationCount, 1)
 }
 
 func (u *federatedUpdaterImpl) clusterOperation(clusterName, op string, opFunc func(util.ResourceClient) (string, error)) {
--- a/pkg/controller/sync/updater.go
+++ b/pkg/controller/sync/updater.go
@@ -38,6 +38,7 @@ type FederatedUpdater interface {
 	Create(clusterName string)
 	Update(clusterName string, clusterObj *unstructured.Unstructured)
 	Delete(clusterName string)
+	RemoveManagedLabel(clusterName string, clusterObj *unstructured.Unstructured)
 }
 
 type updaterResult struct {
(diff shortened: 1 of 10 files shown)


marun:
@shashidharatd @irfanurrehman I realized while working on this that it will fix #549 without further work.  We had discussed making adoption optional, but I don't think that will be necessary.  Resources have always only been deleted if they appear in the informer cache, and as of this PR only labeled resources will appear in that cache.  

I'll add a test to this PR that validates that a non-labeled resource in a given member cluster will not be deleted if a federated resource is created without placement for that cluster.

marun:
Rebased

marun:
Rebased

marun:
Added a new e2e test validating the behavior required by #549.

pkg/controller/sync/updater.go
```diff
@@ -117,15 +115,29 @@ func (u *federatedUpdaterImpl) Create(clusterName string) {
...
 		}
-		return util.ObjectVersion(createdObj), err
+
+		// Attempt to update the existing resource to ensure that it
+		// is labeled as a managed resource.
+		clusterObj, err := client.Resources(obj.GetNamespace()).Get(obj.GetName(), metav1.GetOptions{})
+		if err != nil {
+			return "", errors.Wrapf(err, "Failed to retrieve object potentially requiring adoption for cluster %q", clusterName)
+		}
+		u.Update(clusterName, clusterObj)
```
irfanurrehman: I think it should be possible for the update on a pre-existing object to also be ignored based on user pre-set configuration. I think a safer option would be that the object is not updated if the configuration is missing or not set.

marun: I agree that configurable adoption is not in scope for this PR.  

I think adoption should remain the default, and I am also supportive of your adding a flag (er, `FederationConfig` option) that would ensure that adoption would never occur (i.e. an existing resource in a member cluster would only become managed by federation if a user manually set the label `federation.k8s.io/managed: true`).

marun:
Rebased on #736

marun:
Rebased

xunpan:
LGTM

test/common/crudtester.go
```diff
@@ -482,8 +483,8 @@ func (c *FederatedTypeCrudTester) waitForResource(client util.ResourceClient, qu
 					}
 					// Lacking type information for the override
 					// field, use string conversion as a cheap way to
-					// determine equality.
-					if fmt.Sprintf("%v", expectedValue) != fmt.Sprintf("%v", value) {
+					// determine equality.  Fall back on reflection.
+					if fmt.Sprintf("%v", expectedValue) != fmt.Sprintf("%v", value) && !reflect.DeepEqual(expectedValue, value) {
```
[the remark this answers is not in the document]
marun: I've updated the comment to provide more detail about why string conversion is attempted.  The tl;dr is that the use of interface types and different deserialization paths breaks comparison by reflection for numeric types, and that comparison of the results of string conversion work around that problem.  Note that this is a test-specific issue, since the sync controller doesn't know enough about any given resource to perform comparison on more than the generation and resource version fields.

test/e2e/crud.go
```diff
@@ -44,49 +50,128 @@ var _ = Describe("Federated", func() {
...
-					tl.Fatalf("Error initializing dynamic client: %v", err)
+				typeConfig, testObjectsFunc := getCrudTestInput(f, tl, typeConfigName, fixture)
+				crudTester, targetObject, overrides := initCrudTest(f, tl, typeConfig, testObjectsFunc)
+				crudTester.CheckLifecycle(targetObject, overrides)
+			})
+
+			// Unlabeled resource handling behavior should not vary between
+			// types, so testing a single type is sufficient.  Picking a
+			// namespaced type minimizes the impact of teardown failure.
+			if typeConfigName != "configmaps" || validatedUnlabeledNotDeleted {
```
irfanurrehman: Why is the additional flag required? Isn't there a single fixture per type as of now.

test/e2e/crud.go
```diff
@@ -44,49 +50,128 @@ var _ = Describe("Federated", func() {
...
+			It("should not be deleted if unlabeled", func() {
+				typeConfig, testObjectsFunc := getCrudTestInput(f, tl, typeConfigName, fixture)
+				crudTester, targetObject, _ := initCrudTest(f, tl, typeConfig, testObjectsFunc)
+
+				testClusters := crudTester.TestClusters()
+
+				By("Selecting a member cluster to create an unlabeled resource in")
+				clusterName := ""
+				for key := range testClusters {
+					if len(clusterName) == 0 {
```
irfanurrehman: Did not understand the need of this. `clusterName := ""` is there two lines above.

test/e2e/crud.go
```diff
@@ -44,49 +50,128 @@ var _ = Describe("Federated", func() {
...
 				}
+				clusterClient := genericclient.NewForConfigOrDie(clusterConfig)
+				defer clusterClient.Delete(context.TODO(), unlabeledObj, unlabeledObj.GetNamespace(), unlabeledObj.GetName())
 
-				if framework.TestContext.LimitedScope && !typeConfig.GetNamespaced() {
-					framework.Skipf("Federation of cluster-scoped type %s is not supported by a namespaced control plane.", typeConfigName)
+				By("Intitializing a federated resource with placement excluding all clusters")
+				fedObject, err := federate.FederatedResourceFromTargetResource(typeConfig, unlabeledObj)
+				if err != nil {
+					tl.Fatalf("Error generated federated resource: %v", err)
```
irfanurrehman: s/generated/generating

test/e2e/crud.go
```diff
@@ -44,49 +50,128 @@ var _ = Describe("Federated", func() {
...
+				By("Creating the federated resource")
+				createdObj, err := common.CreateResource(f.KubeConfig(), typeConfig.GetFederatedType(), fedObject)
+				if err != nil {
+					tl.Fatalf("Error creating federated resource: %v", err)
+				}
+				hostClient := genericclient.NewForConfigOrDie(f.KubeConfig())
+				defer hostClient.Delete(context.TODO(), createdObj, createdObj.GetNamespace(), createdObj.GetName())
+
+				waitDuration := 10 * time.Second // Arbitrary amount of time to wait for deletion
+				By(fmt.Sprintf("Checking that the unlabeled resource remains is not deleted within %v", waitDuration))
```
irfanurrehman: s/remains /""/

test/e2e/crud.go
```diff
@@ -44,49 +50,128 @@ var _ = Describe("Federated", func() {
...
+				wait.PollImmediate(framework.PollInterval, waitDuration, func() (bool, error) {
+					obj := unlabeledObj.DeepCopy()
```
irfanurrehman: [Query]For my ignorance; reason of using this in place of an empty object.

irfanurrehman:
@marun few nits and queries in test code, otherwise lgtm.

test/e2e/crud.go
marun: There is no need to validate the unlabeled handling for more than a single type, since the behavior is common across all types.

test/e2e/crud.go
marun: Expediency - I needed an object with the correct group/version/resource and this is simpler than creating a new empty object.

test/e2e/crud.go
marun: Done

test/e2e/crud.go
marun: Done

test/e2e/crud.go
marun: Done

test/e2e/crud.go
marun: Done

test/e2e/crud.go
irfanurrehman: What I meant was, is a check like below not sufficient? (why is `validatedUnlabeledNotDeleted` needed?) :

```
if typeConfigName != "configmaps" {
    return
}
```

irfanurrehman: Am merging this as this is anyways non-functional and don't see a point wasting time on it.

irfanurrehman:
@marun Just realised that this needs a rebase. Please rebase and I will merge.

test/e2e/crud.go
marun: Ah, gotcha.  I recall now that the flag was an attempt to run the test on any type before I realized that a namespaced type was preferable.  Done.

marun:
Rebased

irfanurrehman:
/lgtm
@marun Thanks for doing this.

PR #716, opened 2019-04-02
Upgrade to kubebuilder v1.0.8

shashidharatd:
This PR upgrades the kubebuilder version from v1.0.4 to v1.0.8. Below changes are done as part of this pr.

- Remove generated client/informers/listers for federation apis.
- Remove script based installation method and keep only helm based installation method (helm based installation is still wrapped in scripts as earlier).
- Remove ./hack/install-latest.yaml from tree, as this will no longer be generated.
- Upgrade changes related to API for kubebuilder-v1.0.8. The controller and manager parts are not upgraded.

/cc @marun


shashidharatd:
@marun, @xunpan, This pr is ready for review. PTAL. Thanks !

marun:
@shashidharatd CI seems to be failing consistently.

shashidharatd:
oops, a silly mistake introduced at the last moment. Have fixed it now. @marun, PTAL again after the CI build passes. Thanks!

test/e2e/framework/managed/kubeapi.go
```diff
@@ -30,7 +25,7 @@ import (
 
 // KubernetesApiFixture manages a kubernetes api server
 type KubernetesApiFixture struct {
-	EtcdUrl   *url.URL
+	Etcd      *integration.Etcd
```
marun: (No action required) Why have you decided to switch to running an etcd instance per api server (vs the old strategy of reusing a single etcd instance)?  

I'm inclined to give this issue a pass anyway (as per #718), but I was curious as to why you decided to make this change.

test/e2e/framework/managed/controller.go
```diff
@@ -89,12 +89,15 @@ func NewIngressDNSControllerFixture(tl common.TestLogger, config *util.Controlle
...
 
 // NewClusterControllerFixture initializes a new cluster controller fixture.
-func NewClusterControllerFixture(config *util.ControllerConfig) *ControllerFixture {
+func NewClusterControllerFixture(tl common.TestLogger, config *util.ControllerConfig) *ControllerFixture {
 	f := &ControllerFixture{
 		stopChan: make(chan struct{}),
 	}
 	monitorPeriod := 1 * time.Second
-	federatedcluster.StartClusterController(config, f.stopChan, monitorPeriod)
+	err := federatedcluster.StartClusterController(config, f.stopChan, monitorPeriod)
```
marun: Nice catch!

test/common/typeconfig.go
```diff
@@ -21,13 +21,12 @@ import (
...
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	client "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
 )
 
 func GetTypeConfig(genericClient client.Client, name, namespace string) (typeconfig.Interface, error) {
 	typeConfig := &fedv1a1.FederatedTypeConfig{}
-	key := client.ObjectKey{Name: name, Namespace: namespace}
-	err := genericClient.Get(context.Background(), key, typeConfig)
+	err := genericClient.Get(context.Background(), typeConfig, namespace, name)
```
marun: (No action required) I'd still like to see the generic client methods avoid requiring a context (and expose other methods that do e.g. `Get` wouldn't require a context, `GetWithContext` would).

test/common/typeconfig.go
```diff
@@ -21,13 +21,12 @@ import (
 
 	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	client "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
```
marun: (No action required) Was this change required for the transition to kubebuilder?  It's nice to have, don't get me wrong, but I'd prefer to review PRs with a cohesive purpose.

marun:
This is LGTM but I'll give @xunpan responsibility for merging once he's satisfied.

Makefile
```diff
@@ -108,6 +108,14 @@ controller: $(CONTROLLER_TARGET)
 
 kubefed2: $(KUBEFED2_TARGET)
 
+# Generate code
+generate: kubefed2
+ifndef GOPATH
+	$(error GOPATH not defined, please define GOPATH. Run "go help gopath" to learn more about GOPATH)
+endif
+	go generate ./pkg/... ./cmd/...
+	./scripts/sync-up-helm-chart.sh
```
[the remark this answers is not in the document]
shashidharatd: Error handling done as a separate commit now.

scripts/pre-commit.sh
```diff
@@ -112,6 +104,12 @@ cd "$base_dir" || {
   exit 1
 }
 
+# Ensure the test binaries are in the path.
```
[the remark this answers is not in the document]
shashidharatd: comment addressed as separate commit. TEST_ASSET_* are only required for `run-e2e-tests-with-managed-fixture`.

shashidharatd:
@marun, @xunpan, Addressed comments as separate commits. PTAL once again. Thanks!

test/e2e/framework/managed/kubeapi.go
shashidharatd: When i upgraded the kubebuilder to v1.0.8, the binaries which came along with it kube-apiserver (v1.13.1) no longer works with `--secure-port=0`. So instead of updating and fixing the arguments, i thought it is best to use the methods from the imported package. Also since it is just the test fixture, so i believe it does not hurt spawning multiple instances of etcd.

test/common/typeconfig.go
shashidharatd: I am still unable to decide whether it is a best practice to store the context in client structure as it is against the recommendation from the package authors. Here is one such discussion about it https://groups.google.com/forum/#!topic/golang-nuts/xRbzq8yzKWI

test/common/typeconfig.go
shashidharatd: After transitioning, i started getting errors like the API type not found. The generic client uses a scheme which includes all the types from federation, k8s and cluster-registry. So replacing the client solved the issue.

shashidharatd:
@gyliu513, this pr also fixes this issue https://github.com/kubernetes-sigs/federation-v2/issues/723

test/common/typeconfig.go
marun: My point wasn't to store a context, but rather to have a client call by default supplied a new context with each call since it's effectively just boilerplate.  Where we actually intend to do things like cancel we can call `<method>WithContext` or something.

test/e2e/framework/managed/kubeapi.go
marun: It's memory overhead and would limit the maximum number of clusters we could test with.  But, not a long-term concern.

test/common/typeconfig.go
shashidharatd: Ack

marun:
Thanks for the fixes, LGTM!  Still waiting on another reviewer for a second opinion on such a large PR.

scripts/deploy-federation.sh
```diff
@@ -49,44 +49,6 @@ set -o pipefail
...
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 
-function deploy-with-script() {
-  # Create a permissive clusterrolebinding to allow federation controllers to run.
-  if [[ "${NAMESPACED}" ]]; then
-    # TODO(marun) Investigate why cluster-admin is required to view cluster registry clusters in a namespace
-    kubectl -n "${NS}" create rolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
-  else
-    # TODO(marun) Make this more restrictive.
-    kubectl create clusterrolebinding federation-admin --clusterrole=cluster-admin --serviceaccount="${NS}:default"
```
xunpan: because we will not create `clusterrolebinding` `federation-admin`, we should clean code to avoid useless deletion of this resource in `scripts/delete-federation.sh`

test/e2e/framework/managed/federation.go
```diff
@@ -17,28 +17,28 @@ limitations under the License.
 package managed
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
+
```
xunpan: I think it is not necessary to have an empty line.

xunpan:
This is awesome. This is the largest diff that I reviewed. Thanks @shashidharatd to contribute such a significant enhancement.

It is not block issue for my comments above. It is a huge diff, I'd like to merge it and then sync with shashi for comments.

/lgtm

PR #717, opened 2019-04-02
apply FederationConfig to set controller options

xunpan:
following up actions after add FederationConfig resources in #685

After it is merged, I will do:
- [ ] change helm chart to use FederationConfig for controller options for both namespaced and cluster scoped clusters.
- [ ] make kubectl disable reject in namespaced federation.
- [ ] update document to reflect the changes


cmd/controller-manager/app/controller-manager.go
```diff
@@ -177,6 +183,45 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
...
+	var featureGates = make(map[string]bool)
+	for _, v := range fedConfig.Spec.FeatureGates {
+		featureGates[v.Name] = v.Enabled
+	}
+	if len(featureGates) < 1 {
+		return
+	}
+
+	opts.FeatureGates = featureGates
+	glog.V(1).Infof("\"feature-gates\" are setting with %v", featureGates)
```
marun: s/are setting with/will be set to/

cmd/controller-manager/app/controller-manager.go
```diff
@@ -177,6 +183,45 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
...
+	opts.LeaderElection.ResourceLock = spec.LeaderElect.ResourceLock
+	opts.LeaderElection.RetryPeriod = spec.LeaderElect.RetryPeriod * time.Second
+	opts.LeaderElection.RenewDeadline = spec.LeaderElect.RenewDeadline * time.Second
+	opts.LeaderElection.LeaseDuration = spec.LeaderElect.LeaseDuration * time.Second
+
+	var featureGates = make(map[string]bool)
+	for _, v := range fedConfig.Spec.FeatureGates {
+		featureGates[v.Name] = v.Enabled
+	}
+	if len(featureGates) < 1 {
```
marun: I think `== 0`  would be preferable

cmd/controller-manager/app/controller-manager.go
```diff
@@ -177,6 +183,45 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
...
+	opts.ClusterMonitorPeriod = spec.ControllerDuration.ClusterMonitorPeriod * time.Second
+
+	opts.Config.ClusterNamespace = spec.RegistryNamespace
+	opts.Config.ClusterAvailableDelay = spec.ControllerDuration.AvailableDelay * time.Second
+	opts.Config.ClusterUnavailableDelay = spec.ControllerDuration.UnavailableDelay * time.Second
+
+	opts.LeaderElection.ResourceLock = spec.LeaderElect.ResourceLock
+	opts.LeaderElection.RetryPeriod = spec.LeaderElect.RetryPeriod * time.Second
+	opts.LeaderElection.RenewDeadline = spec.LeaderElect.RenewDeadline * time.Second
+	opts.LeaderElection.LeaseDuration = spec.LeaderElect.LeaseDuration * time.Second
```
marun: Rather than assuming seconds, would it be possible to parse any given duration (e.g. `ms`, `s`, `m`, `h`)?

cmd/controller-manager/app/controller-manager.go
```diff
@@ -177,6 +183,45 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
...
 
+func setOptionsByFederationConfig(opts *options.Options) {
+	client := genericclient.NewForConfigOrDieWithUserAgent(opts.Config.KubeConfig, "federationconfig")
+
+	name := util.FederationConfigName
+	namespace := opts.Config.FederationNamespace
+	fedConfig := &corev1a1.FederationConfig{}
+	err := client.Get(context.Background(), fedConfig, namespace, name)
+	if err != nil {
+		glog.V(1).Infof("Cannot retrieve FederationConfig %q in namespace %q: %v. Command line options are used.", name, namespace, err)
```
marun: (No action required) Are you intending to remove the cli args at some point?

xunpan: Yes. It is agreement during our discuss as far as I remember. I will update when clean up options.

cmd/controller-manager/app/controller-manager.go
xunpan: done

cmd/controller-manager/app/controller-manager.go
xunpan: done

cmd/controller-manager/app/controller-manager.go
xunpan: I use `time.Duration` in the API to try to make it parse unit of time. However, it does not work as I expected. I will investigate and fix this.

xunpan: Good news is that this case is supported in Feb.: https://github.com/kubernetes-sigs/controller-tools/pull/138

Addressed.

marun:
Thank you!  Please also make sure to add an entry to `CHANGELOG.md` when the feature is fully implemented since this is a user-facing change (especially once kubefed2 can rely on the config to detect the scope of a control plane). 

/approve
/lgtm

Issue #718, opened 2019-04-02
Replace managed e2e with unmanaged in-memory testing


marun:
Currently federation e2e tests can be run with 'managed' (test creates a federation) and 'unmanaged' (a deployed federation is targeted) fixture.  At one point managed fixture was cheaper in terms of time and resource usage and this was helpful in accelerating the early development of fedv2.  With the availability of kind, unmanaged clusters are now arguably the cheaper option.

The remaining reason to continue testing against managed fixture has been that it runs controllers in-process, allowing them to be targeted when under test by golang's race detector.  Continuing to maintain the managed fixture is not free, however, and a cheaper option for ensuring race detection exists.  It should be possible to run an e2e suite against unmanaged fixture, scale down the controller manager, and run a race-detecting suite with the controllers running in-process (via the `-in-memory-controllers` flag).  This would remove the need to support managed fixture going forward (with the exception of the bits that enable in-process controllers).

cc: @shashidharatd @font

shashidharatd:
@marun, agree with your observations about reducing the maintenance burden.

PR #719, opened 2019-04-03
Fix race conditions detected with in-memory execution of controllers

marun:
I noticed these when testing in hybrid mode (unmanaged + in-memory controllers), prompting me to file #718.


shashidharatd:
Thanks for fixing.
/lgtm

PR #724, opened 2019-04-04
Ensure managed label is removed from unmanaged resources

### LINKED ISSUE #612: Minimize the member cluster resources tracked by the control plane

marun:
This PR ensures that the managed label (`federation.k8s.io/managed: true`) is removed from orphaned resources or namespaces in the host cluster or when a managed resource is seen without a corresponding federated resource. 

Fixes #612

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Unreleased
+ - [#612](https://github.com/kubernetes-sigs/federation-v2/pull/612) -
+   Label managed resources in member clusters and only watch resources
+   so labeled to minimize the memory usage of the federated control
+   plane.
  - [#721](https://github.com/kubernetes-sigs/federation-v2/issues/721) -
    kubefed2 disable now deletes the FederatedTypeConfig rather than set
    propagationEnabled, waits for the sync controller to shut down, and
--- a/pkg/controller/sync/accessor.go
+++ b/pkg/controller/sync/accessor.go
@@ -36,7 +36,7 @@ import (
 type FederatedResourceAccessor interface {
 	Run(stopChan <-chan struct{})
 	HasSynced() bool
-	FederatedResource(qualifiedName util.QualifiedName) (FederatedResource, error)
+	FederatedResource(qualifiedName util.QualifiedName) (federatedResource FederatedResource, possibleOrphan bool, err error)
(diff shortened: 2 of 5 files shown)


marun:
Rebased

marun:
Rebased

marun:
Rebased

marun:
Rebased and ready for review.

marun:
Rebased

irfanurrehman:
This lgtm. @font @shashidharatd up to you to review and/or merge.

shashidharatd:
LGTM too

PR #727, opened 2019-04-04
Remove mention of script-based install from the user guide

### LINKED ISSUE #565: User guide shoud state implicit GOPATH dependency

marun:
Helm is the only supportable deployment option, so documenting a script-based way of deploying is just asking for user confusion.

Fixes #565

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -8,7 +8,6 @@
     - [Deployment Image](#deployment-image)
     - [Create Clusters](#create-clusters)
   - [Helm Chart Deployment](#helm-chart-deployment)
-  - [Automated Deployment](#automated-deployment)
   - [Operations](#operations)
     - [Join Clusters](#join-clusters)
     - [Check Status of Joined Clusters](#check-status-of-joined-clusters)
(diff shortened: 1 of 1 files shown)


shashidharatd:
LGTM

docs/userguide.md
```diff
@@ -8,7 +8,6 @@
     - [Deployment Image](#deployment-image)
     - [Create Clusters](#create-clusters)
   - [Helm Chart Deployment](#helm-chart-deployment)
```
irfanurrehman: I think this title could now be `Automated Deployment` rather then `Helm Chart Deployment`, given that helm is the only mechanism of automated deployment now.

docs/userguide.md
```diff
@@ -31,7 +30,7 @@
   - [Cleanup](#cleanup)
     - [Deployment Cleanup](#deployment-cleanup)
```
irfanurrehman: Is this section left as is because of the issue with [uninstalling the charts](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/README.md#uninstalling-the-chart)?

xunpan: I think make reference information for both install/uninstall is better. So, either of below is fine to me.
1.  make a clean up section to refer to unstallation the charts.
2. delete this section and make Helm Chart Deployment section as
```diff
- You can refer to [helm chart installation guide](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/README.md)
- to install federation-v2.
+ You can refer to [helm chart guide](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/README.md)
+ to install/uninstall federation-v2.
```

docs/userguide.md
marun: Automated Deployment vs what - non-automated?  I think being explicit about the use of Helm is appropriate here.

docs/userguide.md
marun: @irfanurrehman Good catch, I just missed it.  I've removed the `Cleanup` section and added mention of uninstall to the helm section as per @xunpan's suggestion.

xunpan:
LGTM

Issue #728, opened 2019-04-05
Request to break relevant sections from User Guide out into separate installation docs module


abrennan89:
**What would you like to be added**:
We should split the sections in the user guide up until 'Operations' into a separate doc for installation docs, and then just provide a link as/if needed to this.

**Why is this needed**:

- As the document grows it will become unwieldy to have just a single user guide, modules of documentation make it easier to find content a user needs than searching through a monolith.
- I've noticed this already in working with our team to plan docs updates for releases.
- Good opportunity to thoroughly review the existing content and ensure that required dependencies are covered.

**Somewhat related issues / PRs for my own reference**:
- https://github.com/kubernetes-sigs/federation-v2/pull/727 [only helm deployment supported]
- https://github.com/kubernetes-sigs/federation-v2/pull/451 [changes to user guide examples?]

I am happy to work on this but raising the issue first for visibility.

/kind feature

shashidharatd:
@abrennan89, Thank you for taking this up. much appreciated.

abrennan89:
PR added; https://github.com/kubernetes-sigs/federation-v2/pull/730

Issue #701, opened 2019-03-28
Provide a versioned helm chart with each release


marun:
- [x] Update [release doc](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/releasing.md) to reflect the need to:
     - Create an archive of the helm chart that defaults to the released image tag
     - Create an sha hash of the chart archive
     - Include the archive and sha hash as release artifacts
     - Create a PR that updates the repo index of the chart to point to the location of the released chart
 - [x] Update the [helm install instructions](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/README.md) 
     - Explain how to add the federation chart repo (i.e. url to the repo index)
     - Explain how to install a specific version of the chart

/kind feature

cc: @sfxworks

Issue #654, opened 2019-03-15
kubefed2 installtion guide missing in userguide


poothia:
**What would you like to be added**:
Going throught the userguide prerequisites, there should be a (optional) guide on how to install "kubefed2".

**Why is this needed**:
"kubefed2" tool is used in many commands mentioned in userguide document.

/kind docs

marun:
The binary is provided as part of a release, and the user can invoke it directly or put it in their path.    This seems pretty basic to me, and not necessarily worth documenting.  Are there additional material you think would be helpful?

digivava:
As someone new to working with open-source software in an alpha state, I was lost for awhile trying to find it as well. I had assumed that the Prerequisites section contained all the info on binaries needed to run the deployment scripts.

It might be helpful just to add the first sentence of your comment above into the prerequisite section of the docs.

poothia:
> The binary is provided as part of a release, and the user can invoke it directly or put it in their path. This seems pretty basic to me, and not necessarily worth documenting. Is there additional material you think would be helpful?

@marun Ya I do agree, but for a new person it is quite confusing. Because there is mention of where to find the "kubefed2" file. So maybe a note in pre-requisites section might be helpful.



> As someone new to working with open-source software in an alpha state, I was lost for awhile trying to find it as well. I had assumed that the Prerequisites section contained all the info on binaries needed to run the deployment scripts.
> 
> In the scripts it seems to be invoked from ./bin. Where did bin come from? Should I be generating it somehow? It might be helpful just to add a little more detail into the prerequisite section of the docs.

@digivava Ya I too got confused. I dont feel it is necessary but a little more clarity would be of no harm.

shashidharatd:
@poothia, are you interested in working on this task?

xunpan:
/assign

Let me handle it for beta

PR #730, opened 2019-04-08
Documentation changes for issue #728

abrennan89:
Made the following changes
- split installation instructions out from user guide
- general review and editing of userguide.md, development.md and installation.md
- removed list of required binaries as this is superseded by Helm deployment
- updated links to installation instructions

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,0 +1,54 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Installing Federation v2](#installing-federation-v2)
+  - [Prerequisites](#prerequisites)
+    - [Required binaries](#required-binaries)
+    - [Creating Clusters](#creating-clusters)
+    - [Deployment Image](#deployment-image)
+  - [Helm Chart Deployment](#helm-chart-deployment)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Installing Federation v2
+
+## Prerequisites
+
+Federation v2 requires Kubernetes v1.13 or newer.
+
+### Required binaries
+The following binaries are required to install and use Federation v2.
+- [`kubebuilder`](https://book.kubebuilder.io/getting_started/installation_and_setup.html) v1.13 or newer.
+- [`etcd`](https://github.com/etcd-io/etcd/blob/master/Documentation/dl_build.md) v1.13 or newer.
+- `kube-apiserver` v1.13 or newer.
+
+All of the binaries listed can be installed by the `download-binaries.sh` script. Use the following commands to clone the repository and run the script.
+
+   `git clone https://github.com/kubernetes-sigs/federation-v2.git`   
+   `./script/download-binaries.sh`   
+   `export PATH=$(pwd)/bin:${PATH}`   
+
+### Creating Clusters
+
+The following is a list of Kubernetes environments that have been tested and are supported by the Federation v2 community.
+
+- [kind](./environments/kind.md)
+- [Minikube](./environments/minikube.md)
+- [Google Kubernetes Engine (GKE)](./environments/gke.md)
+- [IBM Cloud Private](./environments/icp.md)
+
+After completing the steps in one of the above guides, return here to continue the deployment.
+
+**IMPORTANT:** You must set the correct context in your cluster(s) using the command below.
+
+```bash
+kubectl config use-context cluster1
+```
+### Deployment Image
+
+If you follow this user guide without any changes you will be using the latest master image tagged as `canary`](development.md#test-latest-master-changes-canary).
+
+## Helm Chart Deployment
+
+You can refer to [helm chart installation guide](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/README.md) for instructions on installing Federation v2.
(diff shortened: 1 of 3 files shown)


abrennan89:
I signed it

marun:
@abrennan89 As per #727, I think all mention of script-based installation should be removed.  As of last week, helm is the only way for someone to deploy fedv2.

abrennan89:
Thanks Maru, I had wondered if that was the case. This has been removed now. Does it make sense just to move the Helm docs that we link out to into this install.md instead, since it is the only way to install?

docs/installation.md
```diff
@@ -0,0 +1,45 @@
...
+After completing the steps in one of the above guides, return here to continue the deployment.
+
+**IMPORTANT:** You must set the correct context in your cluster(s) using the command below.
+
+```bash
+kubectl config use-context cluster1
+```
+### Deployment Image
+
+If you follow this user guide without any changes you will be using the latest
```
marun: I think the helm chart actually defaults to `canary` rather than `latest`.  #701 is intended to ensure a released chart that sets the appropriate image tag rather than requiring it to be manually set, but in advance of that it might make sense to mention specifying a released tag (e.g. `v0.0.7`).

docs/userguide.md
```diff
@@ -372,18 +297,25 @@ for c in cluster1 cluster2; do
...
 ```
-
-If you were able to verify the resources removed and added back then you have
+If you were able to verify that the resources were removed and re-added, you have
 successfully verified a working federation-v2 deployment.
 
-#### Using Cluster Selector
+#### Cleaning up
+
+To cleanup the example simply delete the namespace:
```
marun: Please add something like 'deletion of the test namespace may be delayed while federation controllers perform the pre-deletion step of removing managed resources from member clusters'.

docs/userguide.md
```diff
@@ -449,29 +381,19 @@ spec:
...
-```bash
-kubectl delete ns test-namespace
-```
-
-### Troubleshooting
+## Troubleshooting
 
 If federated resources are not propagated as expected to the member clusters, you can
-use the following command to view `Events` which may aid in diagnosing the problem.
+use the following command to view `Events` which may helpto diagnose the problem.
```
marun: s/helpto/help to/

docs/installation.md
```diff
@@ -0,0 +1,45 @@
...
+  - [Prerequisites](#prerequisites)
+    - [Creating Clusters](#creating-clusters)
+    - [Deployment Image](#deployment-image)
+  - [Helm Chart Deployment](#helm-chart-deployment)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Installing Federation v2
+
+## Prerequisites
```
marun: Please add reference to sourcing a kubefed2 binary from a recent release, as per #654.

marun:
@abrennan89 Do you want to rebase on #727 or incorporate those changes in this PR and I can close the other one?  The goal is removing all mention of 'automated installation' in favor of helm.

marun:
@abrennan89 There is a proposal to add a glossary entry for federation to the kubernetes website: https://github.com/kubernetes/website/pull/13520

Would it be possible to ensure the links proposed in the website PR be valid for the fedv2 docs?

abrennan89:
> @abrennan89 There is a proposal to add a glossary entry for federation to the kubernetes website: [kubernetes/website#13520](https://github.com/kubernetes/website/pull/13520)
> 
> Would it be possible to ensure the links proposed in the website PR be valid for the fedv2 docs?

I think this is a separate issue - it looks like we should be integrating into their website content repo rather than linking to an external github page, waiting to confirm if that's the issue. This is what we do for Knative docs.
If it is, we can raise an issue to rework our content to fit that structure and figure out where our information modules should fit into the kubernetes docs.

@pmorie this is related to our 'where should the docs live' discussion and probably applies to any content that lives upstream.

docs/installation.md
abrennan89: I think this should stay as a separate issue for now under #654 because it will require adding a generic kubefed2 installation doc somewhere else. 1. we would need details for this, and 2. I'm not sure where it would live if it's not specific to Federation only?

docs/userguide.md
abrennan89: Done now, see NOTE.

marun:
Thank you for the updates!   It looks like this PR needs to be rebased against master.

docs/installation.md
marun: I'm not sure what you mean by 'not specific to federation'.  `kubefed2` is very much specific to federation.  And the instructions are trivial:

 - download binary from the [desired release](https://github.com/kubernetes-sigs/federation-v2/releases/)
 - ensure the binary is in a directory referenced by the PATH argument

abrennan89: This makes way more sense now we're renaming things - might make sense to add this in whatever PR we use to rename to kubefed in the docs?

marun: Instructions for `kubefedctl` installation have been added to the userguide since my original comment:

https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/userguide.md#binaries

abrennan89:
Rebased since #865 - will review for renaming under #891 when this PR is closed.

marun:
@abrennan89 Should I be reviewing this now or are you still intending to update this PR to reflect the rename?

abrennan89:
@marun no, as I had mentioned to Paul I've already had to rebase this twice at least due to other changes made in the docs repo.
We need to merge this first and then clean up the naming after. Renaming has already undone some of my changes; it's easier to rename a product than it is to rewrite or re-edit a whole section over again.
I can do #891 to clean up after but this needs to be merged first.

marun:
@abrennan89 Please squash the 2 commits so we can merge.

PR #684, opened 2019-03-22
build improvements

yamt:
TODO:

* ~~maybe `imagebuild.sh` should use `make container` as well~~ test this

PR #734, opened 2019-04-09
Make deploy-federation.sh more robust

michaelfig:
Allow running deploy-federation.sh when:
* There is already a working tiller installation, and the user doesn't have the ability to create cluster-wide roles.
* There are no additional specified JOIN_CLUSTERS.

### DIFF
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -53,7 +53,7 @@ function deploy-with-helm() {
   # Don't install tiller if we already have a working install.
   if ! helm version --server 2>/dev/null; then
     # RBAC should be enabled to avoid CI fail because CI K8s uses RBAC for Tiller
-    cat <<EOF | cat # kubectl apply -f -
+    cat <<EOF | kubectl apply -f -
 apiVersion: v1
 kind: ServiceAccount
 metadata:
--- a/scripts/deploy-federation.sh
+++ b/scripts/deploy-federation.sh
@@ -50,8 +50,10 @@ set -o pipefail
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 
 function deploy-with-helm() {
-  # RBAC should be enabled to avoid CI fail because CI K8s uses RBAC for Tiller
-  cat <<EOF | kubectl apply -f -
+  # Don't install tiller if we already have a working install.
+  if ! helm version --server; then
+    # RBAC should be enabled to avoid CI fail because CI K8s uses RBAC for Tiller
+    cat <<EOF | cat # kubectl apply -f -
 apiVersion: v1
 kind: ServiceAccount
 metadata:
@@ -72,8 +74,9 @@ subjects:
     namespace: kube-system
 EOF
 
-  helm init --service-account tiller
-  util::wait-for-condition "Tiller is ready" "helm version --server &> /dev/null" 120
+    helm init --service-account tiller
+    util::wait-for-condition "Tiller to become ready" "helm version --server &> /dev/null" 120
+  fi
 
   REPOSITORY=${IMAGE_NAME%/*}
   IMAGE_TAG=${IMAGE_NAME##*/}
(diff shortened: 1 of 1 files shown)


michaelfig:
I signed it!

scripts/deploy-federation.sh
```diff
@@ -128,7 +131,8 @@ to ensure credentials are available for push:
 fi
 
 shift
-JOIN_CLUSTERS="${*}"
+# Allow for no specific JOIN_CLUSTERS: they probably want to kubefed2 themselves.
+JOIN_CLUSTERS="${*-}"
```
xunpan: The original code works for no argument case. I do not understand the concern here, could you please explain it?

michaelfig: On my Mac, I have:
```
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

That version of Bash returns: 

```
$ NAMESPACED=y FEDERATION_NAMESPACE=dev bash scripts/deploy-federation.sh quay.io/kubernetes-multicluster/federation-v2:latest
scripts/deploy-federation.sh: line 131: *: unbound variable
$
```

So maybe it's only a problem in older Bash?

Otherwise the script works fine on my Mac.

xunpan: I guess it is a problem in older bash. Linux has no such issue with newer version of bash. I tried on my Mac. It has similar issue.

We have 
```
set -o nounset
```
$* is a special parameter. It seems not defined in old bash but defined in new bash even it has no arguments specified.

As we start building MAC binaries from #684, and the logic here is fine for both platforms. I think it is OK.

scripts/deploy-federation.sh
```diff
@@ -99,10 +99,16 @@ IMAGE_NAME="${1:-}"
 NAMESPACED="${NAMESPACED:-}"
 
 LATEST_IMAGE_NAME=quay.io/kubernetes-multicluster/federation-v2:latest
+CANARY_IMAGE_NAME=quay.io/kubernetes-multicluster/federation-v2:canary
```
xunpan: In https://github.com/kubernetes-sigs/federation-v2/pull/716/commits/b270b3c5c3ae3f74d8737db3a7e2be22fa4297a6, `quay.io/kubernetes-multicluster/federation-v2:canary` is changed to `<containerregistry>/<username>/federation-v2:canary`. I think it makes sense to build your own image to test latest master changes instead of using existing one. 

So, I guess it is better to keep old logic without `canary` part here. @shashidharatd

scripts/deploy-federation.sh
```diff
@@ -50,8 +50,10 @@ set -o pipefail
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 
 function deploy-with-helm() {
-  # RBAC should be enabled to avoid CI fail because CI K8s uses RBAC for Tiller
-  cat <<EOF | kubectl apply -f -
+  # Don't install tiller if we already have a working install.
+  if ! helm version --server; then
```
xunpan: This is correct from logic to have `if` for helm initialization. But it shows an error message `Error: could not find tiller`. It is not an error condition here. So, it should be better to clean this message to avoid misunderstanding of the message.

xunpan:
@michaelfig 
Thanks for your PR to fix this. Please help to check CI failed.

scripts/deploy-federation.sh
michaelfig: Thanks!

scripts/deploy-federation.sh
michaelfig: Okay, I've reverted the `canary` part.

scripts/deploy-federation.sh
michaelfig: I just removed the error output, as it will be displayed lower down.

michaelfig:
The failure is at this output, which seems unrelated to my patch:

```
Configuring container registry on host
Error: /etc/docker/daemon.json exists and CONFIGURE_INSECURE_REGISTRY=y. This script needs to add an 'insecure-registries' entry with host '172.17.0.1:5000' to /etc/docker/daemon.json. Please make the necessary changes or backup and try again.
The command "test $DOC_ONLY && echo "Skipped!" || DOWNLOAD_BINARIES=y bash -x ./scripts/pre-commit.sh" exited with 1.
```

xunpan:
@michaelfig 
Please rebase. I fix the CI issue in #757

marun:
Thank you!

/approve
/lgtm

PR #735, opened 2019-04-09
Refactor deletion handling

marun:
This PR adds the functionality formerly implemented by the deletion helper to the sync controller and rewrites it to use a new operation dispatcher.  This allows the old deletion helper and federated updater to be removed.

9th in a series targeting #612


marun:
Rebased

marun:
Rebased

marun:
Rebased and ready for review.

pkg/controller/sync/controller.go
```diff
@@ -373,3 +364,123 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
+	for _, cluster := range clusters {
+		clusterName := cluster.Name
+		rawClusterObj, _, err := s.informer.GetTargetStore().GetByKey(clusterName, targetKey)
+		if err != nil {
+			runtime.HandleError(errors.Wrapf(err, "failed to retrieve managed resource for cluster %q", clusterName))
+			retrievalErrorClusterNames = append(retrievalErrorClusterNames, clusterName)
+			continue
+		}
+		// Already removed
+		if rawClusterObj == nil {
```
yamt: i think this can also mean that the informer has not observed it yet.
isn't it safer to issue delete to confirm notfound?

pkg/controller/sync/controller.go
```diff
@@ -373,3 +364,123 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
+		glog.V(2).Infof("Initiating the removal of the label %q from resources previously managed by %s %q.", util.ManagedByFederationLabelKey, kind, key)
+		return nil
+		// TODO(marun) Implement removal of labels
+	}
+
+	glog.V(2).Infof("Deleting resources managed by %s %q from member clusters.", kind, key)
+
+	// First retrieve a list of unready clusters to maximize the
+	// changes of detecting when it's not possible to guarantee
+	// removal from member clusters.
```
yamt: chances?

either ways it isn't clear to me what this comment means.

pkg/controller/sync/controller.go
```diff
@@ -373,3 +364,123 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
+			// Host cluster namespace needs to have the managed
+			// label removed so it won't be cached anymore.
+			updater.RemoveManagedLabel(clusterName, clusterObj)
+		} else {
+			updater.Delete(clusterName)
+			deletedFromClusterNames = append(deletedFromClusterNames, clusterName)
+		}
+	}
+	if updater.NoChanges() {
+		if len(retrievalErrorClusterNames) > 0 {
```
yamt: i guess retrievalErrorClusterNames check is also needed for !NoChanges case

pkg/controller/sync/controller.go
```diff
@@ -373,3 +364,123 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
+		}
+
+		if len(unreadyClusters) == 0 {
+			// Managed resources no longer exist in any member cluster.
+			return s.removeFinalizer(fedResource)
+		}
+
+		// It is not possible to know if resources have been removed
+		// from all clusters if one or more clusters is not ready.
+		var unreadyClusterNames []string
```
yamt: i guess unreadyClusters check is also needed for !NoChanges case

pkg/controller/sync/controller.go
marun: If the informer hasn't observed it yet, the resource will be reconciled again when it does observe it.  What would be safer about issuing a deletion on the basis of potentially incomplete information?

marun: Without the resource in the informer cache, it is entirely possible that the resource in the member cluster is not managed.  As per #549, only managed resources should be deleted.

yamt: do you plan to make it delete labeled objects when federated object is not there? then it makes sense.

marun: Yes, that's the plan

pkg/controller/sync/controller.go
marun: I've switched to testing for unready clusters in the loop, so this comment is no longer relevant.

pkg/controller/sync/controller.go
marun: I've rewritten the logic such that this comment should be resolved.

pkg/controller/sync/controller.go
marun: I've rewritten the logic such that this comment should be resolved.

marun:
@yamt I've created a new deletion dispatcher and incorporated your feedback into a revised deletion handling function.  My intention is to revise the existing updater to embed the deletion dispatcher in a follow-on PR.  One notable improvement of the latest iteration is that waiting for deletion will no longer result in error messages if the informer cache is current.

marun:
Updated to include test of label removal from namespaces in the host cluster.

marun:
Refactored `deleteFromClusters` to better support #724.

marun:
Rebased

yamt:
i feel this PR is incomplete, or at least tricky to review, because many of relevant logic will be changed in the later PRs. https://github.com/kubernetes-sigs/federation-v2/pull/784 https://github.com/kubernetes-sigs/federation-v2/pull/724
it might be simpler to make them a single PR. how do you think?

marun:
@yamt Subsequent PRs are additive - they add functionality rather than changing logic.  I don't see how combining everything into a single PR is going to make it easier to review.

marun:
This PR has been updated to be a strict refactor - there is no change in behavior.

marun:
Rebased and ready for review!

pkg/controller/sync/controller.go
```diff
@@ -377,3 +367,145 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
+	}
+	if len(remainingClusters) > 0 {
+		fedKind := fedResource.FederatedKind()
+		fedName := fedResource.FederatedName()
+		glog.V(2).Infof("Waiting for resources managed by %s %q to be removed from the following clusters: %s", fedKind, fedName, strings.Join(remainingClusters, ", "))
+		return nil
+	}
+	// Managed resources no longer exist in any member cluster
+	return s.removeFinalizer(fedResource)
+}
```
shashidharatd: I am thinking is there a possibility that reconciliation is never triggered for the resource as we return nil and keep waiting for next reconciliation. The finalizer on the resource stays and federated resource deletion never runs to completion.

marun: The only nil return in this function is from line 439, and that path is only followed if one more more operations was executed successfully against one or more member clusters.  I've been assuming that ensures that one or more reconcile-triggering events would be pending, but is that not correct?

shashidharatd: Correct me if I'm wrong, In case of deletion, the reconcile may not be triggered as now in federated informer we do watch based on `managed` label.

marun:
@shashidharatd I've added a new commit (`Ensure reconcilliation..`) that addresses your comment, PTAL.

shashidharatd:
/lgtm
Thanks !

PR #736, opened 2019-04-09
Replace managed fixture test suite with in-memory suite

### LINKED ISSUE #718: Replace managed e2e with unmanaged in-memory testing

marun:
I've had to try to debug enough failures in the managed fixture test suite while working on #612 that I decided it was time to get rid of the managed fixture. To ensure that the race detector still gets a chance to run against controllers, the managed test suite is replaced with an e2e suite running with in-memory controllers.

Fixes #718


shashidharatd:
This LGTM. Thanks for doing this.
would merge it in a day, if no other comments from reviewers.

marun:
Rebased

font:
Glad to see this -- thanks @marun!

/lgtm

PR #737, opened 2019-04-09
apply federationconfig in chart and remove useless command line options of controller

### LINKED ISSUE #729: FederationConfig enhancement
following up actions after add FederationConfig resources in #685:
- [ ] change helm chart to use FederationConfig for controller options for both namespaced and cluster scoped clusters.
- [ ] remove command line options
- [ ] make kubectl disable reject in namespaced federation.
- [ ] update document to reflect the changes
- [ ] --federation-config=myfile.yaml

**Why is this needed**:

/kind feature

xunpan:
1st task to fix #729:
change helm chart to use FederationConfig for controller options for both namespaced and cluster scoped clusters.


charts/federation-v2/README.md
```diff
@@ -101,11 +101,8 @@ chart and their default values.
...
 | controllermanager.pullPolicy          | Image pull policy.                                                                                                                                                                                          | IfNotPresent                                                                                          |
-| controllermanager.featureGates        | Feature gates are a set of `key=value` pairs that describe alpha or experimental features. An administrator can use the `--feature-gates` command line flag on each component to turn a feature on or off.  | PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true |
-| controllermanager.federationNamespace | The namespace the federation control plane is deployed in.                                                                                                                                                  | federation-system                                                                                     |
```
marun: I'm just noticing that the federation namespace can be specified as a helm argument.  Why would that ever be done vs using the downward API to determine the namespace the deployment is running in?

xunpan: It is removed now. And I use downward API to set the hosted namespace in environment variables of helm definition and get in controller during startup.

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
```diff
@@ -22,20 +22,8 @@ spec:
     spec:
       containers:
       - args:
-{{- if .Values.featureGates }}
-        - --feature-gates={{ .Values.featureGates }}
-{{- end }}
 {{- if .Values.global.limitedScope }}
         - --limited-scope={{ .Values.global.limitedScope }}
-        - --federation-namespace=$(FEDERATION_NAMESPACE)
```
marun: Please restore this line and use it to configure the controller manager without condition.  The `--federation-namespace` arg should likely be retained to allow this.  

My understanding is that it is preferable to communicate values from the downward API via an arg rather than an environment variable.  Initial implementations of the controller manager used environment variables until @pmorie taught me the errors of my ways.

marun: Please restore this line and use it to configure the controller manager without condition.  The `--federation-namespace` arg should be retained to allow this.  

My understanding is that it is preferable to communicate values from the downward API via an arg rather than an environment variable.  Initial implementations of the controller manager used environment variables until @pmorie taught me the errors of my ways.

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
```diff
@@ -45,13 +33,13 @@ spec:
...
 {{- end }}
-{{- if .Values.global.limitedScope }}
         env:
-        - name: POD_NAMESPACE
+        - name: FEDERATION_NAMESPACE
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        - name: FEDERATION_NAMESPACE
+{{- if .Values.global.limitedScope }}
```
marun: Is this block necessary, given that `POD_NAMESPACE` appears unused and `CLUSTER_REGISTRY_NAMESPACE` is intended to be set by `FederationConfig`?

charts/federation-v2/values.yaml
```diff
@@ -19,18 +19,8 @@ controllermanager:
...
-  ## Federation V2 now have the following default feature gates:
-  ## PushReconciler=true,SchedulerPreferences=true,CrossClusterServiceDiscovery=true,FederatedIngress=true
-  featureGates:
-  ## The namespace the federation control plane is deployed in. If unset,
-  ## will default to federation-system.
-  federationNamespace:
-  ## The cluster registry namespace. If unset, will default to kube-multicluster-public.
-  registryNamespace:
-  
+    ## The cluster registry namespace. If unset, will default to kube-multicluster-public.
```
marun: Given the logic in `federationconfig.yaml`, would it be more correct to say:

```
...will default to kube-multicluster-public for a cluster-scoped deployment, and the federation namespace for namespace-scoped deployment.
```

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: FederationConfig
+metadata:
+  name: federation-v2
+  namespace: {{ .Release.Namespace }}
+spec:
+  limited-scope: {{ .Values.global.limitedScope }}
+{{- if .Values.registryNamespace }}
+  registry-namespace: {{ .Values.registryNamespace }}
+{{- else }}
+{{- if .Values.global.limitedScope }}
```
marun: What do you think about a new flag (e.g. `registryInFederationNamespace: true|false`) to avoid having to hard-code a registry namespace in the `FederationConfig` in the case of a limited-scope deployment without `registryNamespace` being specified?

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
xunpan: @marun, 
What is the issue for using environment variable for namespace? I think in helm chart, it is fine to use environment variable instead of command line options. Because the dependency (environment variable and federation config) should be always there. And it is way to use downward API.

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
xunpan: I will clean this because we donot use such 2 environment variables anymore.

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
xunpan: As we discussed before, `registry-namespace` should be always support for potential `cluster` registry sharing case. 

Therefore, we should make `registry-namespace` configurable. It is the reason to put the variable into `FederationConfig`. 
And in such case, we still need a default namespace if it is cluster federation and with no `registry-namespace` specified. So, `kube-multicluster-public` is hardcoded here.

I think it is clear in feature. 
I do not understand why flag `registryInFederationNamespace` helps in this case:   
    If it is true, we've already handled it with `registry-namespace: {{ .Release.Namespace}}` 
    If it is false, we still need a `resigtry-namespace`.

charts/federation-v2/values.yaml
xunpan: make sense

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
marun: It's not the helm chart we're talking about - it's the `controller-manager` binary.  Environment variables are not as easily discoverable as explicit command-line arguments and easy to screw up (since they can be set separate from command invocation).  Unless we have a specific reason to use environment variables - and we arguably do not - we shouldn't use them.

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
marun: I'm not sure how I gave you the impression that I wanted to avoid having `registry-namespace` be configurable.  That was not my intent.

In all cases, the federation namespace is the namespace in which the controller manager is deployed.  If the registry namespace is not provided by the user, it is defaulted to `kube-multicluster-public` if cluster-scoped and to the federation namespace if namespace-scoped.  

I think is strange to supply the federation namespace to the controller manager via the downward api, but supply the same value for the registry namespace via a hard-coded value.  My suggestion for `registryInFederationNamespace` would be to indicate that the controller manager should use the federation namespace, and it would only be respected if `registry-namespace` was empty.  The helm chart would provide this configuration for a limited-scope deployment where the user did not specify `registry-namespace`.  This would ensure the discoverability of the registry namespace without having to hard-code it.

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
xunpan: I'm ok to add it back because both way works.
 
My question is : what is the reason that downward api with using environment variable is arguable. 
It is a proposed way 
> to expose Pod and Container fields to a running Container.

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
marun: Ah, nevermind.  I'm splitting hairs.

charts/federation-v2/charts/controllermanager/templates/deployment.yaml
xunpan: Had a sync-up with marun: environment variable is not preferred as controller configuration from command line. `mybinary --my-var=${MY_VAR}` is better than `MY_VAR=foo mybinary`

xunpan:
The CI failed because leader elector test failed. 

The root cause is that leader elector test starting controllers as process. However, as we use `FederationConfig` to override command line options. It break assumption in the test code. The command line options assumes it is cluster scope controller, but our `FederationConfig` make it namespaced. So the new process controller never be a master because we had controller in the kubernetes cluster already.


To fix it, I will continue working on remove command line options and make leader elector works propoerly with '--federation-config fedconfig.yaml'.

marun:
@xunpan The leader election test should be run in a test namespace to ensure that it has no federated type configs that prompt sync controllers to be launched.  That means it should only run when `framework.TestContext.LimitedScope == false`.

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
```diff
@@ -0,0 +1,35 @@
...
+{{- end }}
+  controller-duration:
+    available-delay: 20s
+    unavailable-delay: 60s
+    cluster-monitor-period: 40s
+  leader-elect:
+    lease-duration: 15s
+    renew-deadline: 10s
+    retry-period: 5s
+    resource-lock: configmaps
```
shashidharatd: In the current format, it looks like the above parameters are hardcoded and there is no option to user to configure them according to their environment. Please expose them in Values.yaml of the chart.

Also, not specific to this pr. we should introduce defaulting the fields in API types, if they are not specified by user (or by helm).

marun: Defaulting support for CRDs is proposed but the PR appears to have been abandoned: https://github.com/kubernetes/kubernetes/pull/63604

That means that defaulting will require webhooks (yuck).

xunpan: I think we can use condition check in helm chart currently to make it work.
If related value is defined, we use it. Or else, use hard coded default value.

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+  registry-namespace: {{ .Values.registryNamespace }}
+{{- else }}
+{{- if .Values.global.limitedScope }}
+  registry-namespace: {{ .Release.Namespace }}
+{{- else }}
+  registry-namespace: kube-multicluster-public
+{{- end }}
+{{- end }}
+  controller-duration:
+{{- if .Values.clusterAvailableDelay }}
```
marun: Please refer to helm best practice for setting defaults:

https://github.com/helm/helm/blob/master/docs/chart_best_practices/values.md#flat-or-nested-values

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    retry-period: {{ .Values.leaderElectRetryPeriod }}
+{{- else }}
+    retry-period: 5s
+{{- end }}
+{{- if .Values.leaderElectResourceLock }}
+    resource-lock: {{ .Values.leaderElectResourceLock }}
+{{- else }}
+    resource-lock: configmaps
+{{- end }}
+  feature-gates:
```
marun: Should it be possible to configure the feature gate to enable?

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
xunpan: done

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
xunpan: done. 

I did not do it because we always enable all features. However, make the feature complete, I add the feature-gates in helm support.

xunpan: By the way, I organize `feature-gates` in nested way because I think it makes sense in `feature-gates` case.

xunpan:
rebased done based on lastest code. 
@marun

test/e2e/leaderelection.go
```diff
@@ -34,6 +34,10 @@ var _ = Describe("Leader Elector", func() {
 	tl := framework.NewE2ELogger()
 
 	It("should chose secondary instance, primary goes down", func() {
+		if framework.TestContext.LimitedScope {
+			framework.Skipf("Leader Elector is not need to be tested in namespaced federation setup")
```
marun: nit: s/Leader Elector.*/Testing of leader election requires an isolated test namespace which is only possible with cluster-scoped federation/

cmd/controller-manager/app/options/options.go
```diff
@@ -39,33 +35,23 @@ type Options struct {
...
+	o.ClusterMonitorPeriod = util.DefaultClusterMonitorPeriod
 
-	fs.BoolVar(&o.LimitedScope, "limited-scope", false, "Whether the federation namespace will be the only target for federation.")
-	fs.DurationVar(&o.ClusterMonitorPeriod, "cluster-monitor-period", time.Second*40, "How often to monitor the cluster health")
+	o.Config.ClusterNamespace = util.MulticlusterPublicNamespace
+	o.Config.ClusterAvailableDelay = util.DefaultClusterAvailableDelay
+	o.Config.ClusterUnavailableDelay = util.DefaultClusterUnavailableDelay
 
-	fs.DurationVar(&o.LeaderElection.LeaseDuration, "leader-elect-lease-duration", util.DefaultLeaderElectionLeaseDuration, ""+
-		"The duration that non-leader candidates will wait after observing a leadership "+
```
marun: (No action required) Would it make sense to retain the help text for removed parameters as comments on `FederationConfig`?

cmd/controller-manager/app/controller-manager.go
```diff
@@ -174,19 +176,46 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
...
-	err := client.Get(context.Background(), fedConfig, namespace, name)
+	if federationConfig == "" {
+		// there is no --federation-config specified, get `federation-v2` FederationConfig from the cluster
+		client := genericclient.NewForConfigOrDieWithUserAgent(opts.Config.KubeConfig, "federationconfig")
+
+		name := util.FederationConfigName
+		namespace := opts.Config.FederationNamespace
+		err := client.Get(context.Background(), fedConfig, namespace, name)
+		if err != nil {
+			glog.Infof("Cannot retrieve FederationConfig %q in namespace %q: %v. Default options are used.", name, namespace, err)
```
marun: nit: Please follow the convention of `<type name> "<qualified name>"` rather than `"<name>" in namespace "<namespace>"`

cmd/controller-manager/app/controller-manager.go
```diff
@@ -174,19 +176,46 @@ func startControllers(opts *options.Options, stopChan <-chan struct{}) {
...
+			return nil
+		}
+
+		glog.Infof("Setting Options with FederationConfig %q in namespace %q", name, namespace)
```
marun: Same nit as above.

charts/federation-v2/values.yaml
```diff
@@ -19,18 +18,28 @@ controllermanager:
...
+  clusterAvailableDelay:
+  clusterUnavailableDelay:
+  clusterMonitorPeriod:
+  leaderElectLeaseDuration:
+  leaderElectRenewDeadline:
+  leaderElectRetryPeriod:
+  ## Supported options are `configmaps` and `endpoints`
+  leaderElectResourceLock:
+  ## Value of feature gates item should be either `true` or `false`
+  featureGates:
```
marun: Given that we're not hard-coding the names of available feature gates in the API, is there a reason to hard-code them in the helm chart?

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+  registry-namespace: {{ .Values.registryNamespace }}
+{{- else }}
+{{- if .Values.global.limitedScope }}
+  registry-namespace: {{ .Release.Namespace }}
+{{- else }}
+  registry-namespace: kube-multicluster-public
+{{- end }}
+{{- end }}
+  controller-duration:
+    available-delay: {{ .Values.clusterAvailableDelay | default "20s" | quote }}
```
marun: (No action required) I think it's ok to set defaults in this PR, but longer-term I'm not sure that we want to set a value here if none were provided vs having the controller manager supply its own default.  I also wonder if the controller manager should update the config with missing defaults to ensure that it is the sole arbiter of what default value should be.

cmd/controller-manager/app/options/options.go
xunpan: I uses these messages in the chart README for related features. I guess it is a good way to add it in `FederationConfig` as comments.

charts/federation-v2/values.yaml
xunpan: These names in `values.yaml` exists to help users who want to enable/disable features on their own. List them here does not impact helm installation by default.

test/e2e/leaderelection.go
```diff
@@ -74,9 +78,6 @@ var _ = Describe("Leader Elector", func() {
 func spawnControllerManagerProcess(kubeConfigPath, namespace string) (*exec.Cmd, io.ReadCloser, error) {
 	args := []string{fmt.Sprintf("--kubeconfig=%s", kubeConfigPath),
 		fmt.Sprintf("--federation-namespace=%s", namespace),
-		"--leader-elect-lease-duration=1500ms",
```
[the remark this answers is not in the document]
xunpan: It does not slow down the `Leader Elector` in my test. So, I just simply remove the arguments.
However, we can do the correct thing for this. I will fix it.

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
xunpan: If there is no `FederationConfig`, controller will set the defaults.

xunpan:
@marun addressed

charts/federation-v2/charts/controllermanager/templates/federationconfig.yaml
marun: Given that a value of zero is invalid for any of these values, that's a good indication to avoid setting a value.

charts/federation-v2/values.yaml
marun: That's fine for this PR, I'm just concerned at the prospect of hard-coding values in the helm chart that are not guaranteed to be supported in the future in the API (since `FederationConfig` exposes a list of name/value pairs).

config/federationconfig.yaml
```diff
@@ -0,0 +1,27 @@
...
+  namespace: federation-system
+spec:
+  limited-scope: false
+  registry-namespace: kube-multicluster-public
+  controller-duration:
+    available-delay: 20s
+    unavailable-delay: 60s
+    cluster-monitor-period: 40s
+  leader-elect:
+    lease-duration: 1500ms
```
marun: Given that these are test-specific values, I'd like to avoid putting this in a location where a user might assume it was something that should be used for non-test purposes.   I also don't think it's necessary to supply non-default values for everything.  That said, I'll fix in a follow-up PR rather than delaying this one.

marun:
Thank you!

/lgtm
/approve

PR #741, opened 2019-04-10
Implement federate namespaces with content

irfanurrehman:
This implements:
  - [x] federate ns --content.

cc @marun @xunpan

### DIFF
--- a/test/e2e/federate.go
+++ b/test/e2e/federate.go
@@ -132,8 +132,6 @@ var _ = Describe("Federate ", func() {
 		// Namespace itself
 		namespaceTypeName := "namespaces"
 
-		defer deleteAllResources(f, tl, testResources)
-
 		testResources, err = containedTestResources(f, client, typeConfigFixtures, containedTypeNames, kubeConfig)
 		if err != nil {
 			tl.Fatalf("Error creating target resources: %v", err)
@@ -208,14 +206,6 @@ func validateTemplateEquality(tl common.TestLogger, fedResource, targetResource 
 	}
 }
 
-func deleteAllResources(f framework.FederationFramework, tl common.TestLogger, testResoures []testResources) {
-	for _, resources := range testResoures {
-		typeConfig := resources.typeConfig
-		testResourceName := util.NewQualifiedName(resources.targetResource)
-		deleteResources(f, tl, typeConfig, testResourceName)
-	}
-}
-
 func deleteResources(f framework.FederationFramework, tl common.TestLogger, typeConfig typeconfig.Interface, testResourceName util.QualifiedName) {
 	client := getFedClient(tl, typeConfig, f.KubeConfig())
 	deleteResource(tl, client, testResourceName, typeConfig.GetFederatedType().Kind)
@@ -262,7 +252,7 @@ func targetResourceFromAPI(tl common.TestLogger, typeConfig typeconfig.Interface
 	client := getTargetClient(tl, typeConfig, kubeConfig)
 	targetResource, err := client.Resources(qualifiedName.Namespace).Get(qualifiedName.Name, metav1.GetOptions{})
 	if err != nil {
-		tl.Fatalf("Federated resource %q not found: %v", qualifiedName, err)
+		tl.Fatalf("Test resource %q not found: %v", qualifiedName, err)
 	}
 	return targetResource
 }
@@ -278,9 +268,9 @@ func getFedClient(tl common.TestLogger, typeConfig typeconfig.Interface, kubeCon
 }
 
 func getTargetClient(tl common.TestLogger, typeConfig typeconfig.Interface, kubeConfig *restclient.Config) util.ResourceClient {
-	fedAPIResource := typeConfig.GetTarget()
-	fedKind := fedAPIResource.Kind
-	client, err := util.NewResourceClient(kubeConfig, &fedAPIResource)
+	apiResource := typeConfig.GetTarget()
+	fedKind := apiResource.Kind
+	client, err := util.NewResourceClient(kubeConfig, &apiResource)
 	if err != nil {
 		tl.Fatalf("Error getting resource client for %s", fedKind)
 	}
--- a/test/e2e/federate.go
+++ b/test/e2e/federate.go
@@ -85,22 +85,22 @@ var _ = Describe("Federate resource", func() {
 				tl.Fatalf("Error creating resource: %v", err)
 			}
 
-			typeName := util.QualifiedName{
-				Name:      typeConfig.GetObjectMeta().Name,
-				Namespace: typeConfig.GetObjectMeta().Namespace,
-			}
+			typeName := typeConfig.GetObjectMeta().Name
+			typeNamespace := typeConfig.GetObjectMeta().Namespace
 			testResourceName := util.NewQualifiedName(createdTargetObject)
 
 			defer deleteResources(f, tl, typeConfig, testResourceName)
 
 			tl.Logf("Federating %s %q", kind, testResourceName)
 			fedKind := typeConfig.GetFederatedType().Kind
-			artifacts, err := federate.GetFederateArtifacts(kubeConfig, typeName, testResourceName, false, false)
+			artifacts, err := federate.GetFederateArtifacts(kubeConfig, typeName, typeNamespace, testResourceName, false, false)
 			if err != nil {
 				tl.Fatalf("Error getting %s from %s %q: %v", fedKind, kind, testResourceName, err)
 			}
 
-			err = federate.CreateFedResource(kubeConfig, artifacts, false)
+			artifactsList := []*federate.FederateArtifacts{}
+			artifactsList = append(artifactsList, artifacts)
+			err = federate.CreateResources(nil, kubeConfig, artifactsList, typeNamespace, false, false)
 			if err != nil {
 				tl.Fatalf("Error creating %s %q: %v", fedKind, testResourceName, err)
 			}
--- a/test/e2e/federate.go
+++ b/test/e2e/federate.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/pkg/errors"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -38,7 +40,12 @@ import (
 	. "github.com/onsi/ginkgo"
 )
 
-var _ = Describe("Federate resource", func() {
+type testResources struct {
+	targetResource *unstructured.Unstructured
+	typeConfig     typeconfig.Interface
+}
+
+var _ = Describe("Federate ", func() {
 	f := framework.NewFederationFramework("federate-resource")
 	tl := framework.NewE2ELogger()
 	typeConfigFixtures := common.TypeConfigFixturesOrDie(tl)
@@ -57,12 +64,12 @@ var _ = Describe("Federate resource", func() {
 		}
 	})
 
-	// Use one cluster scoped and one namespaced type to test complete flow
+	// Use one cluster scoped and one namespaced type to test federate a single resource
 	toTest := []string{"clusterroles.rbac.authorization.k8s.io", "configmaps"}
 	for _, testKey := range toTest {
 		typeConfigName := testKey
 		fixture := typeConfigFixtures[testKey]
-		It(fmt.Sprintf("for %q should create an equivalant federated resource in federation", typeConfigName), func() {
+		It(fmt.Sprintf("resource %q, should create an equivalant federated resource in federation", typeConfigName), func() {
 			typeConfig := &fedv1a1.FederatedTypeConfig{}
 			err := client.Get(context.Background(), typeConfig, f.FederationSystemNamespace(), typeConfigName)
 			if err != nil {
@@ -75,23 +82,24 @@ var _ = Describe("Federate resource", func() {
 
 			kind := typeConfig.GetTarget().Kind
 			targetAPIResource := typeConfig.GetTarget()
-			targetObject, err := common.NewTestTargetObject(typeConfig, f.TestNamespaceName(), fixture)
+			targetResource, err := common.NewTestTargetObject(typeConfig, f.TestNamespaceName(), fixture)
 			if err != nil {
-				tl.Fatalf("Error creating test object: %v", err)
+				tl.Fatalf("Error creating test resource: %v", err)
 			}
 
-			createdTargetObject, err := common.CreateResource(kubeConfig, targetAPIResource, targetObject)
+			createdTargetResource, err := common.CreateResource(kubeConfig, targetAPIResource, targetResource)
 			if err != nil {
 				tl.Fatalf("Error creating resource: %v", err)
 			}
 
 			typeName := typeConfig.GetObjectMeta().Name
 			typeNamespace := typeConfig.GetObjectMeta().Namespace
-			testResourceName := util.NewQualifiedName(createdTargetObject)
+			testResourceName := util.NewQualifiedName(createdTargetResource)
 
 			defer deleteResources(f, tl, typeConfig, testResourceName)
 
-			tl.Logf("Federating %s %q", kind, testResourceName)
+			By(fmt.Sprintf("Federating %s %q", kind, testResourceName))
+
 			fedKind := typeConfig.GetFederatedType().Kind
 			artifacts, err := federate.GetFederateArtifacts(kubeConfig, typeName, typeNamespace, testResourceName, false, false)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/kubefed2/federate/federate.go
```diff
@@ -308,12 +345,86 @@ func CreateFedResource(hostConfig *rest.Config, resources *fedResources, dryrun
...
+	}
+
+	glog.Infof("Successfully created %s %q from %s", fedKind, qualifiedFedName, typeConfig.GetTarget().Kind)
+	return nil
+}
+
+func createResources(cmdOut io.Writer, hostConfig *rest.Config, resourcesList []FedResources, namespace string, dryrun, enableType bool) error {
+	for _, resources := range resourcesList {
+		if enableType {
+			err := enable.CreateResources(cmdOut, hostConfig, resources.TypeResources, namespace)
```
marun: Have you haven't tested this with more than one resource of a given type?  `enable.CreateResources` is not idempotent when it comes to creating the FederatedTypeConfig and will fail if it already exists.

pkg/kubefed2/federate/util.go
```diff
@@ -47,3 +54,62 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
 		resource.SetNamespace(namespace)
 	}
 }
+
+func getAPIResourceList(config *rest.Config) ([]metav1.APIResource, error) {
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/kubefed2/federate/util.go
```diff
@@ -47,3 +54,62 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+				continue
+			}
+			resources = append(resources, resource)
+		}
+	}
+
+	return resources, nil
+}
+
+type namespacedResource struct {
```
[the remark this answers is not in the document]
irfanurrehman: Nothing, updated name accordingly!

pkg/kubefed2/federate/util.go
```diff
@@ -47,3 +54,62 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+
+func getNamespacedResources(config *rest.Config, namespace string) ([]namespacedResource, error) {
+	apiResources, err := getAPIResourceList(config)
+	if err != nil {
+		return nil, err
+	}
+
+	resourcesInNamespace := []namespacedResource{}
+	for _, apiResource := range apiResources {
+		client, err := ctlutil.NewResourceClient(config, &apiResource)
```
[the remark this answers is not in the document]
irfanurrehman: Skipping this for now!

pkg/kubefed2/federate/util.go
```diff
@@ -47,3 +54,62 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+			resources = append(resources, resource)
+		}
+	}
+
+	return resources, nil
+}
+
+type namespacedResource struct {
+	targetResource *unstructured.Unstructured
+	apiResource    metav1.APIResource
```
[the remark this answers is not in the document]
irfanurrehman: This is what is directly used in place of looking the apiResource  again later. The resources are already looked up here in `getResourcesInNamespace` via `getAPIResourceList`.

pkg/kubefed2/federate/federate.go
```diff
@@ -347,12 +363,68 @@ func CreateFedResource(hostConfig *rest.Config, resources *FedResources, dryrun
 	return nil
 }
 
-func createResources(cmdOut io.Writer, hostConfig *rest.Config, resources *FedResources, namespace string, dryrun, enableType bool) error {
-	if enableType {
-		err := enable.CreateResources(cmdOut, hostConfig, resources.TypeResources, namespace)
+func createResources(cmdOut io.Writer, hostConfig *rest.Config, resourcesList []FedResources, namespace string, dryrun, enableType bool) error {
+	for _, resources := range resourcesList {
+		if enableType {
```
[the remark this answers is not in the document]
irfanurrehman: This is understood and updated the code accordingly!

pkg/kubefed2/federate/federate.go
irfanurrehman: Yes you are right, this wasn't even tested for multiple resources. The intention of this WIP was exactly to get your early feedback on my proposed impl to minimise rework.

irfanurrehman:
@marun this is ready for review now. PTAL!

pkg/kubefed2/enable/util.go
```diff
@@ -141,6 +119,37 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
 	return nil, errors.Errorf("Unable to find api resource named %q.", key)
 }
 
+func NameMatchesResource(name string, apiResource metav1.APIResource, group string) bool {
+	lowerCaseName := strings.ToLower(name)
```
marun: (No action required) Given that the key was previously lowercased outside of the loop, is there a reason not to continue that behavior?

pkg/kubefed2/federate/federate.go
```diff
@@ -44,9 +44,13 @@ import (
 const (
 	createResourceRetryTimeout  = 10 * time.Second
 	createResourceRetryInterval = 1 * time.Second
+	federationKindPrefix        = "Federated"
 )
 
 var (
+	// Controller created resources should always be skipped while federating content
+	controllerCreatedAPIResourceNames = []string{"events", "endpoints", "propagatedversions"}
```
marun: I don't think it's sufficient to use an unqualified name.  There is nothing stopping a 3rd party type from using the same name with a different group, so group-qualified is suggested.

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +58,104 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+		}
+
+		group := gv.Group
+		for _, apiResource := range apiResourceList.APIResources {
+			if !apiResource.Namespaced || isAlreadyFederated(apiResource) ||
+				apiResourceMatchesSkipNames(apiResource, skipAPIResourceNames, group) {
+				continue
+			}
+
+			// TODO(irfanurrehman): Define a strategy to federate the latest available version
```
marun: I would like to see this strategy implemented before this PR merges to avoid throwing away the care that  `enable` takes to avoid enabling types for ambiguous input.

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +58,104 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error listing resources for %s", apiResource.Kind)
+		}
+
+		targetResources := resources{apiResource: apiResource}
+		for _, resource := range resourceList.Items {
+			targetResources.resources = append(targetResources.resources, &resource)
+		}
+
+		// It would be a waste of cycles to iterate through empty slices while federating resource
```
marun: (No action required) Consider checking the length of `resourceList.Items` after retrieval and continuing if 0.

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +58,104 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
 		resource.SetNamespace(namespace)
 	}
 }
+
+func getAPIResourceList(config *rest.Config, skipAPIResourceNames string) (map[string]metav1.APIResource, error) {
```
marun: As per a previous comment, the name of this function doesn't seem to match what it does.  Consider renaming to something more appropriate (e.g. `apiResourceMap` or `containedAPIResources`).

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +58,104 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+			continue
+		}
+		if enable.NameMatchesResource(name, apiResource, group) {
+			return true
+		}
+	}
+	return false
+}
+
+func isAlreadyFederated(apiResource metav1.APIResource) bool {
```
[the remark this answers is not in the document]
irfanurrehman: In what scenario would an API Resource with federation default prefix and federation default group not be considered a federated API resource, irrespective of it being configured by FTC or not? Probable the name I used is confusing, I think the more appropriate name would be `isFederatedAPIResource`. Updated to that.
Additionally querying the FTC for each API Resource incurs additional cost which makes the command slow (think 1 API call per API Resource) and super slow for remote (for example accessing gke cluster) API servers. (I am assuming the number of actual resources to federate in a namespace being lesser then all the listed API Resources).
Having said that, I don't mind adding FTC query as the first check and then fall back to checking the name if FTC not found, if there is a reason to do that. Am I missing something?

pkg/kubefed2/federate/federate.go
```diff
@@ -63,23 +67,26 @@ var (
...
+	federateContents     bool
+	skipAPIResourceNames string
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.resourceNamespace, "namespace", "n", "default", "The namespace of the resource to federate.")
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
+	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
+	flags.StringVarP(&j.skipAPIResourceNames, "skip-resources", "s", "", "Comma separated names of the api resources to skip when federating contents in a namespace.")
```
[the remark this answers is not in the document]
irfanurrehman: Deferred this change to another PR.

pkg/kubefed2/federate/federate.go
irfanurrehman: Updated.

pkg/kubefed2/federate/util.go
irfanurrehman: Updated!

pkg/kubefed2/federate/util.go
irfanurrehman: Updated! Thanks for the suggestion.

pkg/kubefed2/federate/util.go
irfanurrehman: Updated to `namespacedAPIResourceMap`

pkg/kubefed2/enable/util.go
irfanurrehman: In that case the caller of the utility function would need to take the responsibility of passing the lower cased name.

irfanurrehman:
@marun updated. PTAL!

irfanurrehman:
@marun its a little difficult to test this without skip resources. What do you say about adding the skip resources based on names alone in this and then implement other possibilities in a follow up.

pkg/kubefed2/enable/util.go
```diff
@@ -141,6 +119,37 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
...
+	if name == apiResource.Name ||
```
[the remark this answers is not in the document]
irfanurrehman: Good catch; Updated!

pkg/kubefed2/federate/federate.go
```diff
@@ -196,27 +220,29 @@ func GetFederateArtifacts(hostConfig *rest.Config, qualifiedTypeName, qualifiedN
...
+	return &FederateArtifacts{
 		typeConfigInstalled: typeConfigInstalled,
 		typeConfig:          typeConfig,
-		fedResource:         fedResource,
+		federatedResources:  federatedResources,
 	}, nil
 }
 
-func getTypeConfig(hostConfig *rest.Config, apiResource metav1.APIResource, typeName, namespace string, enableType, outputYAML bool) (bool, typeconfig.Interface, error) {
+func getTypeConfig(hostConfig *rest.Config, apiResource metav1.APIResource, typeNamespace string, enableType, outputYAML bool) (bool, typeconfig.Interface, error) {
```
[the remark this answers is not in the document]
irfanurrehman: Seems logical. Updated!

pkg/kubefed2/federate/federate.go
```diff
@@ -351,27 +411,57 @@ func CreateFedResource(hostConfig *rest.Config, artifacts *federateArtifacts, dr
...
+
+func writeUnstructuredObjsToYaml(unstructuredObjs []*unstructured.Unstructured, w io.Writer) error {
+	for _, unstructuredObj := range unstructuredObjs {
+		if _, err := w.Write([]byte("---\n")); err != nil {
+			return errors.Wrap(err, "Error encoding object to yaml")
+		}
+		err := util.WriteUnstructuredToYaml(unstructuredObj, w)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "Error encoding unstructured object to yaml")
```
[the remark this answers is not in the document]

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +59,126 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+
+func compareGroupVersions(newGV, oldGV schema.GroupVersion) int {
+	// The strategy involved to choose a Group higher in order here is to
+	// consider 'extensions' as the outdated group [This seems to be true for
+	// all k8s APIResources, so far]. For example deployments exists in
+	// 'extensions' and 'apps'; 'deployments.apps' will be chosen.
+	// This doesn't apply to events but events are listed in controllerCreatedAPIResourceNames
+	// and skipped always.
+	// TODO(irfanurrehman) Are there any other known groups which don't match this?
+	if oldGV.Group != newGV.Group {
```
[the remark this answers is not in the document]
irfanurrehman: 1. Before implementing this, I did go through the list of k8s APIs and found the below to be available across more then 1 group:
```
events      events.k8s.io
events       "" (core)

daemonsets   apps
daemonsets   extensions 
deployments   apps 
deployments   extensions 
replicasets      apps 
replicasets      extensions

networkpolicies   extensions 
networkpolicies   networking.k8s.io 

podsecuritypolicies  extensions   
podsecuritypolicies  policy 
 ```
Except `events` all others were earlier in `extensions` but evolved into their own groups; so the strategy chosen applies (Please correct me if I am wrong). 
However it seems logical to add code to error out in case any other API (say a CRD in future) which does not map to this strategy is discovered withing multiple groups. Updated to this effect.

> Note that in the case of extensions and apps, only apps has `v1`, so a straight version comparison might suffice.

2. Does not seem like the right strategy.. :).  Some more context -> We use `ServerPreferredResources` API on `discovery client` to get all API resources, which returns one version per group in almost all scenario. The code does not reach `CompareKubeAwareVersionStrings` thus. However I have still used the version comparison because `ServerPreferredResources` does not guarantee that it will always have a preferred version available for a given API resource.

File: test/e2e/federate.go [hunk 18]
```diff
@@ -75,60 +82,142 @@ var _ = Describe("Federate resource", func() {
...
+	}
+
+	if !reflect.DeepEqual(expectedResource, targetResource) {
+		tl.Fatalf("Federated object template and target object don't match for %s %q; expected: %v, target: %v", fedKind, qualifiedName, expectedResource, targetResource)
+	}
+}
 
-	if !reflect.DeepEqual(expectedObj, targetObj) {
-		tl.Fatal("Federated object template and target object don't match for %s %q", fedKind, qualifiedName)
+func deleteAllResources(f framework.FederationFramework, tl common.TestLogger, testResoures []testResources) {
```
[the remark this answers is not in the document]
irfanurrehman: Not really, removed!

test/e2e/federate.go
```diff
@@ -161,15 +250,24 @@ func deleteResource(tl common.TestLogger, client util.ResourceClient, qualifiedN
...
 	}
 	return fedResource
 }
 
-func getClient(tl common.TestLogger, typeConfig typeconfig.Interface, kubeConfig *restclient.Config) util.ResourceClient {
+func targetResourceFromAPI(tl common.TestLogger, typeConfig typeconfig.Interface, kubeConfig *restclient.Config, qualifiedName util.QualifiedName) *unstructured.Unstructured {
+	client := getTargetClient(tl, typeConfig, kubeConfig)
+	targetResource, err := client.Resources(qualifiedName.Namespace).Get(qualifiedName.Name, metav1.GetOptions{})
+	if err != nil {
+		tl.Fatalf("Federated resource %q not found: %v", qualifiedName, err)
```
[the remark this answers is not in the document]
irfanurrehman: Leaving `Federated` in the message was a copy paste mistake. Updated!

test/e2e/federate.go
```diff
@@ -178,3 +276,53 @@ func getClient(tl common.TestLogger, typeConfig typeconfig.Interface, kubeConfig
 	}
 	return client
 }
+
+func getTargetClient(tl common.TestLogger, typeConfig typeconfig.Interface, kubeConfig *restclient.Config) util.ResourceClient {
+	fedAPIResource := typeConfig.GetTarget()
```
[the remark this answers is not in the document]
irfanurrehman: Copy paste mistake again; Updated!

irfanurrehman:
@marun Updated PTAL. Also I have pushed another commit for review fixes to save time updating and rebasing the relevant commit wrt change.

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +59,134 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
 	}
 }
+
+func namespacedAPIResourceMap(config *rest.Config, skipAPIResourceNames string) (map[string]metav1.APIResource, error) {
+	apiResourceLists, err := enable.GetServerPreferredResources(config)
+	if err != nil {
+		return nil, err
+	}
+
+	apiResources := make(map[string]metav1.APIResource)
```
marun: The keys of this map should be `<group-qualified plural name>` rather than kind, since kind is not guaranteed to be globally unique.

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +59,134 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+	// This doesn't apply to events but events are listed in controllerCreatedAPIResourceNames
+	// and skipped always.
+	if oldGV.Group != newGV.Group && (oldGV.Group == "extensions" || newGV.Group == "extensions") {
+		if oldGV.Group == "extensions" {
+			// New resource is considered bigger by virtue of its group.
+			return 1, nil
+		}
+		return -1, nil
+	}
+	if oldGV.Group != newGV.Group {
```
marun: This block won't be necessary when using group-qualified names.

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +59,134 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+}
+
+func compareGroupVersions(newGV, oldGV schema.GroupVersion, kind string) (int, error) {
+	// The strategy involved to choose a Group higher in order here is to
+	// consider 'extensions' as the outdated group [This seems to be true for
+	// all k8s APIResources, so far]. For example if deployments exists in
+	// 'extensions' and 'apps'; 'deployments.apps' will be chosen.
+	// This doesn't apply to events but events are listed in controllerCreatedAPIResourceNames
+	// and skipped always.
+	if oldGV.Group != newGV.Group && (oldGV.Group == "extensions" || newGV.Group == "extensions") {
```
marun: I think it would be preferable to simply ignore the `extensions` group in `namespacedAPIResourceMap`:

```bash
group = gv.Group
// The `apps` group has been superseded by the `extensions` group 
if group == "extensions" {
    continue
}
```

pkg/kubefed2/federate/util.go
```diff
@@ -49,3 +59,134 @@ func SetBasicMetaFields(resource *unstructured.Unstructured, apiResource metav1.
...
+		}
+		return -1, nil
+	}
+	if oldGV.Group != newGV.Group {
+		// An unknown API Resource found in multiple groups
+		// TODO(irfanurrehman) Update this to suggest usage of --skip-resources to resolve this with skip resources impl.
+		return 0, errors.Errorf("Found API Resource %s in multiple groups, %s and %s with no known strategy of resolution.", kind, oldGV.Group, newGV.Group)
+	}
+
+	return versionhelper.CompareKubeAwareVersionStrings(newGV.Version, oldGV.Version), nil
```
marun: This call can probably be performed inline given that it will be the only remaining call in `compareGroupVersions`.

irfanurrehman:
> As per my inline comments, I think there's a bit more cleanup suggested for the collection of namespaced api resources. Provided the extensions group is ignored (and the events type is skipped), group-qualifying the map should ensure correct behavior for known types.

I am not sure if I understand you well. Wouldn't this mean that I would end up getting multiple resources for a given kind?
(For a little more context, if say a deployment is created in API with `GV=apps/v1`, a client query would return the same resource against `apps/v1` and `extensions/v1beta1`, in effect giving me two resources and 2 `federatedDeployments` as the result of federate)
The whole reason we are trying to implement this strategy is to pick up a single latest group and version for an API kind and if we cannot decide which should be the latest GV then error out to be on the correct side (i.e. let user decide which one to pick, by skipping the other).

pkg/kubefed2/federate/util.go
irfanurrehman: This wont work in the scenario if there is only one group exposed on an API server and that be `extensions`. Some API installation might choose to do that, isn't it?

pkg/kubefed2/federate/util.go
irfanurrehman: Please see [this comment](https://github.com/kubernetes-sigs/federation-v2/pull/741#issuecomment-488247153).

pkg/kubefed2/federate/util.go
irfanurrehman: Please see [this comment](https://github.com/kubernetes-sigs/federation-v2/pull/741#issuecomment-488247153).

pkg/kubefed2/federate/util.go
irfanurrehman: Please see [this comment](https://github.com/kubernetes-sigs/federation-v2/pull/741#issuecomment-488247153).

pkg/kubefed2/federate/util.go
marun: `extensions` is deprecated.  We are under no obligation to support deprecated APIs.

marun:
@irfanurrehman `Kind` does not make sense unqualified, any more than an apiresource name unqualified by group makes sense.  That's why groups exist - to allow a given kind or name to be defined in more than one API.  The context here is discovering all the types that exist in a namespace, and in a case where a namespace contains kind `MyType` with group-qualified name `mytypes.foo.com` and `MyType` with group-qualified name `mytypes.bar.com`, your proposed approach of keying only by unqualified kind would completely ignore one of those types.

irfanurrehman:
@marun After much thinking and discussion offline, I have update the code as per your suggestion. This does not handle the case where say a k8s server has exposed only `extensions` for kinds like `deployments` or `replicasets`. I did choose to not handle this case with some messy logic, as the `extensions` (as per our offline discussion) is anyways going to be deprecated. I do intend to update documentation for it though.

irfanurrehman:
@marun rebased on latest also.

marun:
@irfanurrehman Please rebase.

irfanurrehman:
@marun rebased

marun:
@irfanurrehman Thank you for your patience! LGTM. Please squash for merge.

irfanurrehman:
@marun merged the last 3 commits into 1 as they were post review and/or review/rebase fixes.
The earlier commits were intentional.

irfanurrehman:
@marun I discovered an issue with the code and pushed it as a separate commit here rather then another PR. I think this can now be merged as is.

PR #744, opened 2019-04-10
kubefed2 enable succeeds if federation of the type already enabled

### LINKED ISSUE #720: kubefed2 enable should succeed if federation of the type is enabled
Currently running `kubefed2 enable` the same type more than once will result in an error on subsequent invocations, e.g.

```
Error creating FederatedTypeConfig "services": federatedtypeconfigs.core.federation.k8s.io "services" already exists
```

Likely the command should not report the error if a FederatedTypeConfig exists.  Or it might make sense to add a flag to support overwriting an existing FTC.

font:
kubefed2 enable now succeeds if federation of the type is already enabled. If the federated type is already enabled, it will update the FederatedTypeConfig to stay consistent with #628 for CRDs.

Fixes #720

### DIFF
--- a/pkg/kubefed2/enable/enable.go
+++ b/pkg/kubefed2/enable/enable.go
@@ -267,14 +267,30 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
 	if err != nil {
 		return errors.Wrap(err, "Failed to get federation clientset")
 	}
+
 	concreteTypeConfig := resources.TypeConfig.(*fedv1a1.FederatedTypeConfig)
 	concreteTypeConfig.Namespace = namespace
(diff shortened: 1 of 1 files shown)


marun:
LGTM.  Will rely on an impartial reviewer to merge.

cc: @xunpan @shashidharatd 

As an aside, I'd really like to see some helper functions to facilitate idempotency factored out of this PR and #555 at some point in the near future.

xunpan:
/lgtm

After `Get`, even it tells not found at the moment, it may still failed in `Create` because others  operate on the same resource to create it. But it is fine. We do not need to guarantee a transction with complex code. And after the potential error above happens, it still succeed for the next execution.

Issue #750, opened 2019-04-10
Improve reliability of cluster health check


shashidharatd:
**What would you like to be added**:
  Currently, the cluster health check does a single probe and based on the result, the cluster state is changed. This could be erroneous. It is preferable to retry the probe in a row to ascertain the cluster condition accurately. This is also how it is done in [liveness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) for pods.

**Why is this needed**:
  A lot of things depends on cluster state and an faulty health check could simply switch cluster condition back and forth.

/kind feature

PR #755, opened 2019-04-11
Improve reliability of cluster health check

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/federation-v2/issues/750


pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,127 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
-			} else {
-				for i := 0; i < len(clusterStatusNew.Conditions); i++ {
-					if !(strings.EqualFold(string(clusterStatusNew.Conditions[i].Type), string(clusterStatusOld.Conditions[i].Type)) &&
-						strings.EqualFold(string(clusterStatusNew.Conditions[i].Status), string(clusterStatusOld.Conditions[i].Status))) {
-						hasTransition = true
-						break
-					}
-				}
-			}
+			store.resultRun = 1
```
irfanurrehman: There seems to be something wrong with setting this counter to 1 here.
Shouldn't this be after checking wrt thresholds?

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,127 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
+
+	wg.Wait()
+	return nil
+}
+
+func updateIndividualClusterStatus(client genericclient.Client, cluster *fedv1a1.FederatedCluster,
+	store *ClusterStore, clusterHealthCheckConfig *util.ClusterHealthCheckConfig, wg *sync.WaitGroup) {
+	clusterClient := store.clusterKubeClient
+
+	clusterStatusNew := clusterClient.GetClusterHealthStatus()
```
irfanurrehman: consider renaming this to `clusterStatusNow` or `currentClusterStatus`. `New` confused me a bit.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,127 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
-			// Don't return err here, as we want to continue processing remaining clusters.
-			continue
+	cluster.Status = *clusterStatusNew
+	if err := client.UpdateStatus(context.TODO(), cluster); err != nil {
+		glog.Warningf("Failed to update the status of cluster: %v, error is : %v", cluster.Name, err)
+	}
+	wg.Done()
+}
+
+func isProbeSuccessful(clusterStatus *fedv1a1.FederatedClusterStatus) bool {
```
irfanurrehman: This should rather be named something like `isClusterOnline` or `isClusterAvailable`, isn't it.

irfanurrehman:
@shashidharatd thanks for doing this.
Couple of naming nits and one flaw in the algorithm; not sure how this was tested and if the test can catch the same.

pkg/controller/federatedcluster/controller.go
shashidharatd: oops, some goof-up happened. Fixed it now. Also have added tests to verify.

pkg/controller/federatedcluster/controller.go
shashidharatd: Refactored. PTAL once again.

pkg/controller/federatedcluster/controller.go
shashidharatd: Have renamed to isClusterReady to reflect the cluster condition. PTAL.

shashidharatd:
Thanks @irfanurrehman for the review. Have worked on your comments. PTAL once again.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,146 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
+
+	wg.Wait()
+	return nil
+}
+
+func (cc *ClusterController) updateIndividualClusterStatus(cluster *fedv1a1.FederatedCluster,
+	store *ClusterStore, wg *sync.WaitGroup) {
+	clusterClient := store.clusterKubeClient
+
+	currentClusterStatus := thresholdAdjustedClusterStatus(clusterClient.GetClusterHealthStatus(), store,
```
marun: I think `clusterClient.GetClusterHealthStatus()` should be on its own line.  It is not a trivial call, and having it on its own line simplifies debugging and avoids obscuring the importance of the call.

pkg/controller/federatedcluster/controller_test.go
```diff
@@ -0,0 +1,107 @@
...
+		TimeoutSeconds:   3 * time.Second,
+	}
+
+	testCases := map[string]struct {
+		clusterStatus         *fedv1a1.FederatedClusterStatus
+		store                 *ClusterStore
+		expectedClusterStatus *fedv1a1.FederatedClusterStatus
+	}{
+		"ClusterReadyAtBegining": {
+			clusterStatus:         clusterStatus(common.ClusterReady, corev1.ConditionTrue, t1, t1),
```
marun: nit: Specifying `common.ClusterReady` seems entirely redundant given that every call to `clusterStatus` provides the same value.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,146 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
+	if store.clusterStatus == nil {
+		return clusterStatus
+	}
+
+	threshold := clusterHealthCheckConfig.FailureThreshold
+	if isClusterReady(clusterStatus) {
+		threshold = clusterHealthCheckConfig.SuccessThreshold
+	}
+
+	if store.resultRun < threshold {
```
yamt: i guess you should update resultRun before checking threshold.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,146 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
-					if !(strings.EqualFold(string(clusterStatusNew.Conditions[i].Type), string(clusterStatusOld.Conditions[i].Type)) &&
-						strings.EqualFold(string(clusterStatusNew.Conditions[i].Status), string(clusterStatusOld.Conditions[i].Status))) {
-						hasTransition = true
-						break
-					}
-				}
-			}
+	}
+
+	if clusterStatusEqual(clusterStatus, store.clusterStatus) {
```
yamt: how about "isClusterReady(clusterStatus) != isClusterReady(store.clusterStatus)"?
keeping resultRun small while flipping between two different non-ready conditions doesn't make much sense.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -201,6 +200,11 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	opts.LeaderElection.RenewDeadline = spec.LeaderElect.RenewDeadline.Duration
 	opts.LeaderElection.LeaseDuration = spec.LeaderElect.LeaseDuration.Duration
 
+	opts.Config.PeriodSeconds = spec.ClusterHealthCheck.PeriodSeconds.Duration
```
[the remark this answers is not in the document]
shashidharatd: Initial thought was that it would have made calling all controller start functions similar. But have updated now.

cmd/controller-manager/app/options/options.go
```diff
@@ -48,7 +47,10 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.Config.ClusterUnavailableDelay, "cluster-unavailable-delay", util.DefaultClusterUnavailableDelay, "Time to wait before giving up on an unhealthy cluster.")
 
 	fs.BoolVar(&o.LimitedScope, "limited-scope", false, "Whether the federation namespace will be the only target for federation.")
-	fs.DurationVar(&o.ClusterMonitorPeriod, "cluster-monitor-period", time.Second*40, "How often to monitor the cluster health")
+	fs.DurationVar(&o.Config.ClusterHealthCheckConfig.PeriodSeconds, "cluster-monitor-period", 10*time.Second, "How often to monitor the cluster health")
```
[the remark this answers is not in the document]
shashidharatd: There should be no impact. On the positive, The cluster state is monitored more frequently. The transition from Ready --> NotReady still takes 3 cycles by default. While the reverse takes 1 cycle by default configuration.

pkg/controller/federatedcluster/clusterclient.go
```diff
@@ -85,30 +87,35 @@ func (self *ClusterClient) GetClusterHealthStatus() *fedv1a1.FederatedClusterSta
...
 		LastTransitionTime: currentTime,
 	}
-	body, err := self.kubeClient.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do().Raw()
-	if err != nil {
-		clusterStatus.Conditions = append(clusterStatus.Conditions, newNodeOfflineCondition)
-	} else {
-		if !strings.EqualFold(string(body), "ok") {
-			clusterStatus.Conditions = append(clusterStatus.Conditions, newClusterNotReadyCondition, newNodeNotOfflineCondition)
+
+	for i := 0; i < maxHealthCheckRetries; i++ {
```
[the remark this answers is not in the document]
shashidharatd: Agree with you on retrying immediately doesn't makes sense. As we are taking care of thresholds, i think this change is not required. reverted now.

pkg/controller/federatedcluster/clusterclient.go
```diff
@@ -85,30 +87,35 @@ func (self *ClusterClient) GetClusterHealthStatus() *fedv1a1.FederatedClusterSta
...
+		clusterStatus.Conditions = clusterStatus.Conditions[:0]
+		body, err := self.kubeClient.DiscoveryClient.RESTClient().Get().AbsPath("/healthz").Do().Raw()
+		if err != nil {
```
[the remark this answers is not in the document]
shashidharatd: probably user could configure wrong cluster credentials, which might return error. Preserving the old behaviour for now

pkg/controller/federatedcluster/controller.go
```diff
@@ -81,17 +87,17 @@ func StartClusterController(config *util.ControllerConfig, stopChan <-chan struc
 }
 
 // newClusterController returns a new cluster controller
-func newClusterController(config *util.ControllerConfig, clusterMonitorPeriod time.Duration) (*ClusterController, error) {
-	client := genericclient.NewForConfigOrDieWithUserAgent(config.KubeConfig, "cluster-controller")
+func newClusterController(config *util.ControllerConfig) (*ClusterController, error) {
+	kubeConfig := restclient.CopyConfig(config.KubeConfig)
```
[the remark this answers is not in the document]
shashidharatd: Ack

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,146 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
+		cc.clusterHealthCheckConfig)
+
+	if utilfeature.DefaultFeatureGate.Enabled(features.CrossClusterServiceDiscovery) {
+		currentClusterStatus = updateClusterZoneAndRegion(currentClusterStatus, store, clusterClient, cluster.Name)
+	}
+
+	store.clusterStatus = currentClusterStatus
+	cluster.Status = *currentClusterStatus
+	if err := cc.client.UpdateStatus(context.TODO(), cluster); err != nil {
+		glog.Warningf("Failed to update the status of cluster: %v, error is : %v", cluster.Name, err)
```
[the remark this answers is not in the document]
shashidharatd: Updated

pkg/controller/federatedcluster/controller.go
```diff
@@ -20,42 +20,48 @@ import (
...
 
+	fedcommon "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/common"
+	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
 	genericclient "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+	"github.com/kubernetes-sigs/federation-v2/pkg/features"
 )
 
+// ClusterStore stores cluster client and previous health check probe results of individual cluster.
+type ClusterStore struct {
```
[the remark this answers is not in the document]
shashidharatd: For lack of better word, i named it ClusterStore. Now i have updated it with ClusterData. Not very happy though.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +150,146 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
-		cc.clusterStatusMap[cluster.Name] = *clusterStatusNew
-		cc.mu.Unlock()
-		cluster.Status = *clusterStatusNew
-		err = cc.client.UpdateStatus(context.TODO(), &cluster)
-		if err != nil {
-			glog.Warningf("Failed to update the status of cluster: %v, error is : %v", cluster.Name, err)
-			// Don't return err here, as we want to continue processing remaining clusters.
-			continue
+func clusterStatusEqual(newClusterStatus, oldClusterStatus *fedv1a1.FederatedClusterStatus) bool {
+	if len(newClusterStatus.Conditions) != len(oldClusterStatus.Conditions) {
```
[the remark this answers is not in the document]
shashidharatd: Updated

pkg/controller/federatedcluster/controller.go
shashidharatd: Done

pkg/controller/federatedcluster/controller_test.go
shashidharatd: Done

pkg/controller/federatedcluster/controller.go
shashidharatd: Taken care now.

pkg/controller/federatedcluster/controller.go
shashidharatd: your comment makes sense. I initially was pondering upon the same for quite some time. Updated to just store Ready or NotReady states. ptal once again.

shashidharatd:
@marun, @yamt, Thanks for the review. Have addressed your comments. PTAL.

pkg/controller/federatedcluster/clusterclient.go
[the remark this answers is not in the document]
marun: I think it would be preferable to not maintain the old behavior given that it hides details that is likely to be useful for troubleshooting.  If you're not intending to fix the problem in this PR I can raise a fix separately.

cmd/controller-manager/app/options/options.go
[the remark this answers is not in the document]
marun: I'm confused as to the value of this change.  If a given cluster is flakey, why should 3 cycles be required to determine `NotReady` but only a single cycle to determine `Ready`?  Is there a precedent for this strategy?

pkg/apis/core/v1alpha1/federationconfig_types.go
```diff
@@ -53,6 +53,13 @@ type FeatureGatesConfig struct {
 	Enabled bool   `json:"enabled,omitempty"`
 }
 
+type ClusterHealthCheckConfig struct {
+	PeriodSeconds    metav1.Duration `json:"period-seconds,omitempty"`
```
marun: If your intention is to require the use of seconds, I think `int` would be a more appropriate data type.  Is there precedent here that suggests assuming seconds vs allowing any value to be used?  Same comment on `TimeoutSeconds` below.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +149,138 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
-			} else {
-				if len(zone) == 0 {
-					zone = cluster.Status.Zone
-				}
-				if len(region) == 0 {
-					region = cluster.Status.Region
-				}
-				clusterStatusNew.Zone = zone
-				clusterStatusNew.Region = region
+func isClusterReady(clusterStatus *fedv1a1.FederatedClusterStatus) bool {
```
marun: Please reuse `util.IsClusterReady` instead of adding a new implementation.

pkg/controller/federatedcluster/controller_test.go
```diff
@@ -0,0 +1,106 @@
...
+			expectedClusterStatus: clusterStatus(corev1.ConditionTrue, t5, t5),
+		},
+	}
+
+	for testName, tc := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			newClusterStatus := thresholdAdjustedClusterStatus(tc.clusterStatus, tc.storedClusterData, config)
+			if !reflect.DeepEqual(tc.expectedClusterStatus, newClusterStatus) {
+				t.Fatalf("Unexpected state, expected: %v, got:%v", tc.expectedClusterStatus, newClusterStatus)
+			}
```
yamt: i guess it's better to check if resultRun is updated as expected.

yamt:
@marun 
afaik it's a common strategy at least for healthcheck for loadbalancers.
i don't know if it's appropriate for our purpose though.

eg. https://kubernetes.io/docs/concepts/services-networking/service/#other-elb-annotations
service.beta.kubernetes.io/aws-load-balancer-healthcheck-healthy-threshold
service.beta.kubernetes.io/aws-load-balancer-healthcheck-unhealthy-threshold

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +149,138 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
+	}
+
+	if storedData.resultRun < threshold {
+		// Success/Failure is below threshold - leave the probe state unchanged.
+		probeTime := clusterStatus.Conditions[0].LastProbeTime
+		clusterStatus = storedData.clusterStatus
+		setLastProbeTime(clusterStatus, probeTime)
+	} else {
+		if clusterStatusEqual(clusterStatus, storedData.clusterStatus) {
+			// preserve the last transition time
```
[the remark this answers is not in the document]
yamt: or maybe get and fill zone/region info only when the cluster is ready.

cmd/controller-manager/app/options/options.go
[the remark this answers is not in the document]
shashidharatd: One such precedent is here while configuring kubernetes liveness probe https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes

pkg/controller/federatedcluster/clusterclient.go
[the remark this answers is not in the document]
shashidharatd: Have added runtime.HandleError. hope this is as you expected.

pkg/controller/federatedcluster/controller.go
shashidharatd: I think it was a mistake and messes up the logic. Have updated PTAL again.

pkg/apis/core/v1alpha1/federationconfig_types.go
shashidharatd: Agree, `int` is more appropriate data type. Have updated now. PTAL

pkg/controller/federatedcluster/controller.go
shashidharatd: Could not use `util.IsClusterReady`, as the parameter it takes is `FederatedCluster` object against the `FederatedClusterStatus` as being used in this PR. It would be lot of unrelated change. so skipping it and hope its fine.

pkg/controller/federatedcluster/controller.go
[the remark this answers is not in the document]
shashidharatd: Makes sense. we should be getting zone region labels only when cluster is Ready. Added a check.

pkg/controller/federatedcluster/controller.go
```diff
@@ -160,82 +149,138 @@ func (cc *ClusterController) Run(stopChan <-chan struct{}) {
...
+	if storedData.clusterStatus == nil {
+		return clusterStatus
+	}
+
+	threshold := clusterHealthCheckConfig.FailureThreshold
+	if isClusterReady(clusterStatus) {
+		threshold = clusterHealthCheckConfig.SuccessThreshold
+	}
+
+	if clusterStatusEqual(clusterStatus, storedData.clusterStatus) {
```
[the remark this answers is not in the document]
shashidharatd: > the intention here is to compare the latest observed status with the one of the previous iteration, right?

yes partially. the intention is to compare the latest observed status with the previous state, not just the previous iteration. it is represented by `resultRun`

pkg/controller/federatedcluster/controller_test.go
shashidharatd: Agree. Added a check in tests.

shashidharatd:
@marun, @yamt, Addressed your comments PTAL

pkg/controller/federatedcluster/controller.go
marun: Why not refactor, then, so that one method takes `FederatedCluster` and the other `FederatedClusterStatus`?  Or update callers of `IsClusterReady` to pass status instead (since passing cluster just to use status violates the law of demeter)?  I'm not clear why duplicating logic like this is ever desirable.

marun:
LGTM, thank you for responding to my concerns.  You will need to rebase and switch to the use of FederationConfig, though, now that #737 has merged.

pkg/controller/federatedcluster/controller.go
shashidharatd: Refactored and reused `util.IsClusterReady` now.

shashidharatd:
@marun, rebased. PTAL. Thanks!

marun:
Thank you muchly!

/lgtm

PR #757, opened 2019-04-12
Fix CI fail when there has daemon.json already

### LINKED ISSUE #756: CI failed in configure registry
**Which test(s) are failing**:
`Configuring container registry on host`

**Since when has it been failing**:
yesterday as far as I know

**Reason for failure**:
It seems CI environment issue.

**Anything else we need to know**:
```
Configuring container registry on host
Error: /etc/docker/daemon.json exists and CONFIGURE_INSECURE_REGISTRY=y. This script needs to add an 'insecure-registries' entry with host '172.17.0.1:5000' to /etc/docker/daemon.json. Please make the necessary changes or backup and try again.
The command "test $DOC_ONLY && echo "Skipped!" || DOWNLOAD_BINARIES=y bash -x ./scripts/pre-commit.sh" exited with 1.
```

/kind failing-test

xunpan:
CI fails because daemon.json exists. It may be related to CI environment change. But logically, it is not good to fail if the file exists. If exists, we only need to add our own configuration instead of reporting error.

Fixes #756

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -38,12 +38,15 @@ function create-insecure-registry() {
 function configure-insecure-registry() {
   local err=
   if sudo test -f "${docker_daemon_config}"; then
-    echo <<EOF "Error: ${docker_daemon_config} exists and \
-CONFIGURE_INSECURE_REGISTRY=${CONFIGURE_INSECURE_REGISTRY}. This script needs \
-to add an 'insecure-registries' entry with host '${CONTAINER_REGISTRY_HOST}' to \
-${docker_daemon_config}. Please make the necessary changes or backup and try again."
+    if sudo grep -q "\"insecure-registries\": \[\"${CONTAINER_REGISTRY_HOST}\"\]" ${docker_daemon_config}; then
+      return 0
+    elif sudo grep -q "\"insecure-registries\": " ${docker_daemon_config}; then
+      echo <<EOF "Error: ${docker_daemon_config} exists and \
+is already configured with an 'insecure-registries' entry but not set to ${CONTAINER_REGISTRY_HOST}. \
+Please make sure it is removed and try again."
 EOF
-    err=true
+      err=true
+    fi
   elif pgrep -a dockerd | grep -q 'insecure-registry'; then
     echo <<EOF "Error: CONFIGURE_INSECURE_REGISTRY=${CONFIGURE_INSECURE_REGISTRY} \
 and about to write ${docker_daemon_config}, but dockerd is already configured with \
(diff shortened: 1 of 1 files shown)


scripts/create-clusters.sh
```diff
@@ -38,12 +38,13 @@ function create-insecure-registry() {
...
   local err=
   if sudo test -f "${docker_daemon_config}"; then
-    echo <<EOF "Error: ${docker_daemon_config} exists and \
-CONFIGURE_INSECURE_REGISTRY=${CONFIGURE_INSECURE_REGISTRY}. This script needs \
-to add an 'insecure-registries' entry with host '${CONTAINER_REGISTRY_HOST}' to \
-${docker_daemon_config}. Please make the necessary changes or backup and try again."
+    if sudo grep -q 'insecure-registries' ${docker_daemon_config}; then
+      echo <<EOF "Error: ${docker_daemon_config} exists and \
+is already configured with an 'insecure-registries' entry. \
+Please make sure it is removed and try again."
```
shashidharatd: (suggestion) Most probably the `insecure-registries entry` is added by our own script in local environment. In Such cases the entry could be correct. if so, just continue the work. otherwise report error about correcting insecure-registries entry.

xunpan: make sense. 

My fix is to check if the `insecure-registries` is set exactly by our own script or not.
If yes, just return success. If not, print error message.

scripts/create-clusters.sh
```diff
@@ -38,12 +38,15 @@ function create-insecure-registry() {
...
   if sudo test -f "${docker_daemon_config}"; then
-    echo <<EOF "Error: ${docker_daemon_config} exists and \
-CONFIGURE_INSECURE_REGISTRY=${CONFIGURE_INSECURE_REGISTRY}. This script needs \
-to add an 'insecure-registries' entry with host '${CONTAINER_REGISTRY_HOST}' to \
-${docker_daemon_config}. Please make the necessary changes or backup and try again."
+    if sudo grep -q "\"insecure-registries\": \[\"${CONTAINER_REGISTRY_HOST}\"\]" ${docker_daemon_config}; then
+      return 0
+    else
+      echo <<EOF "Error: ${docker_daemon_config} exists and \
+is already configured with an 'insecure-registries' entry but not set to ${CONTAINER_REGISTRY_HOST}. \
```
shashidharatd: A small problem with the error message now. This branch may also be entered when there is no 'insecure-registries' entry.

xunpan:
CI environment has its own "registry-mirrors" definition in `daemon.json which make our CI failed.

scripts/create-clusters.sh
xunpan: done.

shashidharatd:
/lgtm
Thanks @xunpan

PR #764, opened 2019-04-15
Fix configuring insecure registry in local environment

shashidharatd:
Recently while fixing the CI failure in https://github.com/kubernetes-sigs/federation-v2/pull/757, we had overlooked into a problem and it caused problem in local environment, but it seems the CI env is good.

This pr will fix the issue for local environment.

/cc @xunpan @font

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -75,16 +75,17 @@ function configure-insecure-registry-and-reload() {
 }
 
 function insecure-registry-config-cmd() {
-  if test -f "${docker_daemon_config}"; then
-    sed -i "1a \    \"insecure-registries\": [\"${CONTAINER_REGISTRY_HOST}\"]" ${docker_daemon_config}
-  else
-    echo "cat <<EOF > ${docker_daemon_config}
+  echo "
+    if test -f ${docker_daemon_config}; then
+      sed -i '1a \    \\\"insecure-registries\\\": [\\\"${CONTAINER_REGISTRY_HOST}\\\"]' ${docker_daemon_config}
+    else
+      cat <<EOF > ${docker_daemon_config}
 {
     \"insecure-registries\": [\"${CONTAINER_REGISTRY_HOST}\"]
 }
 EOF
+    fi
 "
-  fi
 }
 
 function reload-docker-daemon-cmd() {


scripts/create-clusters.sh
```diff
@@ -75,16 +75,17 @@ function configure-insecure-registry-and-reload() {
 }
 
 function insecure-registry-config-cmd() {
-  if test -f "${docker_daemon_config}"; then
-    sed -i "1a \    \"insecure-registries\": [\"${CONTAINER_REGISTRY_HOST}\"]" ${docker_daemon_config}
-  else
-    echo "cat <<EOF > ${docker_daemon_config}
+  echo "
```
marun: What is your intent in echoing a command instead of executing it?

scripts/create-clusters.sh
```diff
@@ -75,16 +75,17 @@ function configure-insecure-registry-and-reload() {
...
+    if test -f ${docker_daemon_config}; then
+      sed -i '1a \    \\\"insecure-registries\\\": [\\\"${CONTAINER_REGISTRY_HOST}\\\"]' ${docker_daemon_config}
```
xunpan: What is the reason to add use `\\\"` instead of `\"`? We only need to add `"insecure-registries"` here.

scripts/create-clusters.sh
xunpan: The returned value of this function `insecure-registry-config-cmd` is to be executed as a command. I just overlooked it before. 
```
${cmd_context} "$(insecure-registry-config-cmd)"
```

xunpan:
LGTM
CI failed because environment issue. I just restart it to have another try.

shashidharatd:
yes, indeed the CI failure is related to the changes in this PR. Need some more digging into the issue.

scripts/create-clusters.sh
font: We need to add a trailing comma at the end of this config option since we have other options already in the file.

scripts/create-clusters.sh
```diff
@@ -75,16 +75,17 @@ function configure-insecure-registry-and-reload() {
 }
 
 function insecure-registry-config-cmd() {
-  if test -f "${docker_daemon_config}"; then
-    sed -i "1a \    \"insecure-registries\": [\"${CONTAINER_REGISTRY_HOST}\"]" ${docker_daemon_config}
-  else
-    echo "cat <<EOF > ${docker_daemon_config}
+  echo "
+    if test -f ${docker_daemon_config}; then
```
font: It does not suffice to just check it on the host because it will assume the host's existence of `/etc/docker/daemon.json` when inside the `clusterX-control-plane` docker containers it does not exist. So this check needs to be done on the host AND inside the cluster because it's called from 2 different contexts.

shashidharatd:
Closing this as it is no longer require.

font:
We can close this for now, but I've seen travis CI failures resulting from an `/etc/docker/daemon.json` file already existing on the host, which was the original intent of #757. However, it doesn't seem to happen often. If/when it does, we'll need to address it again based on comments in this PR.

Issue #680, opened 2019-03-21
Releases should use make targets to push images to quay


pmorie:
Created after #679; we should use a make target instead of a separate script to push images during a release.

qinpingli:
Pull request #766 has been merged, @pmorie could you help close this issue?

marun:
@qinpingli Thanks!

PR #766, opened 2019-04-16
Bug fix for issue #680

qinpingli:
Use `make push` to replace scripts/imagebuild.sh


marun:
@qinpingli Thank you for doing this!   Please update `.travis.yml` to ensure your change is tested and remove `scripts/imagebuild.sh`.

qinpingli:
@marun updated the .travis.yml file and removed the scripts/imagebuild.sh file, please help check, thx.

qinpingli:
@font Seems travis has logged into the quay.io (https://api.travis-ci.org/v3/job/521063358/log.txt),
and no QUAY_USERNAME env in travis(https://api.travis-ci.org/v3/job/520924639/log.txt).

qinpingli:
@font Thank you! Added docker login into Makefile.

Makefile
```diff
@@ -116,8 +118,14 @@ endif
 	go generate ./pkg/... ./cmd/...
 	./scripts/sync-up-helm-chart.sh
 
-push:
-	$(DOCKER) push $(REGISTRY)/$(TARGET):$(GIT_VERSION)
+push: container
+	docker login -u "$(QUAY_USERNAME)" -p "$(QUAY_PASSWORD)" quay.io
```
font: Two issues here:

1. We should use the global `$(DOCKER)` variable.

2. This will be executed by users that may not be using quay and won't have these user/pass variables defined. I suggest passing a flag from `.travis.yml` when invoking `make push` and adding a check for that flag before running this `docker login` command. Or, you can create a new make target such as `travis_login` that is invoked by `.travis.yml` before invoking `make push`.

qinpingli: Added a new target `travis_login`, but seems there is no `QUAY_USERNAME` and `QUAY_PASSWORD` env params.

```
$ test $DOC_ONLY && echo "Skipped!" || make travis_login && make push
docker login -u "" -p "" quay.io
Username: 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```
@font  Could you help take a look? Thank you!

font: @qinpingli That's expected as those credential environment variables are only defined whenever travis runs an official build from master or a new tagged release of federation. So we only need to run the `docker login` command when we need to push to the official federation registry repo. For example from the `imagebuild.sh` script, we could invoke the new `travis_login` target (or rename to `travis_image_build` or similar), but within it we would need to check `TRAVIS_TAG` and `TRAVIS_BRANCH` exists and match `master` or a tagged release version to see if we need to login and push, otherwise we don't bother.

qinpingli:
Added a new target for travis.

marun:
@qinpingli Why is this PR failing CI?

qinpingli:
> @qinpingli Why is this PR failing CI?
@marun failed on `docker login into quay.io`, for there is no `QUAY_USERNAME` and `QUAY_PASSWORD` env params.
This is the log:
```
if [  ] && [ x = x ] || [ masterx = masterx ]; \
then \
	docker login -u "" -p "" quay.io; \
fi
Username: 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```
In this PR, doing `docker login` cmd only when the `TRAVIS_TAG` is one of git tag, or `TRAVIS_BRANCH` is `master`.

marun:
@qinpingli @font Is this still being worked on?

Makefile
```diff
@@ -126,6 +126,27 @@ push:
 		$(DOCKER) push $(REGISTRY)/$(TARGET):latest; \
 	fi
 
+travis_image_push: container
+
+	if [ $(GIT_TAG) ] && [ $(GIT_TAG)x = $(TRAVIS_TAG)x ] || [ $(TRAVIS_BRANCH)x = masterx ]; \
```
font: `TRAVIS_BRANCH` is always going to be set here as for builds triggered by a pull request this is the name of the branch targeted by the pull request i.e. `master`. The old script even had a bug here as can be seen in a recent travis run https://travis-ci.org/kubernetes-sigs/federation-v2/builds/527879538#L3120. I suggest using the `TRAVIS_PULL_REQUEST` environment variable to determine whether `docker login` is necessary as it is:

> The pull request number if the current job is a pull request, “false” if it’s not a pull request.

Then you can move this logic within the `push` target and just add the `master` check to it for the `canary` `tag` and `push` commands.

Also, is the trailing `x` necessary? Presumably we should be using `bash`.

qinpingli: Thank you @font . Updated the PR, please help take a look.

Makefile
```diff
@@ -126,6 +132,12 @@ push:
 		$(DOCKER) push $(REGISTRY)/$(TARGET):latest; \
 	fi
 
+	if [ "$(GIT_BRANCH)" = "master" ]; \
```
font: This can still be `master` for a pull request if the user is using their own `master` branch even though it wouldn't be advised. I suggest putting this check within the `TRAVIS_PULL_REQUEST` check so that it is only run for official travis builds from `master`.

Makefile
```diff
@@ -116,8 +117,13 @@ endif
 	go generate ./pkg/... ./cmd/...
 	./scripts/sync-up-helm-chart.sh
 
-push:
-	$(DOCKER) push $(REGISTRY)/$(TARGET):$(GIT_VERSION)
```
font: This line should not be removed so that users can still push to whatever registry they're using when invoking `make push`. But it should **not** be run if running in the `after_success` `travis` phase.

We basically need to support `make push` for users, and for travis to build official tagged images. The idea is that we could use the same `push` target for both and just add checks within it for both cases. Feel free to split it out if it would help simplify things.

qinpingli: Thank you @font !
Now, 
1. for users, not the Travis CI call the `make push`, it will keep the same with the original.
2. for Travis CI, it will do not push anything for PRs and will build and push official tagged images.

Makefile
```diff
@@ -116,8 +117,17 @@ endif
...
+push: container
+
+	if [ "$(TRAVIS_PULL_REQUEST)" = "false" ]; \
+	then \
+		$(DOCKER) login -u "$(QUAY_USERNAME)" -p "$(QUAY_PASSWORD)" quay.io; \
+	fi
+
+        ifeq ($(TRAVIS_PULL_REQUEST),)
```
font: How come you've chosen a make conditional here but use shell conditionals elsewhere?

Makefile
```diff
@@ -126,6 +136,12 @@ push:
 		$(DOCKER) push $(REGISTRY)/$(TARGET):latest; \
 	fi
 
+	if [ "$(TRAVIS_PULL_REQUEST)" = "false" ] && [ "$(GIT_BRANCH)" = "master" ]; \
```
font: You can insert this `master` branch conditional check inside the previous `TRAVIS_PULL_REQUEST` check.

Makefile
qinpingli: Updated to use shell.

Makefile
qinpingli: Done.

shashidharatd:
Thanks @qinpingli !
/lgtm

PR #767, opened 2019-04-16
Give controller cluster role ability to patch

Joseph-Irving:
The controller currently lacks the permission to patch events which causes some errors. To rectify this I've added patch to the cluster role
```
'events "test-deployment.1595eea7da7b10cc" is forbidden: User "system:serviceaccount:federation-system:default" cannot patch resource "events" in API group "" in the namespace "test-namespace"' (will not retry!)
```

### DIFF
--- a/charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
+++ b/charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
@@ -66,4 +66,5 @@ rules:
   - list
   - create
   - update
+  - patch
 {{- end }}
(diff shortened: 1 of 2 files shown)


xunpan:
@Joseph-Irving Thanks for your PR to fix this. It is valid to me.
Could you help to fix this in role.yaml? It has similar issue in namespaced deployment.

Joseph-Irving:
@xunpan yeah of course, I missed that, I've updated it to include both now

marun:
/approve

@xunpan I guess this wasn't causing CI failures, just errors in the log?

xunpan:
@marun 
I think so because CI not failed or with message in many PRs. 
I saw that `k8s.io/client-go/tools/record/event.go::recordEvent()` uses `Patch` to update events. So, there should be a case, which this PR fixed, to update existing events with `Patch`.

marun:
@xunpan I think there should be a new rule that is limited to patching events, rather than updating the rule for resources that afaik we don't need that permission for: namespace, configmap, secret.

Issue #721, opened 2019-04-03
kubefed2 disable should delete the FTC rather than set propagationEnabled


marun:
`kubefed2 enable` adds the crd and ftc necessary for federation of a type.  Currently `kubefed2 disable` sets `propagationEnabled: false` on the target FTC (and deletes the crd and ftc if `--delete-from-api=true` is provided).  Instead, `kubefed2 disable` should be the reverse of `enable` and delete the target FTC and optionally the crd.  A user can edit the FTC if they want to disable propagation.

/kind bug

marun:
As part of this bug, the `disable` command should return without error if the type is already disabled.

marun:
When optionally deleting the crd, `kubefed2` should delete the ftc (which will ensure the sync controller is stopped) and then iterate over all instances of the crd and remove the federation finalizer to ensure that garbage collection is possible.  Only then should deletion of the crd be attempted.

font:
Closing this as fixed by #769. I've created a new issue https://github.com/kubernetes-sigs/federation-v2/issues/829 to track handling the deletion of custom resources.

PR #769, opened 2019-04-16
kubefed2 disable should delete the FTC rather than set propagationEnabled

font:
With this change `kubefed2 disable` now deletes the `FederatedTypeConfig` instead of just setting `propagationEnabled` to `false`. It will also wait for the sync controller to shut down before deleting it, and optionally delete the federated type CRD as well.

Partial fix for #721.

### DIFF
--- a/pkg/kubefed2/disable.go
+++ b/pkg/kubefed2/disable.go
@@ -63,8 +63,8 @@ var (
 		kubefed2 disable deployments.apps
 
 		# Disable propagation of the kubernetes API type 'Deployment', named
-		in FederatedTypeConfig as 'deployments.apps', and delete corresponding
-		Federated API resource
+		in FederatedTypeConfig as 'deployments.apps', and delete the
+		corresponding Federated API resource
 		kubefed2 disable deployments.apps --delete-crd`
 )
 
@@ -75,14 +75,14 @@ type disableType struct {
 }
 
 type disableTypeOptions struct {
-	delete              bool
+	deleteCRD           bool
 	enableTypeDirective *enable.EnableTypeDirective
 }
 
 // Bind adds the disable specific arguments to the flagset passed in as an
 // argument.
 func (o *disableTypeOptions) Bind(flags *pflag.FlagSet) {
-	flags.BoolVar(&o.delete, "delete-crd", false, "Whether to remove the API resource added by 'enable'.")
+	flags.BoolVar(&o.deleteCRD, "delete-crd", false, "Whether to remove the API resource added by 'enable'.")
 }
 
 // NewCmdTypeDisable defines the `disable` command that
@@ -119,13 +119,13 @@ func NewCmdTypeDisable(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
 // Complete ensures that options are valid and marshals them if necessary.
 func (j *disableType) Complete(args []string) error {
 	j.enableTypeDirective = enable.NewEnableTypeDirective()
-	fd := j.enableTypeDirective
+	directive := j.enableTypeDirective
 
 	if err := j.SetName(args); err != nil {
 		return err
 	}
 
-	if !j.delete {
+	if !j.deleteCRD {
 		if len(j.TargetVersion) > 0 {
 			return errors.New("--version flag valid only with --delete-crd")
 		} else if j.FederationGroup != options.DefaultFederationGroup {
@@ -134,10 +134,10 @@ func (j *disableType) Complete(args []string) error {
 	}
 
 	if len(j.TargetVersion) > 0 {
-		fd.Spec.TargetVersion = j.TargetVersion
+		directive.Spec.TargetVersion = j.TargetVersion
 	}
 	if len(j.FederationGroup) > 0 {
-		fd.Spec.FederationGroup = j.FederationGroup
+		directive.Spec.FederationGroup = j.FederationGroup
 	}
 
 	return nil
@@ -167,11 +167,11 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		Name:      name,
 	}
 	j.enableTypeDirective.Name = typeConfigName.Name
-	return DisableFederation(cmdOut, hostConfig, j.enableTypeDirective, typeConfigName, j.delete, j.DryRun)
+	return DisableFederation(cmdOut, hostConfig, j.enableTypeDirective, typeConfigName, j.deleteCRD, j.DryRun)
 }
 
 func DisableFederation(cmdOut io.Writer, config *rest.Config, enableTypeDirective *enable.EnableTypeDirective,
-	typeConfigName ctlutil.QualifiedName, delete, dryRun bool) error {
+	typeConfigName ctlutil.QualifiedName, deleteCRD, dryRun bool) error {
 	client, err := genericclient.New(config)
 	if err != nil {
 		return errors.Wrap(err, "Failed to get federation clientset")
@@ -199,7 +199,7 @@ func DisableFederation(cmdOut io.Writer, config *rest.Config, enableTypeDirectiv
 
 	if ftcExists {
 		if typeConfig.Spec.PropagationEnabled {
-			err = disablePropagationController(client, typeConfig, typeConfigName, write)
+			err = disablePropagation(client, typeConfig, typeConfigName, write)
 			if err != nil {
 				return err
 			}
@@ -211,7 +211,7 @@ func DisableFederation(cmdOut io.Writer, config *rest.Config, enableTypeDirectiv
 		}
 	}
 
-	if delete {
+	if deleteCRD {
 		if !ftcExists {
 			typeConfig, err = generatedFederatedTypeConfig(config, enableTypeDirective)
 			if err != nil {
@@ -241,7 +241,7 @@ func checkFederatedTypeConfigExists(client genericclient.Client, typeConfig *fed
 	return false, errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", typeConfigName)
 }
 
-func disablePropagationController(client genericclient.Client, typeConfig *fedv1a1.FederatedTypeConfig, typeConfigName ctlutil.QualifiedName, write func(string)) error {
+func disablePropagation(client genericclient.Client, typeConfig *fedv1a1.FederatedTypeConfig, typeConfigName ctlutil.QualifiedName, write func(string)) error {
 	if typeConfig.Spec.PropagationEnabled {
 		typeConfig.Spec.PropagationEnabled = false
 		err := client.Update(context.TODO(), typeConfig)
@@ -283,7 +283,7 @@ func deleteFederatedTypeConfig(client genericclient.Client, typeConfig *fedv1a1.
 	if err != nil {
 		return errors.Wrapf(err, "Error deleting FederatedTypeConfig %q", typeConfigName)
 	}
-	write(fmt.Sprintf("FederatedTypeConfig %q deleted\n", typeConfigName))
+	write(fmt.Sprintf("federatedtypeconfig %q deleted\n", typeConfigName))
 	return nil
 }
 
--- a/pkg/kubefed2/disable.go
+++ b/pkg/kubefed2/disable.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
@@ -28,6 +29,8 @@ import (
 	"github.com/spf13/pflag"
 
 	apiextv1b1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 
 	"github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
@@ -56,8 +59,8 @@ var (
 
 		# Disable propagation of the kubernetes API type 'Deployment', named
 		in FederatedTypeConfig as 'deployments.apps', and delete corresponding
-		Federated API resources
-		kubefed2 disable deployments.apps --delete-from-api`
+		Federated API resource
+		kubefed2 disable deployments.apps --delete-crd`
 )
 
 type disableType struct {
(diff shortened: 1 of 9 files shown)


pkg/kubefed2/disable.go
```diff
@@ -37,11 +37,17 @@ import (
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
 	genericclient "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
 	ctlutil "github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+
```
xunpan: Is this convention to add a space line here? It seems not necessary to me.

pkg/kubefed2/disable.go
```diff
@@ -252,6 +284,15 @@ func deleteFederatedTypeConfig(client genericclient.Client, typeConfig *fedv1a1.
 	return nil
 }
 
+func generatedFederatedTypeConfig(config *rest.Config, enableTypeDirective *enable.EnableTypeDirective) (*fedv1a1.FederatedTypeConfig, error) {
+	apiResource, err := enable.LookupAPIResource(config, enableTypeDirective.Name, enableTypeDirective.Spec.TargetVersion)
+	if err != nil {
+		return nil, err
+	}
+	typeConfig := enable.GenerateTypeConfigForTarget(*apiResource, enableTypeDirective).(*fedv1a1.FederatedTypeConfig)
```
xunpan: This works. But as caller functions only needs `typeconfig.Interface`, it does not needs such type conversion here.

font: This type assertion is necessary because this function returns a `*fedv1a1.FederatedTypeConfig`. It returns this because the `DisableFederation` caller is re-using its `typeConfig` variable (for simplicity) that is defined as a `*fedv1a1.FederatedTypeConfig` [here](https://github.com/kubernetes-sigs/federation-v2/pull/769/files#diff-fc79e5fb1808a541a91f4994f5f02254R187). `typeConfig` is defined this way because it is passed to multiple functions that expect different interface implementations e.g. `runtime.Object`, `typeconfig.Interface`.

pkg/kubefed2/disable.go
font: Not a convention, just an artifact left over from a previous `goimports` issue. I've removed.

pkg/kubefed2/disable.go
xunpan: I'm fine with the code anyway. 
I guess it is compromise for 2 case:
 - use a common variable in caller function and has type conversion in the function
 - use a different temparary variable in caller function and has no type conversion internally
Outter logic makes more sense than inner logic for code reading.

font: It is a compromise as you say. I prefer this implementation for a couple reasons:
1. It provides the most flexibility to the caller because it returns the same type passed in so that the caller is free to choose what interface implementation to use without performing any type assertion.
2. If I used a temporary variable in `DisableFederation`, I'd have to make 2 different calls to `deleteFederatedType` using 2 different `typeConfig` variables. Instead, this approach simplifies the code in `DisableFederation`.

I am happy to change it if you and others feel strongly.

pkg/kubefed2/disable.go
```diff
@@ -148,48 +151,104 @@ func DisableFederation(cmdOut io.Writer, config *rest.Config, typeConfigName ctl
 	if err != nil {
 		return errors.Wrap(err, "Failed to get federation clientset")
 	}
+
+	write := func(data string) {
```
irfanurrehman: [Not necessary to be done in this PR] This could be a common utility function both `enable` and `disable` could use.

pkg/kubefed2/disable.go
```diff
@@ -148,48 +151,104 @@ func DisableFederation(cmdOut io.Writer, config *rest.Config, typeConfigName ctl
...
+
+	if apierrors.IsNotFound(err) {
+		write(fmt.Sprintf("FederatedTypeConfig %q already removed\n", typeConfigName))
+		return false, nil
+	}
+
+	return false, errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", typeConfigName)
+}
+
+func disablePropagationController(client genericclient.Client, typeConfig *fedv1a1.FederatedTypeConfig, typeConfigName ctlutil.QualifiedName, write func(string)) error {
```
irfanurrehman: Isn't this `stopPropagationController`, meaning the controller for this type would actually be stopped; right?

pkg/kubefed2/disable.go
```diff
@@ -104,17 +110,35 @@ func NewCmdTypeDisable(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
...
 
 	return cmd
 }
 
 // Complete ensures that options are valid and marshals them if necessary.
 func (j *disableType) Complete(args []string) error {
-	if len(args) == 0 {
-		return errors.New("NAME is required")
+	j.enableTypeDirective = enable.NewEnableTypeDirective()
+	fd := j.enableTypeDirective
```
irfanurrehman: `fd` is probably confusing, `etd` or just `td` would be better for a reader with no prior code knowledge.. :-)

pkg/kubefed2/disable.go
irfanurrehman: I think its fine the way it is. Type conversion is anyways needed at some place, because of what @font mentioned [here](https://github.com/kubernetes-sigs/federation-v2/pull/769/commits/2f9112de2089e8f6b58415c7686b91df784d6cc8#r276788968). Also I think it makes sense to have this function return the concrete type as the caller can receive it in either the concrete type or an interface.
Also, let this not lead to bikeshedding.. :)

irfanurrehman:
@font apologies for delayed look at this. Few naming nits, otherwise good to go.

pkg/kubefed2/disable.go
```diff
@@ -139,60 +166,132 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
...
+}
+
+func checkPropagationControllerStopped(client genericclient.Client, typeConfigName ctlutil.QualifiedName, write func(string)) {
+	write(fmt.Sprintf("Checking propagation controller is stopped for FederatedTypeConfig %q\n", typeConfigName))
+
+	typeConfig := &fedv1a1.FederatedTypeConfig{}
+	err := wait.PollImmediate(100*time.Millisecond, 10*time.Second, func() (bool, error) {
+		err := client.Get(context.TODO(), typeConfig, typeConfigName.Namespace, typeConfigName.Name)
+		if err != nil {
+			return false, err
```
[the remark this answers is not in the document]
font: We can timeout instead. I've updated.

pkg/kubefed2/disable.go
```diff
@@ -139,60 +166,132 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
...
+		}
+
+		return false, nil
+	})
 
-	// TODO(marun) consider waiting for the sync controller to be stopped before attempting deletion
-	err = deleteFederatedType(config, typeConfig, write)
 	if err != nil {
-		return err
+		write(fmt.Sprintf("WARNING: unable to verify propagation controller for FederatedTypeConfig %q is stopped: %v\n", typeConfigName, err))
```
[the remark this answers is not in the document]
font: It is intentional. I don't think `check` is as strong as something like `verify`, which would suggest making sure something is done. I'm open to other suggestions for names.

pkg/kubefed2/disable.go
```diff
@@ -139,60 +166,132 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
...
 
-	err = client.Delete(context.TODO(), typeConfig, typeConfig.Namespace, typeConfig.Name)
+func deleteFederatedTypeConfig(client genericclient.Client, typeConfig *fedv1a1.FederatedTypeConfig, typeConfigName ctlutil.QualifiedName, write func(string)) error {
+	err := client.Delete(context.TODO(), typeConfig, typeConfig.Namespace, typeConfig.Name)
 	if err != nil {
 		return errors.Wrapf(err, "Error deleting FederatedTypeConfig %q", typeConfigName)
 	}
-	write(fmt.Sprintf("federatedtypeconfig %q deleted\n", typeConfigName))
-
+	write(fmt.Sprintf("FederatedTypeConfig %q deleted\n", typeConfigName))
```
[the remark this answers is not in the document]
font: Ah, I was wondering about that as it seemed inconsistent with the rest of the logs using camel case.

pkg/kubefed2/disable.go
```diff
@@ -202,8 +301,13 @@ func deleteFederatedType(config *rest.Config, typeConfig typeconfig.Interface, w
 	crdName := typeconfig.GroupQualifiedName(typeConfig.GetFederatedType())
 	err = client.CustomResourceDefinitions().Delete(crdName, nil)
 	if err != nil {
-		return errors.Wrapf(err, "Error deleting crd %q", crdName)
+		if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
font: Thanks, simplified to something similar.

pkg/kubefed2/disable.go
```diff
@@ -202,8 +301,13 @@ func deleteFederatedType(config *rest.Config, typeConfig typeconfig.Interface, w
 	crdName := typeconfig.GroupQualifiedName(typeConfig.GetFederatedType())
 	err = client.CustomResourceDefinitions().Delete(crdName, nil)
```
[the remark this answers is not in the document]
font: The latter is not implemented in this PR due to the following per our offline conversation:

1. Needing to discover the federation config (it has since been merged).
2. Possibly resolving it with docs instead of code due to member cluster resources.

With that in mind, are you suggesting we implement this in code before merging this or that we should at least add a `--force` option prior to merging?

font:
Thanks @irfanurrehman and @marun for the review! I've pushed a new commit addressing most comments and will need to discuss a couple more.

font:
@irfanurrehman @marun I've rebased and added a check to avoid deleting the federated type CRD if any of its CRs exist. Pending your review, I think that should allow this to merge and we can think about how to support a possible `--force` option to also remove the CRs prior to deleting the CRD while considering cluster and namespaced scoped deployments.

pkg/kubefed2/disable.go
```diff
@@ -298,12 +299,45 @@ func generatedFederatedTypeConfig(config *rest.Config, enableTypeDirective *enab
 }
 
 func deleteFederatedType(config *rest.Config, typeConfig typeconfig.Interface, write func(string)) error {
+	federatedTypeAPIResource := typeConfig.GetFederatedType()
+	crdName := typeconfig.GroupQualifiedName(federatedTypeAPIResource)
+	exists, err := federatedTypeCustomResourcesExist(config, &federatedTypeAPIResource)
+	if err != nil {
+		return err
+	} else if exists {
+		return errors.Errorf("Cannot delete CRD %q while resource instances exist. Please delete them and try again or remove --delete-crd option to disable only\n", crdName)
```
irfanurrehman: I found the language a little confusing.
Consider rewording to "Please try `kubefed2 disable` again after removing the resource instances or without the option `--delete-CRD`"
Or something similar. The usage of the word disable is a little tricky, I think.

irfanurrehman:
> @irfanurrehman @marun I've rebased and added a check to avoid deleting the federated type CRD if any of its CRs exist. Pending your review, I think that should allow this to merge and we can think about how to support a possible `--force` option to also remove the CRs prior to deleting the CRD while considering cluster and namespaced scoped deployments.

+1 to implementing `--force` in a follow up.

pkg/kubefed2/disable.go
font: Updated, PTAL.

irfanurrehman:
Thanks @font. lgtm. Would wait for @marun to verify (and merge).

pkg/kubefed2/disable.go
```diff
@@ -139,75 +167,184 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
...
+}
+
+func checkPropagationControllerStopped(client genericclient.Client, typeConfigName ctlutil.QualifiedName, write func(string)) {
+	write(fmt.Sprintf("Checking propagation controller is stopped for FederatedTypeConfig %q\n", typeConfigName))
+
+	var typeConfig *fedv1a1.FederatedTypeConfig
+	err := wait.PollImmediate(100*time.Millisecond, 10*time.Second, func() (bool, error) {
+		typeConfig = &fedv1a1.FederatedTypeConfig{}
+		err := client.Get(context.TODO(), typeConfig, typeConfigName.Namespace, typeConfigName.Name)
+		if err != nil {
```
marun: Is there a good reason not to log this error?

pkg/kubefed2/disable.go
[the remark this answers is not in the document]
marun: Why is it desirable to wait 10s for the controller to stop but not fail if it takes longer than that?  Is that an indication that disabling propagation is unnecessary and the ftc can just be deleted without first being disabled?

pkg/kubefed2/disable.go
```diff
@@ -139,75 +167,184 @@ func (j *disableType) Run(cmdOut io.Writer, config util.FedConfig) error {
...
-		if cmdOut == nil {
-			return
+	if deleteCRD {
+		if !ftcExists {
+			typeConfig, err = generatedFederatedTypeConfig(config, enableTypeDirective)
+			if err != nil {
+				return err
+			}
 		}
+		err = deleteFederatedType(config, typeConfig, write)
```
marun: I'm recalling the original reason why the propagation was being disabled.  Deleting the crd while propagation is still being performed is likely to generate a bunch of spurious error messages in the controller manager log as the watches are terminated and can't be reestablished.  Deleting the FTC before the CRD risks breaking idempotent behavior if deletion of the CRD is subsequently unsuccessful, since a subsequent disable attempt may not be able to find the CRD again (if it has non-standard group/version/name/etc).  Disabling, deleting CRD, then deleting FTC would seem the cleanest path.

pkg/kubefed2/disable.go
font: Avoiding dumping 100 errors to the console buffer in the event the error persists for the entire duration of the poll. I went ahead and added it since it's not in the normal path we expect.

pkg/kubefed2/disable.go
font: I see, makes sense. Updated.

font:
@marun I've rebased and updated based on your comments. PTAL.

pkg/kubefed2/disable.go
marun: (No action required) Hmm, maybe we need a wait loop that does exponential backoff on error?

marun:
@font LGTM.   

@irfanurrehman  Over to you for a final review.

irfanurrehman:
@font the CI failure looks like a flake but seems legit. Not sure if increasing the timeout would help anything. 10 seconds seems to be a good enough time for the  controller to be stopped.

font:
Thanks @irfanurrehman! I've addressed the e2e issue and this is now ready for your re-review.

irfanurrehman:
@font thanks for doing this.
/lgtm

PR #771, opened 2019-04-17
support optional values in FederationConfig 

### LINKED ISSUE #770: support optional values in FederationConfig
support optional filed in `FederationConfig` yaml.

Currently, default values of `FederationConfig` is not supported in yaml file. 
If a field in yaml file is not specified, the default value in yaml but not in controller manager is used. E.g. `cluster-monitor-period` is not specified, it is not `40s` as default but `0s` as default value.

**Why is this needed**:
better usability

/kind feature

xunpan:
Use default values in controller manager if the `FederationConfig` yaml file not specified the field.

As mentioned in #770, the configuration value used by controller will be updated to `FederationConfig`

Fixes #770

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -24,14 +24,17 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/apiserver/pkg/util/logs"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/kubernetes-sigs/federation-v2/cmd/controller-manager/app/leaderelection"
@@ -190,9 +193,12 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
 		}
 
 		err := client.Get(context.Background(), fedConfig, namespace, name)
-		if err != nil {
+		if apierrors.IsNotFound(err) {
 			glog.Infof("Cannot retrieve FederationConfig %q: %v. Default options are used.", qualifiedName.String(), err)
 			return nil
+		}
+		if err != nil {
+			glog.Fatalf("Error retrieving FederationConfig %q: %v.", qualifiedName.String(), err)
 		}
 
 		glog.Infof("Setting Options with FederationConfig %q", qualifiedName.String())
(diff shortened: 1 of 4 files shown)


marun:
@xunpan Please rebase.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -239,6 +317,8 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	opts.ClusterHealthCheckConfig.FailureThreshold = spec.ClusterHealthCheck.FailureThreshold
 	opts.ClusterHealthCheckConfig.SuccessThreshold = spec.ClusterHealthCheck.SuccessThreshold
 
+	updateFederationConfig(opts.Config.KubeConfig, fedConfig)
```
marun: As proposed, `FederationConfig` will only be written the API if it already existed in the API or was provided by command line.  It should probably always be written to ensure that `kubefed2` can rely on it existing.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -212,16 +215,91 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
...
+}
+
+func getInt(specValue, defaultValue int) int {
+	if specValue == 0 {
+		return defaultValue
+	}
+	return specValue
+}
+
+func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
```
marun: Setting defaults via the federation config likely removes the need to set defaults [here](https://github.com/kubernetes-sigs/federation-v2/blob/master/cmd/controller-manager/app/options/options.go#L40).

cmd/controller-manager/app/controller-manager.go
xunpan: Yes. No matter specified by command line `--federation-config` or exist `FederationConfig`, we will create/update `FederationConfig` for kubefed2 use.

cmd/controller-manager/app/controller-manager.go
xunpan: Yes. It does not need anymore.

xunpan: yes. Addressed.

xunpan:
@marun 
For `--federation-namespace`, did you mean it does not need to be changed? Your comments just information for references?

marun:
@xunpan Not saying you need to address it in this PR, but I think it would be desirable for a follow-on that would set `federation-namespace` via the config mechanism if `scope == Namespace`.

cmd/controller-manager/app/controller-manager.go
marun: It appears that if `getFederationConfig` returns nil because `FederationConfig` is not provided by flag or does not exist in the API, this function will exit on line 303 without setting defaults and without writing them back to the API.

xunpan:
@marun got it. I will log a bug issue to trace it.

cmd/controller-manager/app/controller-manager.go
xunpan: I assumed that configuration is the hard dependecy for federation before. Now I realize it does not with your comments. 

If there is no configuration at all, we can make all by defaults and create `FederationConfig`.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -190,10 +193,13 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
 		}
 
 		err := client.Get(context.Background(), fedConfig, namespace, name)
-		if err != nil {
+		if err != nil && !apierrors.IsNotFound(err) {
```
marun: This statement will be true for any error that isn't `IsNotFound`.  Likely `IsNotFound` should trigger this code path instead.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -212,16 +218,98 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
...
+			glog.Fatalf("Error updating FederationConfig %q: %v", qualifiedName.String(), err)
+		}
+	}
+}
+
 func setOptionsByFederationConfig(opts *options.Options) {
 	fedConfig := getFederationConfig(opts)
 	if fedConfig == nil {
-		return
+		// Both `--federation-config` flag and `FederationConfig` resource are not used
```
marun: -> `FederationConfig` could not be sourced from `--federation-config` or from the API.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -212,16 +218,98 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
...
+	qualifiedName := util.QualifiedName{
+		Namespace: namespace,
+		Name:      name,
+	}
+
+	configResource := &corev1a1.FederationConfig{}
+	client := genericclient.NewForConfigOrDieWithUserAgent(config, "federationconfig")
+	err := client.Get(context.Background(), configResource, namespace, name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		glog.Fatalf("Error retrieving FederationConfig %q: %v", qualifiedName.String(), err)
```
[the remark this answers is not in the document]
xunpan: Yes. It is redundant. I just clean it.

cmd/controller-manager/app/controller-manager.go
xunpan: Obvious logic error. Thanks!

cmd/controller-manager/app/controller-manager.go
```diff
@@ -212,16 +215,91 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
...
+	}
+	return specValue
+}
+
+func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
+	spec := &fedConfig.Spec
+	spec.RegistryNamespace = getString(spec.RegistryNamespace, util.MulticlusterPublicNamespace)
+
+	duration := &spec.ControllerDuration
+	duration.AvailableDelay.Duration = getDuration(duration.AvailableDelay, util.DefaultClusterAvailableDelay)
```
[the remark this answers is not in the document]
xunpan: Yes. It is simpler. 

I did not recheck after I changed the related code. The function is used for different variables in checking and assignement when it is introduced in my old code. It is not good. After I update related code to make it simpler. The function itself is not changed.

addressed

cmd/controller-manager/app/controller-manager.go
```diff
@@ -190,10 +193,13 @@ func getFederationConfig(opts *options.Options) *corev1a1.FederationConfig {
 		}
 
 		err := client.Get(context.Background(), fedConfig, namespace, name)
-		if err != nil {
+		if err != nil && apierrors.IsNotFound(err) {
```
marun: `err != nil` appears unnecessary here.

xunpan: OK.

I use it because other places use it in similar way. 
However, I checked related code, and test locally. if `err` is `nil`, `apierrors.IsNotFound(err)` returns false.

marun:
Thank you!

/lgtm

PR #772, opened 2019-04-17
user federation-controller sa instead of the default one

### LINKED ISSUE #265: Federation control plane should not use default service account
As per a [comment](https://github.com/kubernetes-sigs/federation-v2/pull/230#discussion_r217504775) from @font, the federation control plane should use a dedicated service account rather than the default service account for the federation namespace.  Use of the default service account dates back to fedv1 when the  federation control plane api and the host cluster api were separate and federated workloads were unlikely to run in the federation namespace.

xunpan:
fixes #265


marun:
Thank you!

/lgtm
/approve

PR #775, opened 2019-04-18
get option from FederationConfig instead of specified in command line

### LINKED ISSUE #748: kubefed2 should configure itself with FederationConfig
With the impending requirement that the federation controller manager configure itself with a `FederationConfig` resource, `kubefed2` should be updated to configure itself where possible from the same resource.  At a minimum, this should remove the need for the following arguments:

 - `--registry-namespace`
 - `--limited-scope`

xunpan:
`kubefed2` should configure itself with `FederationConfig`
Currently, `limited-scope` and `registry-namespace` can be get from `FederationConfig`

Fixes #748

### DIFF
--- a/pkg/kubefed2/options/options.go
+++ b/pkg/kubefed2/options/options.go
@@ -17,10 +17,15 @@ limitations under the License.
 package options
 
 import (
+	"context"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 
+	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
+	genericclient "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+	"k8s.io/client-go/rest"
 )
 
 // GlobalSubcommandOptions holds the configuration required by the subcommands of
@@ -37,7 +42,7 @@ func (o *GlobalSubcommandOptions) GlobalSubcommandBind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
 	flags.StringVar(&o.HostClusterContext, "host-cluster-context", "", "Host cluster context")
 	flags.StringVar(&o.FederationNamespace, "federation-namespace", util.DefaultFederationSystemNamespace,
-		"Namespace in the host cluster where the federation system components are installed.  This namespace will also be the target of propagation if the controller manager is configured with --limited-scope and clusters are joined with --limited-scope.")
+		"Namespace in the host cluster where the federation system components are installed. This namespace will also be the target of propagation if the controller manager is running with namespaced scope.")
 	flags.BoolVar(&o.DryRun, "dry-run", false,
 		"Run the command in dry-run mode, without making any server requests.")
 }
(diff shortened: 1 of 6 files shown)


marun:
@xunpan Should this be dependent on #771?

xunpan:
@marun, this is independent from #771. 
Without #771, all items in `FederationConfig` is set by helm or example/config yaml file. So, `limited-scope` and `registry-namespace` should exist in `FederationConfig`.

pkg/kubefed2/options/options.go
```diff
@@ -37,26 +43,23 @@ func (o *GlobalSubcommandOptions) GlobalSubcommandBind(flags *pflag.FlagSet) {
 	flags.StringVar(&o.Kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
 	flags.StringVar(&o.HostClusterContext, "host-cluster-context", "", "Host cluster context")
 	flags.StringVar(&o.FederationNamespace, "federation-namespace", util.DefaultFederationSystemNamespace,
-		"Namespace in the host cluster where the federation system components are installed.  This namespace will also be the target of propagation if the controller manager is configured with --limited-scope and clusters are joined with --limited-scope.")
+		"Namespace in the host cluster where the federation system components are installed.  This namespace will also be the target of propagation if the controller manager is configured with limitedScope of FederationConfig.")
```
marun: `s/is configured with.*/is running with namespaced scope/`

pkg/kubefed2/options/options.go
```diff
@@ -71,3 +74,31 @@ func (o *CommonSubcommandOptions) SetName(args []string) error {
 	o.ClusterName = args[0]
 	return nil
 }
+
+type FederationConfigOptions struct {
+	ClusterNamespace string
+}
+
+func GetOptionsByFederationConfig(hostConfig *rest.Config, namespace string) (limitedScope bool, clusterNamespace string, err error) {
```
marun: `s/GetOptionsByFederationConfig/GetOptionsFromFederationConfig/`

pkg/kubefed2/options/options.go
```diff
@@ -71,3 +74,31 @@ func (o *CommonSubcommandOptions) SetName(args []string) error {
...
+	}
+
+	fedConfig := &fedv1a1.FederationConfig{}
+	err = client.Get(context.TODO(), fedConfig, namespace, util.FederationConfigName)
+	if err != nil {
+		confName := util.QualifiedName{
+			Namespace: namespace,
+			Name:      util.FederationConfigName,
+		}
+		errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", confName)
```
marun: The result of this statement is unused.

pkg/kubefed2/options/options.go
marun: Also, please prefer returning variables explicitly setting values defined in the signature.  As per the logical error noted below with `errors.Wrapf`, there is less room for error when returning explicitly.

xunpan: @marun, what did you mean about `explicitly setting values defined in the signature`? I set each returning varible explicitely.

pkg/kubefed2/options/options.go
xunpan: removed

pkg/kubefed2/options/options.go
marun: My intent was not to have you remove the error wrapping:

```golang
// This is what appeared, and the result is thrown away:
errors.Wrapf(err, "...")
// This is what was likely intended
err = errors.Wrapf(err, "...")
```

Notice, however, that almost no code in this repo uses 'named return values' as this function does.  For consistency, please rewrite this function to use regular unnamed return values.

xunpan: Got it. Addressed.

pkg/kubefed2/options/options.go
```diff
@@ -71,3 +74,28 @@ func (o *CommonSubcommandOptions) SetName(args []string) error {
...
 }
+
+type FederationConfigOptions struct {
+	ClusterNamespace string
+}
+
+func GetOptionsFromFederationConfig(hostConfig *rest.Config, namespace string) (bool, string, error) {
+	client, err := genericclient.New(hostConfig)
+	if err != nil {
+		glog.V(2).Infof("Failed to get federation clientset: %v", err)
```
marun: Should the error be wrapped with this message and returned, rather than logged?  Otherwise the error will be logged redundantly (and with no context).

xunpan: done.

marun:
Thank you!

/lgtm

PR #777, opened 2019-04-18
feat: refactor dnsendpoint common file

draveness:


xunpan:
@draveness 
Thanks for you PR.

My qustions are
 - what is the convention for your refector? I check the code, if you want to shorten `Implementation`, we use `Impl` in many place but no `Imp` found.
 - what is the reason to remove `Default` from your consideration?

draveness:
> @draveness
> Thanks for you PR.
> 
> My qustions are
> 
> * what is the convention for your refector? I check the code, if you want to shorten `Implementation`, we use `Impl` in many place but no `Imp` found.
> * what is the reason to remove `Default` from your consideration?

Sorry I didn't check the code in other files. This is the first file I checked out in this project. And my editor prompts with many lint err, ex: function exported should have comment start with the function name.

I think `defaultNetWrapper` or `defaultNetWrapperImpl` maybe a better name for this struct. And I don't see quite many `Impl` in go code. It's something like the style of Java code. But it's not a big deal I could change it back to `NetWrapperDefaultImplementation`. 

Despite this, the PRs improves the code in the ways below:

+ fix the wrong approach of documenting the exported method.
+ remove `init` function when seems not necessary for now.

Anyway, I respect your options and ideas on this, and what do you think? :)

xunpan:
@draveness 
I appreciate your work on this. There is no functional change. So I just try to understand related thoughts of your change.
- removing `init` makes code concise, I think it makes sense.
- though there has no convention, I think use same code name style is better in a project. I thing `defaultNetWrapperImpl` is fine
- I don't know this part. @marun what do you think?
> fix the wrong approach of documenting the exported method.

draveness:
Just updated the commits with `defaultNetWrapperImpl`.

There is an official blog [Godoc: documenting Go code](https://blog.golang.org/godoc-documenting-go-code) which introduces the approach to documenting go code. Below is the golint error which prompts by my editor.

image

Kubernetes main project also follows this pattern to generate docs. Here is an example [config.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/config.go#L39-L50) I picked from the project.

marun:
@draveness I'm not clear on the value of maintaining code manually to a standard enforced by the golint configuration of your editor.  Unless golint is used to automatically validate changes in our CI, maintaining the proposed change will require manual effort on the part of reviewers.

draveness:
> @draveness I'm not clear on the value of maintaining code manually to a standard enforced by the golint configuration of your editor. Unless golint is used to automatically validate changes in our CI, maintaining the proposed change will require manual effort on the part of reviewers.

Don't get me wrong :). It's not about my editor. It's about the way to document the code, and since we have a lot of comments in this project, why don't we follow the go official recommendation provided by the official? 

In my opinion, we may need some sort of CI lint check, which would help us to identify the defects and problems in advance. There are many choices for them.

+ Kubernetes use golint as the static check approach, which gives all the warnings, ex: documenting, etc.
+ [golangci-lint](https://github.com/golangci/golangci-lint) is also a good choice, which provides configuration.

If we decide to use this, I'm happy to open a PR to add golint or golangci-lint to travis CI and fix the static analysis error in this project. However, it's just a style improvement, and the decision of whether to take it is all yours. If you think this isn't useful, I will roll back the part of comment change in this PR. And thanks for your comments and reviews.

> And besides, Is there anything else I could do to contribute to this project? I checked many good first issues and found out all of them were actually fixed in some PRs. Do you think to fix TODOs is a good way to start? 😆

shashidharatd:
@draveness, please go ahead and add a lint check to CI. `golangci-lint` is probably a good choice and thanks for taking it up.

marun:
@draveness Can this be closed in favor of #790?

draveness:
@marun the init function is not necessary, maybe we could remove all the unnecessary init funciton in the project to make the code more concise.

marun:
@draveness That sounds like a reasonable change.  I'm not sure I'm clear on when `init` not preferable to static initialization, though.  Would you care to elaborate?

draveness:
> @draveness That sounds like a reasonable change. I'm not sure I'm clear on when `init` not preferable to static initialization, though. Would you care to elaborate?

Yeah, I think it is a good approach to avoid the `init` function since it implicit runs in each pkg which could be hard to control especially for the developers who use the package.

> [The init function is called after all the variable declarations in the package have evaluated their initialisers.](https://golang.org/doc/effective_go.html#init)

**If we could do the initialise without `init` function, we are supposed to use the static initialisation instead of `init` for the concise of the code and to avoid the `init`.**

But we do have the time which `init` function is required when the initialisation process is quite complicated, ex: `if` or switch` statement involved, etc. The examples below are directly from [Effective Go](https://golang.org/doc/effective_go.html#init) which should be a good example of how to initialise variables in packages:



```go
ar user = os.Getenv("USER")

func init() {
    if user == "" {
        panic("no value for $USER")
    }
}
```

```go
func init() {
    if user == "" {
        log.Fatal("$USER not set")
    }
    if home == "" {
        home = "/home/" + user
    }
    if gopath == "" {
        gopath = home + "/go"
    }
    // gopath may be overridden by --gopath flag on command line.
    flag.StringVar(&gopath, "gopath", gopath, "override default GOPATH")
}
```

marun:
@draveness Thank you for explaining.  If there are uses of `init` in this repo that you think can be simplified, please do so.

PR #779, opened 2019-04-18
Add imagePullPolicy to helm chart controllermanager deployment template

font:
Add missing `imagePullPolicy` option to the helm chart's controllermanager deployment template and add support for it in the deploy federation script. If you want to change the policy to `Always` for example, just set `IMAGE_PULL_POLICY=Always` when invoking the `deploy-federation.sh` script.

This also adds a `.helmignore` file to avoid helm parsing errors while editing the chart.


xunpan:
LGTM

PR #780, opened 2019-04-19
[Makefile] fix `make` compilation error on macOS

iamzhout:
`make` will result into below below error without fix on macOS

```
cp -f bin/hyperfed-linux images/federation-v2/hyperfed
cp: bin/hyperfed-linux: No such file or directory
make: *** [container] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ vet:
 fmt:
 	$(shell ./hack/update-gofmt.sh)
 
-container: hyperfed
+container: $(HYPERFED_TARGET)-linux
 	cp -f $(HYPERFED_TARGET)-linux images/federation-v2/hyperfed
 	$(DOCKER) build images/federation-v2 -t $(IMAGE_NAME)
 	rm -f images/federation-v2/hyperfed


Makefile
```diff
@@ -77,7 +77,7 @@ fmt:
 	$(shell ./hack/update-gofmt.sh)
 
 container: hyperfed
-	cp -f $(HYPERFED_TARGET)-linux images/federation-v2/hyperfed
+	cp -f $(HYPERFED_TARGET)-$(HOST_OS) images/federation-v2/hyperfed
```
irfanurrehman: @iamzhout thanks for your PR, but I think this change is not desirable as the container is based on linux. Please have a look at the Dockerfile [here](https://github.com/kubernetes-sigs/federation-v2/blob/master/images/federation-v2/Dockerfile)

iamzhout: I get your point which I just missed and submitted previous quick fix.

I have updated the fix and verified locally on my macOS laptop, please check if this new fix make sense.

Thanks!

irfanurrehman: @iamzhout thanks again. @yamt can you please verify the same, given you did the macos relevant changes. The fix seems right to me, as the container needs to always have the linux bin only.

yamt: yes, it should always be a linux binary.

PR #784, opened 2019-04-21
WIP: Ensure deletion of uncached managed resources

marun:
**NOTE:** This PR is not ready for review until #724 merges.

Since a federated resource may be deleted before all the resources it manages appear in the informer cache, it is necessary to manually check each member cluster that does not have a cached resource before removing the finalizer.

This PR addresses a [TODO item in the old deletion helper](https://github.com/kubernetes-sigs/federation-v2/blob/v0.0.8/pkg/controller/util/deletionhelper/deletion_helper.go#L125).


marun:
Rebased

marun:
This has been folded into #724

PR #785, opened 2019-04-21
Refactor updater for extensibility

marun:
The new updater was written to support immediate dispatch (vs precomputing all operations) to simplify adoption of unmanaged resources.  This commit rewrites the updater to be more extensible to support the future replacement of the deletion handler.  As part of the rewrite, the dispatcher and retain logic have been moved to the `dispatch` package to ensure the sync controller only relies on exported interfaces.

7th in a series targeting #612


marun:
Rebased

marun:
Rebased and ready for review.

shashidharatd:
I could not find any flaws. the changes LGTM.

pkg/controller/sync/dispatch/managed.go
```diff
@@ -0,0 +1,225 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+)
+
+// DispatchResource is the subset of the FederatedResource interface
+// required for dispatching operations to managed resources.
+type DispatchResource interface {
```
irfanurrehman: [Suggestion] What say about naming this `ResourceToDispatch`. I had to wrap my head around some portion of the rest of the code to understand that `DispatchResource` is not an action (literally speaking).

irfanurrehman: Also if this is a subset of FederatedResource, then why not embed this in FederatedResource rather then duplicating the method definitions.

pkg/controller/sync/resource.go
```diff
@@ -244,17 +242,13 @@ func (r *federatedResource) RecordEvent(reason, messageFmt string, args ...inter
...
-}
-
 func (r *federatedResource) overridesForCluster(clusterName string) (util.ClusterOverridesMap, error) {
 	r.Lock()
 	defer r.Unlock()
 	if r.overridesMap == nil {
 		overridesMap, err := util.GetOverrides(r.federatedResource)
 		if err != nil {
-			return nil, errors.Errorf("Error reading cluster overrides for %s %q", r.federatedKind, r.federatedName)
+			return nil, errors.Wrapf(err, "Error reading cluster overrides")
```
irfanurrehman: This change seems not relevant to this PR. Is this because the Updater (now dispatcher) deciphers and appends this information from the resource?

irfanurrehman:
@marun few nits, otherwise lgtm.

pkg/controller/sync/resource.go
marun: This was a latent bug discovered in the process of implementing this PR.  Since this error is only ever logged as an event on the resource, there is zero value in including the kind and name.

pkg/controller/sync/dispatch/managed.go
marun: Done.

marun:
Responded to comments.

marun:
The latest push ensures that `ManagedDispatcher.Create` returns `StatusError` on creation if the error that doesn't indicate that the resource already exists. Previously the error was being logged but then execution continued.

pkg/controller/sync/resource.go
irfanurrehman: Ack.

irfanurrehman:
Thanks @marun.
/lgtm

PR #787, opened 2019-04-22
Remove host-only docker daemon config file check

font:
This reverts a check for the docker daemon config file that was performed only on the host. Currently this breaks local testing as well as CI whenever the host's docker daemon config file already exists since this code is called from 2 different contexts: the host and the cluster. We'll need to address it differently per https://github.com/kubernetes-sigs/federation-v2/pull/764#discussion_r276029510.


marun:
Merging to fix master. 

/lgtm

PR #789, opened 2019-04-22
Factor unmanaged dispatcher out of the managed dispatcher 

marun:
This will enable reuse for deletion handling.  The primary difference between the 2 contexts (managed vs unmanaged) is that events cannot be recorded against a federated resource if it is marked for deletion, so logging and error handling need to be used instead.

8th in a series targeting #612

### DIFF
--- a/pkg/controller/sync/dispatch/managed.go
+++ b/pkg/controller/sync/dispatch/managed.go
@@ -45,11 +45,9 @@ type FederatedResourceForDispatch interface {
 // ManagedDispatcher dispatches operations to member clusters for resources
 // managed by a federated resource.
 type ManagedDispatcher interface {
-	OperationDispatcher
+	UnmanagedDispatcher
 
 	Create(clusterName string)
-	Delete(clusterName string)
-	RemoveManagedLabel(clusterName string, clusterObj *unstructured.Unstructured)
 	Update(clusterName string, clusterObj *unstructured.Unstructured)
 	VersionMap() map[string]string
 }
@@ -57,9 +55,10 @@ type ManagedDispatcher interface {
 type managedDispatcherImpl struct {
 	sync.RWMutex
 
-	dispatcher  *operationDispatcherImpl
-	fedResource FederatedResourceForDispatch
-	versionMap  map[string]string
+	dispatcher          *operationDispatcherImpl
+	unmanagedDispatcher *unmanagedDispatcherImpl
+	fedResource         FederatedResourceForDispatch
+	versionMap          map[string]string
 }
 
 func NewManagedDispatcher(clientAccessor clientAccessorFunc, fedResource FederatedResourceForDispatch) ManagedDispatcher {
@@ -68,6 +67,7 @@ func NewManagedDispatcher(clientAccessor clientAccessorFunc, fedResource Federat
 		versionMap:  make(map[string]string),
 	}
 	d.dispatcher = newOperationDispatcher(clientAccessor, d)
+	d.unmanagedDispatcher = newUnmanagedDispatcher(d.dispatcher, d, fedResource.TargetKind(), fedResource.TargetName())
 	return d
 }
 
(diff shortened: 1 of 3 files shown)


marun:
Rebased and ready for review.

pkg/controller/sync/dispatch/unmanaged.go
```diff
@@ -0,0 +1,129 @@
...
+				d.recorder.RecordError(clusterName, op, err)
+			}
+			return util.StatusError
+		}
+		return util.StatusAllOK
+	})
+}
+
+func (d *unmanagedDispatcherImpl) wrapOperationError(err error, clusterName, operation string) error {
+	return errors.Wrapf(err, "Failed to %s %s %q in cluster %q", operation, d.targetKind, d.targetName, clusterName)
```
font: You can use `eventTemplate` here.

pkg/controller/sync/dispatch/unmanaged.go
```diff
@@ -0,0 +1,129 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/runtime"
+
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+)
+
+const eventTemplate = "%s %s %q in cluster %q"
```
font: Consider also using this in `managed.go`.

pkg/controller/sync/dispatch/managed.go
```diff
@@ -45,21 +45,20 @@ type FederatedResourceForDispatch interface {
 // ManagedDispatcher dispatches operations to member clusters for resources
 // managed by a federated resource.
 type ManagedDispatcher interface {
-	OperationDispatcher
+	UnmanagedDispatcher
```
font: There's probably a good reason to introduce this indirection, but it's difficult to contextualize the need for it based solely on this review.

pkg/controller/sync/dispatch/unmanaged.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		// Avoid mutating the resource in the informer cache
+		updateObj := clusterObj.DeepCopy()
+
+		util.RemoveManagedLabel(updateObj)
+
+		_, err := client.Resources(updateObj.GetNamespace()).Update(updateObj, metav1.UpdateOptions{})
+		if err != nil {
+			if d.recorder == nil {
```
font: Consider adding a helper function for these if/else checks.

marun: I’m explicitly avoiding the use of a helper because it would change the location of the thrown error.

pkg/controller/sync/dispatch/managed.go
marun: If you need context, please look at subsequent PRs for #612

pkg/controller/sync/dispatch/managed.go
```diff
@@ -68,6 +67,7 @@ func NewManagedDispatcher(clientAccessor clientAccessorFunc, fedResource Federat
 		versionMap:  make(map[string]string),
 	}
 	d.dispatcher = newOperationDispatcher(clientAccessor, d)
+	d.unmanagedDispatcher = newUnmanagedDispatcher(d.dispatcher, d, fedResource.TargetKind(), fedResource.TargetName())
```
shashidharatd: I think you should be using `NewUnmanagedDispatcher` instead of `newUnmanagedDispatcher`, otherwise the recorder is non-nil and deviates from the main intention of the pr.

marun: @shashidharatd This is intentional.  In the managed case (i.e. when using the managed dispatcher), there is a federated resource to log events against even for deletion and unlabeling operations.   That's why the unmanaged dispatcher decides how to log events/errors depending on whether a recorder is available.  In the follow-on PR to refactor deletion handling, the sync controller uses `NewUnmanagedDispatcher` (i.e. without a recorder) to handle deletion and unlabeling when a federated resource is not capable of logging events.

pkg/controller/sync/dispatch/unmanaged.go
marun: Done

pkg/controller/sync/dispatch/unmanaged.go
marun: Done

marun:
Updated, PTAL!

pkg/controller/sync/dispatch/managed.go
shashidharatd: Okie, SGTM

irfanurrehman:
/lgtm to unblock the next PR and be opened for review (given @shashidharatd would be asleep already :) )
thanks for doing this @marun.

PR #790, opened 2019-04-23
add golangci lint to travis

draveness:
Related to #777

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,12 @@
+issues:
+  exclude-rules:
+    - path: pkg/kubefed2/enable/enable.go
+      linters:
+        - unused
+        - structcheck
+    - path: pkg/kubefed2/join.go
+      linters:
+        - structcheck
+    - path: pkg/kubefed2/disable.go
+      linters:
+        - structcheck
--- a/pkg/controller/util/planner/planner.go
+++ b/pkg/controller/util/planner/planner.go
@@ -65,31 +65,45 @@ func NewPlanner(preferences *fedschedulingv1a1.ReplicaSchedulingPreference) *Pla
 // * a map that contains information how many extra replicas would be nice to schedule in a cluster so,
 //   if by chance, they are scheduled we will be closer to the desired replicas layout.
 func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[string]int64,
-	estimatedCapacity map[string]int64, replicaSetKey string) (map[string]int64, map[string]int64) {
+	estimatedCapacity map[string]int64, replicaSetKey string) (map[string]int64, map[string]int64, error) {
 
 	preferences := make([]*namedClusterPreferences, 0, len(availableClusters))
 	plan := make(map[string]int64, len(preferences))
 	overflow := make(map[string]int64, len(preferences))
 
-	named := func(name string, pref fedschedulingv1a1.ClusterPreferences) *namedClusterPreferences {
+	named := func(name string, pref fedschedulingv1a1.ClusterPreferences) (*namedClusterPreferences, error) {
 		// Seems to work better than addler for our case.
 		hasher := fnv.New32()
-		hasher.Write([]byte(name))
-		hasher.Write([]byte(replicaSetKey))
+		if _, err := hasher.Write([]byte(name)); err != nil {
+			return nil, err
+		}
+		if _, err := hasher.Write([]byte(replicaSetKey)); err != nil {
+			return nil, err
+		}
 
 		return &namedClusterPreferences{
 			clusterName:        name,
 			hash:               hasher.Sum32(),
 			ClusterPreferences: pref,
-		}
+		}, nil
 	}
 
 	for _, cluster := range availableClusters {
 		if localRSP, found := p.preferences.Spec.Clusters[cluster]; found {
-			preferences = append(preferences, named(cluster, localRSP))
+			preference, err := named(cluster, localRSP)
+			if err != nil {
+				return nil, nil, err
+			}
+
+			preferences = append(preferences, preference)
 		} else {
 			if localRSP, found := p.preferences.Spec.Clusters["*"]; found {
-				preferences = append(preferences, named(cluster, localRSP))
+				preference, err := named(cluster, localRSP)
+				if err != nil {
+					return nil, nil, err
+				}
+
+				preferences = append(preferences, preference)
 			} else {
 				plan[cluster] = int64(0)
 			}
@@ -216,7 +230,7 @@ func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[strin
 	}
 
 	if p.preferences.Spec.Rebalance {
-		return plan, overflow
+		return plan, overflow, nil
 	} else {
 		// If rebalance = false then overflow is trimmed at the level
 		// of replicas that it failed to place somewhere.
@@ -227,7 +241,7 @@ func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[strin
 				newOverflow[key] = value
 			}
 		}
-		return plan, newOverflow
+		return plan, newOverflow, nil
 	}
 }
 
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -17,6 +17,7 @@ limitations under the License.
 package features
 
 import (
+	"github.com/golang/glog"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
@@ -54,7 +55,9 @@ const (
 )
 
 func init() {
-	utilfeature.DefaultFeatureGate.Add(defaultFederationFeatureGates)
+	if err := utilfeature.DefaultFeatureGate.Add(defaultFederationFeatureGates); err != nil {
+		glog.Fatalf("Unexpected error: %v", err)
+	}
 }
 
 // defaultFederationFeatureGates consists of all known Federation-specific feature keys.
--- a/pkg/kubefed2/disable.go
+++ b/pkg/kubefed2/disable.go
@@ -159,8 +159,12 @@ func DisableFederation(cmdOut io.Writer, config *rest.Config, typeConfigName ctl
 	}
 
 	write := func(data string) {
-		if cmdOut != nil {
-			cmdOut.Write([]byte(data))
+		if cmdOut == nil {
+			return
+		}
+
+		if _, err := cmdOut.Write([]byte(data)); err != nil {
+			glog.Fatalf("Unexpected err: %v\n", err)
 		}
 	}
 
--- a/pkg/kubefed2/enable/enable.go
+++ b/pkg/kubefed2/enable/enable.go
@@ -229,7 +229,9 @@ func GetResources(config *rest.Config, enableTypeDirective *EnableTypeDirective)
 func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResources, namespace string) error {
 	write := func(data string) {
 		if cmdOut != nil {
-			cmdOut.Write([]byte(data))
+			if _, err := cmdOut.Write([]byte(data)); err != nil {
+				glog.Fatalf("Unexpected err: %v\n", err)
+			}
 		}
 	}
 
@@ -334,17 +336,18 @@ func GenerateTypeConfigForTarget(apiResource metav1.APIResource, enableTypeDirec
 }
 
 func federatedTypeCRD(typeConfig typeconfig.Interface, accessor schemaAccessor, shortNames []string) *apiextv1b1.CustomResourceDefinition {
-	var templateSchema map[string]apiextv1b1.JSONSchemaProps
-	templateSchema = accessor.templateSchema()
+	templateSchema := accessor.templateSchema()
 	schema := federatedTypeValidationSchema(templateSchema)
 	return CrdForAPIResource(typeConfig.GetFederatedType(), schema, shortNames)
 }
 
 func writeObjectsToYAML(objects []pkgruntime.Object, w io.Writer) error {
 	for _, obj := range objects {
-		w.Write([]byte("---\n"))
-		err := writeObjectToYAML(obj, w)
-		if err != nil {
+		if _, err := w.Write([]byte("---\n")); err != nil {
+			return errors.Wrap(err, "Error encoding object to yaml")
+		}
+
+		if err := writeObjectToYAML(obj, w); err != nil {
 			return errors.Wrap(err, "Error encoding object to yaml")
 		}
 	}
@@ -358,6 +361,9 @@ func writeObjectToYAML(obj pkgruntime.Object, w io.Writer) error {
 	}
 
 	unstructuredObj := &unstructured.Unstructured{}
-	unstructured.UnstructuredJSONScheme.Decode(json, nil, unstructuredObj)
+	if _, _, err := unstructured.UnstructuredJSONScheme.Decode(json, nil, unstructuredObj); err != nil {
+		return err
+	}
+
 	return util.WriteUnstructuredToYaml(unstructuredObj, w)
 }
--- a/pkg/kubefed2/unjoin.go
+++ b/pkg/kubefed2/unjoin.go
@@ -290,14 +290,12 @@ func deleteFederatedClusterAndSecret(hostClientset kubeclient.Interface, client 
 func deleteRBACResources(unjoiningClusterClientset kubeclient.Interface,
 	namespace, unjoiningClusterName, hostClusterName string, dryRun bool) bool {
 
-	var deletionSucceeded = true
-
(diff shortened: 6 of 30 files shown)


pkg/features/features.go
```diff
@@ -54,7 +54,10 @@ const (
 )
 
 func init() {
-	utilfeature.DefaultFeatureGate.Add(defaultFederationFeatureGates)
+	if err := utilfeature.DefaultFeatureGate.Add(defaultFederationFeatureGates); err != nil {
+		// TODO:
```
draveness: What should I do when `Add` return error

pkg/kubefed2/disable.go
```diff
@@ -65,6 +65,7 @@ type disableType struct {
 	disableTypeOptions
 }
 
+// nolint[:structcheck]
```
draveness: is this a good approach to disable false positive, another option is to use a `golangci.yml` file to disable the structcheck in the file and `enable.go[1]`

pkg/kubefed2/enable/enable.go
```diff
@@ -72,6 +72,7 @@ type enableType struct {
 	enableTypeOptions
 }
 
+// nolint[:structcheck]
```
draveness: [1]

pkg/kubefed2/disable.go
marun: What is golint complaining about in this instance?

pkg/kubefed2/enable/enable.go
```diff
@@ -227,10 +228,14 @@ func GetResources(config *rest.Config, enableTypeDirective *EnableTypeDirective)
 // already been enabled for federation.  This would likely involve
 // updating the version of the target type and the validation of the schema.
 func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResources, namespace string) error {
-	write := func(data string) {
+	write := func(data string) error {
```
marun: This is intended to be a braindead helper method.  It should not return an error and necessitate that every client checks for an error in turn.  Panicking should be sufficient.

pkg/kubefed2/enable/enable.go
```diff
@@ -334,17 +344,18 @@ func GenerateTypeConfigForTarget(apiResource metav1.APIResource, enableTypeDirec
...
 	return CrdForAPIResource(typeConfig.GetFederatedType(), schema, shortNames)
 }
 
 func writeObjectsToYAML(objects []pkgruntime.Object, w io.Writer) error {
 	for _, obj := range objects {
-		w.Write([]byte("---\n"))
-		err := writeObjectToYAML(obj, w)
-		if err != nil {
+		if _, err := w.Write([]byte("---\n")); err != nil {
+			return err
```
marun: The error should be wrapped with the same message as line 359.

test/e2e/crud.go
```diff
@@ -92,7 +92,12 @@ var _ = Describe("Federated", func() {
 					tl.Fatalf("Failed to create unlabeled resource in cluster %q: %v", clusterName, err)
 				}
 				clusterClient := genericclient.NewForConfigOrDie(clusterConfig)
-				defer clusterClient.Delete(context.TODO(), unlabeledObj, unlabeledObj.GetNamespace(), unlabeledObj.GetName())
+				defer func() {
+					err := clusterClient.Delete(context.TODO(), unlabeledObj, unlabeledObj.GetNamespace(), unlabeledObj.GetName())
+					if err != nil {
+						tl.Fatalf("Unexpected error: %v", err)
```
marun: Will `Fatalf` allow other deferred calls to execute?

test/e2e/crud.go
```diff
@@ -111,11 +116,16 @@ var _ = Describe("Federated", func() {
...
+					err := hostClient.Delete(context.TODO(), createdObj, createdObj.GetNamespace(), createdObj.GetName())
+					if err != nil {
+						tl.Fatalf("Unexpected error: %v", err)
+					}
+				}()
 
 				waitDuration := 10 * time.Second // Arbitrary amount of time to wait for deletion
 				By(fmt.Sprintf("Checking that the unlabeled resource is not deleted within %v", waitDuration))
-				wait.PollImmediate(framework.PollInterval, waitDuration, func() (bool, error) {
+				err = wait.PollImmediate(framework.PollInterval, waitDuration, func() (bool, error) {
```
marun: Ignoring this error was intentional, as timing out is success, and it should continue to be ignored (e.g. `_ = wait.PollImmediate`).

test/e2e/framework/unmanaged.go
```diff
@@ -213,8 +212,8 @@ func (f *UnmanagedFramework) ClusterDynamicClients(apiResource *metav1.APIResour
 		// Check if this cluster is the same name as the host cluster name to
 		// make it the primary cluster.
 		testClusters[clusterName] = common.TestCluster{
-			clusterConfig,
-			client,
+			TestClusterConfig: clusterConfig,
```
marun: What rule is this violating exactly?

pkg/controller/federatedcluster/clusterclient.go
```diff
@@ -125,21 +125,24 @@ func (self *ClusterClient) GetClusterZones() (zone, region string, err error) {
...
-			region, err = getRegionNameForNode(node)
-			if err != nil {
-				return "", "", err
-			}
-		}
-		// TODO: Optimize this flow. All nodes will have the same zone label.
-		// So just considering first node for now.
-		break
+
+	if len(nodes.Items) <= 0 {
```
marun: This code is being rewritten in #797 - please revert.

pkg/features/features.go
marun: This is a pretty low value fix, I suggest leaving it for now.

pkg/kubefed2/unjoin.go
```diff
@@ -285,14 +285,12 @@ func deleteFederatedClusterAndSecret(hostClientset kubeclient.Interface, client
 func deleteRBACResources(unjoiningClusterClientset kubeclient.Interface,
 	namespace, unjoiningClusterName, hostClusterName string, dryRun bool) bool {
 
-	var deletionSucceeded = true
```
marun: What is golint complaining about in this instance?

test/e2e/framework/framework.go
```diff
@@ -306,7 +306,7 @@ func createNamespace(client kubeclientset.Interface, baseName string) (string, e
 	// TODO(marun) should all api calls be made 'robustly'?
 	var namespaceName string
 	if err := wait.PollImmediate(PollInterval, TestContext.SingleCallTimeout, func() (bool, error) {
-		namespace, err := client.Core().Namespaces().Create(namespaceObj)
+		namespace, err := client.CoreV1().Namespaces().Create(namespaceObj)
```
marun: I get that V1 should be preferred, but how does golint know that?

test/e2e/crud.go
marun: Note that this change is responsible for the CI failure.

pkg/kubefed2/disable.go
draveness: some variables in the struct were unused. Excluded with a golangci lint file

test/e2e/crud.go
draveness: Yes, I have tested it.

pkg/kubefed2/unjoin.go
draveness:
```
ineffectual assignment to `deletionSucceeded`
```

the statement to initialzie `deletionSucceeded` var is not useful, which would be overwritten after calling `deleteClusterRoleAndBinding`

test/e2e/framework/unmanaged.go
draveness: Use the style above to initialise structs would reduce the possibility of unexpected initialises error, suppose we have something like:

```go
type Test struct {
  a int
  b int
}
```

Initialise the struct above use `Test{1, 2}` may lead to problems when the order of vars changes in the `Test` struct.

And there are only two places which use this approach to initialise the struct. It's better to maintain consistency in the project.

test/e2e/framework/framework.go
draveness: The function gets a deprecated comment.

```go
	// Deprecated: please explicitly pick a version if possible.
	Core() corev1.CoreV1Interface
```

pkg/kubefed2/unjoin.go
marun: I would prefer that this check be disabled.  An 'ineffectual assignment' may well be useful for communicating intent to future maintainers even if it has no value to the golang runtime.

test/e2e/framework/unmanaged.go
marun: I get the point of the check, but I don't think it's worth enforcing at this time.  Please disable it for now.

pkg/kubefed2/disable.go
marun: I'm confused.  If they were truly unused, why not just remove them?  Or if they are used, and golint isn't smart enough to figure that out, I'm not sure the check is worth enabling.

pkg/features/features.go
marun: I realize I was confused about what you were asking.  I think `glog.Fatalf` is likely the right answer.

pkg/controller/util/planner/planner.go
```diff
@@ -71,25 +72,41 @@ func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[strin
...
-		}
+		}, nil
 	}
 
 	for _, cluster := range availableClusters {
 		if localRSP, found := p.preferences.Spec.Clusters[cluster]; found {
-			preferences = append(preferences, named(cluster, localRSP))
+			preference, err := named(cluster, localRSP)
+			if err != nil {
+				fmt.Printf("Unexpected err: %s\n", err.Error())
```
marun: `glog.Errorf` should be used instead of `Printf`

pkg/controller/util/planner/planner.go
```diff
@@ -71,25 +72,41 @@ func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[strin
...
+				continue
+			}
+
+			preferences = append(preferences, preference)
 		} else {
 			if localRSP, found := p.preferences.Spec.Clusters["*"]; found {
-				preferences = append(preferences, named(cluster, localRSP))
+				preference, err := named(cluster, localRSP)
+				if err != nil {
+					fmt.Printf("Unexpected err: %s\n", err.Error())
```
marun: ditto

pkg/kubefed2/disable.go
```diff
@@ -159,9 +159,11 @@ func DisableFederation(cmdOut io.Writer, config *rest.Config, typeConfigName ctl
 	}
 
 	write := func(data string) {
-		if cmdOut != nil {
-			cmdOut.Write([]byte(data))
+		if cmdOut == nil {
+			return
 		}
+
+		_, _ = cmdOut.Write([]byte(data))
```
marun: I apologize for not being clear.  The error should not be ignored.  It should be logged with `glog.Fatalf`.

pkg/kubefed2/enable/enable.go
```diff
@@ -229,7 +229,9 @@ func GetResources(config *rest.Config, enableTypeDirective *EnableTypeDirective)
 func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResources, namespace string) error {
 	write := func(data string) {
 		if cmdOut != nil {
-			cmdOut.Write([]byte(data))
+			if _, err := cmdOut.Write([]byte(data)); err != nil {
+				panic(err)
```
marun: Same comment as for disable -`glog.Fatalf`

pkg/kubefed2/enable/enable.go
```diff
@@ -255,6 +257,7 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
 		if err != nil {
 			return errors.Wrapf(err, "Error creating CRD %q", resources.CRD.Name)
 		}
+
```
marun: Please remove the inserted newlines - unrelated and nonfunctional formatting changes do not belong in this PR.

test/common/crudtester.go
```diff
@@ -195,7 +195,10 @@ func (c *FederatedTypeCrudTester) CheckUpdate(fedObject *unstructured.Unstructur
 			}
 			clusterOverrides[key] = value
 		}
-		util.SetOverrides(obj, overrides)
+
+		if err := util.SetOverrides(obj, overrides); err != nil {
+			c.tl.Fatalf("Unexpected error: %s", err.Error())
```
marun: Please prefer `"%v", err`

pkg/kubefed2/disable.go
draveness: > I have already disabled the check in this golangci file.

The struct and its vars are all private and not used directly with the form - `disableTypeOptions.xxx`, so the golint complains about that, though it was used in `disableType`. I think `disableTypeOptions` isn't useful and seems quite redundant, move all its vars and method into `disableType` still works the same. But we could keep that if you prefer that way.

pkg/kubefed2/enable/enable.go
draveness: removed

draveness:
Hi @marun, thanks for the review. And I just resolved all the conversations.

.golangci.yml
```diff
@@ -0,0 +1,9 @@
+issues:
+  exclude-rules:
+    - path: pkg/kubefed2/enable/enable.go
+      linters:
+        - unused
+        - structcheck
+    - path: pkg/kubefed2/disable.go
+      linters:
+        - structcheck
```
shashidharatd: (suggestion) I would wish there are no exceptions and all code follow similar pattern

draveness: I came up with commit to remove the exceptions in golangci.yml, maybe you could take a look. https://github.com/kubernetes-sigs/federation-v2/pull/790/commits/9634a69c28cafaef091df727297d9b3e3d2c68e3

pkg/kubefed2/disable.go
marun: `disableTypeOptions` is intended to isolate options that will be set via flags.  Where there are values that are supplied via flags (`targetName` in this case), I am supportive of them being added to the command type (i.e. `disableType`).  Would that satisfy the golint check?

pkg/controller/util/planner/planner.go
```diff
@@ -71,25 +72,41 @@ func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[strin
...
-		}
+		}, nil
 	}
 
 	for _, cluster := range availableClusters {
 		if localRSP, found := p.preferences.Spec.Clusters[cluster]; found {
-			preferences = append(preferences, named(cluster, localRSP))
+			preference, err := named(cluster, localRSP)
+			if err != nil {
+				glog.Errorf("Unexpected err: %v\n", err)
```
marun: Looking at this in a wider context this function should return the error rather than logging and continuing.  My apologies for not figuring this out sooner.

cc: @irfanurrehman

pkg/controller/util/planner/planner.go
```diff
@@ -71,25 +72,41 @@ func (p *Planner) Plan(availableClusters []string, currentReplicaCount map[strin
...
+				continue
+			}
+
+			preferences = append(preferences, preference)
 		} else {
 			if localRSP, found := p.preferences.Spec.Clusters["*"]; found {
-				preferences = append(preferences, named(cluster, localRSP))
+				preference, err := named(cluster, localRSP)
+				if err != nil {
+					glog.Errorf("Unexpected err: %v\n", err)
```
marun: ditto

pkg/kubefed2/disable.go
```diff
@@ -62,17 +62,13 @@ var (
 
 type disableType struct {
 	options.GlobalSubcommandOptions
-	disableTypeOptions
```
marun: As per my previous comment, I would prefer that the options set by flag remain in the type for consistency with the other kubefed commands.  We can revisit this for all commands in the future.

draveness: OK, I'll revert the latest commit

pkg/controller/util/planner/planner.go
draveness: Please take a look at the latest commit, it returns errors in both method - `plan` and `schedule`

marun:
@draveness lgtm.  Please squash to one commit and I'll approve.

draveness:
> @draveness lgtm. Please squash to one commit and I'll approve.

done

marun:
Nice work, thank you!

/approve
/lgtm

draveness:
Hi @xunpan @marun @shashidharatd, the previous lgtm label was removed because of code rebasing, PTAL. :)

PR #791, opened 2019-04-23
Ensure finalizer is not removed if uncached managed resources exist

marun:
Since a federated resource may be deleted before all the resources it manages appear in the informer cache, it is necessary to manually check each member cluster that does not have a cached resource before removing the finalizer.

This PR addresses a [TODO item in the old deletion helper](https://github.com/kubernetes-sigs/federation-v2/blob/v0.0.8/pkg/controller/util/deletionhelper/deletion_helper.go#L125). 

This is a repost of #784 that I unintentionally closed.

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -473,8 +473,45 @@ func (s *FederationSyncController) deleteFromClusters(fedResource FederatedResou
 		glog.V(2).Infof("Waiting for resources managed by %s %q to be removed from the following clusters: %s", fedKind, fedName, strings.Join(remainingClusters, ", "))
 		return true, nil
 	}
+	err = s.ensureRemovedOrUnmanaged(fedResource)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to verify that managed resources no longer exist in any cluster")
+	}
 	// Managed resources no longer exist in any member cluster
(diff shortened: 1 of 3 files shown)


marun:
Rebased and ready for review.

pkg/controller/sync/controller.go
```diff
@@ -473,10 +473,47 @@ func (s *FederationSyncController) deleteFromClusters(fedResource FederatedResou
...
+			continue
+		}
+		dispatcher.CheckRemovedOrUnlabeled(cluster.Name, fedResource.IsNamespaceInHostCluster)
+	}
+	ok, timeoutErr := dispatcher.Wait()
+	if timeoutErr != nil {
+		return timeoutErr
+	}
+	if len(unreadyClusters) > 0 {
+		return errors.Errorf("the following clusters were not ready: %s", strings.Join(unreadyClusters, ", "))
```
shashidharatd: (observation) Any faulty cluster in federation would block deleting the federated resource now. Nothing wrong with this and probably the right thing to do and faulty cluster situation is mostly temporary.

marun: The main deletion loop will similarly block removal of the finalizer if any cluster is not ready., so I don't think this PR represents a change in behavior.  More documentation, as always, is suggested.

Issue #678, opened 2019-03-21
Member cluster's nodes may deployed in muti availability zone


Dingshujie:
What would you like to be added:
FederatedClusterStatus.Zone use []string instead of string, and generate multiple AZ level dns name.

Why is this needed:
Now, FederatedCluster status Region and Zone use first node Region and Zone.
A lot of Cluster's nodes may deployed in muti availability zones, this field may in influence dns enpoint controller to generate dns record name,  dnsname of az-level dns name may be incorrect.

/kind feature

marun:
@shashidharatd What are your thoughts on this issue?

shashidharatd:
IMHO, the cross-cluster service discovery feature should be able to work with multi-zone clusters. logically, there should be no issues. But we have to try and test it. I would try to put up a PR in coming days.

PR #797, opened 2019-04-24
Extend service discovery for multi-zone clusters

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/federation-v2/issues/678

This PR adds additional zone level DNS record to take care of cross-cluster service discovery in case of multi-zone clusters

### DIFF
--- a/pkg/controller/dnsendpoint/service.go
+++ b/pkg/controller/dnsendpoint/service.go
@@ -58,56 +58,52 @@ func getServiceDNSEndpoints(obj interface{}) ([]*feddnsv1a1.Endpoint, error) {
 		commonPrefix = strings.Join([]string{dnsObject.Name, dnsObject.Namespace, dnsObject.Spec.DomainRef, "svc"}, ".")
 	}
 
+	ttl := dnsObject.Spec.RecordTTL
+	if ttl == 0 {
+		ttl = defaultDNSTTL
+	}
+
 	for _, clusterDNS := range dnsObject.Status.DNS {
-		zone := clusterDNS.Zone
-		region := clusterDNS.Region
+		var zoneDNSName string
+		regionDNSName := strings.Join([]string{commonPrefix, clusterDNS.Region, dnsObject.Status.Domain}, ".") // region level, one up from zone level
+		globalDNSName := strings.Join([]string{commonPrefix, dnsObject.Status.Domain}, ".")                    // global level, one up from region level
 
-		dnsNames := []string{
-			strings.Join([]string{commonPrefix, zone, region, dnsObject.Status.Domain}, "."), // zone level
-			strings.Join([]string{commonPrefix, region, dnsObject.Status.Domain}, "."),       // region level, one up from zone level
-			strings.Join([]string{commonPrefix, dnsObject.Status.Domain}, "."),               // global level, one up from region level
-			"", // nowhere to go up from global level
-		}
-
-		var zoneTargets, regionTargets, globalTargets feddnsv1a1.Targets
-		for _, clusterDNS := range dnsObject.Status.DNS {
-			if clusterDNS.Zone == zone {
-				zoneTargets = append(zoneTargets, ExtractLoadBalancerTargets(clusterDNS.LoadBalancer)...)
-			}
-		}
-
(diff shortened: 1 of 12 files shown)


pkg/controller/federatedcluster/controller.go
```diff
@@ -233,28 +233,26 @@ func thresholdAdjustedClusterStatus(clusterStatus *fedv1a1.FederatedClusterStatu
...
 	clusterClient *ClusterClient) *fedv1a1.FederatedClusterStatus {
 
 	if !util.IsClusterReady(clusterStatus) {
 		return clusterStatus
 	}
 
-	zone, region, err := clusterClient.GetClusterZones()
+	zones, region, err := clusterClient.GetClusterZones()
 	if err != nil {
 		glog.Warningf("Failed to get zones and region for cluster %q: %v", clusterClient.clusterName, err)
```
marun: (No action required) Consider returning after this warning to remove the need for protecting the non-error behavior with the else clause.

test/e2e/servicedns.go
```diff
@@ -215,3 +214,21 @@ func createClusterServiceAndEndpoints(f framework.FederationFramework, name, nam
 
 	return serviceDNSStatus
 }
+
+func ensureClustersHasRegionZoneAttributes(client genericclient.Client, federatedClusters *fedv1a1.FederatedClusterList) map[string]fedv1a1.FederatedClusterStatus {
```
marun: nit: `ensureClustersHaveRegionAndZoneAttributes`

pkg/controller/federatedcluster/clusterclient.go
```diff
@@ -119,48 +120,44 @@ func (self *ClusterClient) GetClusterHealthStatus() *fedv1a1.FederatedClusterSta
...
 // Find the name of the zone in which a Node is running.
-func getZoneNameForNode(node corev1.Node) (string, error) {
+func getZoneNameForNode(node corev1.Node) string {
 	for key, value := range node.Labels {
 		if key == LabelZoneFailureDomain {
-			return value, nil
+			return value
 		}
 	}
-	return "", errors.Errorf("Zone name for node %s not found. No label with key %s",
```
marun: (No action required) Why is an error no longer desirable to raise an error if a zone is not found?

pkg/controller/dnsendpoint/service.go
```diff
@@ -58,56 +58,52 @@ func getServiceDNSEndpoints(obj interface{}) ([]*feddnsv1a1.Endpoint, error) {
...
-			ttl = defaultDNSTTL
-		}
-		for i, target := range targets {
-			endpoint, err := generateEndpointForServiceDNSObject(dnsNames[i], target, dnsNames[i+1], ttl, labels)
-			if err != nil {
-				return nil, err
-			}
-			endpoints = append(endpoints, endpoint)
-		}
-		if dnsObject.Spec.DNSPrefix != "" {
```
marun: (No action required) Oops!  I guess this wasn't a problem because of the dedup?

test/e2e/servicedns.go
shashidharatd: Updated.

pkg/controller/federatedcluster/controller.go
shashidharatd: Done.

pkg/controller/federatedcluster/clusterclient.go
shashidharatd: The error was not much useful here as there could be env where the said lables were not available. I have added reporting similar error in `ServiceDNS` controller.

pkg/controller/dnsendpoint/service.go
shashidharatd: yes, this was masked before because of dedup.

marun:
@shashidharatd If there is no doc change implied by this PR, that suggests the existing documentation for the feature could use some work.

shashidharatd:
@shashidharatd If there is no doc change implied by this PR, that suggests the existing documentation for the feature could use some work.

Agree, that the existing docs around the feature needs some more love. Would work on that in near future and thanks for the lgtm.

PR #801, opened 2019-04-26
Removal of the FederatedTypeConfig for namespaces disables all namespaced sync controllers

### LINKED ISSUE #722: Removal of the FederatedTypeConfig for namespaces should disable all namespaced sync controllers
Sync controllers for namespaced types currently require that an FTC for namespaces exist as a condition of starting, but if the FTC is removed running sync controllers for namespaced types will continue running.  The FTC controller should detect the removal of the namespace FTC and stop all namespaced sync controllers.

As part of fixing this bug, the error messages logged by the FTC controller when the namespace FTC is missing should clearly indicate the nature of the problem and how to fix it. 

/kind bug

font:
~~**NOTE**: This PR is not reviewable until #821 merges.~~

This change modifies the `FederatedTypeConfig` controller to start/stop all enabled namespaced sync controllers if the namespace `FederatedTypeConfig` resource is created/deleted. It will also fail to start any namespaced sync controller if the namespaced `FederatedTypeConfig` resource does not exist.

This also reverts a change to add and remove a finalizer for the FTC, requiring some E2E test changes, and discovering a couple race conditions along the way.

As a future follow-up soon, I will refactor the FTC controller, particularly the `reconcile` method.

Fixes #722

Edited:
I think we should add E2E tests for the FTC controller in the near future.

### DIFF
--- a/pkg/controller/federatedtypeconfig/controller.go
+++ b/pkg/controller/federatedtypeconfig/controller.go
@@ -194,6 +194,11 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 			c.stopController(statusKey, statusStopChan)
 		}
 
+		if typeConfig.GetTarget().Kind == util.NamespaceKind {
+			glog.Infof("Reconciling all namespaced FederatedTypeConfig resources on deletion of %q", key)
+			c.reconcileOnNamespaceFTCDelete()
+		}
+
 		err := c.removeFinalizer(typeConfig)
 		if err != nil {
 			runtime.HandleError(errors.Wrapf(err, "Failed to remove finalizer from FederatedTypeConfig %q", key))
@@ -209,7 +214,7 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	}
 
 	startNewSyncController := !syncRunning && syncEnabled
-	stopSyncController := syncRunning && !syncEnabled
+	stopSyncController := syncRunning && (!syncEnabled || (typeConfig.GetNamespaced() && !c.namespaceFTCExists()))
 	if startNewSyncController {
 		if err := c.startSyncController(typeConfig); err != nil {
 			runtime.HandleError(err)
@@ -360,6 +365,11 @@ func (c *Controller) removeFinalizer(tc *corev1a1.FederatedTypeConfig) error {
 	return err
 }
 
+func (c *Controller) namespaceFTCExists() bool {
+	_, err := c.getFederatedNamespaceAPIResource()
+	return err == nil
+}
+
 func (c *Controller) getFederatedNamespaceAPIResource() (*metav1.APIResource, error) {
 	// TODO(marun) Document the requirement to restart the controller
 	// manager if the federated namespace resource changes.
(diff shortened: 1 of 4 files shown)


font:
Rebased and ready for review.

pkg/controller/util/typeconfig.go
```diff
@@ -0,0 +1,26 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/typeconfig"
+
+func IsNamespaceKind(t typeconfig.Interface) bool {
```
marun: Should this method be added to the interface?

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -289,11 +291,11 @@ func (c *Controller) startSyncController(tc *corev1a1.FederatedTypeConfig) error
 	// cluster-scoped federation control plane.  A namespace-scoped
 	// control plane would still have to use a non-shared informer due
 	// to it not being possible to limit its scope.
+	kind := tc.Spec.FederatedType.Kind
 	fedNamespaceAPIResource, err := c.getFederatedNamespaceAPIResource()
 	if err != nil {
-		return err
+		return errors.WithMessagef(err, "Unable to start sync controller for %q due to missing FederatedTypeConfig for namespaces", kind)
```
marun: Why is `WithMessagef` preferable in this instance to `Wrapf`?

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -202,14 +201,17 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 		return util.StatusAllOK
 	}
 
-	err = c.ensureFinalizer(typeConfig)
+	updated, err := c.ensureFinalizer(typeConfig)
 	if err != nil {
 		runtime.HandleError(errors.Wrapf(err, "Failed to ensure finalizer for FederatedTypeConfig %q", key))
 		return util.StatusError
+	} else if updated && util.IsNamespaceKind(typeConfig) {
```
marun: I'm assuming you are detecting the addition of the namespace ftc by noting when the finalizer is added.  I encourage you to document your intent so that future maintainers don't have to wonder about the following question:

Why is the successful addition of the finalizer to the namespace ftc a reason to reconcile namespaced FTCs?  Sync controller start is predicated on being about to source the namespace ftc from the cache, and does not depend on that ftc having a finalizer.

pkg/controller/util/typeconfig.go
font: I originally thought and wanted to do this, but I don't want to rely on the `util.NamespaceKind` constant within the `v1alpha1` package nor hard code the string `Namespace` again. But I can add it there instead if you'd prefer, and either move the common util constant there (many things depend on it) or just hard-code it again.

pkg/controller/federatedtypeconfig/controller.go
font: When I looked at the `Wrapf` API call, it seemed to want to dump a stack and felt that was unnecessary. But it looks like that's only with the `%+v` format specifier. I've fixed.

pkg/controller/federatedtypeconfig/controller.go
font: Done.

pkg/controller/util/typeconfig.go
marun: +1 to adding it to the interface (and an implementation to `FederatedTypeConfig`).  There is already an implementation of checking that the target is a namespace that could be refactored into the implementation:

https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/apis/core/v1alpha1/federatedtypeconfig_types.go#L215

marun:
I think some squashing is suggested before this merges.

font:
@marun Thanks for your review! Updated and rebased. PTAL.

marun:
LGTM.   Would you please add an issue to track implementation of a test for this functionality?

font:
> LGTM. Would you please add an issue to track implementation of a test for this functionality?

@marun you bet. I've created #837.

marun:
Test flake due to #839, I've restarted it.

irfanurrehman:
Thanks for doing this @font.
/lgtm

PR #802, opened 2019-04-29
if FTC is not same version/group, reject request

### LINKED ISSUE #743: Attempting to enable federation of an already-enabled type should fail fast
Attempting to enable federation of an already-enabled type with a non-default group will create the crd for the federated type and then fail in attempting to create the `FederatedTypeConfig` (since it already exists).   `kubefed2 enable` should check for the presence of the FTC for the type and fail immediately if it already exists, since a type can be enabled at most once for a control plane.

/kind bug

xunpan:
Fixes #743 

if FTC exists, ensure that the group and version of the crd that will be generated match what exists (so it can update instead of creating a new one)

If group and version don't match - error.

### DIFF
--- a/pkg/kubefedctl/enable/enable.go
+++ b/pkg/kubefedctl/enable/enable.go
@@ -244,6 +244,29 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
 		return errors.Wrapf(err, "Error attempting to determine whether federation system namespace %q exists", namespace)
 	}
 
+	client, err := genericclient.New(config)
+	if err != nil {
+		return errors.Wrap(err, "Failed to get federation clientset")
+	}
+
+	concreteTypeConfig := resources.TypeConfig.(*fedv1a1.FederatedTypeConfig)
+	existingTypeConfig := &fedv1a1.FederatedTypeConfig{}
+	err = client.Get(context.TODO(), existingTypeConfig, namespace, concreteTypeConfig.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", concreteTypeConfig.Name)
+	}
+	if err == nil {
+		fedType := existingTypeConfig.GetFederatedType()
+		target := existingTypeConfig.GetTarget()
+		concreteType := concreteTypeConfig.GetFederatedType()
+		if fedType.Name != concreteType.Name || fedType.Version != concreteType.Version || fedType.Group != concreteType.Group {
+			return errors.Errorf("Federation is already enabled for %q with federated type %q. Changing the federated type to %q is not supported.",
+				qualifiedAPIResourceName(target),
+				qualifiedAPIResourceName(fedType),
+				qualifiedAPIResourceName(concreteType))
+		}
+	}
+
 	crdClient, err := apiextv1b1client.NewForConfig(config)
 	if err != nil {
 		return errors.Wrap(err, "Failed to create crd clientset")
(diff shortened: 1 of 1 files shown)


marun:
Note that the group/version here is for the federated type CRD.  As per #804, changing the group/version of the target should be supported as a way of 'upgrading' the API that is targeted in member clusters.

xunpan:
This is really a code error. I fixed it.

pkg/kubefed2/enable/enable.go
```diff
@@ -244,6 +244,26 @@ func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResou
...
+	existingTypeConfig := &fedv1a1.FederatedTypeConfig{}
+	err = client.Get(context.TODO(), existingTypeConfig, namespace, concreteTypeConfig.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "Error retrieving FederatedTypeConfig %q", concreteTypeConfig.Name)
+	}
+	if err == nil {
+		fedType := existingTypeConfig.GetFederatedType()
+		concreteType := concreteTypeConfig.GetFederatedType()
+		if fedType.Version != concreteType.Version || fedType.Group != concreteType.Group {
+			return errors.Errorf("Existing FederatedTypeConfig %q has different group version name %s/%s. FederatedTypeConfig is not allowed to change the group/version name",
```
marun: -> `Federation is already enabled for "<group qualified type name>/<version>" with federated type "<group qualified name of federated type>/<version>".  Changing the federated type to "<new group>/<new version>" is not supported.`

e.g. `Federation is already enabled for "deployments.apps" with federated type "federateddeployments.types.federation.k8s.io/v1alpha1".  Changing the federated type to  "federateddeployments.mytypes.example.com/v1" is not supported.`

It think this suggests adding a check on the plural name of the configured federated type as well (e.g. `fedType.Name != concreteType.Name`). Even though we don't support setting the name via `kubefed2`, it is possible for a user to manually configure a federated type and type config and then subsequently attempt to run `kubefed enable`.

xunpan: @marun
I use `GroupVersionResource` to display the resource full name. I think it may be used for other places and more common for such output display. 
```
$ bin/kubefed2 enable --federation-group example.com deployments.apps
F0503 07:42:34.127072    9056 enable.go:115] Error: Federation is already enabled for "apps/v1, Resource=deployments" with federated type "types.federation.k8s.io/v1alpha1, Resource=federateddeployments". Changing the federated type to "example.com/v1alpha1, Resource=federateddeployments" is not supported.
```

marun: Regardless of use in other instances, federation surfaces group qualification more often than most kube operations, and we already use group-qualified names pervasively in enable.  I'm not sure I see a good reason for diverging from that approach when we need to include version (especially given that `apiVersion` appends it in the same way).  Which do you think is more readable and consistent?

`federateddeployments.types.federation.k8s.io/v1alpha1`

or

`"types.federation.k8s.io/v1alpha1, Resource=federateddeployments"`

xunpan: @marun I did not see any output UI for this format but internal key `func resourceKey`.  I'm ok if we use this as federation convention to display.

My question is do we need to follow `func resourceKey` for core resources. E.g. for `services`
 - `func resourceKey()`
    It displays `services.core/v1`
 - `GroupVersionKind.String()`
    It displays `/v1, Resource=services`

I think it is better to display necessary information in UI
  `services/v1`

marun: @xunpan Anything/anyone interacting with kube APIs must assume that an empty group means `core`, so `services/v1` and `services.core/v1` are effectively equivalent.  Note that `kubectl api-resources` omit the core group, and so does federation in its naming of FTC.

xunpan: @marun the code was done for `services/v1`. Please help to review.

pkg/kubefedctl/enable/enable.go
```diff
@@ -333,6 +349,13 @@ func GenerateTypeConfigForTarget(apiResource metav1.APIResource, enableTypeDirec
 	return typeConfig
 }
 
+func printNameResourceVersion(resource metav1.APIResource) string {
```
marun: `s/printNameResourceVersion/qualifiedAPIResourceName/`

`print` is a verb implying that this function will take an action.  A noun (e.g. name) indicating the output is more appropriate.

xunpan: That sounds better. Fixed.

marun:
Thank you!

/approve
/lgtm

Issue #402, opened 2018-11-07
Enable LeaderElect for federation controller 


fisherxu:
Now the federation-controller is deployed through statefulset and without LeaderElect only can in single instance. But in the following scenarios, we need to deploy the federation-controller in multiple instances for high availability with LeaderElect:

1. Need to deploy multiple instances in several nodes, in case of node crash then the slave controller can acquire leadership and serve soon.
2. If the leader controller lost connection to the kube-apiserver but still running, the healthy slave controller can acquire leadership and serve soon.

And if using multiple instances with LeaderElect, I think deploy the controller through deployment is fine.
BTW, the LeaderElect code is taken from kubernetes, and haven't made any changes yet :)

PTAL and request for your comments
/cc @font @irfanurrehman @marun @pmorie @shashidharatd

marun:
I'm not sure why leader election is worth pursuing in advance of users validating the general direction of federation.   The complexity is real, the benefits are not (at least not yet).  I'm also concerned that the challenges of high availability federation are likely to be related to the involvement of multiple clusters, and that leader election as proposed is really only intended to achieve HA within a single cluster.

Even a federation control plane that stays up in the host cluster due to a mechanism like leader election can be a single point of failure when connectivity between the host cluster and one or more member clusters is interrupted for whatever reason.  The best guess at ensuring federation control plane reliability is 'pull reconciliation', in which the federation control plane writes configuration to a distributed backing store shared with the member clusters, and those clusters use local agents to apply the configuration.

So my cone

fisherxu:
> that leader election as proposed is really only intended to achieve HA within a single cluster.

and yes, this is target to the HA within a single cluster, and it's just a base framwork, user can also set it to false so disable it and benefits is to users.

And to the member clusters health checking, that's should be another feature. They are all necessary and important.

marun:
@fisherxu I hear that adding leader election is desirable for your use case, and thank you for putting up with my uncertaintly.  I think there need to be 2 paths:
 
  - a statefulsets-based option that doesn't enable leader election
  - a deployment-based option that does enable leader election

The helm installer would ideally allow the user to select either option.  I'm not sure the deployment script should be deploying leader election, but unless @gyliu513 is making progress adding ci validation of the helm chart there may be no choice.

fisherxu:
@marun Thanks for reminder, and there is a enableLeaderElection option in the PR, so users can disable it.

marun:
[...]

PR #803, opened 2019-04-29
Add leader election documentation to user guide

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/federation-v2/issues/402

/cc @marun


docs/userguide.md
```diff
@@ -712,3 +713,15 @@ A and B are offline:
 ```
 Replica layout: C=20
 ```
+
+## Controller-Manager Leader Election
+
+The federation controller manager is always deployed with leader election feature
```
marun: We deploy with leader election by default, but replica count is [defaulted to 1](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/values.yaml#L9).  Should this be updated to 2 by default?  Otherwise I'm not sure this doc is very accurate since we might as well not be enabling by default.

shashidharatd: yes, i think we should update default replica count to 2, as that's how one is supposed to deploy in production. I would add a commit to update default replica count.

docs/userguide.md
```diff
@@ -712,3 +713,15 @@ A and B are offline:
...
+to ensure high availability of the control plane. Leader election module ensures
+there is always a leader elected among multiple instances which takes care of
+running the controllers. In case an active instance goes down, one of the standby
```
marun: s/an/the/

docs/userguide.md
```diff
@@ -712,3 +713,15 @@ A and B are offline:
...
+gets elected as leader, ensuring minimum downtime. Leader election is necessary so that
+only one instance is responsible for reconciliation. You can refer to
+[helm chart configuration](https://github.com/kubernetes-sigs/federation-v2/tree/master/charts/federation-v2#configuration)
+to configure parameters for leader election module to tune to your environment
```
marun: s/module to tune to/to tune for/
s/environment although the defaults are sane/environment (the defaults should be...)/

shashidharatd:
@marun, have incorporated your comments. PTAL.

PR #806, opened 2019-04-29
FederatedCluster should use strongly-typed reference to 'Cluster'

### LINKED ISSUE #759: FederatedCluster should use strongly-typed reference to 'Cluster'
### Background

Kube API convention discourages the use of general object references that are currently being used in `FederatedCluster` `ClusterRef` via `LocalObjectReference`. The reason is that the use of strongly-typed references provides the following advantages:

- it enables you to have specific doc for your type
- prevents additional fields being added by other stakeholders that may not make sense for your API
- allows you to add new fields if the need arises

### Proposal

`FederatedCluster` should use a strongly-typed reference to a 'Cluster' instead of using `LocalObjectReference` by creating a new type such as `LocalClusterReference`. An example would look like the following:

```go
	// Name of the cluster registry Cluster resource from which to source api
	// endpoints.
	ClusterRef LocalClusterReference `json:"clusterRef,omitempty"`

	// Other fields omitted
}

// Doc comment to explain what type of reference this is.
type LocalClusterReference struct {
    Name string `json:"name,omitempty"`
}

```

font:
Fixes #759 

See issue for details.


xunpan:
LGTM

PR #807, opened 2019-04-29
FederatedCluster should use strongly-typed reference to 'Secret'

### LINKED ISSUE #760: FederatedCluster should use strongly-typed reference to 'Secret'
### Background

Kube API convention discourages the use of general object references that are currently being used in `FederatedCluster` `SecretRef` via `LocalObjectReference`. The reason is that the use of strongly-typed references provides the following advantages:

- it enables you to have specific doc for your type
- prevents additional fields being added by other stakeholders that may not make sense for your API
- allows you to add new fields if the need arises

### Proposal
`FederatedCluster` should use a strongly-typed reference to a 'Secret' instead of using `LocalObjectReference` by creating a new type such as `LocalSecretReference`. An example would look like the following:

```go
	// Other fields omitted

	// This can be left empty if the cluster allows insecure access.
	// +optional
	SecretRef *LocalSecretReference `json:"secretRef,omitempty"`
}

// Doc comment to explain what type of reference this is and to explain the
// API semantics of the secret content that is referenced.
type LocalSecretReference struct {
    Name string `json:"name,omitempty"`
}
```

Additionally, the contents of the referenced secret are themselves an API. So we need to clearly define what that API is in the doc comment of `LocalSecretReference`.

sohankunkerkar:
Fixes #760 

See the issue for details.


pkg/apis/core/v1alpha1/federatedcluster_types.go
```diff
@@ -43,7 +43,13 @@ type FederatedClusterSpec struct {
 	//
 	// This can be left empty if the cluster allows insecure access.
 	// +optional
-	SecretRef *apiv1.LocalObjectReference `json:"secretRef,omitempty"`
+	SecretRef *LocalSecretReference `json:"secretRef,omitempty"`
+}
+
+// LocalSecretReference contains information to reference secret in the
+// cluster registry.
```
xunpan: It seems not correct in comments. It is used to refer a `secret` instead of `registry`

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -269,7 +268,7 @@ func (in *FederatedClusterSpec) DeepCopyInto(out *FederatedClusterSpec) {
 	out.ClusterRef = in.ClusterRef
 	if in.SecretRef != nil {
 		in, out := &in.SecretRef, &out.SecretRef
-		*out = new(v1.LocalObjectReference)
+		*out = new(LocalSecretReference)
 		**out = **in
```
xunpan: I assume this is generated code.

sohankunkerkar: @xunpan it's not generated code. I'd manually changed this to make it work.

pkg/apis/core/v1alpha1/federatedcluster_types.go
sohankunkerkar: @xunpan my bad. Let me change `registry` to `secret`

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
font: @sohankunkerkar This should be generated code. Let's not manually change it.

sohankunkerkar: ack

pkg/apis/core/v1alpha1/federatedcluster_types.go
```diff
@@ -43,7 +43,13 @@ type FederatedClusterSpec struct {
...
+type LocalSecretReference struct {
+	Name string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
sohankunkerkar: done!

pkg/apis/core/v1alpha1/federatedcluster_types.go
```diff
@@ -43,7 +43,13 @@ type FederatedClusterSpec struct {
 	//
 	// This can be left empty if the cluster allows insecure access.
 	// +optional
-	SecretRef *apiv1.LocalObjectReference `json:"secretRef,omitempty"`
+	SecretRef *LocalSecretReference `json:"secretRef,omitempty"`
+}
+
+// LocalSecretReference contains information to reference secret in the
```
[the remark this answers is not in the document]
sohankunkerkar: ack

sohankunkerkar:
@pmorie As suggested, I have updated the code to address your comments.

marun:
@sohankunkerkar Please rebase

PR #814, opened 2019-04-30
namespaced leader elector RBAC for cluster scope federation

### LINKED ISSUE #687: Ensure that the federated control plane is minimally privileged
#670 revealed that the helm chart was not granting sufficient privileges to the control plane.  While it is tempting to just use the cluster-admin role, it should be possible to grant only the privileges required.  For some types, the scope of privilege will depend on how the control plane is deployed (namespaced or cluster-scoped):

 - `get/watch/list/update`
   - `(multiclusterdns|scheduling|types).federation.k8s.io` 
      - namespaced: federation namespace
      - cluster-scoped: *
   - `FederatedTypeConfig` 
       - federation namespace
   - `FederatedCluster` 
       - federation namespace
   - `PropagatedVersion` 
       - namespaced: federation namespace
       - cluster-scoped: *
   - `ClusterPropagatedVersion` 
       - cluster-scoped: *
 - `get/list/watch`
   - `Namespace`
      - cluster-scoped: *
 - `get`
   - `Cluster` 
       - cluster registry namespace
   - `Secret` (for cluster auth)
       - federation namespace

For member clusters, the permissions required by the local service account would depend on the enabled types:

 - `get/watch/list/update`
   - any type intended to be managed by federation
      - namespaced: federation namespace
      - cluster-scoped: *

If the goal was avoiding the use of the cluster-admin role in member clusters, it would be necessary to coordinate the rbac permissions of member cluster service accounts with the types enabled for federation.

cc: @xunpan @pmorie

xunpan:
A new role for leader elector is created in federation controller namespace for bindings.
`secret` and `configmap` has minimum privilege

Fixes #687

### DIFF
--- a/charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
+++ b/charts/federation-v2/charts/controllermanager/templates/clusterrole.yaml
@@ -64,20 +64,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - configmaps
-  verbs:
-  - get
-  - create
-  - update
-- apiGroups:
-  - ""
-  resources:
-  - secrets
-  verbs:
-  - get
-- apiGroups:
-  - ""
-  resources:
   - events
   verbs:
   - get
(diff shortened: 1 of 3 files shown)


xunpan:
@marun the name is better. Fixed.

marun:
Nice, we're now minimally privileged.  Thank you!

/approve
/lgtm

PR #817, opened 2019-04-30
Document how to configure helm rbac

marun:
I attempted to follow the helm install guide but against a kind cluster helm lacked the permissions necessary to deploy the chart. The deploy script was working, which suggested the need to update the install guide to include creation of the privileged service account as per the script.


shashidharatd:
LGTM

PR #818, opened 2019-04-30
Update CI to test with Kubernetes 1.14.0

marun:


scripts/create-clusters.sh
```diff
@@ -107,10 +112,12 @@ function create-clusters() {
     unset KUBECONFIG
   fi
 
-  # TODO(font): Configure insecure registry on kind host cluster. Remove once
-  # https://github.com/kubernetes-sigs/kind/issues/110 is resolved.
-  echo "Configuring insecure container registry on kind host cluster"
-  configure-insecure-registry-on-cluster 1
+  if [[ "${CONFIGURE_INSECURE_REGISTRY}" ]]; then
```
font: I'm not sure I understand the need for this change. Since this flag is intended to control configuring the host for an insecure registry, it needs to be separate from whether or not we need to do it in the cluster. The default has always been to explicitly configure it in the cluster since it's always bootstrapped anew and doesn't hurt even if the user won't use it.

scripts/pre-commit.sh
```diff
@@ -132,7 +132,7 @@ echo "Downloading e2e test dependencies"
 ./scripts/download-e2e-binaries.sh
 
 CREATE_INSECURE_REGISTRY=y CONFIGURE_INSECURE_REGISTRY=y OVERWRITE_KUBECONFIG=y \
-    ./scripts/create-clusters.sh
+    KIND_TAG="v1.14.0" ./scripts/create-clusters.sh
```
font: (no action required) Looks like `v1.14.1` is available.

scripts/create-clusters.sh
marun: I'm leery of having to run code if I don't need to, and I've stopped using a local registry after the recent  breakage.  I'll remove it from this PR and add a flag that supports disabling this behavior explicitly.

scripts/pre-commit.sh
marun: Nice!  It wasn't there when I looked earlier today.

marun:
The only change now is to switch to using Kube 1.14.1.

font:
@marun Is the `v1.14.1` image broken?

marun:
> @marun Is the `v1.14.1` image broken?

Apparently.  Let's update to 1.14 for now.  We'll need to update to kind 0.3 soon enough anyway and can revisit the kube version then.

PR #819, opened 2019-04-30
Set default values for FederationConfig

### LINKED ISSUE #816: Creating default FederationConfig fails
**What happened**:

TLDR: the controller does not correctly create a default `FederationConfig` if one does not exist.

We ran the federation controller thusly:

```
$ oc logs -n federation-test federation-controller-manager-7677c44864-44d52
Federation v2 controller-manager version: version.Info{Version:"v0.0.9", GitCommit:"442e8868c6bd02bb61871af02269cb18a58804d4", GitTreeState:"clean", BuildDate:"2019-04-30T05:36:16Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
I0430 21:27:02.844168       1 controller-manager.go:345] FLAG: --alsologtostderr="false"
I0430 21:27:02.844218       1 controller-manager.go:345] FLAG: --federation-config=""
I0430 21:27:02.844226       1 controller-manager.go:345] FLAG: --federation-namespace="federation-test"
I0430 21:27:02.844233       1 controller-manager.go:345] FLAG: --help="false"
I0430 21:27:02.844241       1 controller-manager.go:345] FLAG: --kubeconfig=""
I0430 21:27:02.844247       1 controller-manager.go:345] FLAG: --log-backtrace-at=":0"
I0430 21:27:02.844255       1 controller-manager.go:345] FLAG: --log-dir=""
I0430 21:27:02.844261       1 controller-manager.go:345] FLAG: --log-flush-frequency="5s"
I0430 21:27:02.844270       1 controller-manager.go:345] FLAG: --log_backtrace_at=":0"
I0430 21:27:02.844276       1 controller-manager.go:345] FLAG: --log_dir=""
I0430 21:27:02.844281       1 controller-manager.go:345] FLAG: --logtostderr="true"
I0430 21:27:02.844286       1 controller-manager.go:345] FLAG: --master=""
I0430 21:27:02.844292       1 controller-manager.go:345] FLAG: --stderrthreshold="2"
I0430 21:27:02.844297       1 controller-manager.go:345] FLAG: --v="5"
I0430 21:27:02.844302       1 controller-manager.go:345] FLAG: --version="false"
I0430 21:27:02.844307       1 controller-manager.go:345] FLAG: --vmodule=""
W0430 21:27:02.844424       1 client_config.go:549] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.                                                            
I0430 21:27:03.242267       1 controller-manager.go:198] Cannot retrieve FederationConfig "federation-test/federation-v2": federationconfigs.core.federation.k8s.io "federation-v2" not found. Default options are
used.
F0430 21:27:03.479683       1 controller-manager.go:284] Error retrieving FederationConfig "": resource name may not be empty                              
```

Notice that in the last line of the log, the error message states `resource name may not be empty` - this is because the flow in setOptionsByFederationConfig does not give the default `FederationConfig` a name

The controller should create a `FederationConfig` named `federation-v2`.

Run the controller without creating a `FederationConfig`.

**Environment**:

- Federation version: 0.0.9

/kind bug

sohankunkerkar:
Fixes #816 

See the issue for details.

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -249,6 +249,20 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+
+	if len(spec.Scope) == 0 {
+		// TODO(sohankunkerkar) Remove when no longer necessary.
+		// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
+		// Its continued existence should not be relied upon.
+		const defaultScopeEnv = "DEFAULT_FEDERATION_SCOPE"
+		defaultScope := os.Getenv(defaultScopeEnv)
+		if len(defaultScope) != 0 {
+			if defaultScope != string(apiextv1b1.ClusterScoped) && defaultScope != string(apiextv1b1.NamespaceScoped) {
+				glog.Fatalf("%s must be Cluster or Namespaced; got %q", defaultScopeEnv, defaultScope)
+			}
+			spec.Scope = apiextv1b1.ResourceScope(defaultScope)
+		}
+	}
 	if spec.Scope == apiextv1b1.NamespaceScoped {
 		setString(&spec.RegistryNamespace, fedConfig.Namespace)
 	} else {
@@ -306,19 +320,19 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	fedConfig := getFederationConfig(opts)
 	if fedConfig == nil {
 		// FederationConfig could not be sourced from --federation-config or from the API.
-		// create a new `FederationConfig` with default values.
-		fedConfig = &corev1a1.FederationConfig{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      util.FederationConfigName,
-				Namespace: opts.Config.FederationNamespace,
-			},
+		qualifiedName := util.QualifiedName{
+			Namespace: opts.Config.FederationNamespace,
+			Name:      util.FederationConfigName,
 		}
 
-		qualifiedName := util.QualifiedName{
-			Namespace: fedConfig.Namespace,
-			Name:      fedConfig.Name,
-		}
 		glog.Infof("Creating FederationConfig %q with default values", qualifiedName)
+
+		fedConfig = &corev1a1.FederationConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      qualifiedName.Name,
+				Namespace: qualifiedName.Namespace,
+			},
+		}
 	}
 
 	setDefaultFederationConfig(fedConfig)
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -307,7 +307,18 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	if fedConfig == nil {
 		// FederationConfig could not be sourced from --federation-config or from the API.
 		// create a new `FederationConfig` with default values.
-		fedConfig = &corev1a1.FederationConfig{}
+		fedConfig = &corev1a1.FederationConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      util.FederationConfigName,
+				Namespace: opts.Config.FederationNamespace,
+			},
+		}
+
+		qualifiedName := util.QualifiedName{
+			Namespace: fedConfig.Namespace,
+			Name:      fedConfig.Name,
+		}
+		glog.Infof("Creating FederationConfig %q with default values", qualifiedName)
 	}
 
 	setDefaultFederationConfig(fedConfig)


marun:
LGTM

cmd/controller-manager/app/controller-manager.go
```diff
@@ -303,7 +303,12 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	if fedConfig == nil {
 		// FederationConfig could not be sourced from --federation-config or from the API.
 		// create a new `FederationConfig` with default values.
-		fedConfig = &corev1a1.FederationConfig{}
+		fedConfig = &corev1a1.FederationConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      util.FederationConfigName,
+				Namespace: util.DefaultFederationSystemNamespace,
```
pmorie: I don't think this is the namespace that we necessarily want; the logic for the namespace should be the same as: https://github.com/kubernetes-sigs/federation-v2/blob/master/cmd/controller-manager/app/controller-manager.go#L115-L121

marun: The config _needs_ to be in the federation namespace.  That's what makes it discoverable by `kubefed2`.

sohankunkerkar: As discussed, I have changed the logic here to compute the right value of the namespace based on how we're planning to deploy federation.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -301,9 +301,23 @@ func updateFederationConfig(config *rest.Config, fedConfig *corev1a1.FederationC
 func setOptionsByFederationConfig(opts *options.Options) {
 	fedConfig := getFederationConfig(opts)
 	if fedConfig == nil {
+		// Computing the right value of the namespace based on how
```
marun: This logic is duplicated from the `Run` method and serves no purpose here.  Please remove.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -301,9 +301,23 @@ func updateFederationConfig(config *rest.Config, fedConfig *corev1a1.FederationC
...
+			opts.Config.TargetNamespace = metav1.NamespaceAll
+			glog.Info("Federation will target all namespaces")
+		}
 		// FederationConfig could not be sourced from --federation-config or from the API.
 		// create a new `FederationConfig` with default values.
-		fedConfig = &corev1a1.FederationConfig{}
+		fedConfig = &corev1a1.FederationConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      util.FederationConfigName,
+				Namespace: opts.Config.TargetNamespace,
```
marun: The correct value for the namespace of the federation config is `opts.Config.FederationNamespace`.  This value is set via the `--federation-namespace` [argument](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/charts/controllermanager/templates/deployment.yaml#L26) which is intended to be [set via the downward api](https://github.com/kubernetes-sigs/federation-v2/blob/master/charts/federation-v2/charts/controllermanager/templates/deployment.yaml#L37) to the namespace the controller manager is running in.

sohankunkerkar: Thanks for this detailed explanation.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: done!

marun:
LGTM

cmd/controller-manager/app/controller-manager.go
```diff
@@ -303,7 +303,12 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	if fedConfig == nil {
 		// FederationConfig could not be sourced from --federation-config or from the API.
 		// create a new `FederationConfig` with default values.
-		fedConfig = &corev1a1.FederationConfig{}
+		fedConfig = &corev1a1.FederationConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      util.FederationConfigName,
+				Namespace: opts.Config.FederationNamespace,
```
pmorie: It seems like `opts.Config.FederationNamespace` will not be correctly set at this point, since it is set in the calling code after this method is called.

marun: `opts.Config.FederationNamespace` is set by flag, not code.

pmorie: OK

cmd/controller-manager/app/controller-manager.go
```diff
@@ -303,7 +303,12 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	if fedConfig == nil {
 		// FederationConfig could not be sourced from --federation-config or from the API.
 		// create a new `FederationConfig` with default values.
-		fedConfig = &corev1a1.FederationConfig{}
+		fedConfig = &corev1a1.FederationConfig{
```
pmorie: Please add a log message saying that the controller is creating a `FederationConfig` with default settings at this point

sohankunkerkar: Done!

sohankunkerkar:
@marun we have tested to establish that the controller is running without any issues and will be performing smoke testing.

sohankunkerkar:
@marun As discussed,I have made necessary changes in the code  I've not finished the testing yet but will let you know once it's done.

sohankunkerkar:
@marun we've performed the testing and it's ready for review.

marun:
LGTM

cmd/controller-manager/app/controller-manager.go
```diff
@@ -89,6 +89,7 @@ member clusters and does the necessary reconciliation`,
 	cmd.Flags().StringVar(&federationConfig, "federation-config", "", "Path to a federation config yaml file. Test only.")
 	cmd.Flags().StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	cmd.Flags().StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	cmd.Flags().StringVar(&defaultScope, "default-scope", "Cluster", "controls the default scope if the FederationConfig is not supplied.")
```
marun: - Please add a TODO comment above the flag definition as follows:

```golang
// TODO(sohankunkerkar) Remove when no longer necessary.  This flag is a temporary addition to support Red Hat's downstream testing efforts.  Its continued existence should not be relied upon.
```

- The first word in the help text should be capitalized, as per the examples of the help text of other flags.  Consistency is very important for collaborative development, and I encourage you to do your best to ensure that changes you make are in keeping with the example of related code.

- Please reuse constants when the exist.  In this case, `string(apiextv1b1.ClusterScoped)` is preferable to `"Cluster"`.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -100,6 +101,12 @@ func Run(opts *options.Options) error {
 
 	stopChan := setupSignalHandler()
 
+	if len(defaultScope) != 0 {
+		if defaultScope != "Cluster" && defaultScope != "Namespaced" {
```
marun: Ditto use of constants

cmd/controller-manager/app/controller-manager.go
```diff
@@ -245,6 +252,9 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+	if spec.Scope == "" {
```
marun: (No action required) It was beaten into me early in my time working on openshift by no less than smarterclayton to prefer a len check vs an equality check to determine if a string is empty (e.g. `len(spec.Scope) == 0`).  I don't think we've been particularly rigorous in this codebase, though, so I can't enforce it.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -303,7 +313,18 @@ func setOptionsByFederationConfig(opts *options.Options) {
...
+				Name:      util.FederationConfigName,
+				Namespace: opts.Config.FederationNamespace,
+			},
+		}
+
+		qualifiedName := util.QualifiedName{
+			Namespace: fedConfig.Namespace,
+			Name:      fedConfig.Name,
+		}
+		glog.Infof("Creating FederationConfig %q with default values", qualifiedName)
```
marun: In addition to providing helpful output at runtime, log statements can serve as informal comments when reading the code.  In the latter capacity they work best when they precede the action they are documenting, e.g.

```golang
// Comment precedes action

doSomething()
// It's time to do something

vs

// It's time to do something
doSomething()

// Logging precedes action

doSomething()
glog.Log("It's time to do something")

vs

glog.Log("It's time to do something")
doSomething()
```

cmd/controller-manager/app/controller-manager.go
```diff
@@ -100,6 +100,17 @@ func Run(opts *options.Options) error {
 
 	stopChan := setupSignalHandler()
 
+	// TODO(sohankunkerkar) Remove when no longer necessary.
```
marun: Please move lines 103-113 to the conditional block starting on line 261.  It should only be executed if the scope in the federation config is empty rather than on every invocation of the controller manager.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -245,6 +256,10 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+	defaultValue := os.Getenv("DEFAULT-SCOPE")
```
marun: This line duplicates 106, effectively ensuring that validation of an environment variable is separate from its use.  As per my comment above, I think it would be preferable to retrieve and validate the scope inside the conditional that requires setting the scope, so this line can simply be removed.  For future, consider that separating the validation and use of an environment variable risks the possibility of the env var changing in the meantime (env vars are mutable).

cmd/controller-manager/app/controller-manager.go
```diff
@@ -303,7 +318,19 @@ func setOptionsByFederationConfig(opts *options.Options) {
 	if fedConfig == nil {
 		// FederationConfig could not be sourced from --federation-config or from the API.
 		// create a new `FederationConfig` with default values.
```
marun: This line is now duplicated by the logging on L326, and can be removed.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -100,6 +100,17 @@ func Run(opts *options.Options) error {
 
 	stopChan := setupSignalHandler()
 
+	// TODO(sohankunkerkar) Remove when no longer necessary.
+	// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
+	// Its continued existence should not be relied upon.
+	defaultValue := os.Getenv("DEFAULT-SCOPE")
```
marun: Please name the env var `DEFAULT_FEDERATION_SCOPE`.  Dashes (`-`) should never be used in the name of an env var - only ascii characters, numbers and underscores (`_`).

cmd/controller-manager/app/controller-manager.go
```diff
@@ -249,6 +254,16 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+
+	if len(spec.Scope) == 0 {
+		defaultValue := os.Getenv(defaultScopeEnv)
+		if len(defaultValue) != 0 {
+			if defaultValue != string(apiextv1b1.ClusterScoped) && defaultValue != string(apiextv1b1.NamespaceScoped) {
+				glog.Fatalf("DefaultScope must be Cluster or Namespaced; got %q", defaultValue)
```
marun: Please include the name of the env var in this message, e.g.

glog.Fatalf("%s must be Cluster or Namespaced; got %q", defaultScopeEnv, defaultValue)

cmd/controller-manager/app/controller-manager.go
```diff
@@ -249,6 +254,16 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+
+	if len(spec.Scope) == 0 {
+		defaultValue := os.Getenv(defaultScopeEnv)
```
marun: s/defaultValue/defaultScope/

cmd/controller-manager/app/controller-manager.go
```diff
@@ -94,6 +94,11 @@ member clusters and does the necessary reconciliation`,
 	return cmd
 }
 
+// TODO(sohankunkerkar) Remove when no longer necessary.
+// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
+// Its continued existence should not be relied upon.
+const defaultScopeEnv = "DEFAULT_FEDERATION_SCOPE"
```
marun: Please define this in the conditional block starting on line 258.  It's not used outside of that block, so defining it up here serves no purpose.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -249,6 +254,16 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+
+	if len(spec.Scope) == 0 {
+		defaultValue := os.Getenv(defaultScopeEnv)
+		if len(defaultValue) != 0 {
+			if defaultValue != string(apiextv1b1.ClusterScoped) && defaultValue != string(apiextv1b1.NamespaceScoped) {
+				glog.Fatalf("DefaultScope must be Cluster or Namespaced; got %q", defaultValue)
+				spec.Scope = apiextv1b1.ResourceScope(defaultValue)
```
marun: This statement needs to appear outside of the conditional block that checks for an invalid scope.  As written, the default scope would never be set.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Done!

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Done!

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: yeah, makes sense. I will place that condition outside of that `if` condition where I'm checking for an invalid scope.

sohankunkerkar:
@marun ready for review

cmd/controller-manager/app/controller-manager.go
```diff
@@ -249,6 +249,19 @@ func setInt(target *int, defaultValue int) {
 
 func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	spec := &fedConfig.Spec
+	// TODO(sohankunkerkar) Remove when no longer necessary.
+	// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
+	// Its continued existence should not be relied upon.
+	const defaultScopeEnv = "DEFAULT_FEDERATION_SCOPE"
```
marun: Please include this inside the scope in which it is used - i.e. after the if conditional (see: limit the scope of declarations to where they are used).

sohankunkerkar: Changed as you suggested.

marun:
@irfanurrehman @shashidharatd @xunpan Please merge, whoever sees this first.  The setting of the default name fixes a bug that deploying with helm masked, and supporting setting the default scope is a temporary measure to support our downstream deployment method.  Both issues are blockers for our downstream testing efforts.

/approve

PR #821, opened 2019-05-01
Fix issues with FederatedTypeConfig controller

font:
This PR has the following changes:
- Reverts a change to add/remove finalizers when a `FederatedTypeConfig` is created/deleted.
- Fixes `FederatedTypeConfig` status to set the correct current status.
- Fixes E2E tests associated with `FederatedTypeConfig` finalizer changes.
- Fixes race conditions associated with the `FederatedTypeConfig` controller.

### DIFF
--- a/pkg/controller/federatedtypeconfig/controller.go
+++ b/pkg/controller/federatedtypeconfig/controller.go
@@ -25,6 +25,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	restclient "k8s.io/client-go/rest"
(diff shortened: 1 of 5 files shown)


pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -230,12 +231,15 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	}
 
 	typeConfig.Status.ObservedGeneration = typeConfig.Generation
-	if syncRunning {
+	syncControllerRunning := startNewSyncController || (syncRunning && !stopSyncController)
```
marun: (No action required) Would it be worth only calling `UpdateStatus` if status had actually changed?  As written it looks like status will be set everytime an ftc is reconciled.  If so, I'm assuming it can wait for the refactor you mentioned in #801.

scripts/pre-commit.sh
```diff
@@ -149,6 +149,9 @@ kubectl scale deployments federation-controller-manager -n federation-system --r
 echo "Running e2e tests with race detector against cluster-scoped federation-v2 with in-memory controllers"
 run-e2e-tests-with-in-memory-controllers
 
+echo "Scaling back up cluster-scoped controller manager prior to deletion"
+kubectl scale deployments federation-controller-manager -n federation-system --replicas=1
```
marun: Please add a comment (not an echo) documenting why this is necessary.

test/e2e/scheduling.go
```diff
@@ -100,6 +100,9 @@ var _ = Describe("Scheduling", func() {
 					framework.Skipf("The scheduling manager can only be tested when controllers are running in-process.")
 				}
 
+				controllerFixture = framework.NewFederatedTypeConfigControllerFixture(tl, f.ControllerConfig())
```
marun: Please add a comment explaining why this is necessary.

marun:
LGTM

shashidharatd:
/lgtm
Thanks !

Issue #805, opened 2019-04-29
Configurable 'adopt pre-existing resource' behaviour of sync controller


shashidharatd:
**What would you like to be added**:
Currently the behaviour of sync controller is to adopt an pre-existing resource in member cluster by adding a `managed` label to it. There is a desire to not adopt and report an error/event on such occurrences in some control plane deployments. We could add a configuration to sync controller to implement the above said behaviour and let the administrator chose the behaviour for their environment.

/cc @irfanurrehman @marun 

/kind feature

PR #823, opened 2019-05-01
Add a config option to control adopting pre-existing resources

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/federation-v2/issues/805

Context: https://github.com/kubernetes-sigs/federation-v2/pull/715#discussion_r273306716

/cc @marun @irfanurrehman


charts/federation-v2/values.yaml
```diff
@@ -34,6 +34,8 @@ controllermanager:
   ## Supported options are `configmaps` and `endpoints`
   leaderElectResourceLock:
   ## Value of feature gates item should be either `true` or `false`
+  syncController:
```
marun: This should be above the comment on 36 which is intended to document the feature gates values.

shashidharatd: Fixed.

shashidharatd:
@marun, Addressed your comment. PTAL.

shashidharatd:
> Is it worth adding a test that validates this behavior? The existing test for adoption could likely be the basis for a non-adoption test, at least as far as setup goes.

I think adding an e2e test case needs lot of effort and involve following things: 
- Start the Federation Control plane (SUT) with SkipAdoptResource config set to true. (probably as a new step in CI).
- Get the FederationConfig from SUT.
- Create exactly the same resource in one of member cluster which you gonna propagate.
- Check that, If SkipAdoptResource config is true, propogation should fail to that cluster.

So now, I am contemplating whether it would be good idea to just add Unit test for dispatcher. wdyt @marun ?

marun:
@shashidharatd I think it could be simpler if the test were implemented for CRDs and only run in cluster-scope.  Then the sync controller could be started with namespace-scope with adopt=false.  I'll let you address that in a subsequent PR.

shashidharatd:
> @shashidharatd I think it could be simpler if the test were implemented for CRDs and only run in cluster-scope. Then the sync controller could be started with namespace-scope with adopt=false. I'll let you address that in a subsequent PR.

Thanks @marun for the merge. I will explore this and submit a pr ASAP. For what it's worth here is some proof of it working when tested manually with `adopt=false`.
```
  Normal   CreateInCluster        23s               federateddeployment-controller  Creating Deployment "default/fr1" in cluster "cluster1"
  Normal   CreateInCluster        2s (x4 over 23s)  federateddeployment-controller  Creating Deployment "default/fr1" in cluster "cluster2"
  Warning  CreateInClusterFailed  2s (x4 over 22s)  federateddeployment-controller  Failed to create Deployment "default/fr1" in cluster "cluster2": Resource pre-exist in cluster
  Normal   UpdateInCluster        2s (x3 over 22s)  federateddeployment-controller  Updating Deployment "default/fr1" in cluster "cluster1"
```

marun:
@shashidharatd I'm glad to see the event logging is proving useful :)

I don't think the test is absolutely required right now, I'm just generally mindful of our need to be more test-focused.

Issue #535, opened 2018-12-26
need healthcheck with federation-controller-manager


tianzhuozhuo:

marun:
@tianzhuozhuo A health-check is certainly suggested as fedv2 becomes something that needs to be supported, but we're still in alpha now and there is lots of other work still to do.  Would you be willing to document your reasons for filing this issue so there are rationale justifying prioritization?

poothia:
I would like to explore this and work on this.

shashidharatd:
This is addressed with https://github.com/kubernetes-sigs/federation-v2/pull/824

PR #824, opened 2019-05-01
Add liveness probe to federation controller-manager

shashidharatd:


marun:
@shashidharatd Nice work!  Can I assume that the addition of the liveness probe to the deployment, and passing tests, is proof that this is working as intended?

Note also that this should fix #535

shashidharatd:
> @shashidharatd Nice work! Can I assume that the addition of the liveness probe to the deployment, and passing tests, is proof that this is working as intended?

yes @marun, it is working as intended.

marun:
Thank you!

/lgtm

PR #830, opened 2019-05-02
Document how to determine that a target type is present in all federated clusters

### LINKED ISSUE #827: Document how to determine that a target type is present in all federated clusters
This is intended to document how to manually do what #314 is intended to automate.

font:
Fixes #827

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -13,6 +13,7 @@
     - [Check Status of Joined Clusters](#check-status-of-joined-clusters)
     - [Unjoin Clusters](#unjoin-clusters)
   - [Enabling federation of an API type](#enabling-federation-of-an-api-type)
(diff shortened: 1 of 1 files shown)


marun:
@font As per the title of this PR (and associated issue), the requirement is to document how to detect whether a target type is present in all member clusters.  The proposed documentation works for CRDs, but a target type could also be a core type that had been disabled or a non-core type that had been added via aggregation.  Is it desirable to have separate paths for crds vs core and aggregated types, or would the following be workable?

```
kubectl api-resources | grep <group> | grep <kind or plural name>
```

font:
@marun While CRDs may be the main issue, you're right it's incomplete. I've updated. Thanks!

marun:
LGTM

PR #832, opened 2019-05-02
Federate resources from input yaml

irfanurrehman:
- Federates resources given in input yaml (emits yaml output).
- Takes input from `stdin` if `-` is specified for file name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Unreleased
+-  [#832](https://github.com/kubernetes-sigs/federation-v2/issues/832)
+   `kubefedctl federate` can take input from a file via `--filename`
+   option and stdin via `--filename -` option.
 -  [#868](https://github.com/kubernetes-sigs/federation-v2/issues/868)
    The default kubefed system namespace has been changed from
    `federation-system` to `kube-federation-system`.  The `kube-`
@@ -6,7 +9,7 @@
    having the kubefed namespace conflict with a user namespace.
 -  [#740](https://github.com/kubernetes-sigs/federation-v2/issues/740)
    Propagation status is now recorded for all federated resources.
-- [#844](https://github.com/kubernetes-sigs/federation-v2/pull/844)
+-  [#844](https://github.com/kubernetes-sigs/federation-v2/pull/844)
    `kubefedctl federate` namespace with its content gets an option to
    skip API Resources via `--skip-api-resources`.
 
--- a/pkg/kubefedctl/federate/federate.go
+++ b/pkg/kubefedctl/federate/federate.go
@@ -26,6 +26,7 @@ import (
 	"github.com/spf13/pflag"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -83,6 +84,7 @@ type federateResource struct {
 	outputYAML           bool
 	enableType           bool
 	federateContents     bool
+	filename             string
 	skipAPIResourceNames []string
 }
 
@@ -91,12 +93,26 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
+	flags.StringVarP(&j.filename, "filename", "f", "", "If specified, the provided yaml file will be used as the input for target resources to federate. This mode will only emit federated resource yaml to standard output. Other flag options if provided will be ignored.")
 	flags.StringSliceVarP(&j.skipAPIResourceNames, "skip-api-resources", "s", []string{}, "Comma separated names of the api resources to skip when federating contents in a namespace. Name could be short name "+
 		"(e.g. 'deploy), kind (e.g. 'deployment'), plural name (e.g. 'deployments'), group qualified plural name (e.g. 'deployments.apps') or group name itself (e.g. 'apps') to skip the whole group.")
 }
 
 // Complete ensures that options are valid.
 func (j *federateResource) Complete(args []string) error {
+	if j.output == "yaml" {
+		j.outputYAML = true
+	} else if len(j.output) > 0 {
+		return errors.Errorf("Invalid value for --output: %s", j.output)
+	}
+
+	if len(j.filename) > 0 {
+		if len(args) > 0 {
+			return errors.Errorf("Flag '--filename' does not take any args. Got args: %v", args)
+		}
+		return nil
+	}
+
 	if len(args) == 0 {
 		return errors.New("TYPE-NAME is required")
 	}
@@ -106,12 +122,6 @@ func (j *federateResource) Complete(args []string) error {
 		return errors.New("RESOURCE-NAME is required")
 	}
 	j.resourceName = args[1]
-
-	if j.output == "yaml" {
-		j.outputYAML = true
-	} else if len(j.output) > 0 {
-		return errors.Errorf("Invalid value for --output: %s", j.output)
-	}
 
 	if j.enableType && j.outputYAML {
 		return errors.New("Flag '--enable-type' cannot be used with '--output [yaml]'")
@@ -157,11 +167,28 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return errors.Wrap(err, "Failed to get host cluster config")
 	}
 
+	if len(j.filename) > 0 {
+		resources, err := decodeUnstructuredFromFile(j.filename)
+		if err != nil {
+			return errors.Wrapf(err, "Failed to load yaml from file %q", j.filename)
+		}
+		federatedResources, err := federateResources(resources)
+		if err != nil {
+			return err
+		}
+
+		err = writeUnstructuredObjsToYaml(federatedResources, cmdOut)
+		if err != nil {
+			return errors.Wrap(err, "Failed to write federated resources to YAML")
+		}
+		return nil
+
+	}
+
 	qualifiedResourceName := ctlutil.QualifiedName{
 		Namespace: j.resourceNamespace,
 		Name:      j.resourceName,
 	}
-
 	artifacts, err := GetFederateArtifacts(hostConfig, j.typeName, j.FederationNamespace, qualifiedResourceName, j.enableType, j.outputYAML)
 	if err != nil {
 		return err
@@ -193,6 +220,38 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 	}
 
 	return CreateResources(cmdOut, hostConfig, artifactsList, j.FederationNamespace, j.enableType, j.DryRun)
+}
+
+func federateResources(resources []*unstructured.Unstructured) ([]*unstructured.Unstructured, error) {
+	var federatedResources []*unstructured.Unstructured
+	for _, targetResource := range resources {
+		// A Group, a Version and a Kind is sufficient for API Resource definition.
+		gvk := targetResource.GroupVersionKind()
+
+		// Construct an API Resource from above info.
+		// TODO(irfanurrehman) Should we depend on the lookup from the
+		// API Server instead, for some specific scenario?
+		plural, singular := apimeta.UnsafeGuessKindToResource(gvk)
+		apiResource := metav1.APIResource{
+			Name:         plural.Resource,
+			SingularName: singular.Resource,
+			Group:        gvk.Group,
+			Version:      gvk.Version,
+			Kind:         gvk.Kind,
+		}
+		apiResource.Namespaced = targetResource.GetNamespace() == ""
+
+		qualifiedName := ctlutil.NewQualifiedName(targetResource)
+		typeConfig := enable.GenerateTypeConfigForTarget(apiResource, enable.NewEnableTypeDirective())
+		federatedResource, err := FederatedResourceFromTargetResource(typeConfig, targetResource)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error getting %s from %s %q", typeConfig.GetFederatedType().Kind, typeConfig.GetTarget().Kind, qualifiedName)
+		}
+
+		federatedResources = append(federatedResources, federatedResource)
+	}
+
+	return federatedResources, nil
 }
 
 type FederateArtifacts struct {
--- a/pkg/kubefedctl/federate/federate.go
+++ b/pkg/kubefedctl/federate/federate.go
@@ -168,16 +168,16 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 	}
 
 	if len(j.filename) > 0 {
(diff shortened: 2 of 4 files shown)


irfanurrehman:
Rebased.

marun:
Please rebase

irfanurrehman:
@marun Rebased. PTAL!

pkg/kubefedctl/federate/federate.go
```diff
@@ -83,17 +85,35 @@ type federateResource struct {
...
+	} else if len(j.output) > 0 {
+		return errors.Errorf("Invalid value for --output: %s", j.output)
+	}
+
+	if len(j.filename) > 0 {
+		if len(args) > 0 {
+			return errors.Errorf("Invalid args: %v, option '--filename' does not take any args", args)
+		}
+		if !j.outputYAML {
+			return errors.New("Flag '--filename' can be used only with '--output [yaml]'")
```
xunpan: I think it should be `s/[yaml]/yaml/`

irfanurrehman: We use `[ ]` to represent options for example `[yaml|json]`; it happens to be only `yaml` in our case.

pkg/kubefedctl/federate/federate.go
```diff
@@ -18,6 +18,7 @@ package federate
 
 import (
 	"context"
+	//"fmt"
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/kubefedctl/federate/federate.go
marun: Using square brackets for a single item suggests that it may be optional (i.e. `--output` should work).

pkg/kubefedctl/federate/federate.go
```diff
@@ -83,17 +84,35 @@ type federateResource struct {
...
 func (j *federateResource) Complete(args []string) error {
+	if j.output == "yaml" {
+		j.outputYAML = true
+	} else if len(j.output) > 0 {
+		return errors.Errorf("Invalid value for --output: %s", j.output)
+	}
+
+	if len(j.filename) > 0 {
+		if len(args) > 0 {
+			return errors.Errorf("Invalid args: %v, option '--filename' does not take any args", args)
```
marun: As per a conversation we've already had on previous PRs, embedding arbitrary user input in a message intended to indicate an error risks obscuring the error.  Please reword so that the args are appended to the error message.

pkg/kubefedctl/federate/federate.go
```diff
@@ -154,11 +167,28 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return errors.Wrap(err, "Failed to get host cluster config")
 	}
 
+	if len(j.filename) > 0 {
+		resources, err := decodeUnstructuredFromFile(j.filename)
+		if err != nil {
+			return errors.Wrapf(err, "Failed to load yamls from file %q", j.filename)
```
marun: Just as the plural of `data` is `data`, the plural of `yaml` is `yaml`.

pkg/kubefedctl/federate/federate.go
```diff
@@ -192,6 +222,42 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
...
 }
 
+func federateResources(resources []*unstructured.Unstructured) ([]*unstructured.Unstructured, error) {
+	var federatedResources []*unstructured.Unstructured
+	for _, targetResource := range resources {
+		// A Group, a Version and a Kind is sufficient for API Resource definition.
+		gvk := targetResource.GroupVersionKind()
+
+		// Construct an API Resource from above info.
+		// TODO(irfanurrehman) Should we depend on the lookup from the
```
marun: I think the answer to this question is yes, for the same reasons that we've previously discussed regarding `kubefedctl federate ... -o yaml`.  If a type config is not available from the api, by all means use a generated one.  But if one is available, ignoring it risks generating yaml that will not be compatible with the FTC in the API.

pkg/kubefedctl/federate/federate.go
```diff
@@ -192,6 +222,42 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
...
+		// API Server instead, for some specific scenario?
+		plural, singular := apimeta.UnsafeGuessKindToResource(gvk)
+		apiResource := metav1.APIResource{
+			Name:         plural.Resource,
+			SingularName: singular.Resource,
+			Group:        gvk.Group,
+			Version:      gvk.Version,
+			Kind:         gvk.Kind,
+		}
+		if targetResource.GetNamespace() == "" {
```
marun:
```golang
apiResource.Namespaced = targetResource.GetNamespace() == ""
```

pkg/kubefedctl/federate/federate.go
```diff
@@ -83,17 +84,35 @@ type federateResource struct {
...
 	federateContents  bool
+	filename          string
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.resourceNamespace, "namespace", "n", "default", "The namespace of the resource to federate.")
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
+	flags.StringVarP(&j.filename, "filename", "f", "", "If specified, the provided yaml file will be used as the input for target resources to federate. Only --output will be accepted from command line with this option. Other flag options if provided will be ignored")
```
marun: Is there a reason to support anything but yaml output when this flag is used?  If a user wanted to apply it to a server, they could pipe it to `kubectl apply`.  I'm pretty sure we already had this conversation.

pkg/kubefedctl/federate/util.go
```diff
@@ -187,3 +192,42 @@ func getResourcesInNamespace(config *rest.Config, namespace, skipAPIResourceName
 
 	return resourcesInNamespace, nil
 }
+
+// decodeUnstructuredFromFile reads a list of yamls into a slice of unstructured objects
+func decodeUnstructuredFromFile(filename string) ([]*unstructured.Unstructured, error) {
+	var f *os.File
+	if filename == "-" {
```
marun: This doesn't seem at all obvious to me.  Do other common commands employ this strategy to accept input from stdin?   And why wouldn't federate just assume stdin input when no args were supplied?

pkg/kubefedctl/federate/federate.go
```diff
@@ -154,11 +167,28 @@ func (j *federateResource) Run(cmdOut io.Writer, config util.FedConfig) error {
...
 	}
 
+	if len(j.filename) > 0 {
+		resources, err := decodeUnstructuredFromFile(j.filename)
+		if err != nil {
+			return errors.Wrapf(err, "Failed to load yamls from file %q", j.filename)
+		}
+		federatedResources, err := federateResources(resources)
+		if err != nil {
+			return err
```
xunpan: any rule to use `errrors.Wrapf` or `return err`? It seems not consistent in code context.

pkg/kubefedctl/federate/util.go
```diff
@@ -187,3 +192,35 @@ func getResourcesInNamespace(config *rest.Config, namespace, skipAPIResourceName
...
+			break
+		} else if err != nil {
+			return nil, err
+		}
+		if len(buf) == 0 {
+			break
+		}
+		if err := yaml.Unmarshal(buf, unstructuedObj); err != nil {
+			return nil, err
+		}
```
xunpan: shall we use `NewYAMLToJSONDecoder` to simplify the logic here?

irfanurrehman:
Rebased!

irfanurrehman:
> This is indeed a welcome feature! I'm assuming you're intending to add test coverage and documentation?

Documentation a certain yes; Was not planning on adding a test for this as inspections of yaml for correctness is quite obvious. Are you indicating at unit testing the new function `federateResources`  or e2e yaml validation, where input is actual yaml file and output is actual yaml output?

pkg/kubefedctl/federate/federate.go
marun: I think output to yaml would be preferable, and there would be no confusion if the help text explained that the flag output to yaml.  `kubectl apply` exists and I don't think it's advantageous for us to duplicate its behavior (see: the best code is no code).

pkg/kubefedctl/federate/federate.go
```diff
@@ -91,12 +93,29 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
...
 func (j *federateResource) Complete(args []string) error {
+	if j.output == "yaml" {
+		j.outputYAML = true
+	} else if len(j.output) > 0 {
+		return errors.Errorf("Invalid value for --output: %s", j.output)
+	}
+
+	if len(j.filename) > 0 {
+		if len(args) > 0 {
+			return errors.Errorf("Flag '--filename' does not take any args. Got args: %v", args)
```
marun: (No action required) This suggests to me that a different command should implement this behavior.

pkg/kubefedctl/federate/util.go
marun: Isn't reading a single yaml object just the degenerate case of reading multiple (just like splitting a string that doesn't contain a separator is effectively a no-op)?

pkg/kubefedctl/federate/federate.go
marun: To point 1, I don't think we have much visibility on whether conversion will be useful for federated types.  I'm not confident that internal API conversion is going to have the same utility for other types as for kubefed, since a resource can only be propagated to member clusters at a single version and I have yet to hear a reason why that version shouldn't be the latest available.

Point 2 suggests that federate should compare the version of a resource targeted for conversion with the version of the type configured in a deployed FTC.  federate should error out if the version of the resource targeted for conversion is greater than that of the type configured for the FTC.  If this occurs, a helpful message could be provided that indicates how to upgrade the type targeted by the FTC.

Point 3 suggests to me that the user should be able to control whether federate uses an apiserver or not to convert file intput to yaml.  If their intent is to create federated content with compatibility for control planes whose types are enabled with defaults, then they don't need to use an apiserver.  If their intent is to create federated content with compatibility for a control plane whose types have non-default naming or groups, then they would want to use the apiserver hosting that control plane.  Maybe `--use-apiserver`?  

Speaking of undocumented complexity, are you working on updating the userguide to reflect the work you've been doing on `federate` and to document issues like these?

irfanurrehman: I am working on documentation of federate feature. I would prefer to work on the details we discussed here outside this PR. Created #864 for that.

pkg/kubefedctl/federate/util.go
irfanurrehman: Theoretically what you mention is right. I did also expect the `decoder` (obtained by `NewYAMLToJSONDecoder`)  to ideally give me a single structure or a slice of structures or an unstructured list based on what it reads from the yaml file.
However the [decode](https://github.com/kubernetes-sigs/federation-v2/blob/master/vendor/k8s.io/apimachinery/pkg/util/yaml/decoder.go#L65) is not implemented this way. It decodes the whole yaml file content in one go which ends up decoding only the first object before the line separater (`---`). It does not return a list.
The utils provide an alternative way to obtain a [reader](https://github.com/kubernetes-sigs/federation-v2/blob/master/vendor/k8s.io/apimachinery/pkg/util/yaml/decoder.go#L262) which understands the separater (`---`)  and provides with individual yaml content on each read, which I have used here.

pkg/kubefedctl/federate/federate.go
irfanurrehman: Updated!

irfanurrehman:
@marun Updated the rest and working on tests now.

irfanurrehman:
@marun @shashidharatd FYI I have been observing a flaky failure on `ServiceDNS` test as on this PR right now and I remember seeing an issue to track it (but could not locate the issue right now). Was it fixed? Also as I could not find the issue, I am not sure that its the same failure. Can you please confirm.

marun:
@irfanurrehman #839 should be fixed

irfanurrehman:
> @irfanurrehman #839 should be fixed

Ok; this looked like a different issue. I will look into it if I observe it again.
Meanwhile update the tests. PTAL!

CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # Unreleased
-- [#844](https://github.com/kubernetes-sigs/federation-v2/pull/844)
+- [#844](https://github.com/kubernetes-sigs/federation-v2/pull/844) -
    `kubefedctl federate` namespace with its content gets an option to
    skip API Resources via `--skip-api-resources`.
-
+-  [#832](https://github.com/kubernetes-sigs/federation-v2/issues/832) -
```
marun: (No action required) I've been inserting new entries at the top, but I guess it doesn't really matter.

pkg/kubefedctl/federate/federate.go
```diff
@@ -91,12 +93,26 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
+	flags.StringVarP(&j.filename, "filename", "f", "", "If specified, the provided yaml file will be used as the input for target resources to federate. This mode will only emit federated resource yaml to standard output. Other flag options if provided will be ignored")
```
marun: Please add a period to the last sentence of the help text.

pkg/kubefedctl/federate/federate.go
marun: Fine to file it separately so long as you document the caveats of what is implemented now.  I would prefer that you don't pursue #864 until post-beta in any case.

test/e2e/federate.go
```diff
@@ -168,11 +172,75 @@ var _ = Describe("Federate ", func() {
 		}
 
 		By("Comparing the test resources with the templates of corresponding federated resources for equality")
-		validateResourcesEquality(tl, testResources, kubeConfig)
+		validateResourcesEqualityFromAPI(tl, createdTargetResources, kubeConfig)
+	})
+
+	It("input yaml from a file, should emit equivalant federated resources", func() {
+		tmpFile, err := os.Create(fmt.Sprintf("/tmp/file-%s", uuid.New()))
```
marun: Please use [ioutil.TempFile](https://golang.org/pkg/io/ioutil/#TempFile) to create a temporary file.  It is not safe to assume that `/tmp` is a safe location to write temporary files.  In my development environment, `/tmp` is not writeable by me and I use the `TMPDIR` env var to point a location I can write to.

CHANGELOG.md
irfanurrehman: Updated! Better to follow a common protocol.

pkg/kubefedctl/federate/federate.go
irfanurrehman: Updated!

test/e2e/federate.go
irfanurrehman: Updated!

irfanurrehman:
@marun Updated. PTAL!

pkg/kubefedctl/federate/federate.go
marun: I don't see the period?

marun:
@irfanurrehman It doesn't appear that the code has been updated to address my comments.

irfanurrehman:
> @irfanurrehman It doesn't appear that the code has been updated to address my comments.

Apologies, it seemed to be some `git push` issue at my end. Pushed the correct branch now.

marun:
Great work, thank you!

/lgtm

PR #842, opened 2019-05-04
Reduce unnecessary error logging on delete

marun:


pkg/controller/sync/dispatch/unmanaged.go
```diff
@@ -66,11 +66,19 @@ func (d *unmanagedDispatcherImpl) Wait() (bool, error) {
...
-func (d *unmanagedDispatcherImpl) Delete(clusterName string) {
+func (d *unmanagedDispatcherImpl) Delete(clusterName string, clusterObj *unstructured.Unstructured) {
 	d.dispatcher.incrementOperationsInitiated()
 	const op = "delete"
 	const opContinuous = "Deleting"
 	go d.dispatcher.clusterOperation(clusterName, op, func(client util.ResourceClient) util.ReconciliationStatus {
+		// It's necessary to have the controller call Delete to ensure
+		// that status is correctly logged, but attempting to delete
+		// an already deleted resource is just going to generate
+		// unnecessary errors in the log.
```
xunpan: If `Delete` resource `IsNotFound`, it assumes no error in this function. 
```
88 err := client.Resources(d.targetName.Namespace).Delete(d.targetName.Name, &metav1.DeleteOptions{})
89 if apierrors.IsNotFound(err) {
90    err = nil
91 }
```

So, what kind of errors log do you mean here? The `RecordEvent` or anything I lost?

If we do not delete a resource which has `GetDeletionTimestamp()`, how it works for `Delete` operation?

marun: I think your confusion is justified.  I'll fold this into #843 and make the status collection explicit.

PR #843, opened 2019-05-04
Write propagation status to reconciled federated resources

### LINKED ISSUE #588: Federation of placement stops when clusterName duplicated in override
**What happened**:
Created a FedreatedConfigMap, FederatedConfigMapPlacement, and FederatedConfigMapOverride.  Initial override settings had unique `clusterName` values listed.  I updated it to include a duplicate `clusterName` and the change was not federated.  Subsequent updates to the config map are not federated either.  I was unable to find an error in the federation-controller-manager log.  None of the CRs have status to indicate anything either.

At least have some error bubbled up so it's clear something failed.  Discussing in https://github.com/kubernetes-sigs/federation-v2/issues/581 that it might pick one of them instead of failing.  Not sure what is really desired.

I have 3 clusters, nmalik-test1, nmalik-test2.

Initial configMap setup:
```
---
apiVersion: v1
kind: Namespace
metadata:
    name: nmalik-config
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedNamespacePlacement
metadata:
  name: nmalik-config
  namespace: nmalik-config
spec:
  clusterSelector: {}
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedConfigMap
metadata:
  name: nmalik-cm
  namespace: nmalik-config
spec:
  template:
    data:
      A: ala ma kota
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedConfigMapPlacement
metadata:
  name: nmalik-cm
  namespace: nmalik-config
spec:
  clusterSelector: {}
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedConfigMapOverride
metadata:
  name: nmalik-cm
  namespace: nmalik-config
spec:
  overrides:
  - clusterName: openshift-hive-nmalik-test1
    clusterOverrides:
    - path: data
      value:
        A: OVERRIDEN
```

Values are federated as expected to target clusters.

Add duplicate `clusterName` and change the config map (which should be updated on nmalik-test2);
```
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedConfigMapOverride
metadata:
  name: nmalik-cm
  namespace: nmalik-config
spec:
  overrides:
  - clusterName: openshift-hive-nmalik-test1
    clusterOverrides:
    - path: data
      value:
        A: OVERRIDEN
  - clusterName: openshift-hive-nmalik-test1
    clusterOverrides:
    - path: data
      value:
        A: OVERRIDEN a second time
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedConfigMap
metadata:
  name: nmalik-cm
  namespace: nmalik-config
spec:
  template:
    data:
      A: ala ma kota UPDATED
```

The config map isn't updated because override has duplicated `clusterName`.

Remove the duplication and the updated config map lands on the cluster without the override.
```
---
apiVersion: primitives.federation.k8s.io/v1alpha1
kind: FederatedConfigMapOverride
metadata:
  name: nmalik-cm
  namespace: nmalik-config
spec:
  overrides:
  - clusterName: openshift-hive-nmalik-test1
    clusterOverrides:
    - path: data
      value:
        A: OVERRIDEN
```

**Environment**:
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

- Federation version
v0.0.4

script

cluster-scoped

- Others

/kind bug

### LINKED ISSUE #740: Propagation status should be recorded on federated resources
There have been a number of issues filed that suggest the value of the sync controller writing per-resource status and per-cluster status to aid in troubleshooting, including but not limited to:

 - Federation of placement stops when clusterName duplicated in override (#588)
 - Jobs are not updated when federated jobs are changed (and status does not show this) (#574)

The consensus seems to be that propagation status be recorded on each reconciliation loop, as per the following design doc:

https://docs.google.com/document/d/15SQ3y4Mx4nV1kWB-7i-bzRzmN8TVmNoJ6_mMSVDSZfk/edit

marun:
This PR ensures that propagation status is written for federated resources.   The PR looks large but the bulk of it is actually generated crd yaml.

TODO

 - [x] Update the user documentation
 - [x] Add checks for status to the crudtester
 - [x]  Truncate `lastProbeTime` and `lastTransitionTime` to the second

Fixes #740
Fixes #588

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -25,11 +25,13 @@ import (
 	"github.com/pkg/errors"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -40,6 +42,7 @@ import (
 	fedv1a1 "github.com/kubernetes-sigs/federation-v2/pkg/apis/core/v1alpha1"
 	genericclient "github.com/kubernetes-sigs/federation-v2/pkg/client/generic"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/sync/dispatch"
+	"github.com/kubernetes-sigs/federation-v2/pkg/controller/sync/status"
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
 	finalizersutil "github.com/kubernetes-sigs/federation-v2/pkg/controller/util/finalizers"
 )
@@ -278,22 +281,22 @@ func (s *FederationSyncController) reconcile(qualifiedName util.QualifiedName) u
 		return util.StatusError
 	}
 
-	clusters, err := s.informer.GetReadyClusters()
-	if err != nil {
-		fedResource.RecordError("ClusterRetrievalError", errors.Wrap(err, "Failed to get cluster list"))
-		return util.StatusNotSynced
-	}
-
-	return s.syncToClusters(fedResource, clusters)
+	return s.syncToClusters(fedResource)
 }
 
 // syncToClusters ensures that the state of the given object is
-// synchronized to the provided clusters.
-func (s *FederationSyncController) syncToClusters(fedResource FederatedResource, clusters []*fedv1a1.FederatedCluster) util.ReconciliationStatus {
+// synchronized to member clusters.
+func (s *FederationSyncController) syncToClusters(fedResource FederatedResource) util.ReconciliationStatus {
+	clusters, err := s.informer.GetClusters()
+	if err != nil {
+		fedResource.RecordError(string(status.ClusterRetrievalFailed), errors.Wrap(err, "Failed to retrieve list of clusters"))
+		return s.setPropagationStatus(fedResource, status.ClusterRetrievalFailed, nil)
+	}
+
 	selectedClusterNames, err := fedResource.ComputePlacement(clusters)
 	if err != nil {
-		fedResource.RecordError("ComputePlacementError", errors.Wrap(err, "Failed to compute placement"))
-		return util.StatusError
+		fedResource.RecordError(string(status.ComputePlacementFailed), errors.Wrap(err, "Failed to compute placement"))
+		return s.setPropagationStatus(fedResource, status.ComputePlacementFailed, nil)
 	}
 
 	kind := fedResource.TargetKind()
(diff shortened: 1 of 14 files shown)


marun:
Ok, ready for review.  This is a big one, but largely mechanical (helm chart update, collecting status everywhere).

marun:
Test issue should be fixed.  Still need to cleanup the timestamp handling.

marun:
TODO items are complete.  Now for the review comments...

marun:
Rebased

pkg/controller/sync/controller.go
```diff
@@ -366,7 +379,45 @@ func (s *FederationSyncController) syncToClusters(fedResource FederatedResource,
...
+
+func (s *FederationSyncController) setPropagationStatus(fedResource FederatedResource,
+	reason status.AggregateReason, statusMap status.PropagationStatusMap) util.ReconciliationStatus {
+
+	kind := fedResource.FederatedKind()
+	name := fedResource.FederatedName()
+	obj := fedResource.Object()
+
+	// If the underlying resource has changed, attempt to retrieve and
+	// update it repeatedly.  Assume that there is not another
```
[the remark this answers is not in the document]
marun: I agree the current text is confusing - I'll remove it.

pkg/controller/sync/status/status.go
```diff
@@ -0,0 +1,203 @@
...
+	err = statusObj.UnmarshalJSON(statusJSON)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to marshall generic status json to unstructured")
+	}
+	fedObject.Object[util.StatusField] = statusObj.Object[util.StatusField]
+
+	return nil
+}
+
+// setPropagationCondition ensures that the Propagation condition of
```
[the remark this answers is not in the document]
marun: Done.

marun:
@shashidharatd Thanks for the review, updated.

shashidharatd:
/lgtm
Thanks

marun:
Rebased to account for the klog PR.

pkg/kubefedctl/enable/validation.go
```diff
@@ -171,6 +171,61 @@ func ValidationSchema(specProps v1beta1.JSONSchemaProps) *v1beta1.CustomResource
 					Type: "object",
 				},
 				"spec": specProps,
+				"status": {
+					Type: "object",
+					Properties: map[string]v1beta1.JSONSchemaProps{
+						"conditions": {
```
irfanurrehman: This is what I was trying to understand in the sync today.

So this looks like that the fields are directly under status which can lead to restructuring the API if we add something similar in hierarchy to propagation status in future.
Would it make any sense to say have it like below:
```
status:
  propagationStatus:
    conditions:
    .......
``` 

To allow a possibility of something like below in future:
```
status:
  propagationStatus:
    conditions:
    .......
  aggregatedStatus:
    .......
``` 
I am not implying that we need to do any change now. I am also not aware of any use case needing all statuses for all objects from all clusters.
Additionally please correct me if I am wrong or assuming something.

marun: I don't think aggregated status will ever belong on a federated resource.  The kube api is a configuration store for the agents that operate on a cluster, not a general-purpose database for storing arbitrary amounts of data.  It can certainly be (and is) done, but I don't think we should directly support doing so on primary resources like federated types.

Are there reasons other than the potential for including aggregated status to suggest adding the status fields proposed by this PR under the `propagationStatus` key?

irfanurrehman: None as of now.

irfanurrehman:
Got held up in some personal stuff today, hence delayed on this. For whatever it might be worth now, I did too have a look at this PR and it all looked good.
Awsome work @marun

PR #844, opened 2019-05-04
Option to user to skip API Resources while federating ns with content

irfanurrehman:
- Provides an option to the user to skip API Resources while federating ns with contents.

### DIFF
--- a/pkg/kubefedctl/federate/federate.go
+++ b/pkg/kubefedctl/federate/federate.go
@@ -76,13 +76,14 @@ var (
 
 type federateResource struct {
 	options.GlobalSubcommandOptions
-	typeName          string
-	resourceName      string
-	resourceNamespace string
-	output            string
-	outputYAML        bool
-	enableType        bool
-	federateContents  bool
+	typeName             string
+	resourceName         string
+	resourceNamespace    string
+	output               string
+	outputYAML           bool
+	enableType           bool
+	federateContents     bool
+	skipAPIResourceNames []string
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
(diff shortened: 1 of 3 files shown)


pkg/kubefedctl/federate/util.go
```diff
@@ -126,7 +131,19 @@ func namespacedAPIResourceMap(config *rest.Config, skipAPIResourceNames string)
 	return apiResources, nil
 }
 
-func apiResourceMatchesSkipNames(apiResource metav1.APIResource, skipAPIResourceNames, group string) bool {
+func apiResourceGroupMatchesSkipName(skipAPIResourceNames, group string) bool {
+	for _, name := range strings.Split(skipAPIResourceNames, ",") {
```
marun: As per my comment on #741, unmarshalling (and validation) of user input should generally be done in `Complete`.  This won't matter, though, if you use `StringSliceP` instead of `StringVarP`.

pkg/kubefedctl/federate/federate.go
```diff
@@ -76,20 +76,22 @@ var (
...
+	federateContents     bool
+	skipAPIResourceNames string
 }
 
 func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.resourceNamespace, "namespace", "n", "default", "The namespace of the resource to federate.")
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
+	flags.StringVarP(&j.skipAPIResourceNames, "skip-api-resources", "s", "", "Comma separated names of the api resources to skip when federating contents in a namespace.")
```
marun: I think a more detailed help text is suggested here, since the 'name' of the api resource could be either a group or a group-qualified plural name.  Examples for each are suggested.

Also, I think `flags.StringSliceP` is probably a better choice.

xunpan: +1 for `flags.StringSliceP`

pkg/kubefedctl/federate/util.go
irfanurrehman: Updated to `StringSliceVarP`.

pkg/kubefedctl/federate/federate.go
irfanurrehman: Updated the help text.

irfanurrehman:
@marun @xunpan Updated. PTAL!

marun:
Thanks!

/lgtm

PR #847, opened 2019-05-05
enable federation of types with federation group

### LINKED ISSUE #746: Document how to enable federation of types with the same plural names
When `kubefed2 enable` is used to enable types whose plural names (e.g. `<deployments>.extensions` 
 and `<deployments>.apps`) match, the crd name of the generated federated type would also match (e.g. `<deployments>.types.federation.k8s.io`).  After the first type with a given plural name is enabled, enabling other types with the same name will fail because a crd with the generated name already exists.

The workaround to document is the use of the `--federation-group` argument to `kubefed2` when enabling a type with a duplicated plural name.   When this strategy is employed, it will also be necessary to update the rbac permissions granted to the federation control plane to include the new group.

xunpan:
Fixes #746

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -232,6 +232,39 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
 Verifying the API type exists on all member clusters will ensure successful
 propagation to that cluster.
 
+### Enabling an API type in a new federation group
+When `kubefedctl enable` is used to enable types whose plural names (e.g. **deployments**.example.com
+and **deployments**.apps) match, the crd name of the generated federated type would also match (e.g.
+**deployments**.types.federation.k8s.io).
+
+`kubefedctl enable --federation-group string` specifies the name of the API group to use for the
+generated federation type. It is `types.federation.k8s.io` by default. If a new federation group is
+enabled, the RBAC permissions for the federation controller manager will need to be updated to include
+permissions for the new group.
+
+For example, after federation deployment, `deployments.apps` is enabled by default. To enable
+`deployments.example.com`, you should:
+```bash
+kubefedctl enable deployments.example.com --federation-group federation.example.com
+kubectl patch clusterrole federation-role --type='json' -p='[{"op": "add", "path": "/rules/1", "value": {
+            "apiGroups": [
+                "federation.example.com"
+            ],
+            "resources": [
+                "*"
+            ],
+            "verbs": [
+                "get",
+                "watch",
+                "list",
+                "update"
+            ]
+        }
+}]'
+```
+This example is for cluster scoped federation deployment. For namespaced federation deployment,
+you can patch role `federation-role` in the federation namespace instead.
+
 ## Disabling federation of an API type
 
 It is possible to disable propagation of a type that is configured for propagation using the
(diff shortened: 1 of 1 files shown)


docs/userguide.md
```diff
@@ -232,6 +233,38 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
...
 propagation to that cluster.
 
+### Enabling an API type in a new federation group
+When `kubefedctl enable` is used to enable types whose plural names (e.g. **deployments**.extensions
+and **deployments**.apps) match, the crd name of the generated federated type would also match (e.g.
+**deployments**.types.federation.k8s.io).
+
+`kubefedctl enable --federation-group string` specifies the name of the API group to use for the
+generated federation type. It is `types.federation.k8s.io` by default. If a new federation group is
+enabled, RBAC for federation controller should be updated accordingly.
```
marun: `s/ RBAC for federation controller should be updated accordingly/the RBAC permissions for the federation controller manager will need to be updated to include permissions for the new group/`

docs/userguide.md
```diff
@@ -232,6 +233,38 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
...
+
+For example, after federation deplyment, `deployments.apps` is enabled by default. To enable
```
marun: s/deplyment/deployment/

docs/userguide.md
```diff
@@ -232,6 +233,38 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
...
+`deployments.extensions`, you should do:
```
marun: `s/do//`

docs/userguide.md
```diff
@@ -232,6 +233,38 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
...
+            "verbs": [
+                "get",
+                "watch",
+                "list",
+                "update"
+            ]
+        }
+}]'
+```
+This example is for cluster scoped fderation deployment. For namespaced federation deployment,
```
marun: `s/fderation/federation/`

xunpan:
@marun done

docs/userguide.md
```diff
@@ -232,6 +233,39 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
...
+
+`kubefedctl enable --federation-group string` specifies the name of the API group to use for the
+generated federation type. It is `types.federation.k8s.io` by default. If a new federation group is
+enabled, the RBAC permissions for the federation controller manager will need to be updated to include
+permissions for the new group.
+
+For example, after federation deployment, `deployments.apps` is enabled by default. To enable
+`deployments.example.com`, you should:
+```bash
+kubefedctl enable deployments.example.com --federation-group example.com
```
marun: `s/--federation-group example.com/--federation-group federation.example.com/`

I don't think it would be desirable to reuse a non-federated group for federated resources because there is no guarantee that the non-federated resource is a CRD.  As per the following comment, mixing CRDs and non-CRDs in the same group is best avoided:

https://github.com/kubernetes/kubernetes/issues/68985#issuecomment-423790348

docs/userguide.md
```diff
@@ -232,6 +233,39 @@ NAME   SHORTNAMES   APIGROUP   NAMESPACED   KIND
...
+kubectl patch clusterrole federation-role --type='json' -p='[{"op": "add", "path": "/rules/1", "value": {
```
marun: As per the above comment suggesting the use of a federated group that is distinct from the resource group, the role would need to be granted permission to both.

xunpan: federation group is managed by `clusterrole`. So, update the name here.

marun:
Thank you!

/approve
/lgtm

marun:
Thank you!

/approve
/lgtm

Issue #373, opened 2018-10-29
Add Cross Compile Support to Deploy Script


danehans:
The deployment [script](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md#automated-deployment) builds the controller-manager bin using GOOS=darwin when running on Mac OSX. This causes problems because the Docker image uses the bin as darwin instead of linux.

gyliu513:
@danehans can you please show more detail for where is using `GOOS=darwin`? I did not found it in the [script](https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/development.md#automated-deployment) you mentioned, am I missing anything?

gyliu513:
Does it because Kubernetes also support running on OSX?

gyliu513:
FYI @xunpan

gyliu513:
@danehans can you help close this one as it was resolved in https://github.com/kubernetes-sigs/federation-v2/pull/416

gyliu513:
@danehans @shashidharatd do we also need to make the `kubefed2` also as cross compile? I found that both `controller-manager` and `hyperfed` support this but `kubefed2` not.

marun:
I think we may want to add a darwin binary to the release artifacts for v1beta1.

shashidharatd:
I can work on this if nobody else is working on this.

shashidharatd:
@yamt, @draveness any of you interested in picking up this task?

draveness:
> @yamt, @draveness any of you interested in picking up this task?

I could pick this up. :)

draveness:
> I think we may want to add a darwin binary to the release artifacts for v1beta1.

Hi, just found out this PR https://github.com/kubernetes-sigs/federation-v2/pull/854/files#diff-935529b9829971ac67b10793cc76e3acR40 which build release artifacts automatically.

Is it correct we should use `kubefedctl-darwin` instead of `kubefedctl-linux`?

marun:
@draveness `make kubefedctl` should continue building for the detected platform.  The cross compile support requested by this issue requires something like `make kubefedctl HOST_OS=<darwin|linux>` to enable building for a host os different from the one that the command is being executed on.

marun:
@draveness Ideally the proposed build script would be able to automatically build `kubefectl` for both darwin and linux:

https://github.com/kubernetes-sigs/federation-v2/pull/854/files#diff-935529b9829971ac67b10793cc76e3acR37

[...]

PR #848, opened 2019-05-05
doc installation for kubefedctl

### LINKED ISSUE #654: kubefed2 installtion guide missing in userguide

xunpan:
For download part, currently, we have `kubefed2`. I don't want to make it here because it is obsolete. I just use new command name for download part for future release download.

Fixes #654

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -27,6 +27,21 @@ help you get started.
 
 ## Prerequisites
 
+### Binaries
+
+The federation deployment depends on `kubebuilder`, `etcd`, `kubectl`, and
+`kube-apiserver` >= v1.13 being installed in the path. The `kubebuilder`
+([v1.0.8](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v1.0.8)
+as of this writing) release packages all of these dependencies together.
+
+These binaries can be installed via the `download-binaries.sh` script, which
+downloads them to `./bin`:
+
+```bash
+./scripts/download-binaries.sh
+export PATH=$(pwd)/bin:${PATH}
+```
+
 ### kubernetes
 
 The federation v2 deployment requires kubernetes version >= 1.11. To see a
--- a/docs/development.md
+++ b/docs/development.md
@@ -4,6 +4,7 @@
 
 - [Development Guide](#development-guide)
   - [Prerequisites](#prerequisites)
+    - [Binaries](#binaries)
     - [kubernetes](#kubernetes)
     - [docker](#docker)
   - [Adding a new API type](#adding-a-new-api-type)
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -60,20 +60,19 @@ is a detailed list of binaries required.
 
 ### Binaries
 
-The federation deployment depends on `kubebuilder`, `etcd`, `kubectl`, and
-`kube-apiserver` >= v1.13 being installed in the path. The `kubebuilder`
-([v1.0.8](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v1.0.8)
-as of this writing) release packages all of these dependencies together.
+`kubectl` is installed by the [guide](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
 
-These binaries can be installed via the `download-binaries.sh` script, which
-downloads them to `./bin`:
-
+`kubefedctl` is the federation command line utility. You can download
+the latest binary from the [release page](https://github.com/kubernetes-sigs/federation-v2/releases).
 ```bash
-./scripts/download-binaries.sh
-export PATH=$(pwd)/bin:${PATH}
+VERSION=<latest-version>
+curl -LO https://github.com/kubernetes-sigs/federation-v2/releases/download/${VERSION}/kubefedctl.tgz
+tar -zxvf kubefedctl.tgz
+chmod u+x kubefedctl
+sudo mv kubefedctl /usr/local/bin/ # make sure the location is in the PATH
 ```
 
-Or you can install them manually yourself using the guidelines provided below.
+**NOTE:** `kubefedctl` is built for Linux only in the release package.
 
 ### Deployment Image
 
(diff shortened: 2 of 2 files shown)


docs/userguide.md
```diff
@@ -117,9 +135,9 @@ Next, you'll want to use the `kubefedctl` tool to join all your
 clusters that you want to test against.
 
 ```bash
-./bin/kubefedctl join cluster1 --cluster-context cluster1 \
```
gyliu513: Why removing `./bin` here?

xunpan: It is useless when we set bin as part of `PATH` in prerequisite.
Remove it from doc to make `kubefedctl` usage consistent across document.

docs/userguide.md
```diff
@@ -69,12 +69,30 @@ These binaries can be installed via the `download-binaries.sh` script, which
...
+If you want to have the latest `kubefedctl`, you can build it by yourself.
+```bash
+make kubefedctl
+```
+Or you can download the latest binary from [release page](https://github.com/kubernetes-sigs/federation-v2/releases).
+```bash
+VERSION=<latest-version>
+wget https://github.com/kubernetes-sigs/federation-v2/releases/download/${VERSION}/kubefedctl.tgz
+tar -zxvf kubefedctl.tgz
+chmod u+x kubefedctl
```
shashidharatd: Do we need to add a note the the tool is built for Linux only?

xunpan: ok

marun: I think we should start releasing a darwin binary with v1beta1.  I've elevated the related issue accordingly:

https://github.com/kubernetes-sigs/federation-v2/issues/373

docs/userguide.md
```diff
@@ -69,12 +69,31 @@ These binaries can be installed via the `download-binaries.sh` script, which
...
+git clone https://github.com/kubernetes-sigs/federation-v2.git $GOPATH/src/github.com/kubernetes-sigs/federation-v2
+cd $GOPATH/src/github.com/kubernetes-sigs/federation-v2
 ./scripts/download-binaries.sh
 export PATH=$(pwd)/bin:${PATH}
 ```
 
 Or you can install them manually yourself using the guidelines provided below.
 
+`kubefedctl` is the federation command line utility. You can built it from
+the federtion source code or download it from the release site.
```
marun: `s/federtion/federation`

xunpan:
@marun 
1. move binaries into dev guide
2. `kubefedctl` and `kubectl` is still kept in user guide

docs/development.md
```diff
@@ -27,6 +28,23 @@ help you get started.
 
 ## Prerequisites
 
+### Binaries
+
+The federation deployment depends on `kubebuilder`, `etcd`, `kubectl`, and
```
marun: With the removal of managed test fixture, `etcd` and `kube-apiserver` are no longer required.  `helm` and `golangci-lint` are new dependencies.  I think it would be preferable to skip naming the dependencies and just refer the user to the download script.  This would avoid having to maintain the docs as our dependencies change.

docs/userguide.md
```diff
@@ -60,20 +60,18 @@ is a detailed list of binaries required.
...
-`kube-apiserver` >= v1.13 being installed in the path. The `kubebuilder`
-([v1.0.8](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v1.0.8)
-as of this writing) release packages all of these dependencies together.
-
-These binaries can be installed via the `download-binaries.sh` script, which
-downloads them to `./bin`:
+`kubectl` is installed by the [guide](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
 
+`kubefedctl` is the federation command line utility. You can download
+the latest binary from [release page](https://github.com/kubernetes-sigs/federation-v2/releases).
```
marun: `s/from/from the/`

docs/userguide.md
marun: - If we're explaining how to download (vs just pointing at the download page), that suggests explaining how to verify the integrity of the archive with sha256sum/openssl.  It's not clear to me that we need to be so detailed though - it may be sufficient to point someone at the release page and tell them to put it in `~/bin`.

- wget is deprecated in favor of curl on many operating systems, so please use curl instead (likely with -LO args as per the example of our download script).

- Is chmod necessary?  It appears to already have that permission when I extract the tgz.

docs/development.md
xunpan: `etcd` and `kube-apiserver` is needed by `sync-up-helm-chart.sh`. So, they are required.

I agree to remove binary list here to decrease the maintenance effort.

docs/userguide.md
xunpan: What did you mean for this?
 - add more for `verify the integrity`?
 - remove all details here and just asks them to download and move to `/bin`?

marun: I would lean towards the 2nd option (unless you feel strongly for the 1st).

xunpan: I prefer option 1, it follows similar way as `kubectl` installation. And easy for a new user to use.
1. `curl` is used instead of `wget`
2. integrity check is not added because it is not mandatory for using `clusterctl` I think.
3. `chmod +x` is harmless and make sure always get what we want.

docs/development.md
```diff
@@ -27,6 +28,23 @@ help you get started.
...
+`kube-apiserver` >= v1.13 being installed in the path. The `kubebuilder`
+([v1.0.8](https://github.com/kubernetes-sigs/kubebuilder/releases/tag/v1.0.8)
+as of this writing) release packages all of these dependencies together.
+
+These binaries can be installed via the `download-binaries.sh` script, which
+downloads them to `./bin`:
+
+```bash
+git clone https://github.com/kubernetes-sigs/federation-v2.git $GOPATH/src/github.com/kubernetes-sigs/federation-v2
```
marun: I don't think these 2 lines belong here.  Checking out the repo is something a developer should be able to figure out.

docs/userguide.md
```diff
@@ -60,20 +60,19 @@ is a detailed list of binaries required.
...
+`kubefedctl` is the federation command line utility. You can download
+the latest binary from the [release page](https://github.com/kubernetes-sigs/federation-v2/releases).
 ```bash
-./scripts/download-binaries.sh
-export PATH=$(pwd)/bin:${PATH}
+VERSION=<latest-version>
+curl -LO https://github.com/kubernetes-sigs/federation-v2/releases/download/${VERSION}/kubefedctl.tgz
+tar -zxvf kubefedctl.tgz
+chmod u+x kubefedctl
+sudo mv kubefedctl /usr/local/bin/ # make sure the locatin is in the PATH
```
marun: `s/locatin/location/`

xunpan:
@marun done

marun:
Thank you!

/approve
/lgtm

PR #850, opened 2019-05-06
feat: make num workers configurable in dnsendpoint

draveness:


marun:
What motivated you to make this change? It's effectively non-functional - no change in behavior or capability.

draveness:
> What motivated you to make this change? It's effectively non-functional - no change in behavior or capability.

I found a TODO listed in the code - consider making numWorkers configurable. Do you have any suggestions on this?

marun:
@draveness I'm afraid you misread the import of the TODO, and I think you're unlikely to find productive work from the code alone.  At the least I suggest engaging with the author of the TODO.  Also, we file issues for things that we identify as priorities - TODOs are mere bookmarks to consider in future maintenance that touches the code.

Please reach out to me on #sig-multicluster on kube slack so that I may assist you in your search for meaningful contribution.

Issue #813, opened 2019-04-30
Automate the release of new versions


font:
**What would you like to be added**:
The release instructions documented at https://github.com/kubernetes-sigs/federation-v2/blob/master/docs/releasing.md should be automated to the extent possible. We could introduce a new Makefile target e.g. `make release` that would invoke a script. If necessary, this may include prompting the user at various steps in order to make a decision. We may also require leveraging the github hub tool https://github.com/github/hub for certain steps.

**Why is this needed**:
Reduce time and effort needed to release as well as minimize human errors.

/kind feature

PR #854, opened 2019-05-06
Add script to automate build of release artifacts

marun:
- Add script to build release artifacts

 - Remove documentation of artifact build (replaced by script)

 - Remove appVersion from chart yaml to avoid having to maintain it

Partial fix for #813.   I tried automating the tag creation/push and it ended up being more trouble than it was worth.

### DIFF
--- a/charts/federation-v2/Chart.yaml
+++ b/charts/federation-v2/Chart.yaml
@@ -1,5 +1,4 @@
 apiVersion: v1
-appVersion: 0.0.2
 description: Kubernetes Federation V2 helm chart
 name: federation-v2
 version: 0.0.2
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -18,25 +18,15 @@ Creating a federation v2 release involves the following steps:
      annotated tag through the github web interface.
 2. Push the tag to master
    - `git push origin <tag>` (this requires write access to the repo)
-3. Build `kubefedctl` for release
-   1. `make kubefedctl`
-   2. `cd bin` (from repo root)
-   3. `kubefedctl version` (check that the output is as expected)
-   4. `tar cvzf kubefedctl.tgz kubefedctl`
-   5. `sha256sum kubefedctl.tgz > kubefedctl.tgz.sha`
-4. Package the helm chart for release
-   1. Update the default image tag in values.yaml (Change the version to match the release)
-   2. Update the chart version in Chart.yaml (Format should be `x.x.x`)
-   3. `cd charts` (from repo root)
-   4. `helm package federation-v2`
-   5. `sha256sum federation-v2-<x.x.x>.tgz > federation-v2-<x.x.x>.tgz.sha`
-   6. `helm repo index . --merge index.yaml --url=https://github.com/kubernetes-sigs/federation-v2/releases/download/v<x.x.x>` (Add the new version to the chart index)
-   7. Check that index.yaml contains the added release
+3. Build the release artifacts (will output to root of repo)
+   - `./scripts/build-release-artifacts.sh <tag>`
 4. Create github release
    1. Copy text from old release and replace old tag references
    2. Add a synopsis of the `Unreleased` section of `CHANGELOG.md`
-   3. Add `kubefedctl.tgz` and `kubefedctl.tgz.sha`
-   4. Add `federation-v2-<x.x.x>.tgz` and `federation-v2-<x.x.x>.tgz.sha`
+   3. Add `kubefedctl-<x.x.x>-linux-amd64.tgz` and `kubefedctl-<x.x.x>-linux-amd64.tgz.sha`
+   4. Add `kubefedctl-<x.x.x>-darwin-amd64.tgz` and `kubefedctl-<x.x.x>-darwin-amd64.tgz.sha`
+   5. Add `federation-v2-<x.x.x>.tgz` and `federation-v2-<x.x.x>.tgz.sha`
 5. Update master
    1. Move the contents of the `Unreleased` section of `CHANGELOG.md` to `v<x.x.x>`
-   2. Propose a PR that includes changes to `charts/index.yaml` (from `4.6`) and `CHANGELOG.md`
+   2. Propose a PR that includes changes to `charts/index.yaml`
+      (updated by the build script) and `CHANGELOG.md`
--- a/scripts/build-release-artifacts.sh
+++ b/scripts/build-release-artifacts.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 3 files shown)


charts/federation-v2/Chart.yaml
```diff
@@ -1,5 +1,4 @@
 apiVersion: v1
-appVersion: 0.0.2
```
xunpan: any comments to remove this?

xunpan:
Overall it LGTM.

charts/federation-v2/Chart.yaml
marun: Yes, it's in the first comment: `Remove appVersion from chart yaml to avoid having to maintain it`

Currently we don't have a reason to set an `appVersion` separate from the chart version, and `appVersion` is optional, so I figured it was preferable to remove it rather than having to update it with each release.

docs/releasing.md
```diff
@@ -18,25 +18,14 @@ Creating a federation v2 release involves the following steps:
...
 4. Create github release
    1. Copy text from old release and replace old tag references
    2. Add a synopsis of the `Unreleased` section of `CHANGELOG.md`
    3. Add `kubefedctl.tgz` and `kubefedctl.tgz.sha`
    4. Add `federation-v2-<x.x.x>.tgz` and `federation-v2-<x.x.x>.tgz.sha`
 5. Update master
    1. Move the contents of the `Unreleased` section of `CHANGELOG.md` to `v<x.x.x>`
-   2. Propose a PR that includes changes to `charts/index.yaml` (from `4.6`) and `CHANGELOG.md`
+   2. Propose a PR that includes changes to `charts/index.yaml`
+      (updated by step 2) and `CHANGELOG.md`
```
font: s/2/3/

scripts/build-release-artifacts.sh
```diff
@@ -0,0 +1,65 @@
...
+    helm package federation-v2
+
+    # Update the repo index (will need to be committed)
+    helm repo index . --merge index.yaml --url="https://github.com/kubernetes-sigs/federation-v2/releases/download/${RELEASE_TAG}"
+
+    sha256sum "federation-v2-${RELEASE_VERSION}.tgz" > "federation-v2-${RELEASE_VERSION}.tgz.sha"
+    mv federation-v2-${RELEASE_VERSION}.tgz* "${ROOT_DIR}/"
+
+    # Revert the chart changes (should not be committed)
+    mv federation-v2/values.yaml.backup federation-v2/values.yaml
```
font: Not part of this PR, but I'm thinking it might make sense to commit these changes as part of the tagged release, possibly once we go beta. The idea is that you can checkout a tagged release and deploy it without having to download any artifacts. This would probably be for dev use primarily.

marun: I don't know that what you're suggesting makes sense in the context of a 'tag'.  A branch, maybe, since that's a persistent context that can be updated, but a tag is just a persistent alias for a commit hash.

docs/releasing.md
marun: Done.

marun:
@font Adding a tag and pushing that tag a) requires correct permissions b) can only be done once.  I opted to skip adding context-sensitive behavior and stick to just building the artifacts (in any case the only part of the release process that has proven tedious and potentially error-prone thus far).

Another manual step in the process that doesn't seem amenable to full automation - at least in the short-term - is the change log.  Maybe at some point we can move to scraping PR content as per the example of k/k, but for now the creation of a concise and user-focused changelog is a manual process taking both the changelog and the commit history as input.

scripts/build-release-artifacts.sh
font: Right, and that commit object would point to code that contains the right version of stuff. It could either be based off `master` as we do today, but that would require a commit dance if we wanted to keep using the `canary` tag. Or a new/temporary branch named the same (or similarly) as the tag.

docs/releasing.md
```diff
@@ -18,25 +18,14 @@ Creating a federation v2 release involves the following steps:
...
 4. Create github release
    1. Copy text from old release and replace old tag references
    2. Add a synopsis of the `Unreleased` section of `CHANGELOG.md`
    3. Add `kubefedctl.tgz` and `kubefedctl.tgz.sha`
    4. Add `federation-v2-<x.x.x>.tgz` and `federation-v2-<x.x.x>.tgz.sha`
 5. Update master
    1. Move the contents of the `Unreleased` section of `CHANGELOG.md` to `v<x.x.x>`
-   2. Propose a PR that includes changes to `charts/index.yaml` (from `4.6`) and `CHANGELOG.md`
+   2. Propose a PR that includes changes to `charts/index.yaml`
+      (updated by when the build script) and `CHANGELOG.md`
```
font: s/when //

marun: le sigh

marun:
Updated

shashidharatd:
@marun, are you planning to include building kubefedctl for macos in this pr or subsequent pr?

marun:
@shashidharatd I'll add it to this PR.

scripts/build-release-artifacts.sh
```diff
@@ -0,0 +1,65 @@
...
+  >&2 echo "usage: $0 <release tag of the form v[0-9]+.[0-9]+.[0-9]+>"
+  exit 1
+fi
+
+ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+RELEASE_VERSION="${RELEASE_TAG:1:${#RELEASE_TAG}-1}"
+
+pushd "${ROOT_DIR}"
+  # Build release artifacts for kubefedctl
+  make kubefedctl
```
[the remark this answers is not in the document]
marun: As per other comments on this PR, the intention is to build binaries for both darwin and linux.

marun:
The script has been updated to build kubefedctl artifacts for both darwin and linux, and the names of the tar files now include the version, distro and architecture.

scripts/build-release-artifacts.sh
```diff
@@ -0,0 +1,79 @@
...
+  exit 1
+fi
+
+ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+RELEASE_VERSION="${RELEASE_TAG:1:${#RELEASE_TAG}-1}"
+
+pushd "${ROOT_DIR}"
+  # Build release artifacts for kubefedctl
+  make kubefedctl HOST_OS=linux
+  make kubefedctl HOST_OS=darwin
```
yamt: HOST_OS is intended to specify the target of the symlink.
as you don't need the symlink here, it's more straightforward to do `make bin/kubefedctl-linux bin/kubefedctl-darwin`

scripts/build-release-artifacts.sh
```diff
@@ -0,0 +1,79 @@
...
+  pushd "${ROOT_DIR}/bin"
+    # Building kubefedctl-[platform] automatically creates symlink to
+    # kubefectl.  Ensure the symlink points to the binary appropraite
```
yamt: s/kubefectl/kubefedctl/

scripts/build-release-artifacts.sh
```diff
@@ -0,0 +1,79 @@
...
+    # to the platform.
+    #
+    # TODO(marun) Fix the makefile to only create the symlink if the
+    # binary suffix matches the host os.
+    HOST_OS="$(uname -s | tr A-Z a-z)"
+    ln -sf "kubefedctl-${HOST_OS}" kubefedctl
```
yamt: i couldn't find what needs this symlink. is it necessary?

marun:
@yamt Thanks for the pointers, I've updated accordingly.

shashidharatd:
/lgtm
Thanks !

PR #695, opened 2019-03-27
WIP Switch from glog to klog

poothia:
closes #694

PR #857, opened 2019-05-08
Switch from glog to klog

### LINKED ISSUE #694: Use a forked version of glog in kubernetes
Use the permanent forked version of glog for kubernetes instead of glog.

ref:
https://github.com/kubernetes/kubernetes/issues/70264
https://github.com/kubernetes/klog/blob/master/README.md

/kind feature

font:
#695 is WIP, but this switch to use `klog` is needed in order to enable webhook dependencies in a follow-up PR.

Posting on behalf of @pmorie who did the work.

Fixes #694

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
 	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -27,7 +27,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
@@ -38,6 +37,7 @@ import (
 	"k8s.io/apiserver/pkg/util/logs"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
 
 	"github.com/kubernetes-sigs/federation-v2/cmd/controller-manager/app/leaderelection"
 	"github.com/kubernetes-sigs/federation-v2/cmd/controller-manager/app/options"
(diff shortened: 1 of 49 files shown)


marun:
@font I guess neither you nor @pmorie noticed that @poothia had already started work on this, and that I made the following suggestion:

https://github.com/kubernetes-sigs/federation-v2/pull/695#issuecomment-477244315

test/e2e/framework/test_context.go
```diff
@@ -22,9 +22,8 @@ import (
 	"os"
 	"time"
 
-	"github.com/golang/glog"
-
 	"github.com/kubernetes-sigs/federation-v2/pkg/controller/util"
+	"k8s.io/klog"
```
xunpan: k8s.io is a third party package. It should be better on top of federation package as indepedent section.

font: Agreed. This is a default `gofmt` placement since there's no space.

font:
> @font I guess neither you nor @pmorie noticed that @poothia had already started work on this in #695?
> 
> @poothia fyi, you can close yours.

@marun I made a comment about it that it looks like it is WIP and hasn't been updated recently. It's not my intent to take the work from @poothia (apologies!), it was more to see it through since a follow-up PR for the webhook framework which depends on it is forthcoming.

test/e2e/framework/test_context.go
marun: The convention we've converged on is:

```golang
import (
  <system imports (e.g. fmt)>

  <3rd party imports (e.g. glog, errors)>

  <k8s imports (e.g. klog, metav1)>

  <fedv2 imports>
 
  [<test-specific imports>]
)
```

font:
@marun @xunpan I've pushed a commit that reorders things based on our converged convention. We should really document that in the development guide.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -113,15 +114,15 @@ func Run(opts *options.Options) error {
 	setOptionsByFederationConfig(opts)
 
 	if err := utilfeature.DefaultFeatureGate.SetFromMap(opts.FeatureGates); err != nil {
-		glog.Fatalf("Invalid Feature Gate: %v", err)
+		klog.Fatalf("Invalid Feature Gate: %v", err)
```
shashidharatd: @font, i think you missed this comment from the @poothia's PR. https://github.com/kubernetes-sigs/federation-v2/pull/695#issuecomment-477451066 . what is your opinion? I think it would reduce much of the unwanted code change.

font: I believe you're referring to using an alias. Having an alias would certainly reduce the number of code changes to just the header imports of each file that uses `glog`. But this is mostly a mechanical change (aside from dependency changes) so I'm not clear on the desire to not want the complete code change if it's already done. Plus using an alias makes `goimports` not work as well when you're trying to add a new import to a file you're editing.

shashidharatd: I was concerned about the downstream users having to absorb all the changes which could be avoidable. But anyway its minor effort as you said. I will leave it to @marun to LGTM please. Thanks !

marun:
Thanks @font and @pmorie!

/lgtm

PR #859, opened 2019-05-08
idempotent unjoin

### LINKED ISSUE #745: Unjoin should be idempotent
Currently, attempting to unjoin an already-unjoined cluster will result in error messages as unjoin tries and fails to remove resources that do not exist in the cluster targeted for unjoin.  Similar to join, unjoin should default to not reporting errors of missing resources unless provided with an argument like `--error-on-missing`.

/king bug

xunpan:
`unjoin`: when resources does not exists, ignore it and delete other resources continuely until try my best. This is almost `--force` by default.

Fixes #745

### DIFF
--- a/pkg/kubefedctl/unjoin.go
+++ b/pkg/kubefedctl/unjoin.go
@@ -189,36 +189,42 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace, hos
 	}
 
 	if clusterClientset != nil {
-		err := deleteRBACResources(clusterClientset, kubefedNamespace, unjoiningClusterName, hostClusterName, dryRun)
+		err := deleteRBACResources(clusterClientset, kubefedNamespace, unjoiningClusterName, hostClusterName, forceDeletion, dryRun)
 		if err != nil {
-			return err
+			if !forceDeletion {
+				return err
+			}
+			klog.V(2).Infof("Failed to delete RBAC resources: %v", err)
 		}
 
 		err = deleteFedNSFromUnjoinCluster(hostClientset, clusterClientset, kubefedNamespace, unjoiningClusterName, dryRun)
 		if err != nil {
-			return err
+			if !forceDeletion {
+				return err
+			}
+			klog.V(2).Infof("Failed to delete kubefed namespace: %v", err)
 		}
 	}
 
 	// deletionSucceeded when all operations in deleteRBACResources and deleteFedNSFromUnjoinCluster succeed.
-	return deleteFederatedClusterAndSecret(hostClientset, client, kubefedNamespace, unjoiningClusterName, dryRun)
+	return deleteFederatedClusterAndSecret(hostClientset, client, kubefedNamespace, unjoiningClusterName, forceDeletion, dryRun)
 }
 
 // deleteKubefedClusterAndSecret deletes a federated cluster resource that associates
 // the cluster and secret.
-func deleteKubefedClusterAndSecret(hostClientset kubeclient.Interface, client genericclient.Client,
-	kubefedNamespace, unjoiningClusterName string, dryRun bool) {
+func deleteFederatedClusterAndSecret(hostClientset kubeclient.Interface, client genericclient.Client,
+	kubefedNamespace, unjoiningClusterName string, forceDeletion, dryRun bool) error {
 	if dryRun {
 		return nil
 	}
 
-	klog.V(2).Infof("Deleting federated cluster resource from namespace %q for unjoin cluster %q",
+	klog.V(2).Infof("Deleting kubefed cluster resource from namespace %q for unjoin cluster %q",
 		kubefedNamespace, unjoiningClusterName)
 
 	fedCluster := &fedv1a1.KubefedCluster{}
 	err := client.Get(context.TODO(), fedCluster, kubefedNamespace, unjoiningClusterName)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to get federated cluster \"%s/%s\"", kubefedNamespace, unjoiningClusterName)
+		return errors.Wrapf(err, "Failed to get kubefed cluster \"%s/%s\"", kubefedNamespace, unjoiningClusterName)
 	}
 
 	err = hostClientset.CoreV1().Secrets(kubefedNamespace).Delete(fedCluster.Spec.SecretRef.Name,
@@ -226,19 +232,27 @@ func deleteKubefedClusterAndSecret(hostClientset kubeclient.Interface, client ge
 	if apierrors.IsNotFound(err) {
 		klog.V(2).Infof("Secret \"%s/%s\" does not exist in the host cluster.", kubefedNamespace, fedCluster.Spec.SecretRef.Name)
 	} else if err != nil {
-		return errors.Wrapf(err, "Failed to delete secret \"%s/%s\" for unjoin cluster %q",
+		wrappedErr := errors.Wrapf(err, "Failed to delete secret \"%s/%s\" for unjoin cluster %q",
 			kubefedNamespace, fedCluster.Spec.SecretRef.Name, unjoiningClusterName)
+		if !forceDeletion {
+			return wrappedErr
+		}
+		klog.V(2).Infof("%v", wrappedErr)
 	} else {
 		klog.V(2).Infof("Deleted secret \"%s/%s\" for unjoin cluster %q", kubefedNamespace, fedCluster.Spec.SecretRef.Name, unjoiningClusterName)
 	}
 
 	err = client.Delete(context.TODO(), fedCluster, fedCluster.Namespace, fedCluster.Name)
 	if apierrors.IsNotFound(err) {
-		klog.V(2).Infof("Federated cluster \"%s/%s\" does not exist in the host cluster.", fedCluster.Namespace, fedCluster.Name)
+		klog.V(2).Infof("KubeFed cluster \"%s/%s\" does not exist in the host cluster.", fedCluster.Namespace, fedCluster.Name)
 	} else if err != nil {
-		return errors.Wrapf(err, "Failed to delete federated cluster \"%s/%s\" for unjoin cluster %q", fedCluster.Namespace, fedCluster.Name, unjoiningClusterName)
+		wrappedErr := errors.Wrapf(err, "Failed to delete kubefed cluster \"%s/%s\" for unjoin cluster %q", fedCluster.Namespace, fedCluster.Name, unjoiningClusterName)
+		if !forceDeletion {
+			return wrappedErr
+		}
+		klog.V(2).Infof("%v", wrappedErr)
 	} else {
-		klog.V(2).Infof("Deleted federated cluster \"%s/%s\" for unjoin cluster %q.", fedCluster.Namespace, fedCluster.Name, unjoiningClusterName)
+		klog.V(2).Infof("Deleted kubefed cluster \"%s/%s\" for unjoin cluster %q.", fedCluster.Namespace, fedCluster.Name, unjoiningClusterName)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


quinton-hoole:
Agreed @marun .  I'm not convinced that this PR is the right approach.

xunpan:
My rules for message:
1.  each resource identity is print out with `%q`
2. `"namespace/resourcename"` is used in message.
3.  as output, I do not use resource name but use description. E.g. `ServiceAccount` is `service account`
4. `dry-run` will have no deleting message showed.
5. `errors.Wrap` are all in the lowest for kubernetes api calling.

irfanurrehman:
@xunpan I think a flag to control the behaviour is still missing; the code will change considerably if you base the behaviour on user choice (similar to `error-on-existing` in `join`).

xunpan:
@irfanurrehman Yes. I did not do it yet.
Did you share your thoughts for what is the case to use `exit-on-missing` like functions? I think a idempotent behavior is to `unjoin` cluster with deleting all existing resources. If it does not exists, it means it is deleted already. Just let it go. what is the reason to break it out?

irfanurrehman:
1. So that the user can take informed decision. 
2. For useful debugging.
User might want to know which resource is/are missing while unjoin, in case they end up with half baked envs.
The default should ideally be unjoin with idempotency.
I am not too particular about this to be implemented necessarily as part of this PR. Idempotency in unjoin is more important as default behaviour, then having an option to choose it.

marun:
@irfanurrehman In the `join` case, the user may not want an existing resource to be overwritten so `--error-on-existing` is a logical addition.  I don't think the same applies to `unjoin`, where no such possibility of negative user impact exists if a resource has already been deleted.  Yes, adding an equivalent flag has the potential to be useful for debugging.  That's not exactly the mission of `kubefedctl`, and insufficient justification for adding yet another knob for for user to consider.

marun:
@xunpan Please rebase.  #876 removed all the previous cluster registry code.

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +189,88 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
...
-		klog.Errorf("Failed to delete Secret resource from namespace: %s for unjoin cluster: %s due to: %v", federationNamespace, unjoiningClusterName, err)
+	if apierrors.IsNotFound(err) {
+		klog.V(2).Infof("Secret \"%s/%s\" does not exist in the host cluster.", federationNamespace, fedCluster.Spec.SecretRef.Name)
+	} else if err != nil {
+		message := fmt.Sprintf("Failed to delete secret \"%s/%s\" for unjoin cluster %q",
+			federationNamespace, fedCluster.Spec.SecretRef.Name, unjoiningClusterName)
+		if !forceDeletion {
+			return errors.Wrap(err, message)
+		}
+		klog.V(2).Infof("%s.", message)
```
marun: The wrapped error should be logged, not just the message.  Also, why is the period not included directly in the message?  Same comment on: 

 - [L248](https://github.com/kubernetes-sigs/federation-v2/pull/859/files#diff-66d64392a61d13d392f12c775c53b62fR248)
 - [L367](https://github.com/kubernetes-sigs/federation-v2/pull/859/files#diff-66d64392a61d13d392f12c775c53b62fR367)
 - [L385](https://github.com/kubernetes-sigs/federation-v2/pull/859/files#diff-66d64392a61d13d392f12c775c53b62fR385)
 - [L404](https://github.com/kubernetes-sigs/federation-v2/pull/859/files#diff-66d64392a61d13d392f12c775c53b62fR404)
 - [L422](https://github.com/kubernetes-sigs/federation-v2/pull/859/files#diff-66d64392a61d13d392f12c775c53b62fR422)

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +189,88 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, federationNamespace,
...
-		return
+		return errors.Wrapf(err, "Failed to get federated cluster \"%s/%s\"", federationNamespace, unjoiningClusterName)
 	}
 
 	err = hostClientset.CoreV1().Secrets(federationNamespace).Delete(fedCluster.Spec.SecretRef.Name,
 		&metav1.DeleteOptions{})
-	if err != nil {
-		klog.Errorf("Failed to delete Secret resource from namespace: %s for unjoin cluster: %s due to: %v", federationNamespace, unjoiningClusterName, err)
+	if apierrors.IsNotFound(err) {
+		klog.V(2).Infof("Secret \"%s/%s\" does not exist in the host cluster.", federationNamespace, fedCluster.Spec.SecretRef.Name)
```
marun: (No action required) Consider defining a `QualifiedName` instance instead of requiring `\"%s/%s\"` in all the messages.  Same comment wherever that awkward formatting is used repeatedly.

pkg/kubefedctl/unjoin.go
xunpan: wrap uses `:` to seperate messages. So, extra `.` is not necessary. I remember related code in controller is already use such kind of convention though a little trivial.

marun: Apologies, I may have confused you.  If `forceDeletion == true`, `err` will not be logged at all.  I think the following would be preferable:

```golang
wrappedErr := errors.Wrapf(err, ...)
if !forceDeletion {
   return wrappedErr
}
klog.V(2).Infof("%v", wrappedErr)
```

xunpan: @marun done.

pkg/kubefedctl/unjoin.go
xunpan: got it.

marun:
Please rebase (the repo rename is complete).

pkg/kubefedctl/unjoin.go
```diff
@@ -295,76 +303,132 @@ func deleteFedNSFromUnjoinCluster(hostClientset, unjoiningClusterClientset kubec
 		return nil
 	}
 
-	klog.V(2).Infof("Deleting kubefed namespace %q from unjoining cluster %q", kubefedNamespace, unjoiningClusterName)
+	klog.V(2).Infof("Deleting federation namespace %q from unjoining cluster %q.", kubefedNamespace, unjoiningClusterName)
```
marun: `s/federation/kubefed/`

pkg/kubefedctl/unjoin.go
```diff
@@ -295,76 +303,132 @@ func deleteFedNSFromUnjoinCluster(hostClientset, unjoiningClusterClientset kubec
...
 	err = unjoiningClusterClientset.CoreV1().Namespaces().Delete(kubefedNamespace, &metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
-		klog.V(2).Infof("The kubefed namespace %q no longer exists in unjoining cluster %q", kubefedNamespace, unjoiningClusterName)
+		klog.V(2).Infof("The federation namespace %q no longer exists in unjoining cluster %q.", kubefedNamespace, unjoiningClusterName)
```
marun: `s/federation/kubefed/`

pkg/kubefedctl/unjoin.go
```diff
@@ -295,76 +303,132 @@ func deleteFedNSFromUnjoinCluster(hostClientset, unjoiningClusterClientset kubec
...
 		return nil
+	} else if err != nil {
+		return errors.Wrapf(err, "Could not delete federation namespace %q from unjoining cluster %q", kubefedNamespace, unjoiningClusterName)
```
marun: `s/federation/kubefed/`

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +188,94 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace, hos
...
-	kubefedNamespace, unjoiningClusterName string, dryRun bool) {
+func deleteFederatedClusterAndSecret(hostClientset kubeclient.Interface, client genericclient.Client,
+	kubefedNamespace, unjoiningClusterName string, forceDeletion, dryRun bool) error {
 	if dryRun {
-		return
+		return nil
 	}
 
-	klog.V(2).Infof("Deleting federated cluster resource from namespace: %s for unjoin cluster: %s",
+	klog.V(2).Infof("Deleting federated cluster resource from namespace %q for unjoin cluster %q",
```
marun: s/federated/kubefed/

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +188,94 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace, hos
...
 		kubefedNamespace, unjoiningClusterName)
 
 	fedCluster := &fedv1a1.KubefedCluster{}
 	err := client.Get(context.TODO(), fedCluster, kubefedNamespace, unjoiningClusterName)
 	if err != nil {
-		klog.Errorf("Failed to get KubefedCluster resource from namespace: %s for unjoin cluster: %s due to: %v", kubefedNamespace, unjoiningClusterName, err)
-		return
+		return errors.Wrapf(err, "Failed to get federated cluster \"%s/%s\"", kubefedNamespace, unjoiningClusterName)
```
marun: s/federated/kubefed/

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +188,94 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace, hos
...
 	} else {
-		klog.V(2).Infof("Deleted Secret resource from namespace: %s for unjoin cluster: %s", kubefedNamespace, unjoiningClusterName)
+		klog.V(2).Infof("Deleted secret \"%s/%s\" for unjoin cluster %q", kubefedNamespace, fedCluster.Spec.SecretRef.Name, unjoiningClusterName)
 	}
 
 	err = client.Delete(context.TODO(), fedCluster, fedCluster.Namespace, fedCluster.Name)
-	if err != nil {
-		klog.Errorf("Failed to delete KubefedCluster resource from namespace: %s for unjoin cluster: %s due to: %v", kubefedNamespace, unjoiningClusterName, err)
+	if apierrors.IsNotFound(err) {
+		klog.V(2).Infof("Federated cluster \"%s/%s\" does not exist in the host cluster.", fedCluster.Namespace, fedCluster.Name)
```
marun: s/Federated/KubeFed/

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +188,94 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace, hos
...
+	} else if err != nil {
+		wrappedErr := errors.Wrapf(err, "Failed to delete federated cluster \"%s/%s\" for unjoin cluster %q", fedCluster.Namespace, fedCluster.Name, unjoiningClusterName)
```
marun: s/federated/kubefed/

pkg/kubefedctl/unjoin.go
```diff
@@ -188,86 +188,94 @@ func UnjoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace, hos
...
+		if !forceDeletion {
+			return wrappedErr
+		}
+		klog.V(2).Infof("%v", wrappedErr)
 	} else {
-		klog.V(2).Infof("Deleted KubefedCluster resource from namespace: %s for unjoin cluster: %s", kubefedNamespace, unjoiningClusterName)
+		klog.V(2).Infof("Deleted federated cluster \"%s/%s\" for unjoin cluster %q.", fedCluster.Namespace, fedCluster.Name, unjoiningClusterName)
```
marun: s/federated/kubefed/

pkg/kubefedctl/unjoin.go
```diff
@@ -295,76 +303,132 @@ func deleteFedNSFromUnjoinCluster(hostClientset, unjoiningClusterClientset kubec
...
+	klog.V(2).Infof("Deleting federation namespace %q from unjoining cluster %q.", kubefedNamespace, unjoiningClusterName)
 	err = unjoiningClusterClientset.CoreV1().Namespaces().Delete(kubefedNamespace, &metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
-		klog.V(2).Infof("The kubefed namespace %q no longer exists in unjoining cluster %q", kubefedNamespace, unjoiningClusterName)
+		klog.V(2).Infof("The federation namespace %q no longer exists in unjoining cluster %q.", kubefedNamespace, unjoiningClusterName)
 		return nil
+	} else if err != nil {
+		return errors.Wrapf(err, "Could not delete federation namespace %q from unjoining cluster %q", kubefedNamespace, unjoiningClusterName)
+	} else {
+		klog.V(2).Infof("Deleted federation namespace %q from unjoining cluster %q.", kubefedNamespace, unjoiningClusterName)
```
marun: s/federation/kubefed/

xunpan:
@marun done

marun:
Thank you!  This was tricky to review (and presumably develop), so much repetition and the logic of the force deletion handling varied by function.  Hopefully we get a chance to cleanup at some point.

/approve
/lgtm

PR #861, opened 2019-05-08
Update changelog to reflact federate resource changes

irfanurrehman:
I realised that I did not update the `CHANGELOG` per PR, that I was supposed to do and @marun did the job of reflecting the updates in the release notes in the last 3 releases he made. Hearty thanks @marun for doing that.
Also not sure if its worth updating the details now; but raising this anyways to keep the `CHANGELOG` in sync with the release notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Unreleased
+- [#844](https://github.com/kubernetes-sigs/federation-v2/pull/844)
+   `kubefedctl federate` namespace with its content gets an option to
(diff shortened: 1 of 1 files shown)


marun:
@irfanurrehman I think the point of the changelog is more to help in creating release notes, so fyi I don't think there is a need for updating for already released version.  The process of creating the release notes is manual for now - it requires comparing the changelog with the commits since the previous release and then judicious editing for cohesiveness and clarity.

/lgtm

PR #862, opened 2019-05-08
Add kubefed validation webhook

font:
~~**NOTE:** This PR is not reviewable until ~~#857~~ #867 is merged.~~

This adds the scaffolding to enable kubefed webhook validation for core API types we plan to move to beta. Particularly the 2 APIs added for validation in this PR:
1. `FederatedTypeConfig`: currently only verifies target kind is set.
1. `FederatedCluster`: no validation at the moment.

This PR can be merged ~~(pending ~~#857~~ #867 merging)~~ as follow-on work/discussion is needed to implement the actual validation pieces.

@pmorie did all the work here. I've just mainly added scaffolding for `FederatedCluster` while cleaning up the history, fixing few chart and deployment issues, etc.

TODO in follow-up PRs:
- [ ] Implement validation pieces for core API types moving to beta
- [ ] Add support for using k8s CA instead of self-generated one
- [ ] Tests


marun:
Before WIP is removed, I'd like to see the reviewable changes squashed down to as few orthogonal commits as possible and highlighted for potential reviewers.  The amount of unreviewable content in this PR makes reviewing via 'Files Changed' all but impossible, and having to review changes commit by commit only works if the commits are few in number and easily identifiable as orthogonal.

font:
@marun Is this more manageable now that deps have been merged separately?

font:
I've rebased and this is now ready for review.

shashidharatd:
@font, is there any particular reason this PR not using the scaffolding provided by kubebuilder for webhooks? https://book.kubebuilder.io/beyond_basics/webhook_installer_generator.html

xunpan:
+1 with @shashidharatd. With kubebuilder, we do not need to maintain an independent webhook binary. It seems simpler.

pmorie:
> @font, is there any particular reason this PR not using the scaffolding provided by kubebuilder for webhooks? https://book.kubebuilder.io/beyond_basics/webhook_installer_generator.html

From what I was able to tell, the kubebuilder support for webhooks is very minimal and depends on you using a particular controller framework which we are not using. We can refactor to it in the future if we want to but ironically using the kubebuilder facility would require significantly more intrusive changes than the approach this PR takes.

>  we do not need to maintain an independent webhook binary. It seems simpler

We're not adding a new binary; the webhook server is integrated with hyperfed so there will not be a new binary present in the image, just another symlink to hyperfed.

charts/federation-v2/charts/webhook/templates/webhook.yaml
```diff
@@ -0,0 +1,58 @@
...
+    service:
+      namespace: {{ .Release.Namespace }}
+      name: kubefed-webhook-server
+      path: /federatedtypeconfigs
+    caBundle: {{ b64enc $ca.Cert }}
+  rules:
+  - operations:
+    - CREATE
+    apiGroups:
+    - project.openshift.io
```
irfanurrehman: This does not seem right..

charts/federation-v2/charts/webhook/templates/webhook.yaml
```diff
@@ -0,0 +1,58 @@
...
+    apiVersions:
+    - "*"
+    resources:
+    - namespaces
+  failurePolicy: Fail
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: openshift-namespace-reservation
```
irfanurrehman: Shouldn't this also be the same namespace as that of the rest of the `webhook` resources.

charts/federation-v2/charts/webhook/templates/webhook.yaml
font: It's not and was updated in a subsequent commit.

irfanurrehman: Oh, I was going commit wise; I would try to look at the whole diff now, which scares me a little.. :-)

charts/federation-v2/charts/webhook/templates/webhook.yaml
font: Yes, and it's updated in a subsequent commit.

pkg/controller/webhook/federatedcluster_webhook.go
```diff
@@ -0,0 +1,126 @@
...
+
+	lock        sync.RWMutex
+	initialized bool
+}
+
+func (a *FederatedClusterValidationHook) ValidatingResource() (plural schema.GroupVersionResource, singular string) {
+	return schema.GroupVersionResource{
+			Group:    "admission.core.federation.k8s.io",
+			Version:  "v1alpha1",
+			Resource: "federatedclusters",
```
irfanurrehman: Some of the strings, for example the resource plural name here could be made as constants for usage everywhere.

pkg/controller/webhook/federatedtypeconfig_webhook.go
```diff
@@ -0,0 +1,126 @@
...
+func (a *FederatedTypeConfigValidationHook) ValidatingResource() (plural schema.GroupVersionResource, singular string) {
+	return schema.GroupVersionResource{
+			Group:    "admission.core.federation.k8s.io",
+			Version:  "v1alpha1",
+			Resource: "federatedtypeconfigs",
+		},
+		"federatedtypeconfig"
+}
+
+func (a *FederatedTypeConfigValidationHook) Validate(admissionSpec *admissionv1beta1.AdmissionRequest) *admissionv1beta1.AdmissionResponse {
```
irfanurrehman: The code here and that for `FederatedClusterValidationHook` looks quite similar. Recommend to move it out to a common code which can take respective `object` (`admittingObject` in the code) and gvk strings as arguments.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/spf13/cobra"
+)
+
+func NewWebhookCommand(stopCh <-chan struct{}) *cobra.Command {
+	admissionHooks := []apiserver.AdmissionHook{
+		&webhook.FederatedTypeConfigValidationHook{},
+		&webhook.FederatedClusterValidationHook{},
+	}
+
+	// done to avoid cannot use admissionHooks (type []AdmissionHook) as type []apiserver.AdmissionHook in argument to "github.com/openshift/kubernetes-namespace-reservation/pkg/genericadmissionserver/cmd/server".NewCommandStartAdmissionServer
```
irfanurrehman: Do not understand this.
The variables at line 29 and that at line 35 are of same type. Also `NewCommandStartAdmissionServer` seems to take the same type. Am I missing something?

irfanurrehman:
@font @pmorie I am not sure how is this tested. Is there a plan to introduce testing for these validations at some point?

pkg/controller/webhook/federatedtypeconfig_webhook.go
pmorie: > I am assuming atleast this portion of code for rest of the types needing validation will also be same.

Can we do that once we have things fully implemented for other types so that we can avoid pre-factoring?

irfanurrehman: Sure.

pmorie:
> @font @pmorie I am not sure how is this tested. Is there a plan to introduce testing for these validations at some point?

Generally validations are tested in unit tests for the validation methods; from an e2e standpoint they're exercised every time an object is created so they are naturally exercised by e2e scenarios. I think it would be fine to add a validation test for the initial validation in this PR in a follow, since the travis job has passed.

WDYT @irfanurrehman

pkg/webhook/webhook.go
marun: This seems entirely unnecessary to me.

charts/federation-v2/charts/webhook/templates/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  namespace: "{{ .Release.Namespace }}"
+  name: kubefed-webhook-server
+  labels:
+    server: "true"
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      server: "true"
```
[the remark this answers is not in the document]
pmorie: I suspect the label is representative of the example it was created from. I'm in favor of using more specific labels.

charts/federation-v2/charts/webhook/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
pmorie: I would be fine to fold it into the existing chart, but it's still currently factored as a 'controller-manager' subchart. Seems like this would probably be better to do in a follow-up to me, wdyt?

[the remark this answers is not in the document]
marun: I'm fine with that so long as it gets done before the next release.

pkg/controller/webhook/federatedcluster_webhook.go
font: Yep there's a lot of room for refactoring of this code after the base functionality is working.

charts/federation-v2/charts/webhook/templates/deployment.yaml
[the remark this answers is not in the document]
font: I have opted for `kubefed-admission-webhook: "true"`.

charts/federation-v2/charts/webhook/Chart.yaml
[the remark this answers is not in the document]
font: I've now moved the chart to be within the `controller-manager` subchart. I think in the future we should just have one `kubefed` chart.

cmd/hyperfed/main.go
```diff
@@ -79,9 +82,13 @@ func NewHyperFedCommand() (*cobra.Command, []func() *cobra.Command) {
 	controller := func() *cobra.Command { return app.NewControllerManagerCommand() }
 	kubefedctlCmd := func() *cobra.Command { return kubefedctl.NewKubeFedCtlCommand(os.Stdout) }
 
+	stopCh := genericapiserver.SetupSignalHandler()
```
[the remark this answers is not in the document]
font: Since each of these anonymous functions are called by hyperfed to establish the subcommand, I've added the `SetupSignalHandler` to hyperfed and passed it into the controller and webhook command functions.

font:
@marun pending CI passing this should be ready to re-review.

Makefile
```diff
@@ -45,6 +45,7 @@ BUILD_IMAGE ?= golang:1.11.2
 HYPERFED_TARGET = bin/hyperfed
 CONTROLLER_TARGET = bin/controller-manager
 KUBEFEDCTL_TARGET = bin/kubefedctl
+WEBHOOK_TARGET = bin/webhook
```
marun: (No action required) I think unique binary names would be preferable (e.g. `kubefed-controller-manager`, `kubefed-webhook`).

charts/kubefed/charts/controllermanager/templates/roles.yaml
```diff
@@ -81,3 +81,34 @@ rules:
   - secrets
   verbs:
   - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    api: federation
```
marun: s/federation/kubefed/

cmd/hyperfed/main.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// A binary that can morph into all of the other kubefed binaries. You can
+// A binary that can morph into all of the other federation-v2 binaries. You can
```
marun: revert

scripts/deploy-federation.sh
```diff
@@ -18,9 +18,9 @@
 # the README - to the current kubectl context.  It also joins the
 # hosting cluster as a member of the federation.
 #
-# WARNING: The service account for the kubefed namespace will be
```
marun: revert

scripts/deploy-federation.sh
```diff
@@ -18,9 +18,9 @@
...
+# WARNING: The service account for the federation namespace will be
 # granted the cluster-admin role.  Until more restrictive permissions
-# are used, access to the kubefed namespace should be restricted to
```
marun: revert

scripts/deploy-federation.sh
```diff
@@ -39,7 +39,7 @@
 #
 # To redeploy federation from scratch, prefix the deploy invocation with the deletion script:
 #
-#   # WARNING: The deletion script will remove federation data
+#   # WARNING: The deletion script will remove federation and cluster registry data
```
marun: revert

charts/kubefed/charts/controllermanager/templates/roles.yaml
font: Fixed. But the others not changed in this PR will also need to be fixed in another PR.

cmd/hyperfed/main.go
font: Fixed. That's strange as this wasn't a conflict.

scripts/deploy-federation.sh
font: strange...

scripts/deploy-federation.sh
font: even more strange

Makefile
font: I think this should eventually be `admission-webhook`. I was going to do this in this PR, but then decided against it because it would require renaming several files and their import paths. The main thing is that the user visible name is `admission-webhook`.

marun:
LGTM

Issue #851, opened 2019-05-06
Fail to run ./scripts/create-clusters.sh


felixhuangau:
After download the binaries and I run `./scripts/create-clusters.sh`, it failed and showed `
./scripts/create-clusters.sh: line 98: kind: command not found`.

How can I fix that?

PR #866, opened 2019-05-09
Fix dangling glog to klog

font:
#791 was merged after #851 was created so this last dangling reference was missed.

### DIFF
--- a/pkg/controller/sync/dispatch/checkunmanaged.go
+++ b/pkg/controller/sync/dispatch/checkunmanaged.go
@@ -17,13 +17,13 @@ limitations under the License.
 package dispatch
 
 import (
-	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


marun:
@font Would it make sense to add a ci check to prevent glog from slipping back in?  It will always be possible to import without error due to vendored dependencies still using it.

`git grep glog | grep -v vendor | grep -v Gopkg.lock`

font:
> @font Would it make sense to add a ci check to prevent glog from slipping back in? It will always be possible to import without error due to vendored dependencies still using it.
> 
> `git grep glog | grep -v vendor | grep -v Gopkg.lock`

It would except #862 completely removes `glog` from being vendored after updating dependencies to accommodate the webhook.

PR #871, opened 2019-05-10
ignore related file for vim and scope development environment

xunpan:
This helps vim users in developing federation project without committing temporary files in careless. 

Yes. I use VIM :-).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@
 /images/federation-v2/hyperfed
 # temporarily added below file to ignore list to avoid someone pushing this file in a pr by accident.
 /images/federation-v2/controller-manager
+
+# ignore vim generated temporary file
+*.swp
+
+# ignore cscope related files
+cscope.*


shashidharatd:
@xunpan, one can ignore files which should not be committed in their local environment by just some simple setting in their local environment and does not warrant change in the project repo `.gitignore` file. refer https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore .

xunpan:
@shashidharatd 
The ignored files seems not used by the project for any purpose. And the changes in `.gitignore` is simple and clear. If we ignore it in the project, it is benefits  people who uses VIM or cscope. I think it is better to make it by default instead of local settings.

marun:
@xunpan I'm with @shashidharatd on this - the appropriate place to ignore editor or ide-specific files is your local global ignore config.  That way it can apply to any repo you work with without having to update the the repo's ignore config.

xunpan:
If you guys insist on that, I can close this ticket.

BTW, I'm still thinking that `.gitignore` is a tool for developer's convenient. Editor files part is one of the use case :-P. However, it is not the urgent and mandatory things need to handle.

PR #873, opened 2019-05-10
Update Makefile to use TRAVIS_BRANCH env var

font:
This should fix travis `master` builds in order to successfully push the `canary` tag.


marun:
LGTM

PR #867, opened 2019-05-09
Add validation webhook dependencies

font:
**NOTE:** This PR is not reviewable until #866 is merged.

This adds the webhook dependencies used by #862.

PR #876, opened 2019-05-10
Replace cluster registry with APIEndpoint field on FederatedCluster

### LINKED ISSUE #869: Storing `Cluster` resources outside of the kubefed namespace is a potential security risk
As per an observation by @deads2k, storing a `Cluster` resource outside of the kubefed namespace (e.g. in the `kube-multicluster-public` namespace) represents a potential security risk. A user with the appropriate permissions for `Cluster` resources in the cluster registry namespace could change the api endpoint of a `Cluster` to a malicious endpoint that could collect the kubefed credentials for that cluster. His suggestion was that the risk could be avoided if `Cluster` resources were always located in the kubefed system namespace, since access would be localized to the kubefed control plane.

If this security risk outweighs the benefits of sharing a cluster registry, though, it's not clear why kubefed needs to rely on the `Cluster` resource at all.  The fields kubefed cares about (i.e. api endpoints and certs) could be added back to the `FederatedCluster` resource, and use of the registry namespace and cluster registry `Cluster` resources discontinued.

/kind bug

marun:
Note that this PR does not update the vendoring for cluster registry to avoid conflicting with #867.  Once #862 merges it will be safe to prune the deps.
 
Fixes #869


pkg/apis/core/v1alpha1/federatedcluster_types.go
```diff
@@ -25,35 +25,22 @@ import (
 
 // FederatedClusterSpec defines the desired state of FederatedCluster
 type FederatedClusterSpec struct {
-	// Name of the cluster registry Cluster resource from which to source api
-	// endpoints.
-	// TODO(marun) should this go away in favor of a 1:1 mapping?
-	ClusterRef LocalClusterReference `json:"clusterRef,omitempty"`
+	// The API endpoint of the member cluster. This can be a hostname,
+	// hostname:port, IP or IP:port.
+	APIEndpoint string `json:"apiEndpoint"`
```
pmorie: While I'm curious to know why the original calculation involving the CIDR was present, I think this is sufficient for now. We can re-add it in the future if there's a need.

pmorie: LGTM otherwise.

shashidharatd:
This LGTM.
@marun, PR needs rebase. Thanks!

pkg/apis/core/v1alpha1/federatedcluster_types.go
pmorie: Seems like we need to add an optional CA bundle here - do we not?

shashidharatd: CA is probably embedded in the secret which is referenced by this object (just few lines below).

pmorie: > CA is probably embedded in the secret which is referenced by this object (just few lines below).

Yeah, I was thinking that after I made the above comment. So, I think that implies that to join a cluster that requires you to configure a CA bundle that the kubeconfig file that you use to join the cluster has to contain the CA bundle information in the relevant context - is that accurate?

marun:
Rebased

pkg/apis/core/v1alpha1/federatedcluster_types.go
marun: That is correct - the CA bundle is stored in the secret.  As per the follow-on PR that cleans up secret handling, the code that builds the kubeconfig for accessing the member cluster sources values for `token` and `ca.crt` keys from the secret.  I'm not sure that the CA bundle needs to be stored in the secret, though, given that it's effectively a public key.  That's just the way it's always been done since fedv1.

PR #877, opened 2019-05-10
Make FederatedCluster.SecretRef mandatory

### LINKED ISSUE #875: Secret ref should be mandatory for FederatedCluster
Given that we lack a use-case that would require insecure access to a member cluster, I think the [secret ref](https://github.com/kubernetes-sigs/federation-v2/blob/master/pkg/apis/core/v1alpha1/federatedcluster_types.go#L44) in `FederatedCluster` should be made mandatory.

/kind bug

marun:
It is no longer desirable to support insecure member clusters even for testing. As part of this change, unused support for legacy authentication (previously used by managed fixture) was also removed.

Fixes #875


marun:
Rebased

marun:
Recent conversations have led me to believe that the ca bundle should be optional rather than mandatory.  Should that change be made as part of this PR, since the secret handling is rewritten anyway?

marun:
The latest push updates the comment on `SecretRef` to match the current implementation.  I will be tackling a switch to storing the ca bundle as an optional field.

shashidharatd:
/lgtm
Thanks !

PR #880, opened 2019-05-11
Set kubefed-namespace via the config mechanism for in-cluster deployment

### LINKED ISSUE #795: use namespace from system in
**What happened**:

It would be desirable to set federation-namespace via the config mechanism if scope == Namespace.

Follow up action for https://github.com/kubernetes-sigs/federation-v2/pull/771

**Environment**:
- Federation version
- Others

/kind bug

sohankunkerkar:
Fixes #795 
/assign @marun @xunpan

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/deployments.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/deployments.yaml
@@ -20,9 +20,7 @@ spec:
         runAsUser: 1001
       serviceAccountName: kubefed-controller
       containers:
-      - args:
-        - --kubefed-namespace=$(KUBEFED_NAMESPACE)
-        command:
+      - command:
         - /hyperfed/controller-manager
         image: "{{ .Values.repository }}/{{ .Values.image }}:{{ .Values.tag }}"
         imagePullPolicy: "{{ .Values.imagePullPolicy }}"
@@ -38,11 +36,6 @@ spec:
 {{- if .Values.resources }}
 {{ toYaml .Values.resources | indent 12 }}
 {{- end }}
-        env:
-        - name: KUBEFED_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
       terminationGracePeriodSeconds: 10
 ---
 apiVersion: apps/v1
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -21,8 +21,10 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -105,6 +107,22 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
 	opts.Config.KubeConfig, err = clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
 	if err != nil {
 		panic(err)
+	}
+
+	runningInCluster := len(masterURL) == 0 && len(kubeconfig) == 0
+	if runningInCluster && len(opts.Config.KubeFedNamespace) == 0 {
+		// For in-cluster deployment set the namespace associated
+		// with the service account token
+		data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		if err != nil {
+			klog.Fatalf("An error occurred while attempting to discover the KubeFed namespace from the service account: %v", err)
+		}
+		opts.Config.KubeFedNamespace = strings.TrimSpace(string(data))
+	}
+
+	// Validate if a kubefed-namespace is configured
+	if len(opts.Config.KubeFedNamespace) == 0 {
+		klog.Fatalf("The KubeFed namespace must be specified via --kubefed-namespace")
 	}
 
 	setOptionsByKubeFedConfig(opts)
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -36,7 +36,7 @@ type Options struct {
 
 // AddFlags adds flags to fs and binds them to options.
 func (o *Options) AddFlags(fs *pflag.FlagSet) {
-	fs.StringVar(&o.Config.KubeFedNamespace, "kubefed-namespace", util.DefaultKubeFedSystemNamespace, "The namespace the KubeFed control plane is deployed in.")
+	fs.StringVar(&o.Config.KubeFedNamespace, "kubefed-namespace", "", "The namespace the KubeFed control plane is deployed in.")
 }
 
 func NewOptions() *Options {


cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +117,17 @@ func Run(opts *options.Options) error {
 		klog.Fatalf("Invalid Feature Gate: %v", err)
 	}
 
+	// for in-cluster deployment set the namespace associated
```
marun: Discovery of the kubefed namespace should only be attempted if it was not supplied explicitly and the configuration indicates that the controller manager is running inside a cluster i.e. the following conditions:

 - `len(opts.Config.FederationNamespace) == 0`
 - `len(masterURL) == 0`
 - `len(kubeconfig) == 0`

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +117,17 @@ func Run(opts *options.Options) error {
...
+	// with the service account token
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
```
marun: We already have a valid config at this point, and the logic for discovering the namespace of the service account doesn't depend on the config at all.  It's just [5 lines of go code](https://github.com/kubernetes-sigs/federation-v2/blob/master/vendor/k8s.io/client-go/tools/clientcmd/client_config.go#L521).  I'd suggest creating a local function with this code or just adding it inline:

```golang
// Fall back to the namespace associated with the service account token, if available
if data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
  if ns := strings.TrimSpace(string(data)); len(ns) > 0 {
    return ns, false, nil
  }
}
```

If the federation namespace is not specified and cannot be discovered via this method, the controller manager should log a fatal error indicating that the federation namespace needs to be manually set via `--federation-config` if the controller manager is not running inside a cluster.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Thanks, @marun. I will update this PR to check for this condition before setting up the value for the `federation namespace`.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Ack. I will make the necessary changes in the code as per your suggestions.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -95,6 +97,17 @@ member clusters and does the necessary reconciliation`,
 	return cmd
 }
 
+// for in-cluster deployment set the namespace associated
+// with the service account token
+func defaultNamespace() (string, error) {
+	if data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
```
marun: `err` needs to be checked and returned if non-nil.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +130,14 @@ func Run(opts *options.Options) error {
 		klog.Fatalf("Invalid Feature Gate: %v", err)
 	}
 
+	if len(masterURL) == 0 && len(kubeconfig) == 0 && len(opts.Config.FederationNamespace) == 0 {
+		fedName, err := defaultNamespace()
```
marun: `s/fedName/fedNamespace/`

Please use clear and unambiguous names.  Avoid brief names that are insufficiently descriptive.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +130,14 @@ func Run(opts *options.Options) error {
...
+		if err != nil {
+			klog.Fatalf("The controller-manager is not running inside the cluster. Use --federation-namespace flag to manually set the fedeation-namespace")
```
marun: The error should be logged and the namespace flag should be checked outside of this conditional, i.e.

```golang
if len(masterURL)... {
  fedNamespace, err := ...
  if err != nil {
    // Include the err in the log message to aid in troubleshooting
    klog.Fatalf("An error occurred while attempting to discover the federation namespace from the service account: %v", err)
  } 
}

// Regardless of whether in-cluster discovery was attempted, validate that a federation namespace is configured 
if len(opts.Config.FederationNamespace) == 0 {
  klog.Fatalf("The federation namespace must be specified via --federation-namespace")
} 
```

Errors should almost always be logged unless you have a good reason not to.  It's entirely possible the error will convey information vital to a troubleshooting exercise.

sohankunkerkar: ack

cmd/controller-manager/app/controller-manager.go
```diff
@@ -56,7 +58,7 @@ import (
 )
 
 var (
-	kubeconfig, federationConfig, masterURL string
+	kubeconfig, federationConfig, masterURL, fedNamespace string
```
marun: Variable definitions should have the minimal scope required. This variable is package global, but is only used within the scope of a conditional in a function.  As per subsequent comments, though, this variable is not  necessary and can be removed.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +119,25 @@ func Run(opts *options.Options) error {
 		klog.Fatalf("Invalid Feature Gate: %v", err)
 	}
 
+	// validate if a federation namespace is configured
+	if len(opts.Config.FederationNamespace) == 0 {
```
marun: This block needs to appear after the block that tries to discover the namespace to ensure that discovery can be attempted if the namespace is not otherwise provided.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +119,25 @@ func Run(opts *options.Options) error {
...
+		klog.Fatalf("The federation namespace must be specified via --federation-namespace")
+	}
+
+	if len(masterURL) == 0 && len(kubeconfig) == 0 && len(opts.Config.FederationNamespace) == 0 {
+		// for in-cluster deployment set the namespace associated
+		// with the service account token
+		data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		if err != nil {
+			klog.Fatalf("An error occurred while attempting to discover the federation namespace from the service account: %v", err)
+		} else {
```
marun: This else block is unnecessary since `klog.Fatalf` will exit the process.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +119,25 @@ func Run(opts *options.Options) error {
...
+			if ns := strings.TrimSpace(string(data)); len(ns) > 0 {
```
marun: There is no need for this conditional or intermediate variable.  The following will suffice:

```golang
opts.Config.FederationNamespace = strings.TrimSpace(string(data))
```

An empty string value is fine.  It won't change anything because a condition of this block is that the federation namespace is not set.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -117,6 +119,25 @@ func Run(opts *options.Options) error {
 		klog.Fatalf("Invalid Feature Gate: %v", err)
 	}
 
+	// validate if a federation namespace is configured
+	if len(opts.Config.FederationNamespace) == 0 {
+		klog.Fatalf("The federation namespace must be specified via --federation-namespace")
+	}
+
+	if len(masterURL) == 0 && len(kubeconfig) == 0 && len(opts.Config.FederationNamespace) == 0 {
```
marun: (No action required) It's often helpful to use a boolean variable to document your intent with a conditional, e.g.

```golang
runningInCluster := len(masterURL) == 0 && len(kubeconfig) == 0
if runningInCluster && len(opts.Config.FederationNamespace) == 0 {
 ...
}
```

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Yeah, I think that makes sense. I will update this PR accordingly.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Done!

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Yeah, I was concerned about the empty string here but as you mentioned it won't affect the overall process for setting the federation namespace.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: I think I like your rationale for clubbing two conditions here and mentioning the exact intent behind it. I will accommodate this change in this PR.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Fixed.

marun:
LGTM

sohankunkerkar:
@marun ready for your review.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -107,12 +109,28 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
...
+		data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		if err != nil {
+			klog.Fatalf("An error occurred while attempting to discover the kubefed namespace from the service account: %v", err)
+		}
+		opts.Config.KubeFedNamespace = strings.TrimSpace(string(data))
+	}
+
+	setOptionsByKubeFedConfig(opts)
+
+	// validate if a kubefed namespace is configured
```
marun: Please follow the convention of capitalizing the first word in a comment e.g. `Validate if a kubefed namespace is configured`

cmd/controller-manager/app/controller-manager.go
```diff
@@ -107,12 +109,28 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
...
+	if len(opts.Config.KubeFedNamespace) == 0 {
```
marun: This check needs to occur before `setOptionsByKubeFedConfig` is called to ensure that the check's error message will appear in the logs instead of a complaint that a namespace must be set from an attempt to load a KubeFedConfig resource.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -107,12 +109,28 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
 		panic(err)
 	}
 
-	setOptionsByKubeFedConfig(opts)
-
 	if err := utilfeature.DefaultFeatureGate.SetFromMap(opts.FeatureGates); err != nil {
```
marun: This check needs to occur after `setOptionsByKubeFedConfig` is called because it depends on `opts.FeatureGates` being set by the call.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Fixed

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Fixed

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Ack

sohankunkerkar:
@marun your comments have been addressed. PTAL

charts/kubefed/charts/controllermanager/templates/deployments.yaml
```diff
@@ -20,9 +20,7 @@ spec:
         runAsUser: 1001
       serviceAccountName: kubefed-controller
       containers:
-      - args:
-        - --kubefed-namespace=$(KUBEFED_NAMESPACE)
```
marun: Please remove the now-unused env entry for KUBEFED_NAMESPACE (L39-L43).

sohankunkerkar: Fixed

sohankunkerkar:
@marun PTAL :)

pkg/controller/util/cluster_util.go
```diff
@@ -32,10 +32,9 @@ import (
 
 const (
 	DefaultKubeFedSystemNamespace = "kube-federation-system"
-
-	KubeAPIQPS   = 20.0
-	KubeAPIBurst = 30
-	TokenKey     = "token"
+	KubeAPIQPS                    = 20.0
+	KubeAPIBurst                  = 30
+	TokenKey                      = "token"
```
marun: Please revert this unrelated change.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -107,6 +109,22 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
...
+		data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		if err != nil {
+			klog.Fatalf("An error occurred while attempting to discover the kubefed namespace from the service account: %v", err)
+		}
+		opts.Config.KubeFedNamespace = strings.TrimSpace(string(data))
+	}
+
+	// Validate if a kubefed-namespace is configured
+	if len(opts.Config.KubeFedNamespace) == 0 {
+		klog.Fatalf("The kubefed namespace must be specified via --kubefed-namespace")
```
marun: nit: Please prefer KubeFed in output seen by the user.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -107,6 +109,22 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
 		panic(err)
 	}
 
+	runningInCluster := len(masterURL) == 0 && len(kubeconfig) == 0
+	if runningInCluster && len(opts.Config.KubeFedNamespace) == 0 {
+		// For in-cluster deployment set the namespace associated
+		// with the service account token
+		data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		if err != nil {
+			klog.Fatalf("An error occurred while attempting to discover the kubefed namespace from the service account: %v", err)
```
marun: nit: Please prefer KubeFed in output seen by the user.

charts/kubefed/charts/controllermanager/templates/deployments.yaml
```diff
@@ -38,11 +36,6 @@ spec:
 {{- if .Values.resources }}
 {{ toYaml .Values.resources | indent 12 }}
 {{- end }}
-        env:
```
marun: 👍

marun:
A few more nits, sorry for not catching earlier!

pkg/controller/util/cluster_util.go
sohankunkerkar: This is related to the `go-fmt` issue. Let me paste the code as-is.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Done.

cmd/controller-manager/app/controller-manager.go
sohankunkerkar: Done

sohankunkerkar:
@marun All the comments have been addressed.

marun:
Thank you!

/approve

PR #882, opened 2019-05-12
script for e2e use case of a simple nginx application

### LINKED ISSUE #104: Basic walkthrough and associated test script

xunpan:
Fixes #104 

Currently, service IP should be get from `minikube` command line. So, it works for `minikube` case and can be modified to work for other cases.

### DIFF
--- a/example/sample1/configmap.yaml
+++ b/example/sample1/configmap.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: web-file
+data:
+  index.html: "Hello from Kubernetes Cluster Federation!"
+
--- a/scripts/deploy-federated-nginx.sh
+++ b/scripts/deploy-federated-nginx.sh
@@ -1,0 +1,105 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script has an end to end steps in deploying a federated deployment
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+
+TEST_NS="test-namespace"
+
+# Prerequisite:
+# - a cluster scoped federation deployment with `cluster1` and
+#   `cluster2` joined as member clusters
+# - kubectl config use-context cluster1
+#
+echo "Validating KubeFed walkthrough."
+
+
+# Create kubernetes resources
+echo
+echo "Creating a test namespace ${TEST_NS}."
+kubectl create ns ${TEST_NS}
+
+echo
+echo "Creating a ngnix application resources in namespace ${TEST_NS}."
+kubectl apply -n ${TEST_NS} -f example/sample1/configmap.yaml
+kubectl apply -n ${TEST_NS} -f example/sample1/deployment.yaml
+kubectl apply -n ${TEST_NS} -f example/sample1/service.yaml
+
+
+# Federate kubernetes resources to member clusters
+echo
+echo "Federating resources in namespace ${TEST_NS} to member clusters."
+kubefedctl federate ns ${TEST_NS} --contents --skip-api-resources 'pods,secrets,serviceaccount,replicaset'
+
+echo
+echo "Checking status of federated resources."
+util::wait-for-condition 'federated deployment status updated in cluster1' \
+    "(kubectl get federateddeployment nginx -n ${TEST_NS} -o jsonpath='{.status.clusters[*].name}' | grep '^cluster1') &> /dev/null" 120
+util::wait-for-condition 'federated deployment status updated in cluster2' \
+    "(kubectl get federateddeployment nginx -n ${TEST_NS} -o jsonpath='{.status.clusters[*].name}' | grep '^cluster2') &> /dev/null" 120
+
+echo
+echo "Querying web server of deployment nginx application from member clusters."
+IP1=$(kubectl get node -o jsonpath="{.items[0].status.addresses[0].address}" --context cluster1)
+IP2=$(kubectl get node -o jsonpath="{.items[0].status.addresses[0].address}" --context cluster2)
+
+util::wait-for-condition 'nodePort ok in cluster1' \
+    "kubectl --namespace ${TEST_NS} --context cluster1 get service -o jsonpath='{.items[0].spec.ports[0].nodePort}' &> /dev/null" 30
+util::wait-for-condition 'nodePort ok in cluster2' \
+    "kubectl --namespace ${TEST_NS} --context cluster2 get service -o jsonpath='{.items[0].spec.ports[0].nodePort}' &> /dev/null" 30
+PORT1=$(kubectl get service -n ${TEST_NS} --context cluster1 -o jsonpath="{.items[0].spec.ports[0].nodePort}")
+PORT2=$(kubectl get service -n ${TEST_NS} --context cluster2 -o jsonpath="{.items[0].spec.ports[0].nodePort}")
+
+util::wait-for-condition 'nginx web server ok in cluster1' "curl ${IP1}:${PORT1} &> /dev/null" 120
+util::wait-for-condition 'nginx web server ok in cluster2' "curl ${IP2}:${PORT2} &> /dev/null" 120
+echo "cluster1: $(curl -s ${IP1}:${PORT1})"
+echo "cluster2: $(curl -s ${IP2}:${PORT2})"
+
+
+# Modify federated resources to update kubernetes resources in member clusters
+echo
+echo "Changing index.html in federated configmap."
+kubectl patch federatedconfigmap web-file -n ${TEST_NS} --type=merge -p '{"spec": {"template": {"data": {"index.html": "Hello from KubeFed!"}}}}'
+
+echo
+echo "Querying web server of deployed nginx application from member clusters."
+util::wait-for-condition 'web content changed in cluster1' "(curl ${IP1}:${PORT1} | grep '^Hello from KubeFed') &> /dev/null" 120
+util::wait-for-condition 'web content changed in cluster2' "(curl ${IP2}:${PORT2} | grep '^Hello from KubeFed') &> /dev/null" 120
+echo "cluster1: $(curl -s ${IP1}:${PORT1})"
+echo "cluster2: $(curl -s ${IP2}:${PORT2})"
+
+echo
+echo "Updating override of federated deployment nginx to increase 'replicas' to 2 in cluster2."
+kubectl patch federateddeployment nginx -n ${TEST_NS} --type=merge --patch \
+    '{"spec" : {"overrides": [{"clusterName" : "cluster2", "clusterOverrides": [{"path": "spec.replicas", "value" : 2}]}]}}'
+echo "Checking 'replicas' of deployment nginx in cluster2."
+util::wait-for-condition 'replicas updated in cluster2' \
+    "(kubectl get deployment nginx -n ${TEST_NS} --context cluster2 -o jsonpath='{.status.replicas}' | grep "^2$") &> /dev/null" 120
+
+echo
+echo "Updating placement to include only 'cluster2' so that the deployment will be removed from 'cluster1'"
+kubectl patch federateddeployment nginx -n ${TEST_NS} --type=merge --patch '{"spec": {"placement": {"clusters": [{"name": "cluster2"}]}}}'
+echo "Checking status of federated deployment nginx."
+util::wait-for-condition 'federated deployment status updated' \
+    "(kubectl get federateddeployment nginx -n ${TEST_NS} -o jsonpath='{.status.clusters[*].name}' | grep "^cluster2$") &> /dev/null" 120
+
+echo
+echo "Successfully completed KubeFed walkthrough."
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -149,6 +149,10 @@ echo "Deploying cluster-scoped kubefed"
 echo "Running e2e tests against cluster-scoped kubefed"
 run-e2e-tests
 
+echo "Validating KubeFed walkthrough"
+./scripts/deploy-federated-nginx.sh
+kubectl delete ns test-namespace
+
 echo "Scaling down cluster-scoped controller manager"
 kubectl scale deployments kubefed-controller-manager -n kube-federation-system --replicas=0
 
--- a/example/sample1/deployment.yaml
+++ b/example/sample1/deployment.yaml
@@ -1,0 +1,26 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginx
+  labels:
+    app: nginx
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: nginx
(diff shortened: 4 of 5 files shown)


example/e2e/deploy-federated-deployment.sh
```diff
@@ -0,0 +1,125 @@
...
+set -o nounset
+set -o pipefail
+
+TEST_NS="test-namespace"
+KUBEFED_NS="kube-federation-system"
+
+# Prerequisite: a cluster scoped federation deployment with `cluster` and
+# `cluster2` joined as member clusters
+#
+# kubectl config use-context cluster1
```
marun: Is this necessary?

xunpan:
@marun

 >   update script to consume example yaml (updated to deploy non-federated resources) from disk rather than defining inline
I'd like to share my original thought. I tried to make the script independent from CWD of the script execution. That is why I make yaml configuration in script. If this is not an issue, I'm fine to create yaml file for related resources.

>  add an example that exercises an override (as per my comment on #104)
I guess it is not for federation override. I think patch for placement is similar to patch for override.
If it is for the `nginx` example configuration, I will do it. Though not sure why this is necessary in this example.

>  update userguide to reflect the steps executed in the script
I will update guide to make this example and its steps described.

example/e2e/deploy-federated-deployment.sh
```diff
@@ -0,0 +1,125 @@
...
+  type: NodePort
+EOF
+
+if [[ $? != 0 ]]; then
+    echo "Failed creating a ngnix service."
+    exit 1
+fi 
+
+echo "Federating resources in namespace ${TEST_NS} to member clusters."
+kubefedctl federate ns --federation-namespace ${KUBEFED_NS} --skip-api-resources 'pods,secrets,serviceaccount,replicaset' --contents ${TEST_NS}
```
irfanurrehman: For readability I would write it as 
```
kubefedctl federate ns ${TEST_NS}  --contents --federation-namespace ${KUBEFED_NS} --skip-api-resources 'pods,secrets,serviceaccount,replicaset'
```

irfanurrehman: Also the script defines `${KUBEFED_NS}` as a constant with value of `kube-federation-system` which is also the default. If the value is always going to be default, specifying the flag `--federation-namespace` is not necessary.

example/e2e/deploy-federated-deployment.sh
```diff
@@ -0,0 +1,125 @@
...
+# limitations under the License.
+
+# This script has an end to end steps in deploying a federated deployment
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+TEST_NS="test-namespace"
+KUBEFED_NS="kube-federation-system"
```
irfanurrehman: Although not necessary but if you are using a variable, you can make it user configurable too.
`KUBEFED_NS="${KUBEFED_NS:-kube-federation-system}"`

example/e2e/deploy-federated-deployment.sh
```diff
@@ -0,0 +1,125 @@
...
+echo "Starting creating a ngnix application in federation cluster."
+
+echo "Creating a test namespace ${TEST_NS}."
+kubectl create ns ${TEST_NS}
+if [[ $? != 0 ]]; then
+    echo "Failed creating ${TEST_NS}."
+    exit 1
+fi 
+
+echo "Creating a ngnix deployment in namespace ${TEST_NS}."
```
irfanurrehman: Did you try `kubectl run nginx --image=nginx --expose=true` or some similar `kubectl run` options to create the deployment and service?

marun: The goal is to have an nginx deployment that serves a message out of a configmap, so that an override can be demonstrated.  This precludes using something like you are suggesting.

example/e2e/deploy-federated-deployment.sh
marun: The goal is not to create an extensible script - it should match the walkthrough exactly.

example/e2e/deploy-federated-deployment.sh
irfanurrehman: Got it, I read your suggested steps after posting this comment.

example/e2e/deploy-federated-deployment.sh
xunpan: done.

example/e2e/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,147 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script has an end to end steps in deploying a federated deployment
+
+set -o nounset
+set -o pipefail
+
+function wait-for-conditions() {
```
marun: This should be imported from the scripts dir rather than redefined.

example/e2e/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,147 @@
...
+echo
+echo "Changing placement of federated deployment nginx to cluster2."
+kubectl --namespace ${TEST_NS} patch federateddeployment nginx --type=merge --patch '{"spec": {"placement": {"clusterNames": ["cluster2"]}}}'
+if [[ $? != 0 ]]; then
+    echo "Failed changing placement of federateddeployments nginx in namespace ${NAMESPACE}."
+    exit 1
+fi 
+
+echo "Checking status of federated deployment nginx."
+sleep 3 # status will be updated soon
```
marun: This check is pointless without evaluation - it can be removed from the script.

example/e2e/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,147 @@
...
+echo "Federating resources in namespace ${TEST_NS} to member clusters."
+kubefedctl federate ns ${TEST_NS} --contents --skip-api-resources 'pods,secrets,serviceaccount,replicaset'
+if [[ $? != 0 ]]; then
+    echo "Failed federating resources to member clusters."
+    exit 1
+fi 
+
+
+echo
+echo "Checking status of federated resources."
```
marun: There is no point in retrieving status if it won't be evaluated.  Either the result of a jsonpath query should be evaluated or this check should be skipped.

example/e2e/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,147 @@
...
+# - a cluster scoped federation deployment with `cluster1` and
+#   `cluster2` joined as member clusters
+# - # kubectl config use-context cluster1
+#
+echo "Starting deploying a ngnix application in federation cluster."
+
+echo
+echo "Creating a test namespace ${TEST_NS}."
+kubectl create ns ${TEST_NS}
+if [[ $? != 0 ]]; then
```
marun: Why have you chosen to check return values instead of `set -o errexit`?

example/e2e/deploy-federated-nginx.sh
xunpan: I implement the customized version:
1. I try to check 2 conditions for different clusters in parallel. Not one by one.
2. example script refer to development tool script seems not good.

marun: The script is _developer only_.  It's intended to validate the instructions of the walkthrough in ci, and is not intended to be run by a user.  Checking conditions in parallel is not required for this use case.

xunpan: when it becomes a developer only scirpt? I'm confused.

https://github.com/kubernetes-sigs/kubefed/issues/104
> There should be a relatively simple walkthrough that guides users through the experience of kicking the tires on federation-v2 by deploying a very simple application. Other attributes of this:

xunpan:
> I asked was for the existing example walkthrough to be rewritten to match the script you are adding. 

Did you mean section `User Guide/Example`? It seems not equivalent, especailly for `Using Cluster Selector`. Did you want all things in `Example` are tested in the script?

example/e2e/deploy-federated-nginx.sh
marun: The point of #104 was originally focused on ensuring that the example walkthrough works by writing a script.  The script wasn't intended to replace the walkthrough.  It was to exist in addition to the walkthrough to automatically validate in CI that the steps described in the walkthrough would work.  My modification to #104 was to ask for the walkthrough to be updated to showcase more kubefed features (like `kubefedctl federate` and propagation status).

So the deliverables for #104 are:

 - an updated walkthrough in the userguide that demonstrates more kubefed features
 - a script that executes the commands documented in the walkthrough and verifies that those steps work as expected

xunpan:
I'd like to clarify what should be done in this PR before continue to coding:
1. this is development script which will be used in CI later.
   So, it is located in `scripts/deploy-federated-nginx.sh `. We do not need to check conditions in parallel, so check each condition of member clusters in sequence with same time out for each member cluster.

2. The script itself is trying to make a step by step case from a local kubernetes resource. 
   It should be started with a local kubernetes resource and then use `kubefedctl federate`.
   We need local resource yaml for this under `example/sample2/` instead of `example/sample1` because `example/sample1` uses federated resource yaml as an example. It is not different from our script demands.

3. The script will verify all command exists in `userguide` and all new `features` exists in the scripts should be documented. 
What should I do among `userguide` and `script` is not clear to me.
 - verify all case in `userguide/example` make the scripts complex. I don't think we need to verify all in CI. Because we should have related case in current test code for the feature.
 - update `userguide` with only `script` used case. This seems not good.

How to balance this? I can not sure of your thoughts.

@marun

xunpan:
After discussion in today's meeting:
1. clean old sample1 and use kubernetes resources
2. add override case
3. remove useless query

Overall is finished. I will work on verify result part.

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,99 @@
...
+util::wait-for-condition 'web content changed in cluster2' "(curl ${IP2}:${PORT2} | grep '^Hi K8s') &> /dev/null" 120
+util::wait-for-condition 'web content changed in cluster1' "(curl ${IP1}:${PORT1} | grep '^Hi K8s') &> /dev/null" 120
+curl ${IP1}:${PORT1}
+echo
+curl ${IP2}:${PORT2}
+echo
+
+echo
+echo "Changing override of federated deployment nginx to cluster1 with 'replicas' to 2."
+kubectl --namespace ${TEST_NS} patch federateddeployment nginx --type=merge --patch \
```
marun: (No action required) We had discussed using an override to change the text returned by nginx, which I think demos better.  Is that still something you are targeting?

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,99 @@
...
+echo "Federating resources in namespace ${TEST_NS} to member clusters."
+kubefedctl federate ns ${TEST_NS} --contents --skip-api-resources 'pods,secrets,serviceaccount,replicaset'
+
+echo
+echo "Checking status of federated resources."
+# TODO: Verify
+
+echo
+echo "Querying status of deployment nginx applications from member clusters."
+IP1=$(minikube -p cluster1 ip)
```
marun: This is going to run in CI, and it won't be an option to use multiple minikube clusters.  You'll need to figure out how to do the same thing with kind.

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,99 @@
...
+curl ${IP2}:${PORT2}
+echo
+
+echo
+echo "Changing override of federated deployment nginx to cluster1 with 'replicas' to 2."
+kubectl --namespace ${TEST_NS} patch federateddeployment nginx --type=merge --patch \
+    '{"spec" : {"overrides": [{"clusterName" : "cluster1", "clusterOverrides": [{"path": "spec.replicas", "value" : 2}]}]}}'
+
+echo
+echo "Changing placement of federated deployment nginx to cluster2 with 'cluster' to 'cluster2'."
```
marun: -> Update placement to include only `cluster2` so that the deployment will be removed from `cluster1`.

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,105 @@
...
+kubectl create ns ${TEST_NS}
+
+echo
+echo "Creating a ngnix application resources in namespace ${TEST_NS}."
+kubectl apply -n ${TEST_NS} -f ../example/sample1/configmap.yaml
+kubectl apply -n ${TEST_NS} -f ../example/sample1/deployment.yaml
+kubectl apply -n ${TEST_NS} -f ../example/sample1/service.yaml
+
+
+# Fedearte kubernetes resources to member clusters
```
marun: `s/Fedearte/Federate/`

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,105 @@
...
+util::wait-for-condition 'nginx web server ok in cluster1' "curl ${IP1}:${PORT1} &> /dev/null" 120
+util::wait-for-condition 'nginx web server ok in cluster2' "curl ${IP2}:${PORT2} &> /dev/null" 120
+echo "cluster1: $(curl -s ${IP1}:${PORT1})"
+echo "cluster2: $(curl -s ${IP2}:${PORT2})"
+
+
+# Modify federated resources to update kubernetes resources in member clusters
+echo
+echo "Changing index.html in federated configmap."
+kubectl patch federatedconfigmap web-file -n ${TEST_NS} --type=merge -p '{"spec": {"template": {"data": {"index.html": "Hi K8s Fed!"}}}}'
```
marun: `/Hi K8s Fed/Hello from KubeFed!/`

example/sample1/configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: web-file
+data:
+  index.html: "Hello Kubernetes Federation!"
```
marun: `s/Hello Kubernetes Federation/Hello from Kubernetes Cluster Federation!/`

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,105 @@
...
+
+echo
+echo "Updating placement to include only 'cluster2' so that the deployment will be removed from 'cluster1'"
+kubectl patch federateddeployment nginx -n ${TEST_NS} --type=merge --patch '{"spec": {"placement": {"clusters": [{"name": "cluster2"}]}}}'
+echo "Checking status of federated deployment nginx."
+util::wait-for-condition 'federated deployment status updated' \
+    "(kubectl get federateddeployment nginx -n ${TEST_NS} -o jsonpath='{.status.clusters[*].name}' | grep "^cluster2$") &> /dev/null" 120
+
+echo
+echo "Federation nginx application is done."
```
marun: `s/Federation nginx application is done/Successfully completed KubeFed walkthrough/`

scripts/deploy-federated-nginx.sh
```diff
@@ -0,0 +1,105 @@
...
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+
+TEST_NS="test-namespace"
+
+# Prerequisite:
+# - a cluster scoped federation deployment with `cluster1` and
+#   `cluster2` joined as member clusters
+# - kubectl config use-context cluster1
+#
+echo "Starting deploying a ngnix application in federation cluster."
```
marun: -> Validating KubeFed walkthrough

xunpan:
@marun done

scripts/pre-commit.sh
```diff
@@ -149,6 +149,10 @@ echo "Deploying cluster-scoped kubefed"
 echo "Running e2e tests against cluster-scoped kubefed"
 run-e2e-tests
 
+echo "Validating KubeFed walkthrough"
+./scripts/deploy-federated-nginx.sh
+kubectl delete ns test-namespace
```
marun: Deletion of the test namespace should be done as part of the script, please fix in a follow-up.

Issue #879, opened 2019-05-11
Align kubefed core APIs with kubernetes API conventions


font:
This issue is to track the work to modify the core API types in `core.federation.k8s.io` that do not abide by the [kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md). These issues include:

- Optional fields should have both the `omitempty` json struct tag and the `+optional` comment tag in Go, as well as be a pointer type (or have built-in nil value such as maps and slices). See [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) for more details.
- Some fields should be required and not be `omitempty`.
- Inconsistent naming of fields that are semantically similar e.g. `target`, `federatedType`, and `status` within the `FederatedTypeConfig` API.
- Fields that use booleans should often use an enumeration of strings instead. This extends the possibility of ever having more than a boolean behavior by making additions backwards compatible. They will also need to be renamed as a result. See [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) for more details.
- All json field names defined by the struct tags should be camelCase compliant. See [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions) for more details.

The following APIs will be updated:
- [x] KubefedCluster (#869 and #915)
- [x] FederatedTypeConfig (#885)
- [x] KubefedConfig (#887)
- [x] PropagatedVersion (#896)
- [x] ClusterPropagatedVersion (#897)
- [x] FederatedServiceStatus (#898)

/kind bug

font:
Closing as all PRs have been merged.

PR #885, opened 2019-05-14
Align FederatedTypeConfig API with K8s API conventions

font:
Partial fix for #879.


config/enabletypedirectives/services.yaml
```diff
@@ -6,4 +6,4 @@ metadata:
 # spec:
 #   status:
 #     kind: FederatedServiceStatus
-#   enableStatus: true
```
marun: Please remove this TODO instead of maintaining it.

pkg/apis/core/typeconfig/util.go
```diff
@@ -51,3 +54,22 @@ func GroupQualifiedName(apiResource metav1.APIResource) string {
 	}
 	return fmt.Sprintf("%s.%s", apiResource.Name, apiResource.Group)
 }
+
+func NamespacedToScope(apiResource metav1.APIResource) apiextv1b1.ResourceScope {
```
marun: Please move this function and the one that follows to the enable package (whether `enable.go` or `util.go`), since that's the only place they are used.  All other code can just use the namespace value of the apiresource verbatim and won't have to consider the discrepancy.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -59,17 +55,35 @@ type APIResource struct {
 	// the fields enforces them as mandatory.
 
 	// Group of the resource.
-	Group string `json:"group,omitempty"`
+	Group string `json:"group"`
```
marun: Core kube types won't have a group.  Also, what is the point of requiring a string when an empty string satisfies the condition?

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -113,7 +125,9 @@ type FederatedTypeConfig struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   FederatedTypeConfigSpec   `json:"spec,omitempty"`
+	// +optional
+	Spec FederatedTypeConfigSpec `json:"spec,omitempty"`
```
marun: Why is the spec optional?

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -134,16 +149,16 @@ func SetFederatedTypeConfigDefaults(obj *FederatedTypeConfig) {
 	// TODO(marun) will name always be populated?
 	nameParts := strings.SplitN(obj.Name, ".", 2)
 	targetPluralName := nameParts[0]
-	setStringDefault(&obj.Spec.Target.PluralName, targetPluralName)
```
marun: (No action required) I'm eagerly awaiting having these defaults being applied by webhook.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -176,52 +191,53 @@ func (f *FederatedTypeConfig) GetObjectMeta() metav1.ObjectMeta {
...
 
-func (f *FederatedTypeConfig) GetEnableStatus() bool {
-	return f.Spec.EnableStatus
+func (f *FederatedTypeConfig) GetStatusEnabled() bool {
+	return f.Spec.StatusCollection != nil && *f.Spec.StatusCollection == StatusCollectionEnabled
 }
 
-// TODO(marun) Remove in favor of using 'true' for namespaces and the
-// value from target otherwise.
 func (f *FederatedTypeConfig) GetFederatedNamespaced() bool {
```
marun: This function should be removed in favor of checking the namespaced property of calling `Namespaced` on the federated apiresource.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -176,52 +191,53 @@ func (f *FederatedTypeConfig) GetObjectMeta() metav1.ObjectMeta {
 	return f.ObjectMeta
 }
 
-func (f *FederatedTypeConfig) GetTarget() metav1.APIResource {
-	return apiResourceToMeta(f.Spec.Target, f.Spec.Namespaced)
+func (f *FederatedTypeConfig) GetTargetType() metav1.APIResource {
+	return apiResourceToMeta(f.Spec.TargetType, f.GetNamespaced())
 }
 
 func (f *FederatedTypeConfig) GetNamespaced() bool {
```
marun: This function should go away (i.e. be removed from the interface) in favor of checking the namespaced property of the appropriate apiresource (federated or target depending on the context).

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -176,52 +191,53 @@ func (f *FederatedTypeConfig) GetObjectMeta() metav1.ObjectMeta {
...
 }
 
-func (f *FederatedTypeConfig) GetStatus() *metav1.APIResource {
-	if f.Spec.Status == nil {
+func (f *FederatedTypeConfig) GetStatusType() *metav1.APIResource {
+	if f.Spec.StatusType == nil {
 		return nil
 	}
-	metaAPIResource := apiResourceToMeta(*f.Spec.Status, f.Spec.Namespaced)
+	metaAPIResource := apiResourceToMeta(*f.Spec.StatusType, f.Spec.StatusType.Namespaced())
```
marun: The `namespaced` paramter to `apiResourceToMeta` is now unnecessary, since that value is now accessible via `APIResource`.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
pmorie: Agree, this should be `omitempty`

config/enabletypedirectives/services.yaml
font: Removed

pkg/apis/core/typeconfig/util.go
font: Fixed.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
font: The core k8s types define the group as the empty string `""`, which is still valid for a required field, and is different from not supplying the field at all. Do we really want it to be `omitempty`?

marun: I don't get the point of requiring a field for which an acceptable value is the default if not supplied.

@pmorie WDDS (what does david say)?

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
marun: Please capture this as a TODO on the issue so it doesn't get misplaced.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
pmorie: Found some [precedent in the RBAC API](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/rbac/v1/types.go#L81)

It looks like the RBAC API has a similar field that is `omitempty` and provides a default.

font: This is also used for `FederatedType` which requires the `Group` field be set. I think that supersedes the requirement to make this field required. Thoughts?

font:
@marun Updated based on our offline conversation and squashed commits as requested. PTAL.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -176,52 +191,60 @@ func (f *FederatedTypeConfig) GetObjectMeta() metav1.ObjectMeta {
...
 	// Special-case the scope of federated namespace since it will
 	// hopefully be the only instance of the scope of a federated
 	// type differing from the scope of its target.
 
-	// TODO(marun) Use the constant in pkg/controller/util
 	if f.IsNamespace() {
-		// Namespace placement is namespaced to allow the control
-		// plane to run with only namespace-scoped permissions.
+		// FederatedNamespace is namespaced to allow the control plane to run
+		// with only namespace-scoped permissions to determine placement.
```
marun: (No action required) I think the addition of 'to determine placement' is not helpful because that's not the only reason for the discrepancy.  Since this function will be shortly be removed, though, I won't ask for any change.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -166,7 +166,11 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 
 		typeConfig.Status.ObservedGeneration = typeConfig.Generation
 		typeConfig.Status.PropagationController = corev1a1.ControllerStatusNotRunning
-		typeConfig.Status.StatusController = corev1a1.ControllerStatusNotRunning
+
+		if typeConfig.Status.StatusController == nil {
+			typeConfig.Status.StatusController = new(corev1a1.ControllerStatus)
```
marun: (No action required) Ugh, I think `StatusController` and `PropagationController` are terrible names for fields indicating whether a controller is running or not.

marun:
Please rebase

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
```diff
@@ -122,6 +136,7 @@ type FederatedTypeConfig struct {
 // FederatedTypeConfigList contains a list of FederatedTypeConfig
 type FederatedTypeConfigList struct {
 	metav1.TypeMeta `json:",inline"`
+	// +optional
```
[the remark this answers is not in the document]
font: It is highly recommended per https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required in the event the backwards compatibility guarantee of `omitempty` alone is not enough.

pkg/apis/core/v1alpha1/federatedtypeconfig_types.go
font: Updated to reflect that it's just an example.

pkg/controller/federatedtypeconfig/controller.go
font: There will probably be another opportunity to update this post API review, but if we want to change these now is also a good time.

font:
@marun Updated and rebased.

marun:
LGTM

PR #887, opened 2019-05-14
Align FederationConfig API with K8s API conventions

### LINKED ISSUE #879: Align kubefed core APIs with kubernetes API conventions

font:
Partial fix for #879 .

**NOTE:** Due to the K8s API conventions doc being ambiguous over the use of pointers for optional fields, I have kept optional fields without a built-in `nil` value as non-pointers for now. This is subject to change pending a more formal API review.

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -232,13 +232,7 @@ func setDuration(target *metav1.Duration, defaultValue time.Duration) {
 	}
 }
 
-func setString(target *string, defaultValue string) {
-	if *target == "" {
-		*target = defaultValue
-	}
-}
-
-func setInt(target *int, defaultValue int) {
+func setInt64(target *int64, defaultValue int64) {
 	if *target == 0 {
 		*target = defaultValue
 	}
@@ -266,17 +260,22 @@ func setDefaultKubefedConfig(fedConfig *corev1a1.KubefedConfig) {
 	setDuration(&duration.UnavailableDelay, util.DefaultClusterUnavailableDelay)
 
 	election := &spec.LeaderElect
-	setString(&election.ResourceLock, util.DefaultLeaderElectionResourceLock)
+	if len(election.ResourceLock) == 0 {
+		election.ResourceLock = util.DefaultLeaderElectionResourceLock
+	}
 	setDuration(&election.RetryPeriod, util.DefaultLeaderElectionRetryPeriod)
 	setDuration(&election.RenewDeadline, util.DefaultLeaderElectionRenewDeadline)
 	setDuration(&election.LeaseDuration, util.DefaultLeaderElectionLeaseDuration)
 
 	healthCheck := &spec.ClusterHealthCheck
-	setInt(&healthCheck.PeriodSeconds, util.DefaultClusterHealthCheckPeriod)
-	setInt(&healthCheck.TimeoutSeconds, util.DefaultClusterHealthCheckTimeout)
-	setInt(&healthCheck.FailureThreshold, util.DefaultClusterHealthCheckFailureThreshold)
-	setInt(&healthCheck.SuccessThreshold, util.DefaultClusterHealthCheckSuccessThreshold)
+	setInt64(&healthCheck.PeriodSeconds, util.DefaultClusterHealthCheckPeriod)
+	setInt64(&healthCheck.TimeoutSeconds, util.DefaultClusterHealthCheckTimeout)
+	setInt64(&healthCheck.FailureThreshold, util.DefaultClusterHealthCheckFailureThreshold)
+	setInt64(&healthCheck.SuccessThreshold, util.DefaultClusterHealthCheckSuccessThreshold)
 
+	if len(spec.SyncController.AdoptResources) == 0 {
+		spec.SyncController.AdoptResources = corev1a1.AdoptResourcesEnabled
+	}
 }
 
 func updateKubefedConfig(config *rest.Config, fedConfig *corev1a1.KubefedConfig) {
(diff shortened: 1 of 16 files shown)


cmd/controller-manager/app/controller-manager.go
```diff
@@ -269,11 +269,13 @@ func setDefaultFederationConfig(fedConfig *corev1a1.FederationConfig) {
 	setDuration(&duration.AvailableDelay, util.DefaultClusterAvailableDelay)
 	setDuration(&duration.UnavailableDelay, util.DefaultClusterUnavailableDelay)
 
-	election := &spec.LeaderElect
-	setString(&election.ResourceLock, util.DefaultLeaderElectionResourceLock)
-	setDuration(&election.RetryPeriod, util.DefaultLeaderElectionRetryPeriod)
-	setDuration(&election.RenewDeadline, util.DefaultLeaderElectionRenewDeadline)
-	setDuration(&election.LeaseDuration, util.DefaultLeaderElectionLeaseDuration)
+	election := spec.LeaderElect
```
marun: (No action required) I'm looking forward to this defaulting becoming the function of the admission webhook.

pkg/apis/core/v1alpha1/federationconfig_types.go
```diff
@@ -26,56 +26,68 @@ type FederationConfigSpec struct {
 	// The scope of the federation control plane should be either `Namespaced` or `Cluster`.
 	// `Namespaced` indicates that the federation namespace will be the only target for federation.
 	Scope              apiextv1b1.ResourceScope `json:"scope,omitempty"`
-	ControllerDuration DurationConfig           `json:"controller-duration,omitempty"`
-	LeaderElect        LeaderElectConfig        `json:"leader-elect,omitempty"`
-	FeatureGates       []FeatureGatesConfig     `json:"feature-gates,omitempty"`
-	ClusterHealthCheck ClusterHealthCheckConfig `json:"cluster-health-check,omitempty"`
-	SyncController     SyncControllerConfig     `json:"sync-controller,omitempty"`
+	ControllerDuration DurationConfig           `json:"controllerDuration,omitempty"`
```
marun: I'm guessing this is WIP for more than just the test failure, and you're going to make these pointers and add `+optional`?

pkg/controller/util/cluster_util.go
```diff
@@ -45,7 +45,7 @@ const (
 	DefaultLeaderElectionLeaseDuration = 15 * time.Second
 	DefaultLeaderElectionRenewDeadline = 10 * time.Second
 	DefaultLeaderElectionRetryPeriod   = 5 * time.Second
-	DefaultLeaderElectionResourceLock  = "configmaps"
+	DefaultLeaderElectionResourceLock  = fedv1a1.ConfigMapResourceLock
```
marun: (No action required) I'm not sure why it ever made sense for these defaults to live here.  Something to revisit when the admission webhook takes responsibility for defaulting.

pkg/controller/util/controllerconfig.go
```diff
@@ -44,7 +46,7 @@ type LeaderElectionConfiguration struct {
 	RetryPeriod time.Duration
 	// resourceLock indicates the resource object type that will be used to lock
 	// during leader election cycles.
-	ResourceLock string
+	ResourceLock fedv1a1.ResourceLockType
```
marun: (No action required) I think this struct predates `FederationConfig`.  I think the use of `ControllerConfig` should either be replaced by `FederationConfig` or at least reuse structs like `LeaderElectConfig` to avoid having to manually synchronize them.

pkg/apis/core/v1alpha1/federationconfig_types.go
font: Correct, this is WIP because there's still more work to do.

font:
@marun thanks for the prelim review! Pending CI passing I think this should be ready for review.

pkg/apis/core/v1alpha1/kubefedconfig_types.go
```diff
@@ -25,59 +25,98 @@ import (
...
+const (
+	ConfigMapResourceLock ResourceLockType = "configmaps"
+	EndpointResourceLock  ResourceLockType = "endpoints"
+)
+
 type FeatureGatesConfig struct {
-	Name    string `json:"name,omitempty"`
-	Enabled bool   `json:"enabled,omitempty"`
+	Name          string            `json:"name"`
+	Configuration ConfigurationMode `json:"configuration"`
```
marun: (No action required) I get that you're trying to be consistent, but this seems odiously excessive to me.  The chances of us having to extend the way we enable a feature gate or not is nil.  We're more likely to bin the very idea of feature gates - it's a terrible way to manage GA features.

font:
@marun I've rebased and pushed 2 new commits to make all fields required and update generated code, which I can subsequently squash. I've also opted to keep the `Options.ClusterHealthCheckConfig` field a pointer to call by reference into `StartController()` functions and other functions further down the path.

pkg/apis/core/v1alpha1/kubefedconfig_types.go
```diff
@@ -25,59 +25,82 @@ import (
...
 	// leader election. Supported options are `configmaps` (default) and `endpoints`.
-	ResourceLock string `json:"resourceLock,omitempty"`
+	ResourceLock ResourceLockType `json:"resourceLock"`
 }
+
+type ResourceLockType string
+
+const (
+	ConfigMapResourceLock ResourceLockType = "configmaps"
+	EndpointResourceLock  ResourceLockType = "endpoints"
```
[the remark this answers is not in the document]
font: Updated. If only those constants were strongly typed we would just use them instead.

pmorie:
I think this moves us forward on the beta candidate release, and it looks good to me as progress.

/lgtm

PR #893, opened 2019-05-15
Run E2E with travis in forked repos

font:
This prevents running `docker login` using the quay credentials if we don't have any travis secure environment vars set e.g. running travis in forked repos.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,25 +122,25 @@ push: container
 
 	if [[ -z "$(TRAVIS_PULL_REQUEST)" ]]; \
 	then \
-		$(DOCKER) push $(REGISTRY)/$(TARGET):$(GIT_VERSION); \
+		$(DOCKER) push $(IMAGE):$(GIT_VERSION); \
 	elif [[ "$(TRAVIS_PULL_REQUEST)" == "false" && "$(TRAVIS_SECURE_ENV_VARS)" == "true" ]]; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -119,27 +120,27 @@ generate: generate-code kubefedctl
...
-		if [ "$(TRAVIS_BRANCH)" = "master" ]; \
+		if [[ "$(TRAVIS_BRANCH)" == "master" ]]; \
 		then \
 			$(DOCKER) tag $(REGISTRY)/$(TARGET):$(GIT_VERSION) $(REGISTRY)/$(TARGET):canary; \
 			$(DOCKER) push $(REGISTRY)/$(TARGET):canary; \
+		fi; \
+		\
+		if git describe --tags --exact-match >/dev/null 2>&1; \
+		then \
+			$(DOCKER) tag $(REGISTRY)/$(TARGET):$(GIT_VERSION) $(REGISTRY)/$(TARGET):$(GIT_TAG); \
```
shashidharatd: (suggestion) Not related to this pr, but can we use $(IMAGE) which is defined at the top of Makefile instead of $(REGISTRY)/$(TARGET)

font: :+1: I've updated.

shashidharatd:
/lgtm
Thanks !

PR #903, opened 2019-05-16
Update docs of charts, minikube, and userguide

qinpingli:


docs/userguide.md
```diff
@@ -59,7 +59,7 @@ Please refer to [Kubefed Concepts](./concepts.md) first before you go through th
 
 ## Prerequisites
 
-The kubefed deployment requires kubernetes version >= 1.11. The following
+The kubefed deployment requires kubernetes version >= 1.13. The following
```
marun: Should this be updated?  Fine to use a newer version, but we want to support older versions of kube if possible.

cc: @pmorie

charts/index.yaml
```diff
@@ -1,28 +1,28 @@
 apiVersion: v1
 entries:
-  federation-v2:
+  kubefed:
```
marun: I don't think any of the previous releases should be renamed. They should remain in the index as federation-v2.

charts/index.yaml
```diff
@@ -1,28 +1,28 @@
...
   - apiVersion: v1
     created: 2019-05-06T15:30:47.795749649-07:00
     description: Kubernetes Federation V2 helm chart
     digest: 9b248eed6f021951653ee98cb255d2d3564748ac04c2484c941f619c39c5dd79
-    name: federation-v2
+    name: kubefed
     urls:
-    - https://sigs.k8s.io/kubefed/releases/download/v0.0.10/federation-v2-0.0.10.tgz
+    - https://github.com/kubernetes-sigs/federation-v2/releases/download/v0.0.10/federation-v2-0.0.10.tgz
```
marun: Good catch.  Looks like the links for old releases aren't changed by the rename.

cc: @pmorie

charts/index.yaml
```diff
@@ -1,28 +1,28 @@
 apiVersion: v1
 entries:
-  federation-v2:
+  kubefed:
   - apiVersion: v1
     created: 2019-05-06T15:30:47.795749649-07:00
     description: Kubernetes Federation V2 helm chart
     digest: 9b248eed6f021951653ee98cb255d2d3564748ac04c2484c941f619c39c5dd79
-    name: federation-v2
+    name: kubefed
```
marun: Same as previous comment - leave as federration-v2.

charts/kubefed/README.md
```diff
@@ -77,7 +77,7 @@ $ helm search kubefed
 Install the chart and specify the version to install with the
 `--version` argument. Replace `<x.x.x>` with your desired version.
 ```bash
-$ helm install fedv2-charts/kubefed --version=<x.x.x> --namespace kube-federation-system
+$ helm install kubefed-charts/kubefed --version=<x.x.x> --namespace kube-federation-system
```
marun: This should be `kubefed-charts/federation-v2` for now.

docs/userguide.md
qinpingli: Thanks, @marun , I saw there is some updating for these docs, kubernetes version is updated to  >= 1.13, I'll merge the new updating, and update this PR.

charts/index.yaml
qinpingli: Got it.

charts/kubefed/README.md
qinpingli: Got it.

charts/kubefed/README.md
```diff
@@ -96,6 +96,7 @@ Delete all kubefed CRDs:
 
 ```bash
 $ kubectl delete crd $(kubectl get crd | grep -E 'kubefed.k8s.io' | awk '{print $1}')
+$ kubectl delete crd $(kubectl get crd | grep -E 'clusterregistry.k8s.io' | awk '{print $1}')
```
marun: This line is no longer necessary, as the cluster registry is no longer installed by the chart.  Please revert.

qinpingli: Done.

PR #904, opened 2019-05-16
Changes to naming from Federation to Kubefed

abrennan89:
I've replaced some initial instances, but this probably needs additional review from @shashidharatd - I'm not sure about federation in the names of scripts or in some of the contexts it has been used?


irfanurrehman:
Seems fine to me.
@abrennan89 thanks for doing this. 
/lgtm

PR #914, opened 2019-05-21
Validation for FederatedTypeConfig API

### LINKED ISSUE #909: Validations for FederatedTypeConfig API
#### Spec
- `APIResource`
  - `Group` must be a valid group name (same as for [CRDs](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L120)), however, the empty string should be allowed in group
  - `Version` must be a valid API version
  - `Kind` must be a valid Kind
  - `Plural` must be a valid [plural name]
(https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L468)
  - `Scope` must be a valid enum value
- `FederatedType.Group` must be set
- `StatusType.Group` must be set if `StatusType` is not `nil`
- `Propagation` must be a valid enum value (can we get this from open API?)
- `StatusCollection` if set, must be a valid enum value

#### Status

- `ObservedGeneration` must be a non-negative `int64`
- `PropagationController` must be a valid enum value
- `StatusController` if set, must be a valid enum value

font:
Fixes #909

### DIFF
--- a/pkg/apis/core/v1beta1/validation/validation.go
+++ b/pkg/apis/core/v1beta1/validation/validation.go
@@ -52,13 +52,15 @@ func ValidateFederatedTypeConfigSpec(spec *v1beta1.FederatedTypeConfigSpec, fldP
 	return allErrs
 }
 
+const domainWithAtLeastOneDot string = "should be a domain with at least one dot"
+
 func ValidateFederatedAPIResource(fedType *v1beta1.APIResource, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if len(fedType.Group) == 0 {
 		allErrs = append(allErrs, field.Required(fldPath.Child("group"), ""))
 	} else if len(strings.Split(fedType.Group, ".")) < 2 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("group"), fedType.Group, "should be a domain with at least one dot"))
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("group"), fedType.Group, domainWithAtLeastOneDot))
 	}
 
 	allErrs = append(allErrs, ValidateAPIResource(fedType, fldPath)...)
--- a/pkg/apis/core/v1beta1/validation/validation.go
+++ b/pkg/apis/core/v1beta1/validation/validation.go
@@ -24,17 +24,29 @@ import (
 	valutil "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
 	"sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 )
 
 func ValidateFederatedTypeConfig(obj *v1beta1.FederatedTypeConfig, statusSubResource bool) field.ErrorList {
 	var allErrs field.ErrorList
 	if !statusSubResource {
-		allErrs = ValidateFederatedTypeConfigSpec(&obj.Spec, field.NewPath("spec"))
+		allErrs = ValidateFederatedTypeConfigName(obj)
+		allErrs = append(allErrs, ValidateFederatedTypeConfigSpec(&obj.Spec, field.NewPath("spec"))...)
 	} else {
 		allErrs = ValidateFederatedTypeConfigStatus(&obj.Status, field.NewPath("status"))
 	}
 	return allErrs
+}
+
+const federatedTypeConfigNameErrorMsg string = "name must be 'TARGET_PLURAL_NAME(.TARGET_GROUP_NAME)'"
+
+func ValidateFederatedTypeConfigName(obj *v1beta1.FederatedTypeConfig) field.ErrorList {
+	expectedName := typeconfig.GroupQualifiedName(obj.GetTargetType())
+	if expectedName != obj.Name {
+		return field.ErrorList{field.Invalid(field.NewPath("name"), obj.Name, federatedTypeConfigNameErrorMsg)}
+	}
+	return field.ErrorList{}
 }
 
 func ValidateFederatedTypeConfigSpec(spec *v1beta1.FederatedTypeConfigSpec, fldPath *field.Path) field.ErrorList {
--- a/pkg/apis/core/v1beta1/validation/validation.go
+++ b/pkg/apis/core/v1beta1/validation/validation.go
@@ -27,10 +27,13 @@ import (
 	"sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 )
 
-func ValidateFederatedTypeConfig(obj *v1beta1.FederatedTypeConfig) field.ErrorList {
-
-	allErrs := ValidateFederatedTypeConfigSpec(&obj.Spec, field.NewPath("spec"))
-	allErrs = append(allErrs, ValidateFederatedTypeConfigStatus(&obj.Status, field.NewPath("status"))...)
+func ValidateFederatedTypeConfig(obj *v1beta1.FederatedTypeConfig, statusSubResource bool) field.ErrorList {
+	var allErrs field.ErrorList
+	if !statusSubResource {
+		allErrs = ValidateFederatedTypeConfigSpec(&obj.Spec, field.NewPath("spec"))
+	} else {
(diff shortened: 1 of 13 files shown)


font:
Rebased and ready for review.

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -0,0 +1,259 @@
...
+	errorCases["spec.propagation: Required value"] = propagation
+
+	// Validate field values
+	validFedGroup := validFederatedTypeConfig()
+	validFedGroup.Spec.FederatedType.Group = "nodomain"
+	errorCases[domainWithAtLeastOneDot] = validFedGroup
+
+	validTargetGroup := validFederatedTypeConfig()
+	validTargetGroup.Spec.TargetType.Group = "invalid#group"
+	errorCases["must start and end with an alphanumeric character"] = validTargetGroup
```
xunpan: it seems the group name already `start and end with an charactor`. I guess you mean `#` should be invalid.

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -17,30 +17,117 @@ limitations under the License.
...
+
+	allErrs := ValidateFederatedTypeConfigSpec(&obj.Spec, field.NewPath("spec"))
+	if statusSubResource {
+		allErrs = append(allErrs, ValidateFederatedTypeConfigStatus(&obj.Status, field.NewPath("status"))...)
+	}
+	return allErrs
+}
+
+func ValidateFederatedTypeConfigSpec(spec *v1beta1.FederatedTypeConfigSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := ValidateAPIResource(&spec.TargetType, fldPath.Child("targetType"))
```
marun: Why does validation of the target type differ from validation of the federated and status types in more than how the group is evaluated?

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -17,30 +17,117 @@ limitations under the License.
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("pluralName"), obj.PluralName, strings.Join(errs, ",")))
+	}
+
+	allErrs = append(allErrs, validateEnumStrings(fldPath.Child("scope"), string(obj.Scope), []string{string(apiextv1b1.ClusterScoped), string(apiextv1b1.NamespaceScoped)})...)
+
 	return allErrs
 }
 
-func ValidateAPIResource(object *v1beta1.APIResource, fldPath *field.Path) field.ErrorList {
+func validateEnumStrings(fldPath *field.Path, value string, accepted []string) field.ErrorList {
```
marun: (No action required) The lack of reusable validation machinery seems problematic.

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	for _, successCase := range successCases {
+		if errs := ValidateFederatedTypeConfigSpec(&successCase.Spec, field.NewPath("spec")); len(errs) != 0 {
+			t.Errorf("expected success: %v", errs)
+		}
+	}
+
+	errorCases := map[string]*v1beta1.FederatedTypeConfig{}
+
+	// Validate required fields
```
marun: Please factor out the common (i.e. everything but the group) validation of the target, federated and status api resources so it can be called once for each type of resource rather than field-by-field.

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -17,30 +17,117 @@ limitations under the License.
...
+	valutil "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 )
 
-func ValidateFederatedTypeConfig(object *v1beta1.FederatedTypeConfig) field.ErrorList {
+func ValidateFederatedTypeConfig(obj *v1beta1.FederatedTypeConfig, statusSubResource bool) field.ErrorList {
+
+	allErrs := ValidateFederatedTypeConfigSpec(&obj.Spec, field.NewPath("spec"))
```
marun: Is it necessary to validate changes to the spec if the `statusSubResource` is true?

marun:
@font Also, please remove the need for `CheckTypeConfigName` by validating this property in the webhook.

pkg/apis/core/v1beta1/validation/validation.go
marun: No, not a comment on you (hence 'no action...').  It exists in apimachinery but it's not exported.

pkg/apis/core/v1beta1/validation/validation_test.go
marun: Your are correct that all 3 fields share a common struct, but that doesn't appear to me to imply that the validation paths are also shared.  

My request to clean up the testing in this method is motivated by my perception that it is not obvious that coverage is complete. There doesn't appear to be any test for the statusType field, for example, and I think the current ordering of errorCases makes it hard to easily determine whether all validation paths for a given field are being checked.

My original suggestion was to test validation of each field completely, but I take your point that it would be excessive.  What are your thoughts on the following strategy, which is already being followed in some cases?

 - test all error conditions in a method like `ValidateAPIResource` in its own test method 
   - i.e. `TestValidateAPIResource`
 - test only a single error condition for a field that is validated with a method like `ValidateAPIResource` to ensure that the validation is wired correctly

I think this would allow coverage to be observed with more clarity than what is currently proposed.

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -17,30 +17,129 @@ limitations under the License.
...
+	} else if len(strings.Split(fedType.Group, ".")) < 2 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("group"), fedType.Group, domainWithAtLeastOneDot))
+	}
 
+	allErrs = append(allErrs, ValidateAPIResource(fedType, fldPath)...)
 	return allErrs
 }
 
-func ValidateKubeFedCluster(object *v1beta1.KubeFedCluster) field.ErrorList {
+func ValidateStatusAPIResource(statusType *v1beta1.APIResource, fldPath *field.Path) field.ErrorList {
```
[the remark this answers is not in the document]
font: Mainly for testing purposes.

pkg/apis/core/v1beta1/validation/validation_test.go
font: I do think the coverage could be made a little more clear without knowledge of how the validation code paths work. I think either approach would work:
1. Testing `ValidateAPIResource` separately while also checking a single error condition for each of the FederatedTypeConfig type fields.
2. For absolute complete coverage: splitting out the common fields between TargetType, FederatedType, and StatusType into a helper test function and invoking this helper function for each of these fields.

If I'm going to revisit this in the future I think I prefer option 2.

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -0,0 +1,290 @@
...
+		Name:       "deployments",
+		Namespaced: true,
+	}
+}
+
+func validFederatedTypeConfig() *v1beta1.FederatedTypeConfig {
+	return federatedTypeConfig(apiResourceWithNonEmptyGroup())
+}
+
+func federatedTypeConfig(apiResource *metav1.APIResource) *v1beta1.FederatedTypeConfig {
```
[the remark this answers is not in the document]
font: Yes, as it's where it originally comes from. :smile: Only thing is I didn't want to add more code changes by reworking the `kubefedctl` one as it incorporates defaulting and this one also includes `Status`. Once defaulting is added it can be revisited.

pkg/apis/core/v1beta1/validation/validation_test.go
marun: I think knowing how the validation paths work is preferable - these are white-box tests after all.  A correlation between the code paths and what is tested ensures that future maintainers have a reference point.

PR #915, opened 2019-05-21
Align KubefedCluster API with K8s API conventions

font:
Partial fix for #879.

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/crds.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/crds.yaml
@@ -439,6 +439,7 @@ spec:
                 required:
                 - type
                 - status
+                - lastProbeTime
                 type: object
               type: array
             region:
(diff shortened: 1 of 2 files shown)


pkg/apis/core/v1alpha1/kubefedcluster_types.go
```diff
@@ -77,7 +76,8 @@ type KubefedCluster struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   KubefedClusterSpec   `json:"spec,omitempty"`
+	Spec KubefedClusterSpec `json:"spec"`
+	// +optional
```
marun: According to @sttts, non-pointer status with omitempty doesn't make sense.

marun:
Merging with the intention of having as good an API as possible for rc1.

/lgtm

PR #923, opened 2019-05-22
s/Kubefed/KubeFed/ in the code

marun:
We don't want more than one release that renames types, so this PR renames `KubefedCluster` to `KubeFedCluster` and `KubefedConfig` to `KubeFedConfig`.  I want to see this merged ASAP so I can cut a release.


marun:
Rebased

Issue #829, opened 2019-05-01
kubefedctl should support deleting CRDs when custom resources exist


font:
**What would you like to be added**:
When optionally deleting the CRD using `--delete-crd`, `kubefedctl` deletes the CRD only if any of its custom resources do not exist. If resources exist, it will report an error suggesting the user to delete them manually and retry. Instead when resources exist, `kubefedctl` should handle having those resources deleted when the user specifies a `--force` option either by:

- [ ] If `PropagationEnabled`, deleting the custom resources and allowing the sync controller to remove the federation finalizer to ensure garbage collection is possible i.e. in-band deletion. This should happen in advance of stopping the sync controller and deleting the FTC.
- [ ] if `!PropagationEnabled`, iterating over all instances of the CRD and remove the federation finalizer to ensure that garbage collection is possible. This would entail having a `--force-orphan` option to allow the user to specify how their templated resources should be dealt with upon deletion e.g. cascade or orphan.

Only then should deletion of the CRD be attempted.

**Why is this needed**:
Ensure proper garbage collection of all CRDs and custom resources regardless of whether propagation is enabled.

/kind feature

PR #924, opened 2019-05-22
Update userguide for disabling federation of an API type

font:


marun:
Please add a new section on disable detailing the deletion of the ftc and optionally deleting instances of the crd.  Right now it reads like it's related to disabling propagation and I think they should be separate.

font:
@marun updated

docs/userguide.md
```diff
@@ -276,19 +277,24 @@ you can patch role `kubefed-role` in the kubefed namespace instead.
 
 ### Disabling federation of an API type
```
marun: This should be 'disabling propagation'

docs/userguide.md
```diff
@@ -276,19 +277,24 @@ you can patch role `kubefed-role` in the kubefed namespace instead.
...
-This command sets the `propagationEnabled` field in the `FederatedTypeConfig`
-associated with this target API type to `false`, which will prompt the sync controller for the target API type to be stopped.
+This patch command sets the `propagation` field in the `FederatedTypeConfig`
+associated with this target API type to `Disabled`, which will prompt the sync
+controller for the target API type to be stopped.
 
-If you want to permanently disable federation of the target API type, pass the
-`--delete-crd` flag to remove the `FederatedTypeConfig` and federated type CRD created by
-`enable`.
+### Deleting federation of an API type
```
marun: This should be 'Disabling federation'

docs/userguide.md
```diff
@@ -276,19 +277,24 @@ you can patch role `kubefed-role` in the kubefed namespace instead.
...
+
+If you want to permanently delete federation of the target API type by
+removing the `FederatedTypeConfig`, and optionally delete the federated type
+CRD created by `enable` using the `--delete-crd` flag, run the following
+command:
 
 ```bash
 kubefedctl disable <FederatedTypeConfig Name> --delete-crd
```
marun: Maybe if --delete-crd is rendered as optional [--delete-crd] a user will be less likely to copy-n-paste this data destroyer?

The warning below should state that '...supplying the --delete-crd flag will remove all custom resources...'

docs/userguide.md
```diff
@@ -276,19 +277,24 @@ you can patch role `kubefed-role` in the kubefed namespace instead.
...
+This patch command sets the `propagation` field in the `FederatedTypeConfig`
+associated with this target API type to `Disabled`, which will prompt the sync
+controller for the target API type to be stopped.
 
-If you want to permanently disable federation of the target API type, pass the
-`--delete-crd` flag to remove the `FederatedTypeConfig` and federated type CRD created by
-`enable`.
+### Deleting federation of an API type
+
+If you want to permanently delete federation of the target API type by
```
marun: The wording of this is a bit awkward.  Suggest separating  from the implementation i.e.

'If you want to permanently disable federation of a target type, use `kubefed disable`. This will remove the FTC that configures federation of the type. If supplied with the optional `--delete-crd` flag, the command will also remove the federated type CRD and all of its instances.'

Maybe the above removes the need for a warning?

font: Deleting all of the instances is not yet implemented and tracked by #829. I've reworded this section to be more clear and removed the warning since it's technically no longer accurate.

docs/userguide.md
font: I can make this change to make it more clear but as of #769 we do not allow deleting the CRD if custom resources exist.

marun:
Thank you!

/lgtm

Issue #936, opened 2019-05-28
Finalize rename from `federation v2` to `kubefed`


marun:
#860 was only partially completed.  This issue is intended to ensure all remaining instances of 'federation' are removed from the repo before the final beta release.  This is not a strictly mechanical process, and is likely to require rewriting of docs and comments.

TODO:
 - [x] - #938 - Rename cleanup code (pkg, cmd)
 - [x] - #939 - Rename cleanup docs (supercedes #907)
 - [x] - #947 - Rename cleanup tests
 - [x] - #946 - Rename cleanup scripts
 - [x] - #937 - Rename cleanup misc (everything else)
 - [x] - #948 - More misc cleanup to reflect kubefed rename

marun:
@shashidharatd Reference to 'federation' remain in the service dns docs and code.  Do you think it should stay as-is or would it be possible to change as per the discussion of 'federation' -> 'KubeFed control plane' in the WG today?

marun:
All PRs targeted for this issue have merged.

PR #937, opened 2019-05-28
Cleanup misc to reflect kubefed rename

marun:
Partial fix for #936

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/crds.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/crds.yaml
@@ -42,7 +42,7 @@ spec:
                     description: The name of the cluster the version is for.
                     type: string
                   version:
-                    description: The last version produced for the resource by a federation
+                    description: The last version produced for the resource by a KubeFed
                       operation.
                     type: string
                 required:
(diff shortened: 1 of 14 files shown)


charts/kubefed/charts/controllermanager/templates/crds.yaml
```diff
@@ -579,9 +579,9 @@ spec:
               - resourceLock
               type: object
             scope:
-              description: The scope of the kubefed control plane should be either
-                `Namespaced` or `Cluster`. `Namespaced` indicates that the kubefed
-                namespace will be the only target for federation.
+              description: The scope of the KubeFed control plane should be either
+                `Namespaced` or `Cluster`.  `Namespaced` indicates that the KubeFed
```
pmorie: nit: i suspect @abrennan89 would have an opinion on two spaces vs. one here, but not a reason to hold the PR

pmorie:
This moves the ball forward I'm sure we'll continually comb over docs so I'd prefer to get this swath merged.

/lgtm

charts/kubefed/charts/controllermanager/templates/crds.yaml
marun: I'll fix in a follow-up.

marun: See #942

PR #938, opened 2019-05-28
Cleanup code to reflect kubefed rename

marun:
Partial fix for #936


xunpan:
LGTM

pmorie:
I couldn't find anything that I think needs to change about this. Whatever traces may be left we can get in another PR.

/lgtm

PR #939, opened 2019-05-28
Cleanup docs to reflect kubefed rename

### LINKED ISSUE #891: Re-examine doc after kubefed rename
This issue is to re-examine the doc content after the merge of #865 to find any places where we may have missed renaming or where grammatical changes may be required.

@abrennan89 is going to take this one, but she's not in the kubernetes-sigs org yet.

marun:
Partial fix for #936 
Fixed #891

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -85,8 +85,8 @@ non-generated code in the commit history.
 
 ## Running E2E Tests
 
-The KubeFed E2E tests must be executed against a deployed
-federation of one or more clusters.  Optionally, the KubeFed
+The KubeFed E2E tests must be executed against a KubeFed control plane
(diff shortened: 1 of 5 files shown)


docs/development.md
```diff
@@ -85,8 +85,8 @@ non-generated code in the commit history.
 
 ## Running E2E Tests
 
-The KubeFed E2E tests must be executed against a deployed
-federation of one or more clusters.  Optionally, the KubeFed
+The KubeFed E2E tests must be executed against a KubeFed control plane
+with one or more registered clusters.  Optionally, the KubeFed
```
pmorie: Just a note, I'm personally ok with the use of 'register' here and think it reads in a way that makes sense. If people feel strongly about it later we can always change it back.

PR #941, opened 2019-05-29
Webhook checking for KubeFedConfig

### LINKED ISSUE #916: Validations for KubefedConfig API
### Field Validations for [KubefedConfig](https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/apis/core/v1alpha1/kubefedconfig_types.go):

- `scope` must be a valid enum value. It should be immutable.
- `ControllerDuration`
    - `availableDelay` is a valid `int64` > 0
    - `unavailableDelay` is a valid `int64` > 0
- `LeaderElectConfig`
    - `leaseDuration` is a valid `int64` > 0 and must be > `renewDeadline`
    - `renewDeadline` is a valid `int64` > 0 and must be > `retryPeriod * JitterFactor`
    - `retryPeriod` is a valid `int64` > 0
    - `resourceLock` should be a valid enum value
- `FeatureGatesConfig`
    - `name` should be a valid feature name
    - `configuration` should be a valid enum value
- `ClusterHealthCheckConfig`
    - All fields should be a valid `int64` > 0
- `SyncControllerConfig`
    - `AdoptResources` should be a valid enum value.

xunpan:
This is not ready to review until #914 is merged and I rebase the code with UT developed.

I also fix a minor non-function helm chart related issue: make clusterrole in clusterrole.yaml. It is not functional change, so I use a single commit instead of a new PR.

Fixes #916


marun:
Please rebase.

xunpan:
rebase done and basic test passed.

I'm adding test now

marun:
CI failure looks like a potential flake.

xunpan:
Not sure details but make CI passed by adding `scope: Cluster`. I'm checking it.

xunpan:
webhook reject scope is empty case. This is because we had no default value setting in webhook and the old default setting does not set value for scope but expected it to be set in .yaml file.

add scope can make it work.

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -139,6 +143,71 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
...
+	allErrs = append(allErrs, validateGreaterThan0(electPath.Child("leaseDuration"), int64(elect.LeaseDuration.Duration))...)
+	allErrs = append(allErrs, validateGreaterThan0(electPath.Child("renewDeadline"), int64(elect.RenewDeadline.Duration))...)
+	allErrs = append(allErrs, validateGreaterThan0(electPath.Child("retryPeriod"), int64(elect.RetryPeriod.Duration))...)
+	if elect.LeaseDuration.Duration <= elect.RenewDeadline.Duration {
+		allErrs = append(allErrs, field.Invalid(electPath.Child("leaseDuration"), elect.LeaseDuration,
+			"leaseDuration must be greater than renewDeadline"))
+	}
+	if elect.RenewDeadline.Duration <= time.Duration(float64(elect.RetryPeriod.Duration)*leaderelection.JitterFactor) {
+		allErrs = append(allErrs, field.Invalid(electPath.Child("renewDeadline"), elect.RenewDeadline,
+			"renewDeadline must be greater than retryPeriod*JitterFactor"))
```
marun: Please include the jitter factor in the message.  

Also, is there a reason the jitter factor isn't configurable?

cc: @shashidharatd

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -139,6 +143,71 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
...
+	}
+	allErrs = append(allErrs, validateEnumStrings(electPath.Child("resourceLock"), string(elect.ResourceLock),
+		[]string{string(v1beta1.ConfigMapsResourceLock), string(v1beta1.EndpointsResourceLock)})...)
+
+	gates := spec.FeatureGates
+	gatesPath := specPath.Child("featureGates")
+	for _, gate := range gates {
```
marun: Please ensure that a gate appears at most once.

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -288,3 +292,111 @@ func federatedTypeConfig(apiResource *metav1.APIResource) *v1beta1.FederatedType
...
+
+func TestValidateKubeFedConfig(t *testing.T) {
+	errs := ValidateKubeFedConfig(validKubeFedConfig())
+	if len(errs) != 0 {
+		t.Errorf("expected success: %v", errs)
+	}
+
+	errorCases := map[string]*v1beta1.KubeFedConfig{}
+
+	validScope := validKubeFedConfig()
```
marun: Please order these in the same order that they appear in the code (which is itself ordered as per the type definition) to simplify future maintenance.

pkg/controller/webhook/kubefedconfig/webhook.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (a *KubeFedConfigValidationHook) Initialize(kubeClientConfig *rest.Config, stopCh <-chan struct{}) error {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	a.initialized = true
+
+	klog.V(2).Infof("KubeFedConfig validation webhook is initialized")
+
+	shallowClientConfigCopy := *kubeClientConfig
```
marun: What makes this a copy?  Is the caller making a copy before calling this function?

pkg/webhook/webhook.go
```diff
@@ -25,12 +25,14 @@ import (
...
 	"sigs.k8s.io/kubefed/pkg/controller/webhook"
 	"sigs.k8s.io/kubefed/pkg/controller/webhook/federatedtypeconfig"
+	"sigs.k8s.io/kubefed/pkg/controller/webhook/kubefedconfig"
 )
 
 func NewWebhookCommand(stopChan <-chan struct{}) *cobra.Command {
 	admissionHooks := []apiserver.AdmissionHook{
 		&federatedtypeconfig.FederatedTypeConfigValidationHook{},
 		&webhook.KubeFedClusterValidationHook{},
+		&kubefedconfig.KubeFedConfigValidationHook{},
```
marun: @font Why are the webhooks for federatedtypeconfig and kubefedconfig in a separate package but that isn't the case for kubefedcluster?

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
```diff
@@ -13,3 +13,29 @@ subjects:
   name: kubefed-controller
   namespace: {{ .Release.Namespace }}
 {{- end }}
+---
```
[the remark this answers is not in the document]
font: The `auth-delegator-kubefed-admission-webhook` creates a binding from the kubefed webhook service account to the `system:auth-delegator` clusterrole in order to delegate authentication and authorization to the k8s API server.

The `anonymous-kubefed-admission-webhook-auth` provides the `system:anonymous` user, which is the incoming user making the admission request, the necessary permissions to post to the kubefed webhook API.

pkg/webhook/webhook.go
font: It's because `kubefedcluster` validations are still WIP and will also be moved to its own separate package when it's complete.

pkg/apis/core/v1beta1/validation/validation.go
shashidharatd: Seems `JitterFactor` is agreed upon constant https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L72 and is not exposed by the leader election library. lets keep it like that for now.

xunpan:
Done:
1. duplicated feature checking in `featureGates`
2. add test case for duplicated features
3. reorder test cases with struct definition order

pkg/controller/webhook/kubefedconfig/webhook.go
xunpan: For this one, I guess we need our own group version and it is not expected to change framework passed config argument.

cc @font

font: > What makes this a copy? Is the caller making a copy before calling this function?

The caller creates a new in-cluster config, whose pertinent details (aside from `GroupVersion`, `APIPath`) are shared by all admission webhooks in this server, before invoking each webhook's `Initialize`.

pkg/apis/core/v1beta1/validation/validation.go
marun: Ok, so it should remain non-configurable.  Is there a reason it should not be exposed in the error message though?  I think the value should either be explicitly mentioned in the error message or mention of `JitterFactor` removed.

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -288,3 +292,118 @@ func federatedTypeConfig(apiResource *metav1.APIResource) *v1beta1.FederatedType
...
+	validElectorDuration := validKubeFedConfig()
+	validElectorDuration.Spec.LeaderElect.RenewDeadline.Duration = 12
+	validElectorDuration.Spec.LeaderElect.RetryPeriod.Duration = 10
+	errorCases["spec.leaderElect.renewDeadline: Invalid value"] = validElectorDuration
+
+	validElectorResourceLock := validKubeFedConfig()
+	validElectorResourceLock.Spec.LeaderElect.ResourceLock = "NeitherConfigmapsOrEndpoints"
+	errorCases["spec.leaderElect.resourceLock: Unsupported value"] = validElectorResourceLock
+
+	validFeatureGateName := validKubeFedConfig()
```
marun: This is a bad feature gate name.

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -288,3 +292,118 @@ func federatedTypeConfig(apiResource *metav1.APIResource) *v1beta1.FederatedType
...
+
+func TestValidateKubeFedConfig(t *testing.T) {
+	errs := ValidateKubeFedConfig(validKubeFedConfig())
+	if len(errs) != 0 {
+		t.Errorf("expected success: %v", errs)
+	}
+
+	errorCases := map[string]*v1beta1.KubeFedConfig{}
+
+	validScope := validKubeFedConfig()
```
marun: This is an invalid scope

PR #943, opened 2019-05-29
Retry adding cluster client if one doesn't exist during cluster healthcheck

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/kubefed/issues/878

/cc @font @marun


marun:
@shashidharatd I think this is a good short-term fix.  I'd appreciate it if you would file an issue to refactor the controller to include first-class update handling. This would ensure logging consistency (update vs delete & add) and remove the special handling added to the loop in `updateClusterStatus` (because the entire update would be performed inside a lock).

/lgtm

Issue #929, opened 2019-05-24
Support running controller and webhook as non-root user


alexbrand:
**What would you like to be added**:
Support running controller and webhook as a non-root user.

**Why is this needed**:
Currently, it is not possible to run kubefed in a cluster that uses PodSecurityPolicy to prevent pods from running as the root user. 

/kind feature

marun:
Fixed by #945

PR #945, opened 2019-05-29
adding non-root user to dockerfile

vsamidurai:
adding non-root user to kubefed controller dockerfile #929

### DIFF
--- a/images/kubefed/Dockerfile
+++ b/images/kubefed/Dockerfile
@@ -17,11 +17,14 @@ FROM alpine:latest
 RUN apk --no-cache add ca-certificates
 RUN adduser -D hyperfed
 
+RUN mkdir -p /hyperfed && \
+    chown -R hyperfed:hyperfed /hyperfed
+     
 WORKDIR /hyperfed/
 COPY /hyperfed .
 RUN ln -s hyperfed controller-manager \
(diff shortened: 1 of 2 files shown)


images/kubefed/Dockerfile
```diff
@@ -15,11 +15,13 @@
 
 FROM alpine:latest
 RUN apk --no-cache add ca-certificates
+RUN adduser -D hyperfed
 
-WORKDIR /root/
+WORKDIR /hyperfed/
```
marun: Should this be /home/hyperfed ?  Or should this directory be created and owned by the hyperfood user?

vsamidurai: directory should be owned by hyperfed user. updated docker file with `chown`

marun:
@vsamidurai You'll also need to update the deployment to account for the new location for the `controller-manager` command:

https://github.com/kubernetes-sigs/kubefed/blob/master/charts/kubefed/charts/controllermanager/templates/deployments.yaml#L27

vsamidurai:
Thank you @marun. I updated the helm chart template `controller-manager` and `webhook` command.

marun:
Thank you!

/approve
/lgtm

PR #946, opened 2019-05-29
Cleanup scripts to reflect kubefed rename

marun:
Partial fix for #936

### DIFF
--- a/scripts/deploy-kubefed.sh
+++ b/scripts/deploy-kubefed.sh
@@ -14,33 +14,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script automates deployment of a federation - as documented in
-# the README - to the current kubectl context.  It also joins the
-# hosting cluster as a member of the federation.
+# This script automates deployment of a KubeFed control plane to the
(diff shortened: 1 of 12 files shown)


scripts/deploy-kubefed.sh
```diff
@@ -14,33 +14,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script automates deployment of a federation - as documented in
-# the README - to the current kubectl context.  It also joins the
-# hosting cluster as a member of the federation.
+# This script automates deployment of a KubeFed control plane - as
+# documented in the README - to the current kubectl context.  It also
```
xunpan: `README` seems not documented control plane. It makes no sense for explaination here. Shall we delete "- as  documented in the README - "

It is not related to KubeFed changes. But it is subtle to do another PR.

marun: Fair point.  The README used to document a manual deployment but only helm is supported now.  I'll fix.

marun: Done

marun:
Rebased

PR #948, opened 2019-05-29
More misc cleanup to reflect kubefed rename

marun:
Partial fix for #936

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -121,7 +121,7 @@ func Run(opts *options.Options, stopChan <-chan struct{}) error {
 		klog.Info("KubeFed will target all namespaces")
 	}
 
-	elector, err := leaderelection.NewFederationLeaderElector(opts, startControllers)
+	elector, err := leaderelection.NewKubeFedLeaderElector(opts, startControllers)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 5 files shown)


cmd/controller-manager/app/controller-manager.go
```diff
@@ -244,7 +244,7 @@ func setDefaultKubeFedConfig(fedConfig *corev1b1.KubeFedConfig) {
 		// TODO(sohankunkerkar) Remove when no longer necessary.
 		// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
 		// Its continued existence should not be relied upon.
-		const defaultScopeEnv = "DEFAULT_FEDERATION_SCOPE"
+		const defaultScopeEnv = "DEFAULT_KUBEFED_SCOPE"
```
marun: cc: @sohankunkerkar @pmorie

cmd/kubefedctl/main.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// kubefedctl is a tool for managing clusters in a federation.
+// kubefedctl is a tool for managing a KubeFed control plane.
```
xunpan: I'm not sure that kubefedctl manages the control plane? It is a tool for federation functions support.

marun: I think it's a reasonable generalization to say that kubefedctl manages a control plane given that it supports cluster registration, type enablement, and federation of resources.  In any case, what I have proposed is more accurate than the previous comment.

irfanurrehman:
This LGTM for me. Thanks @marun

Issue #955, opened 2019-05-31
Separate Cluster Registration documentation into its own .md topic


abrennan89:
**What would you like to be added**: 
Similarly to #728 / #730  - I would like to continue breaking the user guide down into content modules, the next being Cluster Registration.

**Why is this needed**:
This would make the documentation around join/unjoin and other commands more manageable and would be a good opportunity to update and add to this content.

This also possibly addresses #613 

/kind feature
/kind documentation

abrennan89:
PR merged

PR #956, opened 2019-05-31
Separated Cluster Registration into a new .md file

abrennan89:
Moved join/unjoin operators into a new doc section and updated links in userguide.md.
Relates to #955 

Will add any new content in an additional PR.

### DIFF
--- a/docs/cluster-registration.md
+++ b/docs/cluster-registration.md
@@ -1,0 +1,80 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Joining Clusters](#joining-clusters)
+- [Checking status of joined clusters](#checking-status-of-joined-clusters)
+      - [Joining kind clusters on MacOS](#joining-kind-clusters-on-macos)
+- [Unjoining clusters](#unjoining-clusters)
+- [Joining additional clusters in a namespace scoped deployment](#joining-additional-clusters-in-a-namespace-scoped-deployment)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Joining Clusters
+
+You can use the `kubefedctl` tool to join clusters as follows.
+
+```bash
+kubefedctl join cluster1 --cluster-context cluster1 \
+    --host-cluster-context cluster1 --v=2
+kubefedctl join cluster2 --cluster-context cluster2 \
+    --host-cluster-context cluster1 --v=2
+```
+
+Repeat this step to join any additional clusters.
+
+**NOTE:** `cluster-context` will default to use the joining cluster name if not
+specified.
+
+# Checking status of joined clusters
+
+Check the status of the joined clusters by using the following command.
+
+```bash
+kubectl -n kube-federation-system get kubefedclusters
+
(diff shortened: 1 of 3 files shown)


docs/cluster-registration.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Joining additional clusters in a namespace scoped deployment
+
+Joining additional clusters to a namespaced control plane requires
+providing additional arguments to `kubefedctl join`:
+
+- `--kubefed-namespace=<namespace>` to ensure the cluster has been registered
+  with the KubeFed control plane running in the specified namespace
+
+To join `mycluster` when `KUBEFED_NAMESPACE=test-namespace` was used for deployment:
```
marun: These instructions appear to be for a control plane deployed by script rather than helm.  Consider replacing 'when \`KUBEFED_NAMESPACE=test-namespace\` was used for deployment' with 'to a control plane deployed in namespace `test-namespace`'.

docs/cluster-registration.md
```diff
@@ -0,0 +1,63 @@
...
+kubefedctl unjoin cluster2 --cluster-context cluster2 --host-cluster-context cluster1 --v=2
+```
+Repeat this step to unjoin any additional clusters.
+
+# Joining additional clusters in a namespace scoped deployment
+
+Joining additional clusters to a namespaced control plane requires
+providing additional arguments to `kubefedctl join`:
+
+- `--kubefed-namespace=<namespace>` to ensure the cluster has been registered
```
marun: The language 'register with a control plane' is in contrast to the 'join to a control plane' language used in the rest of this file. Is this an intentional deviation or would it make sense to use the latter language for consistency?

abrennan89: Can we add any additional changes to a separate issue to review and update the cluster registration doc as a whole? I am wary of creating scope creep if we start making additional changes since this PR was just to get it separated and update the user guide links :)

marun: So long as you're taking notes of the comments and ensure that they are subsequently addressed, I'm fine with that. In code it's common to refactor as needed to best ensure that issues are fixed when discovered.

marun:
My apologies, you'll need to rebase due to my update to the kind/join docs merged with #959.  Likely that content belongs in the new file as well.

docs/cluster-registration.md
abrennan89: I don't have a problem taking note of the comments. I think the difference is that with code it's usually a bug change or some update that has a definite solution and is related to the original PR - with docs if we don't draw the line for what is being covered for each issue, we risk a small PR that was only intended to deal with a small issue being turned into a full on doc review PR, and that makes it very hard to track individual tasks, and also introduces the chance of unnecessary conflicts. For example, once this is separated, if someone makes a change to another section of the userguide.md, it will no longer conflict with my cluster-registration.md changes, but if this PR goes unmerged and someone makes changes, we're letting unnecessary conflicts slip in and risking additional rebases and duplicated work.

docs/cluster-registration.md
```diff
@@ -0,0 +1,80 @@
...
+```bash
+kubectl -n kube-federation-system get kubefedclusters
+
+NAME       READY   AGE
+cluster1   True    1m
+cluster2   True    1m
+
+```
+
+#### Joining kind clusters on MacOS
```
marun: Should this be `##`?

abrennan89: Yeah it should, I just copied it for quickness since I have more editing to do for the other changes and didn't really review for formatting. Will fix in the next PR.

PR #706, opened 2019-03-29
Make installation compatible with macos under docker 18 with kind.

aoxn:
Found some tiny problems when deploy with MacOS.
1. `sed` in place is not compatible with MacOS.
2. create cluster with `kind` name error. should prefix with `kind` as the context showed.
3. Fix cluster.registry address, change from `localhost` to `container ip address`.

Signed-off-by: Aoxn <yaoyao.aoxn@hotmail.com>

PR #959, opened 2019-06-02
Fix macos deployment compatibility

marun:
This is a follow-up to #706 to fix compatibility with the current state of the API and ensure the kind fixup script conforms to the bash conventions used by other scripts.

This PR also:

 - ensures that the `create-servers.sh` script does not attempt to configure a container registry if `CONFIGURE_CONTAINER_REGISTRY` is not set to ensure compatibility with a MacOS deployment environment that hosts the docker daemon in a VM.
 - fixes the download of `golangci-lint` on macos
 - adds support to the kind fixup script for e2e testing on macos with in-memory controllers

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -23,7 +23,8 @@ set -o pipefail
 
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 CREATE_INSECURE_REGISTRY="${CREATE_INSECURE_REGISTRY:-}"
-CONFIGURE_INSECURE_REGISTRY="${CONFIGURE_INSECURE_REGISTRY:-}"
+CONFIGURE_INSECURE_REGISTRY_HOST="${CONFIGURE_INSECURE_REGISTRY_HOST:-}"
+CONFIGURE_INSECURE_REGISTRY_CLUSTER="${CONFIGURE_INSECURE_REGISTRY_CLUSTER:-y}"
 CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
 NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
 OVERWRITE_KUBECONFIG="${OVERWRITE_KUBECONFIG:-}"
@@ -50,7 +51,7 @@ EOF
       err=true
     fi
   elif pgrep -a dockerd | grep -q 'insecure-registry'; then
-    echo <<EOF "Error: CONFIGURE_INSECURE_REGISTRY=${CONFIGURE_INSECURE_REGISTRY} \
+    echo <<EOF "Error: CONFIGURE_INSECURE_REGISTRY_HOST=${CONFIGURE_INSECURE_REGISTRY_HOST} \
 and about to write ${docker_daemon_config}, but dockerd is already configured with \
 an 'insecure-registry' command line option. Please make the necessary changes or disable \
 the command line option and try again."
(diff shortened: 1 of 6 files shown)


scripts/create-clusters.sh
```diff
@@ -113,10 +113,16 @@ function create-clusters() {
     unset KUBECONFIG
   fi
 
-  # TODO(font): Configure insecure registry on kind host cluster. Remove once
-  # https://github.com/kubernetes-sigs/kind/issues/110 is resolved.
-  echo "Configuring insecure container registry on kind host cluster"
-  configure-insecure-registry-on-cluster 1
+  # Avoid configuring the insecure registry unless explicitly
+  # configured to do so. If the docker daemon is running on another
```
font: Can we clarify this setup? For example, is kind being run inside a VM? If so, isn't this script running inside the VM as well? I ask because this logic configures an insecure registry for the cluster i.e. the docker daemon running inside of the control plane container.

scripts/create-clusters.sh
```diff
@@ -113,10 +113,16 @@ function create-clusters() {
...
+  # host (e.g. docker running in a VM on MacOS) the attempt to
+  # configure the registry on the local host may fail.
+  if [[ "${CONFIGURE_INSECURE_REGISTRY}" ]]; then
```
font: This flag was originally intended to configure the insecure registry on the host, which is different from configuring it in the cluster and should be mutually exclusive. That is, you may already have your host configured for an insecure registry already, and do not want the script touching your config, but the clusters' docker config still needs to be set up with the insecure registry option.

If we really need to make it an option to configure an insecure registry in the cluster, I suggest we use a new option e.g. `CONFIGURE_INSECURE_REGISTRY_CLUSTER` and possibly rename this option to `CONFIGURE_INSECURE_REGISTRY_HOST`.

Also, can we turn this on by default? I think if we're doing dev testing, it doesn't hurt to always set up the clusters with an insecure registry config regardless if you end up using it. However, I think we're more than likely to want to use one, even if we don't want the script to automatically configure the host for an insecure container registry (by modifying your config file) either because you don't trust the script or you've already previously set it up. In any case, I'm biased because I always use this. :smile: I'm not sure about others.

marun: This step was failing because kind 0.3.0 switched to containerd. I've added #964 to ensure we update at some point.  For now I've added `CONFIGURE_INSECURE_REGISTRY_HOST` as per your suggestion and defaulted it to true.

My testing workflow involves deploying once and then running in-memory, so the local registry doesn't buy me much and after the ci configuration problem I stopped using it entirely.

font: Ah, ok. Yes we should switch to 0.3.0 at some point in the near future which may make these comments moot.

scripts/create-clusters.sh
```diff
@@ -113,10 +114,12 @@ function create-clusters() {
     unset KUBECONFIG
   fi
 
-  # TODO(font): Configure insecure registry on kind host cluster. Remove once
-  # https://github.com/kubernetes-sigs/kind/issues/110 is resolved.
-  echo "Configuring insecure container registry on kind host cluster"
-  configure-insecure-registry-on-cluster 1
+  if [[ "${CONFIGURE_INSECURE_REGISTRY_HOST}" ]]; then
```
font: This one should be `CONFIGURE_INSECURE_REGISTRY_CLUSTER` and the original `CONFIGURE_INSECURE_REGISTRY` can be renamed to `CONFIGURE_INSECURE_REGISTRY_HOST`.

marun: Done

PR #966, opened 2019-06-04
Update KubeFedConfig API fields to be optional

font:
For fields we plan to provide a default such that we won't require the user to specify them, we need to make them optional in order to pass `kubectl` client side validation. CRD client side validation will be enabled by default in k8s v1.15. See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#publish-validation-schema-in-openapi-v2 for more details.

Additionally, optional fields are made pointers. This is to distinguish the user not specifying a value (since it's not required), from the user specifying an invalid value of `0` (for Duration and time) or `""` for the empty string (should produce a validation error). It also makes serialized objects cleaner.


pkg/apis/core/v1beta1/kubefedconfig_types.go
```diff
@@ -80,19 +91,24 @@ const (
 
 type ClusterHealthCheckConfig struct {
 	// How often to monitor the cluster health (in seconds).
-	PeriodSeconds int64 `json:"periodSeconds"`
```
marun: Not the fault of this PR, but why should time in the health check config be specified as an integer representing seconds but in the leader election config as a duration?  This seems inconsistent to me.

cc: @shashidharatd

shashidharatd: yes, there is inconsistency. it was thought like if we explicitly put unit in the option/flag name itself there will be no confusion while providing the config. Otherwise user has to explicitly provide something like "10s" to specify 10 seconds. If user misses the "s" it is still a valid configuration (in nanoseconds) which leads to undesired value getting configured.

marun: Fair enough. Why, though, is this important for cluster health check but not leader election?  And which is more consistent with user expectations?

shashidharatd: Mostly, it is like that because leader election module uses the `Duration` type and if we change them to `int` we have to convert and pass the parameters https://github.com/kubernetes-sigs/kubefed/blob/master/cmd/controller-manager/app/leaderelection/leaderelection.go#L72

Leader election module is used by multiple components and i think whoever configures them might already be aware of such an issue :) Now we need to make a choice ;)

marun: I think if leader election uses duration across components, is there a reason we shouldn't be consistent and also use duration for leader election?  And if we use duration for that, why wouldn't we use duration for the health check config?

shashidharatd: I am fine with that choice.

font: If we move to use `Duration` I think the confusing thing for users is the use of the unit in the name. So if we switch, I think we should remove the unit in the name for this and for `TimeoutSeconds` with a comment expressing it should be in seconds. Then we'll need to perform validation to make sure it is in seconds. WDYT?

marun: I'm +1 on using duration and removing the unit from the name.  I'm less clear on why we'd need to validate that the value was in seconds, is that a CRD thing?  Whatever duration is provided could be converted to seconds.

font: It's not a CRD thing. It's more of a granularity thing. I figured the reason for requiring seconds is to allow a sane amount of interval delay granularity between cluster health checks as anything less than 1 second may not be reasonable.

marun: I think it's entirely dependent on context.  Maybe partial seconds (e.g. 1.5) would be useful in some circumstances?  I think setting sane defaults and documenting how and why to change those defaults is a reasonable alternative to strict validation.

font: I have no issues with being less opinionated about the level of granularity we allow. We can always enforce a certain level in validation. I think this change should be done in a follow-up PR though since this PR focuses on optional fields.

marun: Please file an issue in the milestone that tracks the transition to use duration for all time fields.

marun:
@font It looks like merging this broke master due to linting errors:

```
+echo Linting
Linting
+golangci-lint run
WARN [runner/megacheck] Can't run megacheck because of compilation errors in packages [sigs.k8s.io/kubefed/pkg/controller/webhook sigs.k8s.io/kubefed/pkg/controller/webhook/federatedtypeconfig sigs.k8s.io/kubefed/pkg/controller/webhook/kubefedconfig sigs.k8s.io/kubefed/pkg/webhook sigs.k8s.io/kubefed/pkg/apis/core/v1beta1/validation [sigs.k8s.io/kubefed/pkg/apis/core/v1beta1/validation.test]]: pkg/apis/core/v1beta1/validation/validation.go:174: cannot convert elect.ResourceLock (variable of type *v1beta1.ResourceLockType) to string and 32 more errors: run `golangci-lint run --no-config --disable-all -E typecheck` to see all errors 
pkg/apis/core/v1beta1/validation/validation.go:174:88: cannot convert elect.ResourceLock (variable of type *v1beta1.ResourceLockType) to string (typecheck)
	allErrs = append(allErrs, validateEnumStrings(electPath.Child("resourceLock"), string(elect.ResourceLock),
	                                                                                      ^
pkg/apis/core/v1beta1/validation/validation.go:198:84: cannot use health.PeriodSeconds (variable of type *int64) as int64 value in argument to validateGreaterThan0 (typecheck)
	allErrs = append(allErrs, validateGreaterThan0(healthPath.Child("periodSeconds"), health.PeriodSeconds)...)
	                                                                                  ^
pkg/apis/core/v1beta1/validation/validation.go:199:87: cannot use health.FailureThreshold (variable of type *int64) as int64 value in argument to validateGreaterThan0 (typecheck)
	allErrs = append(allErrs, validateGreaterThan0(healthPath.Child("failureThreshold"), health.FailureThreshold)...)
	                                                                                     ^
pkg/apis/core/v1beta1/validation/validation_test.go:341:59: cannot convert "NeitherConfigmapsOrEndpoints" (untyped string constant) to *v1beta1.ResourceLockType (typecheck)
	validElectorResourceLock.Spec.LeaderElect.ResourceLock = "NeitherConfigmapsOrEndpoints"
	                                                         ^
pkg/apis/core/v1beta1/validation/validation_test.go:361:73: cannot convert 0 (untyped int constant) to *int64 (typecheck)
	validPeriodSecondsGreaterThan0.Spec.ClusterHealthCheck.PeriodSeconds = 0
	                                                                       ^
pkg/apis/core/v1beta1/validation/validation_test.go:377:59: cannot convert "NeitherEnableOrDisable" (untyped string constant) to *v1beta1.ResourceAdoption (typecheck)
	validAdoptResources.Spec.SyncController.AdoptResources = "NeitherEnableOrDisable"
```

I'm very much confused as to how CI managed to pass for this PR.

marun:
Lint errors are addressed in #973

PR #970, opened 2019-06-05
Use html for too wide tables

databill:
It is hard for reading and maintenance.

Signed-off-by: Julien <julienjut@gmail.com>

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -8,14 +8,66 @@
 
 ### KubeFed Concepts
 
-| Concept              | Description                                                                                                                                                                                                                                                                                                         |
-| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| Federate             | Federating a set of Kubernetes clusters means, effectively creating a common interface to the pool of these clusters which can be used to deploy Kubernetes applications across those clusters.                                                                                                                     |
-| KubeFed           | Kubernetes Cluster Federation enables users to federate multiple Kubernetes clusters for resources distribution, service discovery, high availability etc across multiple clusters.                                                                                                                                 |
-| Host Cluster         | A cluster which is used to expose the KubeFed API and run the KubeFed control plane.                                                                                                                                                                                                                          |
-| Cluster Registration | A cluster join the Host Cluster via command `kubefedctl join`.                                                                                                                                                                                                                                                        |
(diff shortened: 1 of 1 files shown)


xunpan:
Agree that the changed code is better in reading and maintenance.
Could you please also help to change wide table in `charts/kubefed/README.md`

marun:
While I appreciate your initiative, this repo's docs use markdown to ensure that contributors can focus on content rather than formatting.  Accordingly, I do not think this proposal is acceptable and would ask that you close it.

If you'd like to assist in docs cleanup, I would ask that you contact @abrennan89.  She is actively working on revising the content and contribution not discussed with her is unlikely to be accepted at this time.

abrennan89:
Agreed - diverging from .md formatting is introducing unnecessary complications to doc maintenance,  and would also likely cause issues for any content reuse from this repo for github pages, kubernetes.io integration or other 'prettier' publishing later, as the source files for that are generally markdown. This would duplicate work on conversion efforts we don't need to do.

databill:
> Agree that the changed code is better in reading and maintenance.
> Could you please also help to change wide table in `charts/kubefed/README.md`

If we can get agreed I will. I just read this doc currently. I want to fix some minor issue but it caused me too much time to read.

databill:
> Agreed -

All the markdown formats(including basic and extended) support embedded HTML format. It is also supported by github, even referred by rst docs. The basic syntax is listed in [basic-syntax](https://www.markdownguide.org/basic-syntax/), and the 
notes are here:

    Using Markdown doesn't mean that you can't also use HTML. You can add HTML tags to any 
    Markdown file. This is helpful if you prefer certain HTML tags to Markdown syntax. For example, 
    some people find that it's easier to use HTML tags for images.

Using markdown is absolutely convenient for maintenance and for developers. Markdown won't support multiple-lines format table content when I contacted the author several years ago.

I usually edit the docs in the terminal, but it is too wide and it is not easy for me to read and fix.

xunpan:
I think html tags is suitable for this case. 
MD format is pure text format. It is not user friendly for wide tables using MD syntax. 
As @databill pointed out, it is acceptable for html tag in MD.
In this case, html table tags makes MD file clear in both text only and rendered result display. 

@marun @abrennan89

marun:
@xunpan @databill The authority on this issue is @abrennan89 and I encourage you to re-read her comment.  Regardless of your personal preference, using html to define tables is not the way forward.

PR #971, opened 2019-06-05
Create default FeatureGateConfig for KubeFedConfig

abhat:
When using the kubefed-operator-v2 to deploy Kubefed, if there is no KubeFedConfig provided by the user, the one that is default generated by KubeFed doesn't initialize the required featuregates array in the spec. resulting in 
```
F0605 16:12:09.927814       1 controller-manager.go:299] Error creating KubeFedConfig "federation-test/kubefed": KubeFedConfig.core.kubefed.k8s.io "kubefed" is invalid: []: Invalid value: map[string]interface {}{"kind":"KubeFedConfig", "apiVersion":"core.kubefed.k8s.io/v1beta1", "metadata":map[string]interface {}{"name":"kubefed", "namespace":"federation-test", "creationTimestamp":"2019-06-05T16:12:09Z", "generation":1, "uid":"abe0c833-87ac-11e9-9d90-02421265cfcd"}, "spec":map[string]interface {}{"scope":"", "controllerDuration":map[string]interface {}{"availableDelay":"20s", "unavailableDelay":"1m0s"}, "leaderElect":map[string]interface {}{"leaseDuration":"15s", "renewDeadline":"10s", "retryPeriod":"5s", "resourceLock":"configmaps"}, "featureGates":interface {}(nil), "clusterHealthCheck":map[string]interface {}{"periodSeconds":10, "failureThreshold":3, "successThreshold":1, "timeoutSeconds":3}, "syncController":map[string]interface {}{"adoptResources":"Enabled"}}}: validation failure list:
spec.featureGates in body must be of type array: "null" 
```

This PR is to fix the validation error and making sure the generated KubeFedConfig satisfies the `required` spec fields.

Note that there is PR #966 by @font that makes some of these fields optional. So this maybe a stop gap fix.

/kind bug


marun:
@abhat You'll need to sign the CLA as per the failing check. 

Also, please explain to potential reviewers what your intention is (i.e. what is being changed and why).  This is best practice for upstream PRs.

abhat:
> @abhat You'll need to sign the CLA as per the failing check.
> 
> Also, please explain to potential reviewers what your intention is (i.e. what is being changed and why). This is best practice for upstream PRs.

Yes, @marun will do, this was meant to be a draft PR. Not ready for review.

xunpan:
Let me add more information for references.

`pkg/features/features.go::init()` enable all features by default in code. This is the reason that the default setting of configuration does not set feature gates. But I think it should be better maintenance if we call default feature settings together in one function. But make sure removing useless code in `init()` after that.

Thoughts?

abhat:
> Let me add more information for references.
> 
> `pkg/features/features.go::init()` enable all features by default in code. This is the reason that the default setting of configuration does not set feature gates. But I think it should be better maintenance if we call default feature settings together in one function. But make sure removing useless code in `init()` after that.
> 
> Thoughts?

@xunpan if the default KubeFedConfig has featuregates as `required` we ought to set it? Agree about better maintenance, however, is there any side effect of not adding the featuregates to DefaultFeatureGate post module init?

marun:
@font I leave it to you as to whether there is a point to this merging given the proposed webhook defaulting.

font:
@abhat Since `featureGates` is now optional, is this still necessary?

abhat:
@font - if your recent changes make featureGates optional, then we can close this PR.

PR #974, opened 2019-06-08
Add admission webhook support for namespace-scoped deployments

### LINKED ISSUE #965: Add admission webhook support for namespace-scoped deployments
Add support for running the kubefed admission webhook on a per-namespace basis.

**Why is this needed**:
To support kubefed namespace-scoped deployments.

/kind feature

font:
Fixes #965.

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
@@ -27,6 +27,8 @@ subjects:
   name: kubefed-admission-webhook
   namespace: {{ .Release.Namespace }}
 ---
+# This clusterrolebinding grants permissions for the admission webhook to create
+# admission reviews on behalf of the system:anonymous user.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
--- a/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
@@ -17,7 +17,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: auth-delegator-kubefed-admission-webhook
+  name: kubefed-admission-webhook:auth-delegator
 roleRef:
   kind: ClusterRole
   apiGroup: rbac.authorization.k8s.io
@@ -32,7 +32,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: anonymous-kubefed-admission-webhook-auth
+  name: kubefed-admission-webhook:anonymous-auth
 roleRef:
   kind: ClusterRole
   apiGroup: rbac.authorization.k8s.io
--- a/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
@@ -17,7 +17,14 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
+# For namespace scoped deployments, create a unique cluster-scoped resource
+# name using the namespace. This is needed because admission-webhooks do not
+# currently support the ability to have namespace-scoped RBAC permissions only.
+{{- if and .Values.global.scope (eq .Values.global.scope "Namespaced") }}
+  name: kubefed-admission-webhook:{{ .Release.Namespace }}:auth-delegator
+{{ else }}
   name: kubefed-admission-webhook:auth-delegator
+{{ end }}
 roleRef:
   kind: ClusterRole
   apiGroup: rbac.authorization.k8s.io
@@ -32,11 +39,24 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
+# TODO(font) For namespace scoped deployments, create a unique cluster-scoped
+# resource name using the namespace. This is needed because helm does not
+# currently support the ability to share resources across multiple
+# installations of the same chart. Additionally, admission-webhooks do not
+# currently support the ability to have namespace-scoped RBAC permissions only.
+{{- if and .Values.global.scope (eq .Values.global.scope "Namespaced") }}
+  name: kubefed-admission-webhook:{{ .Release.Namespace }}:anonymous-auth
+{{ else }}
   name: kubefed-admission-webhook:anonymous-auth
+{{ end }}
 roleRef:
   kind: ClusterRole
   apiGroup: rbac.authorization.k8s.io
+{{- if and .Values.global.scope (eq .Values.global.scope "Namespaced") }}
+  name: system:kubefed:{{ .Release.Namespace }}:admission-requester
+{{ else }}
   name: system:kubefed:admission-requester
+{{ end }}
 subjects:
 - apiGroup: rbac.authorization.k8s.io
   kind: User
--- a/charts/kubefed/charts/controllermanager/templates/roles.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/roles.yaml
@@ -82,6 +82,8 @@ rules:
   verbs:
   - get
 ---
+# Only need access to these core namespaced resources in the KubeFed system
+# namespace regardless of kubefed deployment scope.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
--- a/charts/kubefed/charts/controllermanager/templates/roles.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/roles.yaml
@@ -96,7 +96,9 @@ rules:
 - apiGroups:
   - core.kubefed.k8s.io
   resources:
-  - '*'
+  - federatedtypeconfigs
+  - kubefedclusters
+  - kubefedconfigs
   verbs:
   - get
   - watch
--- a/charts/kubefed/charts/controllermanager/templates/webhook.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/webhook.yaml
@@ -6,7 +6,7 @@
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: "federatedtypeconfigs.core.kubefed.k8s.io"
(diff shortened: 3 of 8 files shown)


charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
```diff
@@ -27,14 +34,29 @@ subjects:
   name: kubefed-admission-webhook
   namespace: {{ .Release.Namespace }}
 ---
+# This clusterrolebindng grants permissions for the admission webhook to create
+# admission reviews on behalf of the system:anonymous user.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: anonymous-kubefed-admission-webhook-auth
+# TODO(font) For namespace scoped deployments, create a unique cluster-scoped
```
marun: Did you intend to leave this TODO here?  The comment looks the same as the one you've added above.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
```diff
@@ -27,14 +34,29 @@ subjects:
   name: kubefed-admission-webhook
   namespace: {{ .Release.Namespace }}
 ---
+# This clusterrolebindng grants permissions for the admission webhook to create
```
marun: s/clusterrolebindng/clusterrolebinding/

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
```diff
@@ -17,7 +17,14 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: auth-delegator-kubefed-admission-webhook
+# For namespace scoped deployments, create a unique cluster-scoped resource
```
marun: (No action required) Given that this comment applies for each of the 3 conditionals that appear in this file, consider referencing it in the other 2 rather than duplicating the comment.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
```diff
@@ -17,7 +17,14 @@ subjects:
...
+# name using the namespace. This is needed because admission-webhooks do not
+# currently support the ability to have namespace-scoped RBAC permissions only.
+{{- if and .Values.global.scope (eq .Values.global.scope "Namespaced") }}
+  name: kubefed-admission-webhook:{{ .Release.Namespace }}:auth-delegator
+{{ else }}
+  name: kubefed-admission-webhook:auth-delegator
```
marun: Is there a reason not to qualify the name of the cluster-scoped delegator with the system namespace, so that the name form would be consistent across deployment modes?  This comment applies to other instances of namespace-qualifying only the resources created for a namespaced deployment rather than namespace-qualifying all resources created for either type of deployment.

charts/kubefed/charts/controllermanager/templates/roles.yaml
```diff
@@ -94,7 +96,9 @@ rules:
 - apiGroups:
   - core.kubefed.k8s.io
   resources:
-  - '*'
+  - federatedtypeconfigs
```
marun: Why is it necessary to desirable to be explicit to a the non-generated core group?  The approach you're proposing will require explicitly changing permissions when new types are added, which is something @xunpan and I tried to avoid in using `*`.

scripts/deploy-kubefed.sh
```diff
@@ -160,10 +160,6 @@ cd "$(dirname "$0")/.."
 make kubefedctl
 cd -
 
-if ! kubectl get ns "${NS}" > /dev/null 2>&1; then
```
marun: Does this mean that helm will create the namespace now?  I'm not clear whether this was required for helm or a vestige of the script-based deployment.

charts/kubefed/charts/controllermanager/templates/webhook.yaml
```diff
@@ -17,22 +23,25 @@ webhooks:
...
-  name: "kubefedclusters.core.kubefed.k8s.io"
-webhooks:
+{{- if and .Values.global.scope (eq .Values.global.scope "Namespaced") }}
+# For namespace scoped deployments: filter admission webhook requests for
+# resources whose namespace matches the default namespace label applied by helm
+# upon creating the namespace. User must set this label manually if namespace
+# is created prior to installing chart with helm.
+  namespaceSelector:
+    matchLabels:
+      name: {{ .Release.Namespace }}
```
marun: Is it possible to use a more specific label?  I'm not sure it will be obvious to a user that a label `name: <namespace name>` is going to be critical to the functioning of the KubeFed API.  Ideally it would be possible to use something more specific (e.g. `kubefed-namespace: <namespace name>`).

scripts/deploy-kubefed.sh
xunpan: Yes. `helm install --namespace` will create the namespace on demand
https://github.com/helm/helm/issues/4456#issuecomment-412134651

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
font: It certainly crossed my mind as it would have been easier, but then it wouldn't be consistent with the other non-webhook cluster-scoped resource names. Also, having the same cluster-scoped resource names helps prevent duplicate cluster-scoped kubefed deployments.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
font: I felt it was cleaner to just copy/paste the comment since it's a manifest file that shouldn't really change instead of providing references across files and lines where context may be lost. Note that there are only 2 conditionals in this file where only part of this message is duplicated.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
font: The comment is slightly different as it talks about the lack of support in helm for sharing resources across charts. I left it as a TODO since helm may eventually support it via https://github.com/helm/helm/issues/3189.

charts/kubefed/charts/controllermanager/templates/roles.yaml
font: This `Role` is specific to the admission webhook and aids in making the admission webhook run with least privileges. We already do not require access to all the core API types in the webhook, so I'm not sure we should enable access to all of them. Also, it doesn't seem too onerous to add an additional core type if/when we need to support defaulting or validating one in the webhook, especially since it's a non-generated group. Is it necessary to have the webhook wildcard this to all resources in the core group?

charts/kubefed/charts/controllermanager/templates/webhook.yaml
```diff
@@ -6,7 +6,13 @@
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: "federatedtypeconfigs.core.kubefed.k8s.io"
+# For namespace scoped deployments, create a unique cluster-scoped resource
+# using the namespace.
+{{- if and .Values.global.scope (eq .Values.global.scope "Namespaced") }}
+  name: validations.admission.core.kubefed.k8s.io-{{ .Release.Namespace }}
```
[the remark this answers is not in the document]
font: I thought about doing as you suggest, but didn't want users confusing it as part of the fully qualified group.

charts/kubefed/charts/controllermanager/templates/webhook.yaml
font: That was my original goal, but unfortunately, it is not currently possible with helm (see https://github.com/helm/helm/issues/4178 for more details). However, this suggests I should add a TODO about this to clarify.

font:
@marun @xunpan Thanks for your review! I've responded to questions and comments and pushed changes to address a couple comments. PTAL.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
marun: I don't think the latter argument is worth worrying about - if we're to avoid duplicate deployments lets make it explicit.

Regarding naming consistency, afaik we already use consistent naming for some resources and maybe we should use that as the norm?  cc: @xunpan

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
marun: Hmm, ok.  I was confused because there are large comments with little information as per my comment about duplication, and it wasn't clear to me that this one was different.  Maybe its worth rewording so that it's clear that it is not just boilerplate like the other comments?

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
marun: As per my comment above, the danger of copy-pasting large blocks of duplicative comments is that they end up not being read exhaustively.  I would prefer a bias towards relevant rather than exhaustive commentary.

charts/kubefed/charts/controllermanager/templates/roles.yaml
marun: In what sense does allowing the webhook access to all core types (validation for each of which should be mandatory going forward) something likely to violate the principle of least privilege?

font: The webhook already doesn't validate all the resources in the group (e.g. `PropagatedVersion`, `ClusterPropagatedVersion`, `FederatedServiceStatus`), so by principle of least privilege, it should only be made to access the necessary resources. Using a wildcard would violate this today.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
font: This one has a TODO so I think it is already different in that regard. But I can change the wording if you feel strongly.

charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml
font: Since we do not currently support explicitly avoiding duplicate deployments, I think blocking it implicitly by way of duplicate resource names inherent in the design of our helm chart, along with some documentation that multiple cluster-scoped deployments should be avoided altogether, seems like a reasonable alternative.

In any case, if we want to change all the cluster-scoped resource names, let's open up a new PR to address it.

marun:
Thank you!

/lgtm

PR #979, opened 2019-06-12
 Update KubeFedConfig API time fields

shashidharatd:
Fixes: https://github.com/kubernetes-sigs/kubefed/issues/975

/cc @font @marun @xunpan


shashidharatd:
> what about successThreshold and failureThreshold?

They are threshold and not time based. their unit in count (int) and does not need change.

> helm chart related item should be updated accordingly.

missed it earlier. Added a new commit to update them.
PTAL again. Thanks !

xunpan:
LGTM

marun:
Thank you!

/lgtm

PR #990, opened 2019-06-14
Fix disabling insecure registry in create clusters script

marun:
`CONFIGURE_INSECURE_REGISTRY_CLUSTER="${CONFIGURE_INSECURE_REGISTRY_CLUSTER:-y}"` will substitute 'y' if the value is unset or null, which prevents disabling this configuration step. Removing the colon ensures that 'y' is substituted only if the value is unset.


font:
LGTM

xunpan:
/lgtm

I found DOCKER_PUSH also has the same issue. I will handle it.

PR #991, opened 2019-06-14
Replace override paths with JSON pointer style paths

shashidharatd:
Please refer to the design doc [here](https://docs.google.com/document/d/1nUC4H7SsxhZPMO2aKap5GG6ZJoqdxbxaUvzlSt_dfuc/edit#)

Ref issue: https://github.com/kubernetes-sigs/kubefed/issues/520

todos:
- [x] Add tests
- [x] Update docs and examples

/cc @marun

### DIFF
--- a/pkg/controller/util/overrides.go
+++ b/pkg/controller/util/overrides.go
@@ -58,24 +58,17 @@ var invalidPaths = sets.NewString(
 	"/metadata/generateName",
 )
 
-// Mapping of qualified path (e.g. spec.replicas) to value
-type ClusterOverridesMap map[string]interface{}
+// Slice of ClusterOverride
+type ClusterOverrides []ClusterOverride
 
 // Mapping of clusterName to overrides for the cluster
-type OverridesMap map[string]ClusterOverridesMap
+type OverridesMap map[string]ClusterOverrides
 
 // ToUnstructuredSlice converts the map of overrides to a slice of
 // interfaces that can be set in an unstructured object.
 func (m OverridesMap) ToUnstructuredSlice() []interface{} {
 	overrides := []interface{}{}
-	for clusterName, clusterOverridesMap := range m {
-		clusterOverrides := []map[string]interface{}{}
-		for path, value := range clusterOverridesMap {
-			clusterOverrides = append(clusterOverrides, map[string]interface{}{
-				PathField:  path,
-				ValueField: value,
-			})
-		}
+	for clusterName, clusterOverrides := range m {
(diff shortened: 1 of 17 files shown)


test/common/crudtester.go
```diff
@@ -564,8 +564,10 @@ func (c *FederatedTypeCrudTester) waitForResource(client util.ResourceClient, qu
 
 			// Validate that the expected override was applied
 			if len(expectedOverrides) > 0 {
+				// TODO: use the json patch library to patch with the overrides and then compare it with cluster object.
```
marun: The CI failure suggests that you may need to implement this strategy to support testing of overriding a path that embeds an index.  The nested field copy won't work for that case.

shashidharatd:
finally the mechanim to test json-patch style override is working. @marun, ptal once again. Thanks !

marun:
Nice work, LGTM!

shashidharatd:
@marun, PTAL. Thanks !

pkg/kubefedctl/enable/validation.go
```diff
@@ -107,6 +107,9 @@ func federatedTypeValidationSchema(templateSchema map[string]v1beta1.JSONSchemaP
 									Schema: &v1beta1.JSONSchemaProps{
 										Type: "object",
 										Properties: map[string]v1beta1.JSONSchemaProps{
+											"op": {
```
marun: Would it make sense to validate for the specific values that are supported?

font: Should we consider making at least `path` and `value` fields required?

pkg/controller/util/overrides.go
```diff
@@ -27,8 +28,9 @@ import (
 )
 
 type ClusterOverride struct {
+	Op    string      `json:"op,omitempty"`
 	Path  string      `json:"path"`
-	Value interface{} `json:"value"`
+	Value interface{} `json:"value,omitempty"`
```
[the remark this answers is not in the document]
marun: I'm assuming the reason for it being optional is that `interface{}` is implicitly a pointer.  I don't think there is much value in worrying about this struct since it is just a convenience to enable typed use rather than a public interface.  I share your concern about the validation schema, though - these fields really should be required.

pkg/kubefedctl/enable/validation.go
marun: Path can be required but value is required by `replace` and `add` but not `remove`.

PR #994, opened 2019-06-16
Add explanation, how to configure PSP for the kubefed namespace on ICP

roytman:
IBM Private Cloud default PSP prevents running pods as `root`. This PR adds explanation how to configure ICP's PSP.

### DIFF
--- a/docs/environments/icp.md
+++ b/docs/environments/icp.md
@@ -36,7 +36,7 @@ For the second cluster, set the following value in `cluster/config.yaml` as foll
 cluster_name: cluster2
 ```
 
-## Post Install Configuration
+## Pre Install Configuration
 
 As IBM Cloud Private is [enforcing container image security](https://www.ibm.com/support/knowledgecenter/SSBS6K_3.1.1/manage_images/image_security.html)
 policy by default, and the default image security policy does not allow pulling the KubeFed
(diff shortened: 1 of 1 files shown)


docs/environments/icp.md
```diff
@@ -36,7 +36,7 @@ For the second cluster, set the following value in `cluster/config.yaml` as foll
 cluster_name: cluster2
 ```
 
-## Post Install Configuration
+## Pre Install Configuration
```
gyliu513: Here should be `post` I think as those operations are after ICP install finished.

roytman: No, the namespace should be created AND CONFIGURED, before helm installation. If the namespace is created by the helm, it will have the default PSP. It can be configure later, but it is more complicated process. 
The `clusterimagepolicies` can be edited before or after helm installation, but better to do it before, so the Pod creation will not fail. Otherwise, it will fail and will be restarted after `clusterimagepolicies` editing.

gyliu513: I see, please run `doctoc` to update ToC of this file.

roytman: To prevent ambiguity, renamed the title to "Pre KubeFed Install Configuration", `ran doctoc`

marun:
@gyliu513 @roytman Please note that with the resolution of #929, KubeFed no longer needs to run as root.  Upon the next release of KubeFed, which will include the fix for #929, you may want to revert the documentation change introduced in this PR to ensure that the KubeFed system namespace remains minimally privileged.

roytman:
Thanks @marun, it was one of my questions, why KubeFed pods run as root. Indeed this PR should be reverted before the next release.

PR #995, opened 2019-06-17
when DOCKER_PUSH= is set, the script should skip docker push

xunpan:
fix issues similar to #990. Use `-` instead of `:-` in variable expansion: when `DOCKER_PUSH` is set empty, the script should not set `y` but keep it as is.


marun:
Great, thanks!

/approve
/lgtm

PR #997, opened 2019-06-17
Add .idea files to gitignore

wilhelmguo:
Add idea ignore files who use goland.


marun:
While I appreciate your initiative in posting this PR, I don't think it desirable or necessary to ignore local developer-specific files.  That's what [git's global ignore configuration is for](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore).

PR #998, opened 2019-06-18
Remove unused cluster registry dep

marun:
As of #876 kubefed no longer depends on cluster registry.


shashidharatd:
Nice Cleanup! Thanks for this.
LGTM.

xunpan:
LGTM

I guess prune rule in `Gopkg.toml` should also be cleaned.

marun:
> I guess prune rule in `Gopkg.toml` should also be cleaned.

Done!

PR #1007, opened 2019-06-19
configure images with a UID and set SC in deployments

andrewmyhre:
- the hyperfed user is create with UID 1001
- the controller and webhook pods are run with a security context setting `runAsUser: 1001` so that kubernetes knows it's not running as a root user (supporting restricted pod security policies)
- also create the `hyperfed` user in a named group which seems to be required for `chown` to apply correctly
- also execute `chown` after copying binaries to hyperfed path as the binaries were ending up owned by `root` in my testing


marun:
Great, thank you!

/approve
/lgtm

Issue #690, opened 2019-03-24
Facilitate adding the annotation on federated object to specify users intent of orphaning/not orphaning target/managed resources.


irfanurrehman:
**What would you like to be added**:
https://github.com/kubernetes-sigs/federation-v2/pull/669 moves away from the usage of `orphanDependents` and uses intent via an annotation on the object to signify whether the dependent resources should be orphaned or not.
This seems to be an additional step, where user will need to create the annotation on the federated object if a non default behaviour is needed.

Suggestion is to facilitate adding the annotation using `kubefed2` before deletion using `kubectl`.
Implement `kubefed2 enable-orphaning-deletion <federated-resource-name>`

**Why is this needed**:

Ease of usage to the user.

/kind feature

irfanurrehman:
A more suitable name then `enable-orphaning-deletion` for the command is welcome.

roytman:
If nobody is working on it, I'll do it

marun:
@roytman Hi!  You're welcome to work on this.  The suggested form would be more like:

```
kubefedctl orphaning-deletion <enable|disable> <resource type> <resource name> [-n namespace]
```

Alternately, you could look at one of the higher-priority items in the [v0.2.0](https://github.com/kubernetes-sigs/kubefed/milestone/5) (beta2) milestone.

marun:
Fixed by #1008

Issue #828, opened 2019-05-01
Enable `kubefedctl` testing


marun:
Currently what little testing we have for `kubefedctl` is largely implicit testing via use in e2e setup (e.g. join/unjoin) or e2e execution of internal methods.  Given the large number of additions and improvements to the cli tool, a proper testing framework is suggested that makes it easy to test binary invocation directly.  The deliverable for this issue would be:

 - a framework for invoking `kubefedctl` and evaluating its output
 - one or more reference tests that use the framework

PR #1008, opened 2019-06-20
Implement orphaning-deletion commands

roytman:
Fix for #690

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -54,12 +54,6 @@ const (
 	// controller will have the opportunity to perform pre-deletion operations
 	// (like deleting managed resources from member clusters).
 	FinalizerSyncController = "kubefed.k8s.io/sync-controller"
-
-	// If this annotation is present on a federated resource, resources in the
-	// member clusters managed by the federated resource should be orphaned.
-	// If the annotation is not present (the default), resources in member
-	// clusters will be deleted before the federated resource is deleted.
-	OrphanManagedResources = "kubefed.k8s.io/orphan"
 )
 
 // KubeFedSyncController synchronizes the state of federated resources
@@ -435,13 +429,12 @@ func (s *KubeFedSyncController) ensureDeletion(fedResource FederatedResource) ut
 		return util.StatusAllOK
 	}
 
-	annotations := obj.GetAnnotations()
-	orphanResources := annotations != nil && annotations[OrphanManagedResources] == "true"
-	if orphanResources {
-		klog.V(2).Infof("Found %q annotation on %s %q. Removing the finalizer.", OrphanManagedResources, kind, key)
+	if util.IsOrphaningEnabled(obj) {
+		klog.V(2).Infof("Found %q annotation on %s %q. Removing the finalizer.",
+			util.OrphanManagedResourcesAnnotation, kind, key)
 		err := s.removeFinalizer(fedResource)
 		if err != nil {
-			wrappedErr := errors.Wrapf(err, "failed to remove finalizer %q from %s %q", OrphanManagedResources, kind, key)
+			wrappedErr := errors.Wrapf(err, "failed to remove finalizer %q from %s %q", FinalizerSyncController, kind, key)
 			runtime.HandleError(wrappedErr)
 			return util.StatusError
 		}
--- a/pkg/controller/util/orphaninganotation.go
+++ b/pkg/controller/util/orphaninganotation.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+const (
+	// If this annotation is present on a federated resource, resources in the
+	// member clusters managed by the federated resource should be orphaned.
+	// If the annotation is not present (the default), resources in member
+	// clusters will be deleted before the federated resource is deleted.
+	OrphanManagedResourcesAnnotation = "kubefed.k8s.io/orphan"
+	OrphanedManagedResourcesValue    = "true"
+)
+
+// IsOrphaningEnabled checks status of "orphaning enable" (OrphanManagedResources: OrphanedManagedResourceslValue')
+// annotation on a resource.
+func IsOrphaningEnabled(obj *unstructured.Unstructured) bool {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		return false
+	}
+	return annotations[OrphanManagedResourcesAnnotation] == OrphanedManagedResourcesValue
+}
+
+// Enables the orphaning mode
+func EnableOrphaning(obj *unstructured.Unstructured) {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	annotations[OrphanManagedResourcesAnnotation] = OrphanedManagedResourcesValue
+	obj.SetAnnotations(annotations)
+}
+
+// Disables the orphaning mode
+func DisableOrphaning(obj *unstructured.Unstructured) {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		return
+	}
+	delete(annotations, OrphanManagedResourcesAnnotation)
+	obj.SetAnnotations(annotations)
+}
--- a/pkg/kubefedctl/federate/util.go
+++ b/pkg/kubefedctl/federate/util.go
@@ -20,7 +20,6 @@ import (
 	"bufio"
 	"io"
 	"os"
-	"strings"
 
 	"github.com/pkg/errors"
 
@@ -36,7 +35,7 @@ import (
 	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
 	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
 	"sigs.k8s.io/kubefed/pkg/kubefedctl/enable"
-	"sigs.k8s.io/kubefed/pkg/kubefedctl/options"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
 )
 
 var systemMetadataFields = []string{"selfLink", "uid", "resourceVersion", "generation", "creationTimestamp", "deletionTimestamp", "deletionGracePeriodSeconds"}
@@ -106,7 +105,7 @@ func namespacedAPIResourceMap(config *rest.Config, skipAPIResourceNames []string
 		}
 
 		for _, apiResource := range apiResourceList.APIResources {
-			if !apiResource.Namespaced || isFederatedAPIResource(apiResource.Kind, group) ||
+			if !apiResource.Namespaced || util.IsFederatedAPIResource(apiResource.Kind, group) ||
 				apiResourceMatchesSkipName(apiResource, skipAPIResourceNames, group) {
 				continue
 			}
@@ -159,10 +158,6 @@ func apiResourceMatchesSkipName(apiResource metav1.APIResource, skipAPIResourceN
 		}
 	}
 	return false
-}
-
-func isFederatedAPIResource(kind, group string) bool {
-	return strings.HasPrefix(kind, federatedKindPrefix) && group == options.DefaultFederatedGroup
 }
 
 // resources stores a list of resources for an api type
--- a/pkg/kubefedctl/orphaning/disable.go
+++ b/pkg/kubefedctl/orphaning/disable.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package orphaning
+
+import (
+	"io"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_disable_long = `
+		Removes previously added "orphaning enable" ('kubefed.k8s.io/orphan: true')
+		annotation from a federated resource.  When the federated resource is subsequently marked for deletion,
+		the resources it manages in member clusters will be removed before the federated resource is removed.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphaning_disable_example = `
+		# Disable the orphaning mode for a federated resource of type FederatedDeployment and named foo 
+		kubefedctl orphaning disable FederatedDeployment foo --host-cluster-context=cluster1`
+)
+
+// newCmdDisableOrphaning removes the 'kubefed.k8s.io/orphan: true' annotation from the federated resource
+func newCmdDisableOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	opts := &orphanResource{}
+	cmd := &cobra.Command{
+		Use:     "disable <resource type> <resource name>",
+		Short:   "Disable orphaning deletion to ensure the removal of managed resources before removing the managing federated resource",
+		Long:    orphaning_disable_long,
+		Example: orphaning_disable_example,
+		Run: func(cmd *cobra.Command, args []string) {
+			err := opts.Complete(args)
+			if err != nil {
+				klog.Fatalf("Error: %v", err)
+			}
+
+			err = opts.RunDisable(cmdOut, config)
+			if err != nil {
+				klog.Fatalf("Error: %v", err)
+			}
+		},
+	}
+
+	flags := cmd.Flags()
+	opts.GlobalSubcommandBind(flags)
+	err := opts.Bind(flags)
+	if err != nil {
+		klog.Fatalf("Error: %v", err)
+	}
+
+	return cmd
+}
+
+// RunDisable implements the `disable` command.
+func (o *orphanResource) RunDisable(cmdOut io.Writer, config util.FedConfig) error {
+	resourceClient, err := o.GetResourceClient(config, cmdOut)
+	if err != nil {
+		return err
+	}
+	fedResource, err := o.GetFederatedResource(resourceClient)
+	if err != nil {
+		return err
+	}
+	if !ctlutil.IsOrphaningEnabled(fedResource) {
+		return nil
+	}
+	ctlutil.DisableOrphaning(fedResource)
+	_, err = resourceClient.Update(fedResource, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Failed to update resource %s %q", fedResource.GetKind(),
+			ctlutil.QualifiedName{Name: fedResource.GetName(), Namespace: fedResource.GetNamespace()})
+	}
+	return nil
+}
--- a/pkg/kubefedctl/orphaning/enable.go
+++ b/pkg/kubefedctl/orphaning/enable.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package orphaning
+
+import (
+	"io"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_enable_long = `
+		Prevents the removal of managed resources from member clusters when their managing federated 
+		resource is removed. This is accomplished by adding 'kubefed.k8s.io/orphan: true' as an annotation to the 
+		federated resource.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphan_enable_example = `
+		# Enable the orphaning mode for a federated resource of type FederatedDeployment and named foo 
+		kubefedctl orphaning enable FederatedDeployment foo --host-cluster-context=cluster1`
+)
+
+// newCmdEnableOrphaning adds 'kubefed.k8s.io/orphan: true' as an annotation to the federated resource
+func newCmdEnableOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	opts := &orphanResource{}
+	cmd := &cobra.Command{
+		Use:     "enable <resource type> <resource name>",
+		Short:   "Enable the orphaning (i.e. retention) of resources managed by a federated resource upon its removal.",
+		Long:    orphaning_enable_long,
+		Example: orphan_enable_example,
+		Run: func(cmd *cobra.Command, args []string) {
+			err := opts.Complete(args)
+			if err != nil {
+				klog.Fatalf("Error: %v", err)
+			}
+
+			err = opts.RunEnable(cmdOut, config)
+			if err != nil {
+				klog.Fatalf("Error: %v", err)
+			}
+		},
+	}
+
+	flags := cmd.Flags()
+	opts.GlobalSubcommandBind(flags)
+	err := opts.Bind(flags)
+	if err != nil {
+		klog.Fatalf("Error: %v", err)
+	}
+
+	return cmd
+}
+
+// RunEnable implements the `enable` command.
+func (o *orphanResource) RunEnable(cmdOut io.Writer, config util.FedConfig) error {
+	resourceClient, err := o.GetResourceClient(config, cmdOut)
+	if err != nil {
+		return err
+	}
+	fedResource, err := o.GetFederatedResource(resourceClient)
+	if err != nil {
+		return err
+	}
+	if ctlutil.IsOrphaningEnabled(fedResource) {
+		return nil
+	}
+	ctlutil.EnableOrphaning(fedResource)
+	_, err = resourceClient.Update(fedResource, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Failed to update resource %s %q", fedResource.GetKind(),
+			ctlutil.QualifiedName{Name: fedResource.GetName(), Namespace: fedResource.GetNamespace()})
+	}
+
+	return nil
+}
--- a/pkg/kubefedctl/orphaning/orphaning.go
+++ b/pkg/kubefedctl/orphaning/orphaning.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package orphaning
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/enable"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/options"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+type orphanResource struct {
+	options.GlobalSubcommandOptions
+	typeName          string
+	resourceName      string
+	resourceNamespace string
+}
+
+// Bind adds the join specific arguments to the flagset passed in as an argument.
+func (o *orphanResource) Bind(flags *pflag.FlagSet) error {
+	flags.StringVarP(&o.resourceNamespace, "namespace", "n", "default", "If present, the namespace scope for this CLI request")
+	err := flags.MarkHidden("kubefed-namespace")
+	if err != nil {
+		return err
+	}
+	err = flags.MarkHidden("dry-run")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewCmdOrphaning the head of orphaning-deletion sub commands
+func NewCmdOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "orphaning-deletion",
+		Short: "Manage orphaning delete policy",
+		Long:  "Manage orphaning delete policy",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := cmd.Help()
+			if err != nil {
+				klog.Fatalf("Error: %v", err)
+			}
+		},
+	}
+	cmd.AddCommand(newCmdEnableOrphaning(cmdOut, config))
+	cmd.AddCommand(newCmdDisableOrphaning(cmdOut, config))
+	cmd.AddCommand(newCmdStatusOrphaning(cmdOut, config))
+
+	return cmd
+}
+
+//  Complete ensures that options are valid and marshals them if necessary.
+func (o *orphanResource) Complete(args []string) error {
+	if len(args) == 0 {
+		return errors.New("resource type is required")
+	}
+
+	o.typeName = args[0]
+
+	if len(args) == 1 {
+		return errors.New("resource name is required")
+	}
+	o.resourceName = args[1]
+
+	return nil
+}
+
+// Returns a Federated Resources Interface
+func (o *orphanResource) GetResourceClient(config util.FedConfig, cmdOut io.Writer) (dynamic.ResourceInterface, error) {
+	hostConfig, err := config.HostConfig(o.HostClusterContext, o.Kubeconfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to load configuration for cluster context %q in kubeconfig %q.`",
+			o.HostClusterContext, o.Kubeconfig)
+	}
+	// Lookup kubernetes API availability
+	apiResource, err := enable.LookupAPIResource(hostConfig, o.typeName, "")
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to find targeted %s type", o.typeName)
+	}
+	klog.V(2).Infof("API Resource for %s/%s found", typeconfig.GroupQualifiedName(*apiResource), apiResource.Version)
+	if !util.IsFederatedAPIResource(apiResource.Kind, apiResource.Group) {
+		fmt.Fprintf(cmdOut, "Warning: %s/%s might not be a federated resource\n",
+			typeconfig.GroupQualifiedName(*apiResource), apiResource.Version)
+	}
+	targetClient, err := ctlutil.NewResourceClient(hostConfig, apiResource)
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error creating client for %s", apiResource.Kind)
+	}
+
+	resourceClient := targetClient.Resources(o.resourceNamespace)
+	return resourceClient, nil
+}
+
+// Returns the Federated resource where the orphaning-deletion will be managed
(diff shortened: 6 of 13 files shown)


marun:
Thank you for doing this! I am reviewing.

pkg/kubefedctl/federate/util.go
```diff
@@ -161,10 +157,6 @@ func apiResourceMatchesSkipName(apiResource metav1.APIResource, skipAPIResourceN
 	return false
 }
 
-func isFederatedAPIResource(kind, group string) bool {
-	return strings.HasPrefix(kind, federatedKindPrefix) && group == options.DefaultFederatedGroup
```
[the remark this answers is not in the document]
roytman: I cannot argue about it, just reuse an existing function. Maybe it should be a separate PR.

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,99 @@
...
+		resource is removed, all relevant managed resources will be removed from member 
+		clusters as well.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphaning_disable_example = `
+		# Removes the "orphaning enable" annotation from a federated resources of Deployment foo, 
+		kubefedctl orphaning disable deployments.apps foo --host-cluster-context=cluster1`
```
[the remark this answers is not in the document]
roytman: It is an example of the command (flag) usage, so it operates with a specific types. In the above explanation mentions a generic "federated resource"
The same convention exists with others `kubefedctl` and `kubectl` commands

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,99 @@
...
+	err := opts.Bind(flags)
+	if err != nil {
+		klog.Fatalf("Error: %v", err)
+	}
+
+	return cmd
+}
+
+// Run is the implementation of the `disable` command.
+func (o *orphanResource) RunDisable(cmdOut io.Writer, config util.FedConfig) error {
```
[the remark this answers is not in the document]
roytman: We cannot define 3 different methods with the same name to the `*orphanResource`

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,99 @@
...
+	fedResource, resources, err := o.GetFederatedResource(config)
+	if err != nil {
+		return err
+	}
+	annotations := fedResource.GetAnnotations()
+	if annotations == nil {
+		return nil
+	}
+	st, ok := annotations[sync.OrphanManagedResources]
+	if !ok || st == "false" {
```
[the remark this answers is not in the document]
roytman: I wanted to save an addition update, but I'd agree with you, better have a consistent behavior.
Fixed.

roytman:
Hi @xunpan , thanks for your review. I fixed part of your comments and added responses to others. Please take a look, if it is OK now. 
Thanks Alexey.

pkg/kubefedctl/orphaning/disable.go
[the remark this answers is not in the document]
xunpan: You're right.
Origionally, I was thinking to use same structure for all subcommands. As there is no more functions expected here for `orphaning` to me, I think it is fine to use different `Run` functions with a single `orphanResource` structure.

pkg/kubefedctl/federate/util.go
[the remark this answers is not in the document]
xunpan: Agreed. It should be a separeted PR.

pkg/kubefedctl/orphaning/disable.go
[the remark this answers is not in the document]
xunpan: I think the command line can operate on a specified federated type with object name specified as documented in user guide.
```
kubefedctl orphaning-deletion enable <federated type> <name>
```
My question is about the example of using `<federated type> `

`deployments.apps` is a `FederatedTypeConfig` object but not a federated type.
```
$ kubectl -n kube-federation-system get federatedtypeconfigs deployments.apps
NAME               AGE
deployments.apps   7m
```
I think a name with `Federated` prefixed makes more sense for the example. E.g. `kubefedctl orphaning enable FederatedDeployment foo`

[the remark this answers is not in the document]
roytman: Got it, thanks for the explanation. Fixed.

pkg/kubefedctl/federate/util.go
```diff
@@ -18,25 +18,21 @@ package federate
...
 	"k8s.io/client-go/rest"
-	"sigs.k8s.io/yaml"
-
+	"os"
 	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
 	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
 	"sigs.k8s.io/kubefed/pkg/kubefedctl/enable"
-	"sigs.k8s.io/kubefed/pkg/kubefedctl/options"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+	"sigs.k8s.io/yaml"
```
xunpan: For better readability, it is import convention in kubefed:
```
import(
    "core-packages"

    "3rd-party-packages"

    "kubefed-packages"
)
```

E.g.:
```
import (
	"io"
	"os"

	"k8s.io/apimachinery/pkg/runtime/schema"

	"sigs.k8s.io/kubefed/pkg/kubefedctl/enable"
	"sigs.k8s.io/kubefed/pkg/kubefedctl/options"
)
```

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		klog.Fatalf("Error: %v", err)
+	}
+
+	return cmd
+}
+
+// RunEnable implements the `enable` command.
+func (o *orphanResource) RunEnable(cmdOut io.Writer, config util.FedConfig) error {
+	fedResource, resources, err := o.GetFederatedResource(config)
```
xunpan: `resources` is actually a `resourceClient` I think. It is also better to have `resources` in `GetFederatedResource` as `resourceClient`. It is a little confuse to use `resource` and `resources` in `GetFederatedResource` for different types of object.

xunpan:
@roytman 
This change impacts user interface. Please also update `CHANGELOG.md` in `Unreleased` section.
E.g.
https://github.com/kubernetes-sigs/kubefed/pull/1010/files

pkg/kubefedctl/federate/util.go
roytman: Fixed it, 
looks like kubefed uses a slightly different import convention:
```
import(
    "core-packages"

    "3rd-party-packages"

    "K8s packages"

    "kubefed-packages"
) ```
Is it correct ?

File: pkg/kubefedctl/orphaning/enable.go (hunk 6 above)
Comment (roytman) [reply pkg/kubefedctl/orphaning/enable.go]: agree, fixed

**Comment by roytman** [comment]
> 2 minor readability comments.
> 
> And for `commits`, could you please squash them. It is not encouraged to keep commits that are only used to fix previous commit in one PR.
Did it, (I was under impression that all squashes are done during merge.)

**Comment by roytman** [comment]
> @roytman
> This change impacts user interface. Please also update `CHANGELOG.md` in `Unreleased` section.
> E.g.
> https://github.com/kubernetes-sigs/kubefed/pull/1010/files

Done, thanks

File: .gitignore [hunk 7]
```diff
@@ -6,3 +6,10 @@
 # Ignore binaries built by the makefile to avoid accidentally committing them
 /images/kubefed/hyperfed
 /images/kubefed/controller-manager
+
+# OSX trash
+.DS_Store
+
+# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
+.idea/
+*.iml
```
xunpan: `.gitignore` is used for project related files maintenance in `kubefed`.

I finally agreed this rule after thinking it over. I saw some projects which use `.gitignore` file with hundreds of lines code with editor specific ignore. It is hard to read and maintain. You can refer to https://github.com/kubernetes-sigs/kubefed/pull/871.

roytman: I see the point. 
Actually, I copied the entries from the [Kubernetes main repository](https://github.com/kubernetes/kubernetes/blob/master/.gitignore), and probably worth to be inline with it. But this discussion is out of this PR scope. 
I removed the entries.

xunpan:
Thanks Roytman! 
/lgtm

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	o.resourceName = args[1]
+
+	return o.SetNamespace(config)
+}
+
+// If namespace was not set as an argument, it will be the current context namespace
+func (o *orphanResource) SetNamespace(config util.FedConfig) error {
+
+	if o.resourceNamespace != "" {
```
marun: (No action required) The kube convention afaik is to check the length of a string rather than comparing with empty string.

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil
+	}
+	ns, _, err := config.GetClientConfig(o.HostClusterContext, o.Kubeconfig).Namespace()
```
marun: I think it would be preferable to follow the convention of other kubefedctl commands (e.g. federate) which default the resource namespace to `default` if it is not provided. The approach here might be nicer, but I think consistency is preferable and common support for discovering the namespace from the context can be implemented in a subsequent PR.

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+// Returns the Federated resource where the orphaning-deletion will be managed, and a corresponding Resources Interface
+func (o *orphanResource) GetFederatedResource(config util.FedConfig) (*unstructured.Unstructured, dynamic.ResourceInterface, error) {
+	hostConfig, err := config.HostConfig(o.HostClusterContext, o.Kubeconfig)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Failed to get host cluster config")
+	}
+	// Lookup kubernetes API availability
+	apiResource, err := enable.LookupAPIResource(hostConfig, o.typeName, "")
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "Failed to find target API resource %s", o.typeName)
```
marun: I think the error message here should reflect a failure to find a type corresponding to the provided type name rather than potentially confusing the user with mention of an api resource.

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	klog.V(2).Infof("API Resource for %s found", o.typeName)
```
marun: This log message doesn't seem very helpful without details about the api resource that was found (e.g. `<plural name>.<group name>/<version>`).

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+	if !util.IsFederatedAPIResource(apiResource.Kind, apiResource.Group) {
```
marun: The point of the orphaning commands is to add/remove/check for an annotation.  Why does the command need to be restricted to a weak check for whether the resource is federated vs just applying/removing the annotation on any resource?

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, nil, errors.Errorf("Looks like the resource is not a Federated object, Kind='%v' Group='%v'",
+			apiResource.Kind, apiResource.Group)
+	}
+	targetClient, err := ctlutil.NewResourceClient(hostConfig, apiResource)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "Failed to create a resource client for %s", o.typeName)
+	}
+	resources := targetClient.Resources(o.resourceNamespace)
```
marun: (No action required) The generic client (from `"sigs.k8s.io/kubefed/pkg/client/generic"`) should be preferred to the legacy resources client.

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if ns == "" {
+		ns = "default"
+	}
+	o.resourceNamespace = ns
+	return nil
+}
+
+// Returns the Federated resource where the orphaning-deletion will be managed, and a corresponding Resources Interface
```
marun: Is there a reason this function needs to retrieve the client and the resource, vs performing these operations separately?

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,99 @@
...
+		# Enable the orphaning mode for federated resources with type FederatedDeployment and named foo 
+		kubefedctl orphaning enable FederatedDeployment foo --host-cluster-context=cluster1`
+)
+
+// newCmdEnableOrphaning adds 'kubefed.k8s.io/orphan: true' as an annotation to the given federated resource
+func newCmdEnableOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	opts := &orphanResource{}
+	cmd := &cobra.Command{
+		Use:     "enable <resource type> <resource name>",
+		Short:   "Enable to remain managed resources after removing the federated resource ",
```
marun: -> Enable the orphaning (i.e. retention) of resources managed by a federated resource upon its removal.

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,100 @@
...
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kubefed/pkg/controller/sync"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_disable_long = `
+		Removes previously added "orphaning enable" ('kubefed.k8s.io/orphan: true')
+		annotation from a given federated resource.  As result, when the federated
```
marun: s/given//

s/As result, when the federated resource removed,/When the federated resource is subsequently removed/

s/as well//

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,100 @@
...
+		annotation from a given federated resource.  As result, when the federated 
+		resource is removed, all relevant managed resources will be removed from member 
+		clusters as well.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphaning_disable_example = `
+		# Disable the orphaning mode for federated resources with type FederatedDeployment and named foo
```
marun: s/federated resources with/a federated resource of/

pkg/kubefedctl/federate/util.go
```diff
@@ -20,9 +20,9 @@ import (
 	"bufio"
 	"io"
 	"os"
-	"strings"
 
 	"github.com/pkg/errors"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
```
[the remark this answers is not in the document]
roytman: fixed

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	_, ok := annotations[sync.OrphanManagedResources]
+	if !ok {
+		return nil
+	}
+	delete(annotations, sync.OrphanManagedResources)
+	fedResource.SetAnnotations(annotations)
+	_, err = resourceClient.Update(fedResource, metav1.UpdateOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Failed to update resource %s", o.resourceName)
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+const (
+	Enabled  = "Enabled"
+	Disabled = "Disabled"
+)
+
+var (
+	orphaning_status_long = `
+		Checks status of  "orphaning enable" ('kubefed.k8s.io/orphan: true')
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,95 @@
...
+		annotation on a given federated resource.
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,95 @@
...
+		Checks status of  "orphaning enable" ('kubefed.k8s.io/orphan: true')
+		annotation on a given federated resource. 
+		Returns "Enabled" or "Disabled"
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphaning_status_example = `
+		# Checks status of the orphaning mode of federated resources with type FederatedDeployment and named foo
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		klog.Fatalf("Error: %v", err)
+	}
+
+	return cmd
+}
+
+// RunStatus implements the `status` command.
+func (o *orphanResource) RunStatus(cmdOut io.Writer, config util.FedConfig) error {
+
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,99 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kubefed/pkg/controller/sync"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_enable_long = `
+		Prevents removal of managed resources from member clusters, when their federated
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,99 @@
...
+		Prevents removal of managed resources from member clusters, when their federated 
+		resource is removed. Adds 'kubefed.k8s.io/orphan: true' as an annotation to the 
+		federated resource.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphan_enable_example = `
+		# Enable the orphaning mode for federated resources with type FederatedDeployment and named foo
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+	err = flags.MarkHidden("dry-run")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewCmdOrphaning the head of orphaning-deletion sub commands
+func NewCmdOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	//	opts := &orphanResource{}
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+	cmd.AddCommand(newCmdEnableOrphaning(cmdOut, config))
+	cmd.AddCommand(newCmdDisableOrphaning(cmdOut, config))
+	cmd.AddCommand(newCmdStatusOrphaning(cmdOut, config))
+
+	return cmd
+}
+
+//  Complete ensures that options are valid and marshals them if necessary.
+func (o *orphanResource) Complete(args []string, config util.FedConfig) error {
+
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,141 @@
...
+		return errors.New("resource name is required")
+	}
+	o.resourceName = args[1]
+
+	return o.SetNamespace(config)
+}
+
+// If namespace was not set as an argument, it will be the current context namespace
+func (o *orphanResource) SetNamespace(config util.FedConfig) error {
+
```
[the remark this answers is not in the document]
roytman: done

pkg/kubefedctl/orphaning/orphaning.go
roytman: I see the same comparison in other `kubefedctl` and `controller` files.

pkg/kubefedctl/orphaning/orphaning.go
roytman: done, will postpone namespace discovery for a separate PR, which is consistent with the `kubectl` behavior.

pkg/kubefedctl/orphaning/orphaning.go
roytman: updated the error message

pkg/kubefedctl/orphaning/orphaning.go
roytman: Update the log message according to the suggested format.

pkg/kubefedctl/orphaning/orphaning.go
roytman: I think we can mislead users if we allow applying/removing the annotation on any resource. In the end of the day, a user wants to change behavior of a federated resource deletion, and not to add or remove the annotation. If a wrong resource was provided the command should fail. 
What do you think ?

pkg/kubefedctl/orphaning/orphaning.go
roytman: I wanted to simplify the RunDisable/Enable functions, but I agree the separation is better. Done.

pkg/kubefedctl/orphaning/enable.go
roytman: done

pkg/kubefedctl/orphaning/disable.go
roytman: done

pkg/kubefedctl/orphaning/disable.go
roytman: done

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kubefed/pkg/controller/sync"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_enable_long = `
+		Prevents removal of managed resources from member clusters, when their federated 
+		resource is removed. Adds 'kubefed.k8s.io/orphan: true' as an annotation to the
```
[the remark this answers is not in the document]
roytman: done

roytman:
@marun , thanks for the review. 

> Without #828 there isn't an easy path to testing of the cli just yet, but I would appreciate some common code paths being created (in the vein of the managed label helpers) to leverage the coverage for the sync controller check and crudtester check.

I updated the code.

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return nil
+}
+
+// Returns a Federated Resources Interface
+func (o *orphanResource) GetResourceClient(config util.FedConfig) (dynamic.ResourceInterface, error) {
+	hostConfig, err := config.HostConfig(o.HostClusterContext, o.Kubeconfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to get HostConfig for host cluster context '%s' and kubeconfig '%s'",
```
marun: Please prefer `%q` to `'%s'`.

Also, I'm not sure we can expect a user to know what 'HostConfig' means in this context.  How about the following?

```
Unable to load configuration for cluster context %q in kubeconfig %q.
```

pkg/kubefedctl/orphaning/orphaning.go
marun: If `IsFederatedAPIResource` was guaranteed to identify a federated resource I would agree with you that preventing addition of the annotation would be desirable, but as per previous comments that is not the case. How about adding a `--force` argument to `enable` and `disable` commands to ensure that a user can still use the command even if the check fails? And have the `status` command always indicate orphaning (rather than requiring `--force` for a read-only operation)?

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,133 @@
...
+			o.HostClusterContext, o.Kubeconfig)
+	}
+	// Lookup kubernetes API availability
+	apiResource, err := enable.LookupAPIResource(hostConfig, o.typeName, "")
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to find targeted %s type", o.typeName)
+	}
+	klog.V(2).Infof("API Resource for %s.%s/%s found", apiResource.Name, apiResource.Group, apiResource.Version)
+	if !util.IsFederatedAPIResource(apiResource.Kind, apiResource.Group) {
+		return nil, errors.Errorf("Looks like the resource is not a Federated object, Kind='%v' Group='%v'",
```
marun: How about using the same format for the kgv as line 110?

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,133 @@
...
+
+	resourceClient := targetClient.Resources(o.resourceNamespace)
+	return resourceClient, nil
+}
+
+// Returns the Federated resource where the orphaning-deletion will be managed
+func (o *orphanResource) GetFederatedResource(resourceClient dynamic.ResourceInterface) (*unstructured.Unstructured, error) {
+	resource, err := resourceClient.Get(o.resourceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error retrieving resource %s %q", o.typeName,
```
marun: Would it be possible to use the kind here rather than the type name the user passed in?

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	o.typeName = args[0]
+
+	if len(args) == 1 {
+		return errors.New("resource name is required")
+	}
+	o.resourceName = args[1]
+
+	if len(o.resourceNamespace) == 0 {
```
marun: How about supplying this default to the `StringVarP` definition?

pkg/controller/util/orphaninganotation.go
```diff
@@ -0,0 +1,58 @@
...
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+const (
+	// If this annotation is present on a federated resource, resources in the
+	// member clusters managed by the federated resource should be orphaned.
+	// If the annotation is not present (the default), resources in member
+	// clusters will be deleted before the federated resource is deleted.
+	OrphanManagedResourcesAnnotation = "kubefed.k8s.io/orphan"
+	OrphanedManagedResourceslValue   = "true"
```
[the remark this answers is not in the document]
roytman: Wow, you was able to catch this typo. Sometimes, autocomplite helps to hide typos.  Fixed it. Thanks.

pkg/controller/util/orphaninganotation.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// IsOrphanedManagedResources checks status of "orphaning enable" (OrphanManagedResources: OrphanedManagedResourceslValue')
+// annotation on a resource.
+func IsOrphanedManagedResources(obj *unstructured.Unstructured) bool {
```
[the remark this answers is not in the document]
roytman: Renamed according to suggestion.

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,98 @@
...
+
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_disable_long = `
+		Removes previously added "orphaning enable" ('kubefed.k8s.io/orphan: true')
+		annotation from a federated resource.  When the federated resource is subsequently removed, 
+        all relevant managed resources will be removed from member clusters.
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,98 @@
...
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphaning_disable_example = `
+		# Disable the orphaning mode for a federated resources of type FederatedDeployment and named foo
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/orphaning.go
roytman: Done

pkg/kubefedctl/orphaning/orphaning.go
roytman: I updated the method behavior. Instead of error returning or `--force` flags requirements, it just prints a warning. What do you think ? 
BTW, the format for the kgv as line 110 can look strange for core objects, without group e.g. for a `service`
"Warning: looks like the resource **services./v1** is not a Federated object"
Or it is OK?

pkg/kubefedctl/orphaning/orphaning.go
roytman: Due to separation of `GetResourceClient` and `GetFederatedResource`, it will be complicated. In the `GetFederatedResource` I don't have a reference to `resourceClient` only to the `dynamic.ResourceInterface`.
In order to use Kind, I have to 

- or return `Kind` from  `GetResourceClient`  and provide it to `GetFederatedResource`
- or move `targetClient.Resources(o.resourceNamespace)` out of `GetResourceClient`.

Both options look problematic for me. What do you think ?

pkg/kubefedctl/orphaning/orphaning.go
roytman: Done.
After that PR I plan to propose to update all `kubefedctl` command to resolve namespace from the Context, similar to `kubectl`. so I will have to remove the default values.

roytman:
> The only functional change request is to allow the addition or removal of the annotation for resources that don't pass the weak check for a federated resource.

If a resource doesn't pass the weak check for a federated resource, I suggest to print a warning and continue the processing. What do you think ?

pkg/controller/sync/controller.go
```diff
@@ -435,13 +429,13 @@ func (s *KubeFedSyncController) ensureDeletion(fedResource FederatedResource) ut
 		return util.StatusAllOK
 	}
 
-	annotations := obj.GetAnnotations()
-	orphanResources := annotations != nil && annotations[OrphanManagedResources] == "true"
+	orphanResources := util.IsOrphaningEnabled(obj)
```
marun: (No action required) Maybe just inline this call? The intent of the `orphanResources` variable was to ensure the conditional was self-documenting and this could be accomplished with the new method call.

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,98 @@
...
+
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_disable_long = `
+		Removes previously added "orphaning enable" ('kubefed.k8s.io/orphan: true')
+		annotation from a federated resource.  When the federated resource is subsequently marked for deletion,
+        the resources it manages in member clusters will be removed before the federated resource is removed.
```
marun: nit: alignment.  Note that this content is intentionally prefixed with tabs rather than spaces.  The strategy I use to maintain it is to copy the prefix from existing lines.

pkg/kubefedctl/orphaning/orphaning.go
marun: I think the first option is a change worth making to ensure the UX consistency.

pkg/kubefedctl/orphaning/orphaning.go
```diff
@@ -0,0 +1,130 @@
...
+			o.HostClusterContext, o.Kubeconfig)
+	}
+	// Lookup kubernetes API availability
+	apiResource, err := enable.LookupAPIResource(hostConfig, o.typeName, "")
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to find targeted %s type", o.typeName)
+	}
+	klog.V(2).Infof("API Resource for %s.%s/%s found", apiResource.Name, apiResource.Group, apiResource.Version)
+	if !util.IsFederatedAPIResource(apiResource.Kind, apiResource.Group) {
+		fmt.Fprintf(cmdOut, "Warning: looks like the resource %s.%s/%s is not a Federated object\n",
```
marun: `s/looks like the resource//`
`s/is not a Federated object/does not appear to be a federated resource/`

As to how to compose name and group to avoid a trailing `.` for core groups, please use the [GroupQualifiedName function](https://github.com/marun/kubefed/blob/fix-namespaced-prop-status/pkg/apis/core/typeconfig/util.go#L34).

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,98 @@
...
+		# Disable the orphaning mode for a federated resources of type FederatedDeployment and named foo 
+		kubefedctl orphaning disable FederatedDeployment foo --host-cluster-context=cluster1`
+)
+
+// newCmdDisableOrphaning removes the 'kubefed.k8s.io/orphan: true' annotation from the federated resource
+func newCmdDisableOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	opts := &orphanResource{}
+	cmd := &cobra.Command{
+		Use:     "disable <resource type> <resource name>",
+		Short:   "Disable to remain managed resources after removing the federated resource",
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,99 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	"sigs.k8s.io/kubefed/pkg/kubefedctl/util"
+)
+
+var (
+	orphaning_enable_long = `
+		Prevents removal of managed resources from member clusters, when their managing federated
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/enable.go
```diff
@@ -0,0 +1,99 @@
...
+		Prevents removal of managed resources from member clusters, when their managing federated 
+		resource is removed. This is accomplished by adding 'kubefed.k8s.io/orphan: true' as an annotation to the 
+		federated resource.
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphan_enable_example = `
+		# Enable the orphaning mode for a federated resources of type FederatedDeployment and named foo
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+const (
+	Enabled  = "Enabled"
+	Disabled = "Disabled"
+)
+
+var (
+	orphaning_status_long = `
+		Checks status of "orphaning enable" ('kubefed.k8s.io/orphan: true')
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,97 @@
...
+		annotation on a federated resource. 
+		Returns "Enabled" or "Disabled"
+
+		Current context is assumed to be a Kubernetes cluster hosting
+		the kubefed control plane. Please use the
+		--host-cluster-context flag otherwise.`
+
+	orphaning_status_example = `
+		# Checks status of the orphaning mode of federated resources of type FederatedDeployment and named foo
```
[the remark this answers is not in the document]
roytman: Done

pkg/kubefedctl/orphaning/orphaning.go
roytman: I implemented the first option. 
For generalizing, `GetResourceClient` returns entire `*metav1.APIResource`, but it can be easily replace by `Kind` only.  
However, there is an additional option: the current error message looks like 
`Error: Error retrieving resource FederatedConfigMap "default/game-config1": federatedconfigmaps.types.kubefed.k8s.io "game-config1" not found`
 
Do we really have to print the resource `Kind` or its user provided type? Anyway the dynamic client prints (at list in the most common case) the resource name and group. Maybe we should just return the dynamic client error.
What do you think?

pkg/controller/sync/controller.go
roytman: Done

pkg/kubefedctl/orphaning/disable.go
roytman: Done

pkg/kubefedctl/orphaning/orphaning.go
roytman: I was not aware about the `GroupQualifiedName` function, thanks.
I changed the warning message according to your suggestion, however, as I understood the function `IsFederatedAPIResource` cannot 100% guarantee that the checked resource is not a Federated resource. Therefore, I'd suggest a  weaker statement. 
What do you think about `might not be a federated resource`

roytman:
Thanks for your reviews and sorry for the missed comments. 
I have 2 suggestions: one about the warning message, maybe the statement should be weaker, and another about the error message in `GetFederatedResource` .
Please take a look.

pkg/kubefedctl/orphaning/orphaning.go
marun: I think `does not appear to be` is equivalent to `might not be` (i.e. `this does not appear to be correct` vs `this is not correct`).

pkg/kubefedctl/orphaning/status.go
```diff
@@ -0,0 +1,97 @@
...
+const (
+	Enabled  = "Enabled"
+	Disabled = "Disabled"
+)
+
+var (
+	orphaning_status_long = `
+		Checks the status of "orphaning enable" ('kubefed.k8s.io/orphan: true')
+		annotation on a federated resource. 
+		Returns "Enabled" or "Disabled"
```
marun: nit: Please move the content on this line to the previous line.

pkg/kubefedctl/orphaning/disable.go
```diff
@@ -0,0 +1,98 @@
...
+		# Disable the orphaning mode for a federated resource of type FederatedDeployment and named foo 
+		kubefedctl orphaning disable FederatedDeployment foo --host-cluster-context=cluster1`
+)
+
+// newCmdDisableOrphaning removes the 'kubefed.k8s.io/orphan: true' annotation from the federated resource
+func newCmdDisableOrphaning(cmdOut io.Writer, config util.FedConfig) *cobra.Command {
+	opts := &orphanResource{}
+	cmd := &cobra.Command{
+		Use:     "disable <resource type> <resource name>",
+		Short:   "Disable the orphaning to ensure the removal of managed resources before removing the managing federated resource",
```
marun: nit: `s/the orphaning/orphaning deletion/`

pkg/kubefedctl/orphaning/orphaning.go
roytman: Done

pkg/kubefedctl/orphaning/status.go
roytman: Hope I understand you, done

pkg/kubefedctl/orphaning/disable.go
roytman: Done

pkg/kubefedctl/orphaning/status.go
marun: I meant add ensuring `Returns "Enabled"...` wasn't separated from the previous sentence with a newline, but this is fine too.

marun:
Nice work, thank you!

/approve
/lgtm

PR #1010, opened 2019-06-20
Always retain metadata.finalizers and metadata.annotations

### LINKED ISSUE #982: Infinite update loop on FederatedDeployment
**What happened**:

I federated a Deployment:
`kubefedctl federate -f file/with/deployment.yaml | kubectl apply -f`

It seems to be updating forever:
```
Events:
  Type     Reason                 Age                   From                            Message
  ----     ------                 ----                  ----                            -------
  Normal   CreateInCluster        31m                   federateddeployment-controller  Creating Deployment "default/dummy" in cluster "ar2a"
  Warning  UpdateInClusterFailed  30m (x4 over 31m)     federateddeployment-controller  Failed to update Deployment "default/dummy" in cluster "ar2a": Operation cannot be fulfilled on deployments.apps "dummy": the object has been modified; please apply your changes to the latest version and try again
  Normal   UpdateInCluster        6m4s (x655 over 31m)  federateddeployment-controller  Updating Deployment "default/dummy" in cluster "ar2a"
  Normal   UpdateInCluster        48s (x15 over 69s)    federateddeployment-controller  Updating Deployment "default/dummy" in cluster "ar2a"
  Warning  UpdateInClusterFailed  48s (x4 over 69s)     federateddeployment-controller  Failed to update Deployment "default/dummy" in cluster "ar2a": Operation cannot be fulfilled on deployments.apps "dummy": the object has been modified; please apply your changes to the latest version and try again
  Normal   UpdateInCluster        2s (x4 over 12s)      federateddeployment-controller  Updating Deployment "default/dummy" in cluster "ar2a"
```

Reconciliation to finish and the status to be updated.

Deploy a FederatedDeployment.

**Environment**:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-26T00:04:52Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.7", GitCommit:"4683545293d792934a7a7e12f2cc47d20b2dd01b", GitTreeState:"clean", BuildDate:"2019-06-06T01:39:30Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```
- KubeFed version: 0.1.0-rc2
- Scope of installation (namespaced or cluster): cluster
- Others

/kind bug

marun:
Now that [overriding via jsonpatch is supported](https://github.com/kubernetes-sigs/kubefed/pull/991), it is possible to restrict modification of the `annotations` and `finalizers` fields to overrides. Modifications to these collection fields can be made using the `add` or `remove` jsonpatch operations to ensure that a given item exists (or does not exist) with less chance of conflicting with local controllers.

```yaml
kind: FederatedDeployment
...
spec:
  ...
  overrides:
    - clusterName: cluster1
      clusterOverrides:
        # Ensure the annotation "foo: bar" exists
        - path: "/metadata/annotations"
          op: "add"
          value:
            foo: bar
        # Ensure an annotation with key "foo" does not exist
        - path: "/metadata/annotations/foo"
          op: "remove"
```

To support this change, it was necessary to move the application of overrides from before retention of local values to after so  that overrides would be applied to the result of retention.

This change suggests implementing [selector-based overrides](https://github.com/kubernetes-sigs/kubefed/issues/581) to simplify applying an annotation across multiple clusters. 

Fixes #982

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -454,6 +454,7 @@ The following table enumerates the possible values for cluster status:
 | Status                 | Description                  |
 |------------------------|------------------------------|
 | AlreadyExists          | The target resource already exists in the cluster, and cannot be adopted due to `adoptResources` being disabled. |
+| ApplyOverridesFailed   | An error occurred while attempting to apply overrides to the computed form of the target resource. |
 | CachedRetrievalFailed  | An error occurred when retrieving the cached target resource. |
 | ClientRetrievalFailed  | An error occurred while attempting to create an API client for the member cluster. |
 | ClusterNotReady        | The latest health check for the cluster did not succeed. |
@@ -687,6 +688,23 @@ spec:
           op: "remove"
 ```
 
+### Overriding retained fields
+
+When computing the form of a managed resource that should appear in a cluster
+registered with KubeFed, the following operations are executed in sequence:
+
+ - A new resource is computed from the template of the federated resource
+ - If an existing resource is present, the contents of fields subject to retention are preserved
+ - Overrides are applied
+ - The managed label is set
+
+This order of operations ensures that [fields subject to
+retention](#local-value-retention) can still be overridden (e.g. adding an
+entry to `metadata.annotations`). Care should be taken in applying overrides
+to retained fields that may be modified by controllers in member clusters or
+a managed resource may end up being continuously updated first by the
+controller in the member cluster and then by KubeFed.
+
 ## Using Cluster Selector
 
 In addition to specifying an explicit list of clusters that a resource should be propagated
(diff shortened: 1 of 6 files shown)


font:
@marun please rebase.

marun:
Rebased

marun:
Rebased and ready for review

docs/userguide.md
```diff
@@ -687,6 +688,23 @@ spec:
           op: "remove"
 ```
 
+### Overriding retained fields
+
+When computing the form of a managed resource that should appear in a cluster
+registered with KubeFed, the following operations are executed in sequence:
+
+ - A new resource is computed from the template of the federated resource
+ - If an existing resource is present, field values are retained
```
xunpan: `field values` are retained. I think it is not accurate because only partial field values are retained.

pkg/controller/sync/resource.go
```diff
@@ -165,6 +165,15 @@ func (r *federatedResource) ObjectForCluster(clusterName string) (*unstructured.
 	}
 	obj := &unstructured.Unstructured{Object: templateBody}
 
+	if len(obj.GetAnnotations()) > 0 {
+		r.RecordError("AnnotationsNotSupported", errors.New("metadata.annotations will not be set to avoid conflicting with local controllers."))
```
xunpan: This message seems misunderstanding. Actually, `metadata.annotations` still can be set. But by `override` instead of by `template`

docs/userguide.md
marun: Updated.  PTAL.

pkg/controller/sync/resource.go
marun: Updated. PTAL

marun:
@font I've replaced references to 'local controller' with 'controllers in member clusters'.  Does that make more sense?

marun:
Rebased.

marun:
Rebased

shashidharatd:
/lgtm
Thanks @marun

dev-e:
Hi! I think annotations still should propagate for non controller-managed resources, e.g. `Ingress` or `Service` because in most cases these resources should be consistent across the entire deployment. If you want to override it for some clusters, it still may be done via `jsonpatch`. Otherwise, you are forced to add the same annotations for all member clusters which is not very convenient.

PR #1011, opened 2019-06-20
WIP Retain the annotations of known controllers

marun:
Retaining the annotations set by known controllers will avoid thrashing (i.e. controller sets an annotation, kubefed clears it,
repeat). This is intended to be a stop-gap until suppport for jsonpatch add/remove is added. Once overrides can non-destructively
add and remove annotations, retention of known annotations will no longer be necessary and can be removed.

Partial fix for #982 

**TODO**

 - [ ] Add test coverage


marun:
Closing in favor of #1010

PR #1012, opened 2019-06-21
Ensure cluster scoped sync controller starts without namespace ftc

marun:
Previously sync controllers for any type required namespaces are enabled for federation. This change ensures that sync controllers for cluster-scoped types can start if namespaces are not enabled.


pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -296,14 +296,25 @@ func (c *Controller) startSyncController(tc *corev1b1.FederatedTypeConfig) error
 	// cluster-scoped KubeFed control plane.  A namespace-scoped
 	// control plane would still have to use a non-shared informer due
 	// to it not being possible to limit its scope.
+
+	ftc := tc.DeepCopyObject().(*corev1b1.FederatedTypeConfig)
 	kind := tc.Spec.FederatedType.Kind
```
font: Consider replacing `tc` with the `ftc` copy since we've moved the copy before this line.

Issue #983, opened 2019-06-13
Deploy and register admission webhook prior to rest of kubefed installation


font:
**What would you like to be added**:
Need a way to deploy admission webhook manifests and make sure the webhook server is registered with the `kube-apiserver` before any of the admission resources it operates against are created. For example for `KubeFedConfig`, we would need the mutating and validating webhook manifests to be installed by Helm, then the webhook server for `KubeFedConfig` to register itself with `kube-apiserver`, and the webhook service to contain a healthy endpoint, prior to installing the `KubeFedConfig` resource from the Helm chart.

At this time, Helm does not provide a way to order installation manifests that involve `ValidatingWebhookConfiguration` or `MutatingWebhookConfiguration` in a way that would guarantee the `Deployment` that uses those configs is properly running and registered. See https://github.com/helm/helm/issues/5642 for more details.

An alternative might be to use a helm `pre-install` hook with a `Job` and/or a separate webhook sub-chart that can verify the webhook is ready and registered.

**Why is this needed**:
This will ensure we properly default (mutate) and validate our own resource manifests upon installation.

/kind feature

PR #1015, opened 2019-06-24
Add KubeFedConfig defaulting via mutating admission webhook

### LINKED ISSUE #1013: Support defaulting KubeFedConfigs
/kind feature

font:
Fixes #1013.

You may choose to review a commit at a time as changes follow a fairly reasonable sequence.

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -196,7 +195,7 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
 
 		err := client.Get(context.Background(), fedConfig, namespace, name)
 		if apierrors.IsNotFound(err) {
-			klog.Infof("Cannot retrieve KubeFedConfig %q: %v. Default options are used.", qualifiedName.String(), err)
+			klog.Infof("Cannot retrieve KubeFedConfig %q: %v. Default options will be used.", qualifiedName.String(), err)
 			return nil
 		}
 		if err != nil {
@@ -225,24 +224,8 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
 	return fedConfig
 }
 
-func setDuration(target **metav1.Duration, defaultValue time.Duration) {
-	if *target == nil {
-		*target = &metav1.Duration{}
-		(*target).Duration = defaultValue
-	}
-}
-
-func setInt64(target **int64, defaultValue int64) {
-	if *target == nil {
-		*target = new(int64)
-		**target = defaultValue
-	}
-}
-
-func SetDefaultKubeFedConfig(fedConfig *corev1b1.KubeFedConfig) {
-	spec := &fedConfig.Spec
-
-	if len(spec.Scope) == 0 {
+func setDefaultKubeFedConfigScope(fedConfig *corev1b1.KubeFedConfig) {
+	if len(fedConfig.Spec.Scope) == 0 {
 		// TODO(sohankunkerkar) Remove when no longer necessary.
 		// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
 		// Its continued existence should not be relied upon.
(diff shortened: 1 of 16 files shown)


test/e2e/defaulting.go
```diff
@@ -0,0 +1,171 @@
...
+				if err != nil && !apierrors.IsNotFound(err) {
+					f.Logger().Errorf("Error deleting %s %q: %v", resourceName, *qualifiedName, err)
+				}
+			})
+		}
+	})
+
+	// TODO(font): This test verifies the controller-manager does not fail
+	// early on an invalid KubeFedConfig that is created by the helm chart
+	// until https://github.com/kubernetes-sigs/kubefed/issues/983 is resolved.
```
xunpan: I'm not sure of this case.

Based on my understanding, there should no default settings in kubefed logic for `controller-manager` and there is no webhook for this case.  If `controller-manager` can be start (as this case expected), it means the system default (not setting by kubefed related logic but by API itself) can make `controller-manager` started up.

The `defaultKubeFedConfig` does not equal to helm installed `KubefedConfig`. And it works. It seems valid in starting a `controller-manager`. In the comments, it tells that this case can test invalid `KubeFedConfig` installed by helm. It seems not test invalid case. And invalid configuration should failed `controller-manager` start up. Any comments?

test/common/wait.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// WaitForObject waits for object to match the desired status.
+func WaitForObject(tl TestLogger, namespace, name string, objectGetter func(namespace, name string) (pkgruntime.Object, error), desired pkgruntime.Object, interval, timeout time.Duration, equivalent func(actual, desired pkgruntime.Object) bool) {
```
xunpan: Is that possible to use less arguments for this function? It is too many arguments to read and reuse it.
For example, the function is used in test context, `interval` and `timeout` are consistent in my simple check. It can be removed from the argument list.

test/e2e/defaulting.go
font: I've updated the comment to hopefully make it more clear.

xunpan:
Thanks Font. It LGTM

cmd/controller-manager/app/controller-manager.go
```diff
@@ -339,9 +274,22 @@ func setOptionsByKubeFedConfig(opts *options.Options) {
...
+		createKubeFedConfig(opts.Config.KubeConfig, fedConfig)
 	}
 
-	SetDefaultKubeFedConfig(fedConfig)
+	qualifedName := util.QualifiedName{
+		Name:      fedConfig.Name,
+		Namespace: fedConfig.Namespace,
+	}
+
+	errs := validation.ValidateKubeFedConfig(fedConfig)
```
marun: I'm a bit confused as to why it's desirable to validate in this particular location. Would a kubefed api created on line 279 or loaded from the API need validation, or is it just the case of when a config was provided via a yaml file? 

(No action required) Does this imply that any of the controllers that can call validation in code should do so rather than making a call to the webhook?

config/kubefedconfig.yaml
```diff
@@ -5,9 +5,13 @@ metadata:
   namespace: kube-federation-system
 spec:
   scope: Cluster
+  controllerDuration:
```
marun: Would it be possible to avoid duplicating these defaults here and have them instead be set by the controller manager when loading config from a file?

pkg/apis/core/v1beta1/defaults/defaults.go
```diff
@@ -0,0 +1,142 @@
...
+		spec.SyncController = &v1beta1.SyncControllerConfig{}
+	}
+
+	if spec.SyncController.AdoptResources == nil {
+		spec.SyncController.AdoptResources = new(v1beta1.ResourceAdoption)
+		*spec.SyncController.AdoptResources = v1beta1.AdoptResourcesEnabled
+	}
+}
+
+func setDefaultKubeFedFeatureGates(fgs []v1beta1.FeatureGatesConfig) []v1beta1.FeatureGatesConfig {
```
marun: (No action required) Why fgs (vs e.g. fgc)?

test/common/wait.go
```diff
@@ -0,0 +1,73 @@
...
+		return equivalent(actual, desired), nil
+	})
+	if err != nil {
+		tl.Fatalf("Timedout waiting for desired state, \ndesired: %#v\nactual:  %#v", desired, actual)
+	}
+}
+
+// WaitUntilLogStreamContains waits for the given stream to contain the
+// substring until the end of the stream or timeout.
+func WaitUntilLogStreamContains(tl TestLogger, stream io.ReadCloser, substr string, timeout time.Duration) bool {
```
marun: Why is this method accepting a timeout if every caller is passing the same value?

Edit: I see it's related to the source of the timeout being the framework package.  Hopefully moving this file to the framework package would allow use of the context value rather than having to pass it in.

test/common/wait.go
marun: Since you're creating this file, how about putting it in `test/e2e/framework` instead?  `test/common` is a legacy of `test/integration` and `test/e2e` needing to share code, but `test/integration` no longer exists and pretty much everything in this path should end up in `test/e2e/framework`.

test/e2e/servicedns.go
```diff
@@ -103,7 +104,7 @@ var _ = Describe("ServiceDNS", func() {
 
 		serviceDNS.Status = serviceDNSStatus
 		By("Waiting for the ServiceDNS object to have correct status")
-		common.WaitForObject(tl, namespace, serviceDNS.Name, objectGetter, serviceDNS, framework.PollInterval, framework.TestContext.SingleCallTimeout)
+		common.WaitForObject(tl, namespace, serviceDNS.Name, objectGetter, serviceDNS, framework.PollInterval, framework.TestContext.SingleCallTimeout, common.Equivalent)
```
marun: (No action required) Rather than requiring that every caller be updated to specify this argument that you want to vary at a new call site, consider adding a new function that the old function can call. This has the advantage of both minimizing code churn and minimizing the verbosity of calls that share arguments.

```golang
func WaitForObjectWithCheck(..., equivalentFunc func(...)) {...}

func WaitForObject(...) {
  WaitForObjectWithCheck(..., common.Equivalent)
}

```

pkg/controller/webhook/kubefedconfig/webhook.go
```diff
@@ -83,32 +89,68 @@ func (a *KubeFedConfigAdmissionHook) Validate(admissionSpec *admissionv1beta1.Ad
...
 
+	if reflect.DeepEqual(admittingObject, defaultedObject) {
+		status.Allowed = true
+		return status
+	}
+
+	// TODO(font) Optimize by generalizing the ability to add only the fields
+	// that have been defaulted. If merge patch is ever supported use that
+	// instead.
+	patchOperations := []patchOperation{
```
marun: Apologies in advance for my ignorance. Is a mutating webhook called on both create and update?  I assumed that it would only be on create, but the use of a patch operation would seem to suggest otherwise.  Is this the canonical way to do it vs just updating the resource that gets passed in?

test/e2e/defaulting.go
```diff
@@ -0,0 +1,174 @@
...
+				fedConfig := webhookDefaultedKubeFedConfig
+				err := client.Delete(context.TODO(), fedConfig, fedConfig.Namespace, fedConfig.Name)
+				if err != nil && !apierrors.IsNotFound(err) {
+					f.Logger().Errorf("Error deleting %s %q: %v", resourceName, *qualifiedName, err)
+				}
+			})
+		}
+	})
+
+	// TODO(font): This test explicitly verifies the controller-manager detects
```
marun: If the crds and admission webhook were deployed via a subchart whose installation was guaranteed to be complete before deployment of the controller manager, would #983 still be a problem?

pkg/apis/core/v1beta1/defaults/defaults_test.go
```diff
@@ -0,0 +1,149 @@
...
+func TestSetDefaultKubeFedConfig(t *testing.T) {
+	type KubeFedConfigComparison struct {
+		original *v1beta1.KubeFedConfig
+		modified *v1beta1.KubeFedConfig
+	}
+
+	successCases := map[string]KubeFedConfigComparison{}
+
+	// ControllerDuration
+	availableDelayKFC := defaultKubeFedConfig()
```
marun: (No action required) I find these kinds of boilerplate-heavy tests really hard to review. I guess there simply aren't enough of these tests to justify a more data-driven approach, but I would prefer to avoid creating any more of these than absolutely necessary.

font:
@marun I've responded to comments and pushed an update incorporating feedback. Since most changes were mechanical and comments I've squashed into existing commits. PTAL.

pkg/controller/webhook/kubefedconfig/webhook.go
marun: Hmm, that's interesting.  I was assuming direct mutation of a request (ala a python wsgi pipeline).  I'm curious what prompted this way of doing things.

config/kubefedconfig.yaml
marun: There was at least one user that was wanting to run the controller manager outside of the cluster for some reason I can't recall, which suggested that providing the config via argument wasn't strictly test-only.  But given that we're already supporting loading `--kubefed-config`, adding a call to set defaults after unmarshaling seems cheaper than to me than having to maintain defaults manually.

marun:
LGTM

marun:
Thank you!

/lgtm

PR #1017, opened 2019-06-25
Fixup override docs and validation

marun:


scripts/sync-up-helm-chart.sh
```diff
@@ -27,6 +27,16 @@ CHART_FEDERATED_CRD_DIR="${CHART_FEDERATED_CRD_DIR:-charts/kubefed/charts/contro
 CHART_FEDERATED_PROPAGATION_DIR="${CHART_FEDERATED_PROPAGATION_DIR:-charts/kubefed/templates}"
 TEMP_CRDS_YAML="/tmp/kubefed-crds.yaml"
 
+OS=`uname`
+SED=sed
+if [ "$OS" == "Darwin" ];then
```
font: nit: s/$OS/${OS}/

marun: Done

marun:
Addressed comment.

pkg/kubefedctl/enable/validation.go
```diff
@@ -137,6 +138,11 @@ func federatedTypeValidationSchema(templateSchema map[string]v1beta1.JSONSchemaP
 												},
 											},
 										},
+										// TODO(marun) Enable defaulting for federated types
```
font: I don't think enabling defaulting for federated types should constitute making `op` required, as there will be CRD client-side validation that would still fail. However, we can still make it required in validation while still leaving the field optional.

marun: Fair point, I'll remove the comment. I find this behavior very counter-intuitive, and question whether client-side validation is more a blessing or a curse.

marun:
Addressed comment.

shashidharatd:
Thanks for taking it up.
/lgtm

PR #1019, opened 2019-06-25
Adding e2e target for generating binaries for the e2e tests

### LINKED ISSUE #1000: Consuming the e2e binaries as release artifacts
The current setup for kubefed repo doesn't provide the e2e binaries as release artifacts for carrying out end-to-end tests. Currently, we're not only downloading a container image or a binary file but also building it locally.

**Why is this needed**:
With this feature, it would be easier to get rid of manual activities such as building tests, adding dependencies, etc. and will avoid vendoring issues. Let's discuss this in the next WG meeting
/kind feature

sohankunkerkar:
Fixes #1000


sohankunkerkar:
@font As discussed, I have handled that business logic in the Makefile. Could you please take a look at it?

Makefile
```diff
@@ -56,17 +58,20 @@ LDFLAG_OPTIONS = -ldflags "-X sigs.k8s.io/kubefed/pkg/version.version=$(GIT_VERS
...
 
+GO_TESTCMD  = CGO_ENABLED=0 go test $(COMPILE_FLAG) $(LDFLAG_OPTIONS)
+
 TESTARGS ?= $(VERBOSE_FLAG) -timeout 60s
 TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
 TEST_CMD = go test $(TESTARGS)
 TEST = $(TEST_CMD) $(TEST_PKGS)
 
 DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
+DOCKER_BINARY_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(GOPATH)/src/$(GOTARGET) -w $(GOPATH)/src/$(GOTARGET) $(BUILD_IMAGE) /bin/sh -c
```
font: Since both are building binaries how about renaming this to `DOCKER_GOPATH_BUILD` or similar?

Makefile
```diff
@@ -97,15 +102,27 @@ $(1)-$(2): bindir
 	$(DOCKER_BUILD) 'GOARCH=amd64 GOOS=$(2) $(GO_BUILDCMD) -o $(1)-$(2) cmd/$(3)/main.go'
 ALL_BINS := $(ALL_BINS) $(1)-$(2)
 endef
+
+define BINARY_template
```
font: Let's remove `BINARY` from this name since these templates are working with binaries and it's not specific to this template. Perhaps something like `E2E_OS_template`?

Makefile
sohankunkerkar: Fixed.

Makefile
sohankunkerkar: yeah, makes sense to me. Fixed!

sohankunkerkar:
@font All comments have been addressed.

sohankunkerkar:
@font Done!

Makefile
```diff
@@ -97,15 +102,27 @@ $(1)-$(2): bindir
 	$(DOCKER_BUILD) 'GOARCH=amd64 GOOS=$(2) $(GO_BUILDCMD) -o $(1)-$(2) cmd/$(3)/main.go'
 ALL_BINS := $(ALL_BINS) $(1)-$(2)
 endef
+
+define E2E_OS_template
+$(1)-$(2): bindir
+	$(DOCKER_GOPATH_BUILD) 'GOPATH=$(GOPATH) GOARCH=amd64 GOOS=$(2) $(GO_TESTCMD) -o $(1)-$(2) ./test/$(3)'
+ALL_BINS := $(ALL_BINS) $(1)-$(2)
+endef
+
```
xunpan: it seems moving `E2E_OS_template` after line 112 is better in code locality. Line 112 handles `OS_template` define above.

Makefile
```diff
@@ -56,17 +58,20 @@ LDFLAG_OPTIONS = -ldflags "-X sigs.k8s.io/kubefed/pkg/version.version=$(GIT_VERS
...
 
+GO_TESTCMD  = CGO_ENABLED=0 go test $(COMPILE_FLAG) $(LDFLAG_OPTIONS)
+
 TESTARGS ?= $(VERBOSE_FLAG) -timeout 60s
 TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
 TEST_CMD = go test $(TESTARGS)
 TEST = $(TEST_CMD) $(TEST_PKGS)
 
 DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
+DOCKER_GOPATH_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(GOPATH)/src/$(GOTARGET) -w $(GOPATH)/src/$(GOTARGET) $(BUILD_IMAGE) /bin/sh -c
```
xunpan: Could you help to explain more on this?  I did not see any needs to use new one than `DOCKER_BUILD`.

sohankunkerkar: @xunpan This was my plan earlier where I wanted to reuse the `DOCKER_BUILD` parameter as is. However, I ran into the following issue when I was trying to run e2e tests binary:
```
INFO: Error loading type config fixtures: Error reading fixture from path "/go/src/sigs.k8s.io/kubefed/test/common/fixtures": open /go/src/sigs.k8s.io/kubefed/test/common/fixtures: no such file or directory
```
Where my `GOPATH` is `/go`. The moment I introduced that value in `BUILDMNT` path, the issue got resolved. I would love to hear a better approach for solving this issue if you have any.

Makefile
sohankunkerkar: Sure, I will fix this right away.

Makefile
```diff
@@ -39,6 +39,7 @@ endif
 ifneq ($(VERBOSE),)
 VERBOSE_FLAG = -v
 endif
+COMPILE_FLAG = -c
```
marun: Why is it necessary to define this if it's only used once?

Makefile
```diff
@@ -56,17 +58,20 @@ LDFLAG_OPTIONS = -ldflags "-X sigs.k8s.io/kubefed/pkg/version.version=$(GIT_VERS
 
 GO_BUILDCMD = CGO_ENABLED=0 go build $(VERBOSE_FLAG) $(LDFLAG_OPTIONS)
 
+GO_TESTCMD  = CGO_ENABLED=0 go test $(COMPILE_FLAG) $(LDFLAG_OPTIONS)
```
marun: Same question here - is there a reason not to localize this to the macro?

Makefile
marun: I'm also confused - is this change required to build the binary or to run the binary that gets built?

The dependency of the test binary on fixture in the working directory for fixture appears to complicate downstream consumption of a release binary since the fixture isn't included with the release. That suggests either building fixture into the binary or reconsidering shipping the binary.

marun:
As per our slack conversation, one of the reasons the fixtures were not defined as code was to make it easier for downstream consumers to test federation of types that are not federated by default without having to involve code. It's not clear to me that building the fixture into the test binary is a desirable evolution, and for me calls into question what problem is being solved by adding support for shipping a test binary as a release artifact.

xunpan:
@marun
My understanding is that `e2e` should be run as independent binary for basic federation test of kubefed users. So, the rule is to release it as an independent binary. It means `e2e` should be run after it is installed into the system (without kubefed source code). So, we may need to do either of below things:
 - releasing binary together with fixtures
 - encoding fixtures in source code

For the 1st solution, it makes management complex. The binary and related fixture files should be managed in release package and deployment OS. So, solution 2 seems better for a real independent binary for use.

marun:
@xunpan While there may be value in enabling a user to validate the health of a control plane via something like a smoke test, I think that is a separate requirement that would likely involve adding the functionality to kubefedctl rather than shipping an e2e binary. I think performing testing with an e2e binary is something a developer would do (or a downstream consumer like Red Hat, in this case) since the technical complexity of running the tests and interpreting their results is non-trivial. Given that, it's not clear to me why a developer or downstream consumer wouldn't simply run the tests from a source tree.

sohankunkerkar:
Closing this PR after having an offline discussion with Maru.

sohankunkerkar:
Sorry for the confusion here but this pull is opened again after having some discussion with the maintainers.

test/e2e/leaderelection.go
```diff
@@ -76,17 +76,19 @@ var _ = Describe("Leader Elector", func() {
 })
 
 func spawnControllerManagerProcess(tl common.TestLogger, kubeConfigPath, namespace string) (*exec.Cmd, io.ReadCloser, error) {
-	// Get the directory of the current executable
-	_, filename, _, _ := runtime.Caller(0)
-	managedPath := filepath.Dir(filename)
-	confFile, err := filepath.Abs(fmt.Sprintf("%s/../../config/kubefedconfig.yaml", managedPath))
+	kubefedconfigAsset := common.MustAsset("config/kubefedconfig.yaml")
+	tmpFile, err := ioutil.TempFile(os.TempDir(), "kubefedconfig.yaml")
```
marun: (No action required) There is no requirement for the file to be named `kubefedconfig.yaml`, it could instead be truly random.  Also, it's not necessary to pass `os.TempDir()` to `ioutil.TempFile` - passing an empty string has the same result.

test/e2e/leaderelection.go
```diff
@@ -76,17 +76,19 @@ var _ = Describe("Leader Elector", func() {
 })
 
 func spawnControllerManagerProcess(tl common.TestLogger, kubeConfigPath, namespace string) (*exec.Cmd, io.ReadCloser, error) {
-	// Get the directory of the current executable
-	_, filename, _, _ := runtime.Caller(0)
-	managedPath := filepath.Dir(filename)
-	confFile, err := filepath.Abs(fmt.Sprintf("%s/../../config/kubefedconfig.yaml", managedPath))
+	kubefedconfigAsset := common.MustAsset("config/kubefedconfig.yaml")
```
marun: Please use a more specific asset name like 'leaderelectionconfig.yaml' to ensure that future maintainers don't reuse it for other purposes accidentally.

(No action required) I'm not sure it's desirable to suffix the name of the variable with 'Assert'.  Once loaded, it's just a byte array containing yaml, and there is no need to indicate where it was loaded from (e.g. I wouldn't expect a variable containing a byte array loaded from disk to be named 'byteArrayFromDisk').

test/common/fixtures.go
```diff
@@ -42,28 +41,20 @@ func TypeConfigFixturesOrDie(tl TestLogger) map[string]*unstructured.Unstructure
...
-		typeConfigName := strings.TrimSuffix(file.Name(), suffix)
-
-		filename := filepath.Join(path, file.Name())
+	prefix := "test/common/fixtures/"
+	for _, file := range AssetNames() {
 		fixture := &unstructured.Unstructured{}
-		err := kfenable.DecodeYAMLFromFile(filename, fixture)
-		if err != nil {
-			return nil, errors.Wrapf(err, "Error reading fixture for %q", typeConfigName)
+		if strings.Contains(file, prefix) {
```
marun: Is it preferable not to use `HasPrefix`?

test/e2e/framework/enable.go
```diff
@@ -17,46 +17,28 @@ limitations under the License.
...
-			continue
-		}
-		filename := filepath.Join(path, file.Name())
-		obj := kfenable.NewEnableTypeDirective()
-		err := kfenable.DecodeYAMLFromFile(filename, obj)
-		if err != nil {
-			tl.Fatalf("Error loading EnableTypeDirective from file %q: %v", filename, err)
+	prefix := "config/enabletypedirectives/"
+	for _, file := range common.AssetNames() {
+		if strings.Contains(file, prefix) {
```
marun: `HasPrefix`?

test/e2e/framework/enable.go
```diff
@@ -17,46 +17,28 @@ limitations under the License.
...
+			fixtureAsset := common.MustAsset(file)
+			obj := kfenable.NewEnableTypeDirective()
+			filename := strings.TrimSuffix(strings.TrimPrefix(file, prefix), suffix)
+			err := kfenable.DecodeYAML(bytes.NewBuffer(fixtureAsset), obj)
+			if err != nil {
+				tl.Fatalf("Error loading EnableTypeDirective from file %q: %v", filename, err)
```
marun: Does it make sense to say 'from this file' given that the data is being loaded from the binary?

sohankunkerkar:
@marun PTAL

Makefile
```diff
@@ -124,6 +135,7 @@ endif
 
 generate: generate-code kubefedctl
 	./scripts/sync-up-helm-chart.sh
+	./scripts/check-generated-bindata-content.sh
```
marun: The script being called here should be updating the bindata content, not checking it.  `check-make-generate-output` in `pre-comment.sh` calls `make generate` and then verifies that the working tree has not changed.

scripts/check-generated-bindata-content.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
marun: As per a previous comment, this file should be updating the generated content (i.e. it's intent should be something like `update-bindata.go`). The check for whether the tree is current is performed by `pre-commit.sh` in ci.

scripts/pre-commit.sh
```diff
@@ -22,14 +22,21 @@ set -o pipefail
 
 source "$(dirname "${BASH_SOURCE}")/util.sh"
 ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+TEMP_DIR="$(dirname $(mktemp -u))"
```
marun: A simpler way to find this would be `${TMPDIR}`.

But would it make more sense to create a new temporary directory with `mktemp -d` to ensure that a clean path was being used?

scripts/pre-commit.sh
```diff
@@ -22,14 +22,21 @@ set -o pipefail
...
+ARCH="$(go env GOARCH)"
+PLATFORM="${OS}-${ARCH}"
 NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
 JOIN_CLUSTERS="${JOIN_CLUSTERS:-}"
 DOWNLOAD_BINARIES="${DOWNLOAD_BINARIES:-}"
 CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
 COMMON_TEST_CMD="go test -v"
+COMMON_BINARY_TEST_CMD="${TEMP_DIR}/e2e-${PLATFORM}"
 COMMON_TEST_ARGS="./test/e2e -args  -kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
+COMMON_BINARY_TEST_ARGS="-kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
```
marun: Given that these arguments are a subset of those found in `COMMON_TEST_ARGS` would it make sense to remove the duplication?

test/e2e/leaderelection.go
marun: Also, `kubeFedConfig`?

test/common/fixtures.go
```diff
@@ -42,28 +41,20 @@ func TypeConfigFixturesOrDie(tl TestLogger) map[string]*unstructured.Unstructure
...
 	suffix := ".yaml"
-	for _, file := range files {
-		if !strings.HasSuffix(file.Name(), suffix) {
-			continue
-		}
-
-		typeConfigName := strings.TrimSuffix(file.Name(), suffix)
-
-		filename := filepath.Join(path, file.Name())
+	prefix := "test/common/fixtures/"
```
marun: (No action required) Food for thought:

```golang 
func decodeYamlFromBindata(filename, prefix string, obj interface{}) (string, error) {
  if !strings.HasPrefix(filename, prefix) {
   return "", nil
  }
  yaml := MustAssert(filename)
  name := strings.TrimSuffix(..., ".yaml")
  err := kfenable.DecodeYAML(.., obj)
  if err != nil {
    return "", err
  } 
  return name, nil
}
```

test/common/fixtures.go
```diff
@@ -42,28 +41,20 @@ func TypeConfigFixturesOrDie(tl TestLogger) map[string]*unstructured.Unstructure
...
+	for _, file := range AssetNames() {
 		fixture := &unstructured.Unstructured{}
-		err := kfenable.DecodeYAMLFromFile(filename, fixture)
-		if err != nil {
-			return nil, errors.Wrapf(err, "Error reading fixture for %q", typeConfigName)
+		if strings.HasPrefix(file, prefix) {
```
marun: It's generally preferable to minimize the scope of indentation where possible.  In this case, that would mean doing the following at the top of the loop block:

```golang
if !strings.HasPrefix(...) {
  continue
} 
...
```

sohankunkerkar:
@marun PTAL

sohankunkerkar:
@font @xunpan Could you please review this PR?

scripts/pre-commit.sh
```diff
@@ -22,23 +22,31 @@ set -o pipefail
...
 # given control plane scope.
-IN_MEMORY_E2E_TEST_CMD="${COMMON_TEST_CMD} -timeout 900s -race ${COMMON_TEST_ARGS} -in-memory-controllers=true -limited-scope-in-memory-controllers=false"
+IN_MEMORY_E2E_TEST_CMD="${COMMON_TEST_CMD} -timeout 900s -race ./test/e2e -args ${COMMON_TEST_ARGS} -in-memory-controllers=true -limited-scope-in-memory-controllers=false"
 
 function build-binaries() {
   ${MAKE_CMD} hyperfed
   ${MAKE_CMD} controller
   ${MAKE_CMD} kubefedctl
+  ${MAKE_CMD} e2e
+  cp ${ROOT_DIR}/bin/e2e-${PLATFORM} ${TEMP_DIR}
```
font: What's the reason for the temp dir? Is it to ensure no dependency on the static files in the path? Please add a comment to clarify.

scripts/update-bindata.sh
```diff
@@ -0,0 +1,44 @@
...
+# the tree without corresponding fixture to ensure federation of the
+# target type will be tested.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+
+function update-bindata(){
```
font: Did you mean to define this as a function? The way it's written the caller needs to source the script and then invoke the function.

scripts/update-bindata.sh
```diff
@@ -0,0 +1,44 @@
...
+  else
+    echo "Downloading go-bindata"
+    go get -u github.com/jteeuwen/go-bindata/go-bindata/...
+  fi
+
+ "go-bindata \
+  -nocompress \
+  -nometadata \
+  -pkg 'common' \
+  -o ${ROOT_DIR}/bindata_generated.go \
```
font: Would it make sense to output to `${ROOT_DIR}/test/common/bindata.go`, the actual location of the generated content, so that users can easily commit the modified changes if there are any?

scripts/pre-commit.sh
sohankunkerkar: Fixed.

docs/development.md
```diff
@@ -188,6 +189,23 @@ KubeFed controller manager:
 
 Follow the [cleanup instructions in the user guide](../charts/kubefed/README.md#uninstalling-the-chart).
 
+## Embedding static files using go-bindata
+This project is using `go-bindata` tool for embedding static files into its e2e test-suite to enable the creation of a self-contained e2e binary.
+
+You can install this utility using the following command:
```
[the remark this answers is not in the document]
sohankunkerkar: Fixed.

scripts/update-bindata.sh
sohankunkerkar: Fixed

docs/development.md
```diff
@@ -188,6 +189,14 @@ KubeFed controller manager:
...
+`go get -u github.com/jteeuwen/go-bindata/go-bindata/...`
+
+Use `make generate` to regenerate the `bindata.go` file in case the bundled content changes. It's necessary to follow this step to ensure that e2e test-suite passes the CI build. Please refer [this](../scripts/update-bindata.sh) script for more information.
```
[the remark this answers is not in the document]
sohankunkerkar: Fixed

scripts/download-binaries.sh
```diff
@@ -67,10 +67,25 @@ curl "${curl_args}O" "${golint_url}" \
...
+    && rm -rf ${gobindata_dir}
+
 echo    "# destination:"
 echo    "#   ${dest_dir}"
 echo    "# versions:"
 echo -n "#   kubectl:        "; "${dest_dir}/kubectl" version --client --short
 echo -n "#   kubebuilder:    "; "${dest_dir}/kubebuilder" version
 echo -n "#   helm:           "; "${dest_dir}/helm" version --client --short
 echo -n "#   golangci-lint:  "; "${dest_dir}/golangci-lint" --version
+echo -n "#   go-bindata:     "; "${dest_dir}/go-bindata" -version
```
marun: Please ensure your editor is configured to add a newline at the end of the file.

scripts/pre-commit.sh
```diff
@@ -22,23 +22,33 @@ set -o pipefail
...
 NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
 JOIN_CLUSTERS="${JOIN_CLUSTERS:-}"
 DOWNLOAD_BINARIES="${DOWNLOAD_BINARIES:-}"
 CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
 COMMON_TEST_CMD="go test -v"
-COMMON_TEST_ARGS="./test/e2e -args  -kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
-E2E_TEST_CMD="${COMMON_TEST_CMD} ${COMMON_TEST_ARGS}"
+COMMON_BINARY_TEST_CMD="${TEMP_DIR}/e2e-${PLATFORM}"
+COMMON_TEST_ARGS="-kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
+E2E_TEST_CMD="${COMMON_TEST_CMD} ./test/e2e -args ${COMMON_TEST_ARGS}"
```
marun: This var appears unused?

scripts/update-bindata.sh
```diff
@@ -0,0 +1,45 @@
...
+# target type will be tested.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+
+if command -v go-bindata > /dev/null; then
+cd "${ROOT_DIR}"
```
marun: Please be consistent in indenting 2 spaces in an if block.

test/common/fixtures.go
```diff
@@ -42,29 +41,30 @@ func TypeConfigFixturesOrDie(tl TestLogger) map[string]*unstructured.Unstructure
...
-		typeConfigName := strings.TrimSuffix(file.Name(), suffix)
-
-		filename := filepath.Join(path, file.Name())
+	for _, file := range AssetNames() {
 		fixture := &unstructured.Unstructured{}
-		err := kfenable.DecodeYAMLFromFile(filename, fixture)
+		typeConfigName, err := decodeYamlFromBindata(file, "test/common/fixtures/", fixture)
 		if err != nil {
-			return nil, errors.Wrapf(err, "Error reading fixture for %q", typeConfigName)
+			return nil, errors.Wrap(err, "Error reading a fixture")
```
marun: -> Error reading a fixture from %q(file)

test/common/fixtures.go
```diff
@@ -42,29 +41,30 @@ func TypeConfigFixturesOrDie(tl TestLogger) map[string]*unstructured.Unstructure
...
+		if len(typeConfigName) != 0 {
+			fixtures[typeConfigName] = fixture
 		}
-		fixtures[typeConfigName] = fixture
 	}
 
 	return fixtures, nil
 }
+
+func decodeYamlFromBindata(filename, prefix string, obj interface{}) (string, error) {
```
marun: Please be consistent in use of casing of acronyms (e.g. `decodeYAMLFromBindata` to be consistent with `DecodeYAML`).

test/e2e/framework/enable.go
```diff
@@ -17,46 +17,28 @@ limitations under the License.
...
-			continue
-		}
-		filename := filepath.Join(path, file.Name())
-		obj := kfenable.NewEnableTypeDirective()
-		err := kfenable.DecodeYAMLFromFile(filename, obj)
-		if err != nil {
-			tl.Fatalf("Error loading EnableTypeDirective from file %q: %v", filename, err)
+	prefix := "config/enabletypedirectives/"
+	for _, file := range common.AssetNames() {
+		if strings.HasPrefix(file, prefix) {
```
marun: (No action required) Would it make sense to rewrite this block with `decodeYAMLFromBindata`?

docs/development.md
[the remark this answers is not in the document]
marun: Now that go-bindata is downloaded by the script, please update this documentation to reference the need to execute the script rather than indicating how to install the dependency directly.

test/e2e/framework/enable.go
sohankunkerkar: Makes sense to me. I have already taken the necessary action here.

test/common/fixtures.go
sohankunkerkar: Fixed

test/common/fixtures.go
sohankunkerkar: Done

scripts/update-bindata.sh
sohankunkerkar: Ack

scripts/pre-commit.sh
sohankunkerkar: Yeah, I will remove it from the script.

scripts/download-binaries.sh
sohankunkerkar: Fixed

sohankunkerkar:
@marun I have addressed your comments.

scripts/pre-commit.sh
```diff
@@ -22,23 +22,32 @@ set -o pipefail
...
+ARCH="$(go env GOARCH)"
+PLATFORM="${OS}-${ARCH}"
 NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
 JOIN_CLUSTERS="${JOIN_CLUSTERS:-}"
 DOWNLOAD_BINARIES="${DOWNLOAD_BINARIES:-}"
 CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
 COMMON_TEST_CMD="go test -v"
-COMMON_TEST_ARGS="./test/e2e -args  -kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
-E2E_TEST_CMD="${COMMON_TEST_CMD} ${COMMON_TEST_ARGS}"
+COMMON_BINARY_TEST_CMD="${TEMP_DIR}/e2e-${PLATFORM}"
```
marun: (No action required) Given that this is used only once for `E2E_BINARY_TEST_CMD`, maybe just inline?

scripts/pre-commit.sh
```diff
@@ -22,23 +22,32 @@ set -o pipefail
...
+TEMP_DIR="$(mktemp -d)"
 MAKE_CMD="make -C ${ROOT_DIR}"
+OS="$(go env GOOS)"
+ARCH="$(go env GOARCH)"
+PLATFORM="${OS}-${ARCH}"
 NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
 JOIN_CLUSTERS="${JOIN_CLUSTERS:-}"
 DOWNLOAD_BINARIES="${DOWNLOAD_BINARIES:-}"
 CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
 COMMON_TEST_CMD="go test -v"
```
marun: (No action required) Given that this is used only once for `IN_MEMORY_E2E_TEST_CMD `, maybe just inline?

scripts/pre-commit.sh
```diff
@@ -22,23 +22,32 @@ set -o pipefail
...
-COMMON_TEST_ARGS="./test/e2e -args  -kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
-E2E_TEST_CMD="${COMMON_TEST_CMD} ${COMMON_TEST_ARGS}"
+COMMON_BINARY_TEST_CMD="${TEMP_DIR}/e2e-${PLATFORM}"
+COMMON_TEST_ARGS="-kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs"
+E2E_BINARY_TEST_CMD="${COMMON_BINARY_TEST_CMD} ${COMMON_TEST_ARGS}"
```
marun: (No action required) Maybe rename to `E2E_TEST_CMD` since the in-memory command is already differentiated?

scripts/update-bindata.sh
```diff
@@ -0,0 +1,45 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+
+if command -v go-bindata > /dev/null; then
+  cd "${ROOT_DIR}"
+  "./bin/go-bindata" \
```
marun: (No action required) Given that the check on 27 doesn't specify a path (implying that the check is against `PATH` rather than in `./bin`, does this invocation need to be qualified?

scripts/download-binaries.sh
```diff
@@ -67,10 +67,25 @@ curl "${curl_args}O" "${golint_url}" \
     && tar xzfP "${golint_tgz}" -C "${dest_dir}" --strip-components=1 "${golint_dir}/golangci-lint" \
     && rm "${golint_tgz}"
 
+gobindata_version="3.1.2"
```
marun: Ugh, I didn't realize the extent of go-bindata's messy history. And contrary to my request, there is no binary release available. Given that what appears here is essentially what `go get` does, I think it would be preferable to just use that and perhaps symlink the resulting binary into `./bin` for consistency.

sohankunkerkar: @marun This (https://github.com/go-bindata/go-bindata/releases) repo is maintained and they recently cut a new release.

scripts/pre-commit.sh
sohankunkerkar: Makes sense to me. I would change it in code too.

scripts/pre-commit.sh
sohankunkerkar: Fixed.

scripts/update-bindata.sh
sohankunkerkar: I feel there's no need to use `if` condition here b/c it's already handled in the `download-binaries.sh` script.

test/common/bindata.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Code generated by go-bindata.
+// Code generated for package common by go-bindata DO NOT EDIT. (@generated)
```
xunpan: My understanding is that this file is a generated file. So, it is generated in previous `commit`. But it seems generate again here in this `commit`. Any information for sharing for this?

docs/development.md
```diff
@@ -188,6 +189,13 @@ KubeFed controller manager:
 
 Follow the [cleanup instructions in the user guide](../charts/kubefed/README.md#uninstalling-the-chart).
 
+## Embedding static files using go-bindata
+This project is using `go-bindata` tool for embedding static files into its e2e test-suite to enable the creation of a self-contained e2e binary.
```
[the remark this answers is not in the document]
marun: I think the formatting is fine because 'e2e binary' is a noun, not the name of an actual binary.  The actual name is `e2e-<os>-<arch>`.

docs/development.md
```diff
@@ -188,6 +189,13 @@ KubeFed controller manager:
...
+
+You can install this utility using the [download-binaries.sh](../scripts/download-binaries.sh#L70) script.
```
marun: Please remove the line number from the link so we don't have to maintain it. It should be sufficient to link to the script.

scripts/download-binaries.sh
marun: As per our offline conversation, please update to `go get` in preference to what appears here.

scripts/update-bindata.sh
```diff
@@ -0,0 +1,40 @@
...
+# the tree without corresponding fixture to ensure federation of the
+# target type will be tested.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
+
+cd "${ROOT_DIR}"
```
marun: My previous comment was a minor question regarding the need to prefix the binary with `./bin/` given that `command` was checking for the binary in the `PATH` (i.e. if the conditional succeeded the binary could be invoked without a path). I wasn't suggesting removing the conditional, since i's nice for the user if the script detects that the command is missing and it can then suggest that the user invoke `scripts/download-binaries.sh`.

scripts/pre-commit.sh
font: What's the difference between running this from a `/tmp/tmp.XXXXXXXXXX/` directory verses `${ROOT_DIR}/bin/`? The original code was using a relative path (constructed from an absolute path to the binary) that would break under both circumstances and the new code just reads the generated data compiled in.

docs/development.md
```diff
@@ -188,6 +189,13 @@ KubeFed controller manager:
 
 Follow the [cleanup instructions in the user guide](../charts/kubefed/README.md#uninstalling-the-chart).
 
+## Embedding static files using go-bindata
+This project is using `go-bindata` tool for embedding static files into its e2e test-suite to enable the creation of a self-contained e2e binary.
+
+You can install this utility using the [download-binaries.sh](../scripts/download-binaries.sh) script.
+
+Use `make generate` to regenerate the `bindata.go` file in case the bundled content changes. It's necessary to follow this step to ensure that e2e test-suite passes the CI build. Please refer to [this](../scripts/update-bindata.sh) script for more information.
```
font: Please adjust your editor to format the lengths of these lines to match the rest of the file.

scripts/pre-commit.sh
sohankunkerkar: https://github.com/kubernetes-sigs/kubefed/pull/1019#pullrequestreview-265056541
After having this discussion with @marun, it's decided to run this test binary outside of the `GOPATH` just to ensure that it's truly a self-contained binary.

pmorie: That is the right way to verify that the binary isn't loading resources from the gopath.

docs/development.md
sohankunkerkar: Fixed.

sohankunkerkar:
@font  Ready for your review.

scripts/download-binaries.sh
```diff
@@ -67,10 +67,15 @@ curl "${curl_args}O" "${golint_url}" \
     && tar xzfP "${golint_tgz}" -C "${dest_dir}" --strip-components=1 "${golint_dir}/golangci-lint" \
     && rm "${golint_tgz}"
 
+# Install go-bindata tool
+go get -u github.com/go-bindata/go-bindata/...
```
font: A couple of issues:
1. The assumption here is that `GOBIN` is included in the user's `PATH` instead of installing or linking in the kubefed `bin` directory as the other downloads are doing in this script, because the user guide instructs to add kubefed `bin` to the `PATH`. This isn't an issue for me personally as I have `GOBIN` in my `PATH`, but others may run into it.
2. We should probably download and install a known version. If there isn't an official asset we can download, I would suggest checking out a specific tag or commit, that way we're always using the same version.

scripts/update-bindata.sh
```diff
@@ -0,0 +1,44 @@
...
+  "go-bindata" \
+    -nocompress \
+    -nometadata \
+    -pkg 'common' \
+    -o test/common/bindata.go \
+    test/common/fixtures \
+    config/kubefedconfig.yaml \
+    config/enabletypedirectives
+
+  cat ./hack/boilerplate.go.txt ./test/common/bindata.go > tmp \
```
font: Let's use a more unique name than `tmp`.

scripts/download-binaries.sh
sohankunkerkar: > We should probably download and install a known version. If there isn't an official asset we can download, I would suggest checking out a specific tag or commit, that way we're always using the same version.

That's what I did intially but chaged my approach after getting a comment from @marun  https://github.com/kubernetes-sigs/kubefed/pull/1019#discussion_r314254790

scripts/update-bindata.sh
sohankunkerkar: I agree with you. Let me change the name.

scripts/download-binaries.sh
sohankunkerkar: @font Let me know your thoughts

font: I think using `go get` is the right thing to do, but we should still use a specific version. Here's an example that was needed with `kind` during a time when there were no releases available. Instead of a commit sha you could use a tag. The only other thing you could add is perhaps a link to it from kubefed's bin directory.
https://github.com/kubernetes-sigs/kubefed/blob/8d0715420cdb4a336a32e5179c3ab1a2024ad309/scripts/download-e2e-binaries.sh#L38-L49

PR #1020, opened 2019-06-26
Update install instructions for kubefedctl

runyontr:
The release binaries have a new naming syntax compared with the old release.

Currently, the version specified in the path differs from the version specified in the file name, it may be something we want to look at/adjust going forward so we don't need to have the `v${VERSION}` in the path:


marun:
Thanks!

/lgtm
/approve

PR #1023, opened 2019-07-01
Removed outdated script reference

### LINKED ISSUE #1022: Dependencies are outdated in installation docs
**What happened**:
The steps at https://github.com/kubernetes-sigs/kubefed/blob/master/docs/installation.md#required-binaries are outdated - we no longer use the script installation so this is not required.

/kind bug

abrennan89:
Fixes #1022


marun:
Good catch!

/approve
/lgtm

PR #1025, opened 2019-07-02
fix dns endpoint controller merged two different type record's targets with same dns name  

Dingshujie:
**What happened:**
we have cluster A in zone A, cluster B in zone A,
now we create a federated service nginx, deploy service only in cluster A, then create a servicednsrecord associate with this federated service,  and dns endpoint will generate two zone-level endpoint with same dns name.
clusterA: nginx.default.dsj-test.svc.cn-north-7b.cn-north-7.jessiading.com:Record A: xxx.xxx.xxx.xxx
clusterB: nginx.default.dsj-test.svc.cn-north-7b.cn-north-7.jessiading.com:Record CNAME: nginx.default.dsj-test.svc.cn-north-7.jessiading.com
finally, controller will merge two endpont into one,  nginx.default.dsj-test.svc.cn-north-7b.cn-north-7.jessiading.com:Record A: xxx.xxx.xxx.xxx,nginx.default.dsj-test.svc.cn-north-7.jessiading.com
 when external-dns create this endpoint in provider,  get error response, because Record A targets must be ipv4 address list.

### DIFF
--- a/pkg/controller/dnsendpoint/common.go
+++ b/pkg/controller/dnsendpoint/common.go
@@ -118,8 +118,18 @@ func DedupeAndMergeEndpoints(endpoints []*feddnsv1a1.Endpoint) (result []*feddns
 	for i := 1; i < len(endpoints); {
 		if endpoints[i].DNSName == endpoints[i-1].DNSName {
 			// Merge targets
-			endpoints[i-1].Targets = append(endpoints[i-1].Targets, endpoints[i].Targets...)
-			endpoints[i-1].Targets = sortAndRemoveDuplicateTargets(endpoints[i-1].Targets)
+			// dns endpoints controller will generate dns endpoint with same dns name when cluster at same region and az.
+			// if cluster does not have endpoints, controller will generate CNAME record, otherwise will generate A record
+			// here merge same dns name endpoints into one, if a dns name has CNAME and A record, just discard CNAME record, and merge A Record targets.
+			// if only has one type Record, just merge target, remove duplicate targets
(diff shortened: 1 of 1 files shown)


pkg/controller/dnsendpoint/common.go
```diff
@@ -118,8 +118,14 @@ func DedupeAndMergeEndpoints(endpoints []*feddnsv1a1.Endpoint) (result []*feddns
 	for i := 1; i < len(endpoints); {
 		if endpoints[i].DNSName == endpoints[i-1].DNSName {
 			// Merge targets
-			endpoints[i-1].Targets = append(endpoints[i-1].Targets, endpoints[i].Targets...)
-			endpoints[i-1].Targets = sortAndRemoveDuplicateTargets(endpoints[i-1].Targets)
+			if endpoints[i-1].RecordType != endpoints[i].RecordType {
+				if endpoints[i-1].RecordType == RecordTypeCNAME {
```
shashidharatd: (Suggestion) Better to add a comment on why we are doing this, which will help future maintainers.

Dingshujie: done~

shashidharatd: s/enpoint/endpoint/
s/at same region or at same az/at same region and az/
s/if cluster no endpoints/if cluster does not have endpoints/

Dingshujie: thanks. 

fixed. ^^

shashidharatd:
/lgtm
/approve
Thanks @Dingshujie

Issue #1002, opened 2019-06-19
Using `adoptResources:  Disabled` fails to propagate FederatedNamespace, but federates consisting resources.


zeronone:
I have a single-cluster, control-plane scoped federation setup. Although it is an edge case, following the steps below results in an interesting behavior. (`global.scope = "Cluster", syncController.adoptResources = "Disabled"`)

1. Create a `test-namespace` namespace
2. Create a `FederatedDeployment` in `test-namespace`. It won't be federated yet, because we haven't created the associated `FederatedNamespace` yet.
3. Create a `FederatedNamespace` in `test-namespace`.
  - The `FederatedNamespace` status shows as: `PropagationStatus: False`, `Status: AlreadyExists`, however the actual `FederatedDeployment` is schedule normally (`PropgationStatus: true`).

**What would you like to be added**:
I think of following solutions, but happy to hear other's opinions and suggestions.
1. Exclude `namespaces` from `adoptResources` rule. (always adopt)
2. Make special case in case of Host Cluster (`adoptResource` always enabled for `Namespace` in host cluster)

/kind feature

marun:
tl;dr Namespaces in a self-joined host cluster should always be adopted.

This is an interesting case.  Note that disabling adoption is intended to prevent accidental deletion of resources, so whether always adopting makes sense depends on whether a resource can be deleted.

- In the case of a self-joined host cluster, the namespace has to exist and can't be deleted, so always adopting would appear to make sense.
- In the case of a non-host member cluster, the namespace doesn't have to exist and can be deleted, so always adopting does not make sense.

In the case of contained resource propagation (i.e. 3), I think it's fine that resources are propagated if the namespace exists in the member cluster but the namespace can't be adopted.

zeronone:
I would like to work on this task.

marun:
Fixed by #1026

PR #1026, opened 2019-07-02
Always adopt resources for FederatedNamespace in host cluster

zeronone:
- Always adapt resources for `FederatedNamespace` in host cluster. Fix for #1002


marun:
Nice work, thank you!

/approve
/lgtm

PR #1027, opened 2019-07-02
Validation for KubeFedCluster API

### LINKED ISSUE #908: Validations for KubefedCluster API
### Field Validations for [KubefedCluster](https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/apis/core/v1alpha1/kubefedcluster_types.go)

- `KubefedClusterSpec`
  - `APIEndpoint` is a valid API endpoint
- `LocalSecretReference`
  - `Name` is a valid resource name
- `ClusterCondition`
  - `Type` is a valid enum value
- `ConditionStatus`
  - `Status` is a valid enum value

font:
Fixes #908.

### DIFF
--- a/pkg/apis/core/v1beta1/validation/validation.go
+++ b/pkg/apis/core/v1beta1/validation/validation.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apimachineryval "k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,6 +32,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
 
+	"sigs.k8s.io/kubefed/pkg/apis/core/common"
 	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
 	"sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 	"sigs.k8s.io/kubefed/pkg/features"
@@ -147,14 +149,28 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
 	return allErrs
 }
 
-func ValidateKubeFedCluster(obj *v1beta1.KubeFedCluster) field.ErrorList {
-	allErrs := validateKubeFedClusterSpec(&obj.Spec, field.NewPath("spec"))
+func ValidateKubeFedCluster(obj *v1beta1.KubeFedCluster, statusSubResource bool) field.ErrorList {
+	var allErrs field.ErrorList
+	if !statusSubResource {
+		allErrs = validateKubeFedClusterSpec(&obj.Spec, field.NewPath("spec"))
+	} else {
+		allErrs = validateKubeFedClusterStatus(&obj.Status, field.NewPath("status"))
+	}
 	return allErrs
 }
 
 func validateKubeFedClusterSpec(spec *v1beta1.KubeFedClusterSpec, path *field.Path) field.ErrorList {
 	allErrs := validateAPIEndpoint(spec.APIEndpoint, path.Child("apiEndpoint"))
 	allErrs = append(allErrs, validateLocalSecretReference(&spec.SecretRef, path.Child("secretRef"))...)
+	return allErrs
+}
+
+func validateKubeFedClusterStatus(status *v1beta1.KubeFedClusterStatus, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	for i, condition := range status.Conditions {
+		allErrs = append(allErrs, validateClusterCondition(&condition, path.Child("conditions").Index(i))...)
+	}
 	return allErrs
 }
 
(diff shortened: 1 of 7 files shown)


pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -144,6 +149,88 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
...
+	if err != nil {
+		return field.ErrorList{field.Invalid(fldPath, endpt, err.Error())}
+	}
+
+	allErrs := validateEnumStrings(fldPath, hostURL.Scheme, []string{"http", "https"})
+
+	hostname := hostURL.Hostname()
+	dnsErrs := valutil.IsDNS1123Subdomain(hostname)
+	ipErrs := valutil.IsValidIP(hostname)
+	if dnsErrs != nil && ipErrs != nil {
```
marun: I can deduce from this logic that the hostname needs to validate either as a valid subdomain or ip, and that a validation error for both paths indicates an invalid hostname.  I'm not sure both sets of errors will make much sense to the user in the form presented, though.  The test checks for one or the other, but the user would be presented by both error conditions.  What do you think of adding a generic error message instead, similar to what `DefaultServerURL()` returns?

font:
@marun thanks for your review! I've pushed an update with a new commit that addresses the feedback with the plan to squash that prior to merge (hence the hold).

> Is a ca bundle, if provided, something that could be validated?

I was originally validating the cabundle was in base64 encoding, but then discovered it is already decoded by the time the webhook gets it. So aside from that base64 encoding check, I don't think there's much else we can do.

/hold

test/e2e/validation.go
```diff
@@ -130,3 +132,89 @@ var _ = Describe("Core API Validation", func() {
...
+			framework.Skipf("Running validation admission webhook outside of cluster not supported")
+		}
+
+		if hostConfig == nil {
+			userAgent := fmt.Sprintf("test-%s-validation", resourceName)
+			hostConfig = f.HostConfig(userAgent)
+			client = f.Client(userAgent)
+		}
+
+		if validKFC == nil {
```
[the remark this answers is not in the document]
marun: I'd vote for kfcluster over kfcls (cls reads as class to me).  I think config/conf/cfg are all readable options.

pkg/apis/core/v1beta1/validation/validation.go
marun: I think that's a good compromise!

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -144,6 +149,88 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
...
+	}
+
+	allErrs := validateEnumStrings(path, hostURL.Scheme, []string{"http", "https"})
+
+	hostname := hostURL.Hostname()
+	dnsErrs := valutil.IsDNS1123Subdomain(hostname)
+	ipErrs := valutil.IsValidIP(hostname)
+	if dnsErrs != nil && ipErrs != nil {
+		combinedErrMsg := fmt.Sprintf("%s; or %s", strings.Join(ipErrs, ","), strings.Join(dnsErrs, ","))
+		allErrs = append(allErrs, field.Invalid(path, endpoint, combinedErrMsg))
```
marun: The value used for this error message would be for the whole endpoint which could be confusing (e.g. `Invalid value: "https://example_com:8080/foo": must be a valid IP address...`). Would it be possible to use `hostname` as the value parameter instead?

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -144,6 +149,88 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
...
+	} // else one of the two succeeded
+
+	port := hostURL.Port()
+	if port != "" {
+		portInt, _ := strconv.Atoi(port)
```
marun: Would it make sense to report an error for a non-integer port and only check if the port is valid if the port were successfully parsed as an integer?

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -144,6 +149,88 @@ func ValidateFederatedTypeConfigStatus(status *v1beta1.FederatedTypeConfigStatus
...
+		if errs := valutil.IsValidPortNum(portInt); errs != nil {
+			allErrs = append(allErrs, field.Invalid(path, endpoint, strings.Join(errs, ",")))
```
marun: Would it be possible to provide an error message specifying that the port was in error rather than the whole endpoint message (e.g. `Invalid value: 0: port must must be between 1 and 65555)` rather than `Invalid value: "https://example_com:0/foo": must must be between 1 and 65555`)?

pkg/apis/core/v1beta1/validation/validation_test.go
```diff
@@ -332,6 +336,278 @@ func apiResource(apiResource *metav1.APIResource) *v1beta1.APIResource {
...
+		"",
+		"/",
+		"/path/to/endpoint",
+	}
+
+	for _, scheme := range successProtocolSchemes {
+		for _, endpt := range successAPIEndpoints {
+			for _, port := range successPortNums {
+				for _, path := range successURLPath {
+					if scheme == "" && path == "/path/to/endpoint" {
```
marun: Is an empty scheme and `/` valid?

font: According to `DefaultServerURL` it is. I assume it's because adding `/` still points to the root of the http server so it's interpreted the same as if it's omitted.

font:
@marun I've pushed a new commit to address comments. I've also updated the unit test error messages for `apiEndpoint` to be very specific, where possible, in order to verify the exact error message. PTAL.

marun:
@font LGTM, thank you for the cleanup!

font:
Squashed commits to prepare for merging.

/hold cancel

shashidharatd:
@font, this LGTM and thanks for all the work.
/lgtm

PR #1028, opened 2019-07-03
controller/kubefedcluster: fixes for delete handling and suppress changes due to status updates

pjferrell:
controller/kubefedcluster: add handling for deletes with tombstone and prevent constant processing of cluster changes when status is updated (by caching the last observed spec and comparing on informer updates)

Related issue:
https://github.com/kubernetes-sigs/kubefed/issues/950

### DIFF
--- a/pkg/controller/kubefedcluster/controller.go
+++ b/pkg/controller/kubefedcluster/controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"sync"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -44,6 +45,9 @@ type ClusterData struct {
 
 	// How many times in a row the probe has returned the same result.
 	resultRun int64
+
+	// cachedObj holds the last observer object from apiserver
+	cachedObj *fedv1b1.KubeFedCluster
 }
 
 // ClusterController is responsible for maintaining the health status of each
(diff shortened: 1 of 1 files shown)


pkg/controller/kubefedcluster/controller.go
```diff
@@ -98,43 +102,71 @@ func newClusterController(config *util.ControllerConfig, clusterHealthCheckConfi
 		&fedv1b1.KubeFedCluster{},
 		util.NoResyncPeriod,
 		&cache.ResourceEventHandlerFuncs{
-			DeleteFunc: cc.delFromClusterSet,
-			AddFunc:    cc.addToClusterSet,
+			DeleteFunc: func(obj interface{}) {
+				castObj, ok := obj.(*fedv1b1.KubeFedCluster)
+				if !ok {
```
marun: I appreciate your attention to detail here, but even this level of rigor won't preclude missing deletion events.  Would you be up for removing these changes from the PR in favor of a separate PR that used a finalizer to ensure reliable cleanup on deletion? That change seems orthogonal to eliminating updates in response to status changes.

pkg/controller/kubefedcluster/controller.go
```diff
@@ -98,43 +102,71 @@ func newClusterController(config *util.ControllerConfig, clusterHealthCheckConfi
...
+				cc.addToClusterSet(castObj)
+			},
 			UpdateFunc: func(oldObj, newObj interface{}) {
-				cc.delFromClusterSet(oldObj)
-				cc.addToClusterSet(newObj)
+				var clusterChanged bool
+				cluster := newObj.(*fedv1b1.KubeFedCluster)
+				cc.mu.Lock()
+				clusterData, ok := cc.clusterDataMap[cluster.Name]
+				if ok && !equality.Semantic.DeepEqual(clusterData.cachedObj.Spec, cluster.Spec) {
```
marun: Is the use of `metadata.generation` not sufficient to detect when non-status changes occur?

font: Instead of a cached obj why not compare with `oldObj` passed in?

marun: I'm assuming it's because there is no guarantee that the oldObj that gets passed in is the one that the controller last reconciled.

marun:
@pjferrell Thank you, this PR represents a nice improvement in the reliability of the cluster controller.

/approve

PR #1030, opened 2019-07-03
Kubefedcluster: support ignore tls

pjferrell:
Implements an optional field in v1beta1/KubeFedCluster to specify which TLS checks to ignore. Disabling TLS checks is implemented by setting tls.Config.InsecureSkipVerify and providing a function closure implementing tls.Config.VerifyPeerCertificate (to selectively ignore SubjectName or ValidityPeriod).

Modifies kubefedctl join to preserve the kubeconfig cluster setting of InsecureSkipTLSVerify by setting the new field in v1beta1/KubeFedCluster.

Related issue:
https://github.com/kubernetes-sigs/kubefed/issues/1029

### DIFF
--- a/pkg/apis/core/v1beta1/kubefedcluster_types.go
+++ b/pkg/apis/core/v1beta1/kubefedcluster_types.go
@@ -23,6 +23,14 @@ import (
 	"sigs.k8s.io/kubefed/pkg/apis/core/common"
 )
 
+type TLSValidation string
+
+const (
+	TLSAll            TLSValidation = "*"
+	TLSSubjectName    TLSValidation = "SubjectName"
+	TLSValidityPeriod TLSValidation = "ValidityPeriod"
+)
+
 // KubeFedClusterSpec defines the desired state of KubeFedCluster
 type KubeFedClusterSpec struct {
 	// The API endpoint of the member cluster. This can be a hostname,
@@ -37,6 +45,12 @@ type KubeFedClusterSpec struct {
 	// member cluster. The secret needs to exist in the same namespace
 	// as the control plane and should have a "token" key.
 	SecretRef LocalSecretReference `json:"secretRef"`
+
+	// DisabledTLSValidations defines a list of checks to ignore when validating
+	// the TLS connection to the member cluster.  This can be any of *, SubjectName, or ValidityPeriod.
+	// If * is specified, it is expected to be the only option in list.
+	// +optional
+	DisabledTLSValidations []TLSValidation `json:"disabledTLSValidations,omitempty"`
 }
 
 // LocalSecretReference is a reference to a secret within the enclosing
--- a/pkg/apis/core/v1beta1/validation/validation.go
+++ b/pkg/apis/core/v1beta1/validation/validation.go
@@ -162,6 +162,7 @@ func ValidateKubeFedCluster(obj *v1beta1.KubeFedCluster, statusSubResource bool)
 func validateKubeFedClusterSpec(spec *v1beta1.KubeFedClusterSpec, path *field.Path) field.ErrorList {
 	allErrs := validateAPIEndpoint(spec.APIEndpoint, path.Child("apiEndpoint"))
 	allErrs = append(allErrs, validateLocalSecretReference(&spec.SecretRef, path.Child("secretRef"))...)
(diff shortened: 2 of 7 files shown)


marun:
@pjferrell What use case(s) do you have that requires disabling tls?  I can imagine test-only deployments that might want to avoid TLS, but the cost to support it is minimal even then.

pkg/apis/core/v1beta1/kubefedcluster_types.go
```diff
@@ -23,6 +23,14 @@ import (
 	"sigs.k8s.io/kubefed/pkg/apis/core/common"
 )
 
+type TLSValidation string
+
+const (
```
font: Could we add a meaningful prefix to these constants e.g. `TLSAll`, `TLSCheckAll`, etc?

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -268,6 +268,18 @@ func validateGreaterThan0(path *field.Path, value int64) field.ErrorList {
 }
 
 func ValidateKubeFedCluster(object *v1beta1.KubeFedCluster) field.ErrorList {
+	spec := object.Spec
+	specPath := field.NewPath("spec")
 	allErrs := field.ErrorList{}
+
+	disabledTLSValidationsPath := specPath.Child("disabledTLSValidations")
+	for i, value := range spec.DisabledTLSValidations {
+		if i > 0 && value == v1beta1.All {
```
font: Per the comment in the API, this only verifies that `v1beta1.All` is first in the list, but not the only option in the list. Can we also validate that it's the first option and the rest of the list is empty?

pkg/controller/util/cluster_util.go
```diff
@@ -101,3 +115,119 @@ func IsPrimaryCluster(obj, clusterObj pkgruntime.Object) bool {
...
+	if err != nil {
+		return errors.Errorf("Cluster %s transport error: %s", fedCluster.Name, err)
+	}
+
+	err = CustomizeCertificateValidation(fedCluster, transportConfig)
+	if err != nil {
+		return errors.Errorf("Cluster %s custom certificate validation error: %s", fedCluster.Name, err)
+	}
+
+	// using the same defaults as http.DefaultTransport
```
font: Why not just copy `http.DefaultTransport` and then set the specific values you're interested in?

pkg/apis/core/v1beta1/kubefedcluster_types.go
pjferrell: I've updated the names of TLSValidations to start with 'TLS'.

pkg/controller/util/cluster_util.go
pjferrell: Kubernetes go-client creates a new Transport using the same values as DefaultTransport to avoid conflicts between different clients.  I am employing the same pattern since we have to create the transport in case of custom TLS options.

Some backing documentation:
https://github.com/hashicorp/go-cleanhttp

> The Go standard library contains a default http.Client called http.DefaultClient. It is a common idiom in Go code to start with http.DefaultClient and tweak it as necessary, and in fact, this is encouraged; from the http package documentation:
> 
> > The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.
> 
> 
> Unfortunately, this is a shared value, and it is not uncommon for libraries to assume that they are free to modify it at will. With enough dependencies, it can be very easy to encounter strange problems and race conditions due to manipulation of this shared value across libraries and goroutines (clients are safe for concurrent use, but writing values to the client struct itself is not protected).
> 
> Making things worse is the fact that a bare http.Client will use a default http.Transport called http.DefaultTransport, which is another global value that behaves the same way. So it is not simply enough to replace http.DefaultClient with &http.Client{}.

pkg/apis/core/v1beta1/validation/validation.go
pjferrell: Updated the check to return err if v1beta1.TSLAll is specified in a list with more than one object.

pjferrell:
@font Rebased and test added for validation.  Thanks.

PR #1038, opened 2019-07-10
Remove generation of the template validation schema

shashidharatd:
This is one of the step to handle federation api upgrade scenario. The openapi schema validation for templates in federated API is removed.

#Ref: https://docs.google.com/document/d/1uAOMzDgczEYwKcBCuSLpdhuQPHHvRYG-w3ke9659qY0/edit#

/cc @marun


pkg/kubefedctl/enable/enable.go
```diff
@@ -358,8 +358,7 @@ func qualifiedAPIResourceName(resource metav1.APIResource) string {
 }
 
 func federatedTypeCRD(typeConfig typeconfig.Interface, accessor schemaAccessor, shortNames []string) *apiextv1b1.CustomResourceDefinition {
-	templateSchema := accessor.templateSchema()
-	schema := federatedTypeValidationSchema(templateSchema)
+	schema := federatedTypeValidationSchema(nil)
```
marun: It's not a good sign that this PR is passing CI, since `templateSchema` should still be provided for scalable resources to ensure that `spec.retainReplicas` is added to the validation schema. It's my fault, though, since I neglected to write a test for that case.

I think the easiest path forward is to still compute `templateSchema` and instead update  `federatedTypeValidationSchema` to avoid setting `Properties` for `spec.template` (i.e. not set the validation schema past the field being of type `object`). It probably makes sense to add a TODO to remind us to find a simpler way to detect that a resource is scalable than having to compute the entire schema.

shashidharatd: @marun, have worked on the comment. PTAL, Thanks

marun:
@shashidharatd This LGTM. Please update the changelog so that the change can be more easily reflected in the notes of the next release.

shashidharatd:
@marun, Done. PTAL.

marun:
Thank you!

/lgtm

PR #1039, opened 2019-07-11
Fix build on other Linux architectures

yselkowitz:


Makefile
```diff
@@ -80,21 +81,21 @@ vet:
...
+	cp -f $(HYPERFED_TARGET)-linux-$(HOST_ARCH) images/kubefed/hyperfed
 	$(DOCKER) build images/kubefed -t $(IMAGE_NAME)
 	rm -f images/kubefed/hyperfed
 
 bindir:
 	mkdir -p $(BIN_DIR)
 
 COMMANDS := $(HYPERFED_TARGET) $(CONTROLLER_TARGET) $(KUBEFEDCTL_TARGET) $(WEBHOOK_TARGET)
-OSES := linux darwin
+OSES := linux-amd64 linux-arm64 linux-ppc64le linux-s390x darwin-amd64
```
font: These aren't really just OSes anymore, they are OS + hardware architecture.

yselkowitz: It will make for a larger patch, but I can rename the variable if you prefer.

font: Since you're modifying the variable's values already, I think it's okay to make the change as the variable is only used on line 101. Thanks!

poothia:
lgtm

shashidharatd:
/lgtm
@yselkowitz, thanks for your contribution.

PR #1040, opened 2019-07-16
Generalize validation e2e tests for each validating resource type

font:
~~**NOTE:** This PR is not reviewable until #1027 is merged.~~

### DIFF
--- a/test/e2e/validation.go
+++ b/test/e2e/validation.go
@@ -37,14 +37,11 @@ import (
 var _ = Describe("Core API Validation", func() {
 	testBaseName := "core-api-validation"
 	f := framework.NewKubeFedFramework(testBaseName)
-	var namespace string
 
 	BeforeEach(func() {
 		if framework.TestContext.InMemoryControllers {
 			framework.Skipf("Running validation admission webhook outside of cluster not supported")
 		}
-
-		namespace = f.TestNamespaceName()
 	})
 
 	resourcesToValidate := []string{ftc.ResourceName, kfcluster.ResourceName}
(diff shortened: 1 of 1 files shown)


font:
Rebased and ready for review.

test/e2e/validation.go
```diff
@@ -22,199 +22,211 @@ import (
...
-		It(fmt.Sprintf("should succeed when an invalid %s is created outside the kubefed system namespace", resourceName), func() {
-			if !framework.TestContext.LimitedScope {
-				framework.Skipf("Cannot run validation admission webhook namespaced test in a cluster scoped deployment")
+	resourcesToValidate := []string{ftc.ResourceName, kfcluster.ResourceName}
+	for i := range resourcesToValidate {
+		resourceName := resourcesToValidate[i]
+		vrt := newValidationResourceTest(resourceName)
+		vrt.initialize()
+		It(fmt.Sprintf("for %s should fail when an invalid %s is created or updated", resourceName, resourceName), func() {
+			// This test also implicitly tests the successful creation of a valid
```
marun: Is this comment accurate?  The block starting on L69 appears to be an explicit check for successful creation of a valid resource.

(No action required) It may be worth minimizing the indentation of this test code by extracting to a separate function.

test/e2e/validation.go
```diff
@@ -22,199 +22,211 @@ import (
...
-				framework.DeleteNamespace(kubeClient, namespace)
-			})
 
-			By(fmt.Sprintf("Creating an invalid %s in the separate test namespace %s", resourceName, namespace))
-			invalidKFC.Namespace = namespace
-			invalidKFC.Spec.APIEndpoint = "https://"
-			err := client.Create(context.TODO(), invalidKFC)
-			if err != nil {
-				f.Logger().Fatalf("Unexpected error creating invalid %s = %+v in another test namespace %s, err: %v", resourceName, invalidKFC, namespace, err)
+			// Immediately delete the created test resource to avoid errors in
```
marun: I'm not sure this is a sufficient guarantee of test isolation.  How about running the tests only for in-process controllers in cluster-scoped mode to allow for resource creation to be isolated to a test namespace? The webhook would still be running. A check for the webhook being correctly configured for a namespaced control plane could consist of dry-run attempts to create valid and invalid resources to avoid impacting controllers targeting the system namespace.

test/e2e/validation.go
```diff
@@ -22,199 +22,211 @@ import (
...
+	}
+	return vrt
+}
+
+func (ftc *ftcValidationTest) initialize() pkgruntime.Object {
+	if ftc.object != nil {
+		return ftc.object
+	}
+
+	// For the target API type, use an existing K8s API resource that
```
marun: Could this be simplified to using a type enabled by default if tests were always run in a test namespace or with dry-run?

test/e2e/validation.go
```diff
@@ -22,199 +22,211 @@ import (
...
+	return ftc.object.DeepCopyObject()
+}
+
+func (ftc *ftcValidationTest) invalidObjectFromValid(obj pkgruntime.Object) pkgruntime.Object {
+	invalidFtc := obj.DeepCopyObject().(*v1beta1.FederatedTypeConfig)
+	ftc.setInvalidField(invalidFtc)
+	return invalidFtc
+}
+
+func (kfc *kfclusterValidationTest) initialize() pkgruntime.Object {
```
marun: nit: kfCluster?

test/e2e/validation.go
font: The comment was to clarify the `It` description in that this `It` will also test creation as a result of having to test if an invalid update fails. I've removed since there's a `By` that describes the valid creation.

test/e2e/validation.go
font: I think I understand the concern, but I'm not sure I understand the proposed solution alternatives.

For the concern: since we run tests serially, there's probably not too much harm in creating temporary resources that could impact the control plane only during the execution of this test, since the resources are immediately deleted before proceeding. Are you thinking about parallelization?

For the proposed solutions: cluster-scoped mode will run in a separate test namespace regardless of in-process controllers. So I think we're okay there. When running against a namespaced control plane, we could attempt to create a separate test namespace just for this test once a singleton webhook becomes a thing. Until then, how would we perform dry-run attempts to verify validation of an update of already created resources? Or are you thinking to forego the update validation check?

test/e2e/validation.go
font: Yes it could, but the namespace-scoped deployment is the only issue. I'm not sure about the dry-run solution for updates per my comment above.

test/e2e/validation.go
font: That's what I originally had and then changed it since it's referred to `kfcluster` in other places such as the import alias. :smile:

font:
@marun I've responded to comments and pushed a new update reflecting feedback. PTAL.

test/e2e/validation.go
marun: I am not concerned with running tests in parallel.  I think it would be preferable to avoid creating resources that will impact controller state. As you say, this isn't a problem for a cluster-scoped control plane since a test namespace can be used. For a namespace-scoped control plane, I think a preferable strategy to the current approach would be to use dry-run validation. In the case of an update validation check, it should be possible to target the FTC of a type enabled by default (e.g. `deployments.app`).

font: I think that's doable and a good approach for a future follow-up.

marun:
LGTM

shashidharatd:
/lgtm
Thanks @font

PR #1042, opened 2019-07-17
event for kubefedcluster

### LINKED ISSUE #809: Cluster controller errors should be logged as events on clusters where possible
/kind feature

xunpan:
fixes #809

### DIFF
--- a/pkg/controller/kubefedcluster/controller.go
+++ b/pkg/controller/kubefedcluster/controller.go
@@ -18,19 +18,27 @@ package kubefedcluster
 
 import (
 	"context"
+	"fmt"
 	"sync"
 
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	kubeclient "k8s.io/client-go/kubernetes"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
 
 	fedv1b1 "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 	genericclient "sigs.k8s.io/kubefed/pkg/client/generic"
+	genscheme "sigs.k8s.io/kubefed/pkg/client/generic/scheme"
 	"sigs.k8s.io/kubefed/pkg/controller/util"
 	"sigs.k8s.io/kubefed/pkg/features"
 )
@@ -70,6 +78,8 @@ type ClusterController struct {
 	// fedNamespace is the name of the namespace containing
 	// KubeFedCluster resources and their associated secrets.
 	fedNamespace string
+
+	eventRecorder record.EventRecorder
 }
 
 // StartClusterController starts a new cluster controller.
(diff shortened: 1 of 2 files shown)


pkg/controller/kubefedcluster/controller.go
```diff
@@ -159,11 +176,14 @@ func (cc *ClusterController) addToClusterSet(obj *fedv1b1.KubeFedCluster) {
 	if clusterData != nil && clusterData.clusterKubeClient != nil {
 		return
 	}
+
 	klog.V(1).Infof("ClusterController observed a new cluster: %v", obj.Name)
+	cc.RecordEvent(obj, "CreatingClient", "Creating a client of the cluster")
```
marun: Is this condition worthy of a user-visible event?

pkg/controller/kubefedcluster/controller.go
```diff
@@ -159,11 +176,14 @@ func (cc *ClusterController) addToClusterSet(obj *fedv1b1.KubeFedCluster) {
...
+
 	// create the restclient of cluster
 	restClient, err := NewClusterClientSet(obj, cc.client, cc.fedNamespace, cc.clusterHealthCheckConfig.Timeout)
 	if err != nil || restClient == nil {
-		klog.Errorf("Failed to create corresponding restclient of kubernetes cluster: %v", err)
+		cc.RecordError(obj, "CreatingClientFailed", errors.Wrap(err, "Failed to create corresponding client of the cluster"))
```
marun: Is this worthy of a user-visible event?

xunpan:
@marun 
code updated.

For your comment `log the specific error to aid the user in resolving the problem`

When healthz failed, the existing code handles the log logic by 
```
runtime.HandleError(errors.Wrapf(err, "Failed to do cluster health check for cluster %q", self.clusterName))   
```
The default behavior is log the error from the code comments.

My question is, what is the rule to use `klog` and `runtime.HandleError()` in our code?

pkg/controller/kubefedcluster/controller.go
```diff
@@ -159,11 +176,13 @@ func (cc *ClusterController) addToClusterSet(obj *fedv1b1.KubeFedCluster) {
...
 		return
 	}
+
 	klog.V(1).Infof("ClusterController observed a new cluster: %v", obj.Name)
+
 	// create the restclient of cluster
 	restClient, err := NewClusterClientSet(obj, cc.client, cc.fedNamespace, cc.clusterHealthCheckConfig.Timeout)
 	if err != nil || restClient == nil {
-		klog.Errorf("Failed to create corresponding restclient of kubernetes cluster: %v", err)
+		cc.RecordError(obj, "MalformedClusterConfig", errors.Wrap(err, "Malformed configuration of the cluster"))
```
marun: -> `The configuration for this cluster may be malformed`

pkg/controller/kubefedcluster/controller.go
```diff
@@ -219,21 +238,35 @@ func (cc *ClusterController) updateIndividualClusterStatus(cluster *fedv1b1.Kube
 	storedData *ClusterData, wg *sync.WaitGroup) {
 	clusterClient := storedData.clusterKubeClient
 
-	currentClusterStatus := clusterClient.GetClusterHealthStatus()
+	cc.RecordEvent(cluster, "UpdatingStatus", "Updating the status of the cluster")
```
marun: I don't think this is worth recording as an event.

pkg/controller/kubefedcluster/controller.go
```diff
@@ -219,21 +238,35 @@ func (cc *ClusterController) updateIndividualClusterStatus(cluster *fedv1b1.Kube
...
 	if utilfeature.DefaultFeatureGate.Enabled(features.CrossClusterServiceDiscovery) {
-		currentClusterStatus = updateClusterZonesAndRegion(currentClusterStatus, cluster, clusterClient)
+		currentClusterStatus = cc.updateClusterZonesAndRegion(currentClusterStatus, cluster, clusterClient)
 	}
 
 	storedData.clusterStatus = currentClusterStatus
 	cluster.Status = *currentClusterStatus
 	if err := cc.client.UpdateStatus(context.TODO(), cluster); err != nil {
-		klog.Warningf("Failed to update the status of cluster %q: %v", cluster.Name, err)
+		cc.RecordError(cluster, "UpdatingStatusFailed", errors.Wrap(err, "Failed to update the status of the cluster"))
```
marun: I'm not sure I understand how this event would be useful to a user.  What error conditions would it indicate, and how would the event allow them to identify or fix the problem?

xunpan: When status update failed, I think the event can tell there has an issue. It tells that the status is not updated correctly. I'm not sure if `err` information helps for fix the problem but it tells the problem exists.

marun: Are there other kube resources for which failure to report status is recorded as an event? I can think of at least 2 reasons for which an event for failure to update status should not be logged:

 - concurrency (i.e. an update was made to the resource by another party) since this is recoverable
 - transitory connectivity problem (since this is likely recoverable as well)

Consider that in both cases the error is likely to be transitory and automatically recovered.  Logging yes, but an event?  It will just clutter the log.

xunpan: agreed. Updated

marun:
Thank you!

/lgtm

PR #1047, opened 2019-07-19
Cleanup generic status handling

marun:
- remove unused type
 - Ensure last transition time matches last probe time


marun:
I've reverted the pointer change. It's an internal detail and its simpler to use pointers when modifying a slice of conditions.

marun:
Rebased

shashidharatd:
LGTM

PR #1048, opened 2019-07-19
Fix typo in the usage message of controller manager

hainesc:
Fix typo in the usage message of controller manager. The word `which` here means `a bunch of controllers`, so we should not use the third person singular.

watches -> watch
does -> do

Thanks.


shashidharatd:
/lgtm
We may find too many such instances. Is there any way this check can be automated and leave this spelling and grammar check to machines, instead of humans spending time on it?

hainesc:
> We may find too many such instances. Is there any way this check can be automated and leave this spelling and grammar check to machines, instead of humans spending time on it?

Yes, machine can help us, there are some tools can do spelling and grammar check. I make this pr when I was reading the source code.

PR #1049, opened 2019-07-19
Update KubeFedCluster API optional fields

font:
This updates the optional fields in `KubeFedClusterStatus` to make them pointers. It also makes `KubeFedClusterSpec` required as it was inadvertently changed previously.

Additionally, a couple more cluster conditions are added to the shared `ValidKubeFedCluster` test object (to complete all 4 possible conditions) as a result of changes for this PR.


shashidharatd:
this LGTM too..
Thanks @font

PR #1051, opened 2019-07-20
Update release script and doc

font:
Update the release doc to verify `helm` is installed prior to executing the release script. Also updates the release script for:
- new Makefile build targets after merging of #1039
- adds a check for `helm` being installed.


shashidharatd:
LGTM

PR #1052, opened 2019-07-23
Support varying the apiVersion of target resources

marun:
Previously the sync controller used `ResourceClient` to make changes to member clusters. This meant that the API endpoint configuration for a target type was always sourced from an FTC.

This PR updates the sync controller to use `GenericClient` so that API endpoint configuration for create and update calls can be provided on a per-call basis. If `apiVersion` is set via template or override for a given federated resource, that value will supersede the configuration sourced from an FTC. This is intended to allow a given federated type to manage more than one version across member clusters. For example, federated resource `foo` could manage a resource in cluster1 at v1, and a resource in cluster2 at v2. It would still be necessary for all clusters to support the version defined in the FTC since that is what would configure the informers for member clusters.

### DIFF
--- a/pkg/controller/sync/dispatch/unmanaged.go
+++ b/pkg/controller/sync/dispatch/unmanaged.go
@@ -17,14 +17,17 @@ limitations under the License.
 package dispatch
 
 import (
+	"context"
+
 	"github.com/pkg/errors"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/klog"
 
+	"sigs.k8s.io/kubefed/pkg/client/generic"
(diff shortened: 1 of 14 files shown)


marun:
Rebased

marun:
Rebased

shashidharatd:
Awesome work. This LGTM. Thanks @marun!

pkg/controller/sync/dispatch/unmanaged.go
```diff
@@ -111,7 +116,7 @@ func (d *unmanagedDispatcherImpl) RemoveManagedLabel(clusterName string, cluster
 
 		util.RemoveManagedLabel(updateObj)
 
-		_, err := client.Resources(updateObj.GetNamespace()).Update(updateObj, metav1.UpdateOptions{})
+		err := client.Update(context.Background(), updateObj)
```
font: Do we not need to set GVK here because we're dealing with a cached informer object that should already have it set?

marun: gvk is only required for operations where an object isn't sourced from the informer or derived from a federated resource (i.e. get and delete). Otherwise apiVersion will be populated (as evidenced by ci running without error).

pmorie: I believe that is correct, based on my reading.

PR #1057, opened 2019-07-26
Update to use Kind v0.4.0 and remove a todo

shashidharatd:
Handle a todo item.

Fixes #
https://github.com/kubernetes-sigs/kind/issues/111

/cc @font 
**Release note**:

```release-note

```

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -141,18 +141,13 @@ function fixup-cluster() {
   local kubeconfig_path="$(kind get kubeconfig-path --name cluster${i})"
   export KUBECONFIG="${KUBECONFIG:-}:${kubeconfig_path}"
 
+  if [ "$OS" != "Darwin" ];then
+    # Set container IP address as kube API endpoint in order for clusters to reach kube API servers in other clusters.
+    kind get kubeconfig --name "cluster${i}" --internal >${kubeconfig_path}
+  fi
+
   # Simplify context name
   kubectl config rename-context "kubernetes-admin@cluster${i}" "cluster${i}"
-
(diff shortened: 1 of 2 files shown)


scripts/create-clusters.sh
```diff
@@ -139,21 +139,14 @@ function fixup-cluster() {
   local i=${1} # cluster num
 
   local kubeconfig_path="$(kind get kubeconfig-path --name cluster${i})"
+  # Set container IP address as kube API endpoint in order for clusters to reach kube API servers in other clusters.
+  kind get kubeconfig --name "cluster${i}" --internal >${kubeconfig_path}
```
font: Will this work for MacOS? The `--internal` flag according to the PR will use the internal container IP address. I believe MacOS relies on the use of `localhost`, which is what kind uses by default i.e. without `--internal` flag. However, I cannot verify this.

shashidharatd: @font, Ideally this should be taken care by kind, to implement a feature consistently across platform. Also theoretically the same problem would be present if `localhost` is used in `MacOS`. I too am not familiar with `MacOS`. Can somebody using `MacOS` confirm this please. @xunpan @aoxn @iamzhout can you please help in verifying this issue.

shashidharatd: Thanks @font for the clarification. I have updated the code. PTAL once again. Thanks!

marun:
@xunpan kind on macos is a [documented special-case](https://github.com/kubernetes-sigs/kubefed/blob/master/docs/cluster-registration.md#joining-kind-clusters-on-macos)

PR #1063, opened 2019-07-30
Add Error Event for FederatedNamespace

### LINKED ISSUE #1037: An error event should be logged when the name of a FederatedNamespace doesn't match its namespace
A `FederatedNamespace` needs to have the same name as its containing namespace in order to be propagated. Currently if the name differs there is no indication of the problem as per #1036.  

/kind bug

qiujian16:
Add an error event when name and namespace of federatednamespace
are not same.

Fixes #1037

**Release note**:

```release-note

```

### DIFF
--- a/pkg/controller/sync/accessor.go
+++ b/pkg/controller/sync/accessor.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
 
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
 	genericclient "sigs.k8s.io/kubefed/pkg/client/generic"
 	"sigs.k8s.io/kubefed/pkg/controller/sync/version"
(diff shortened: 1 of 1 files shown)


pkg/controller/sync/accessor.go
```diff
@@ -240,6 +236,13 @@ func (a *resourceAccessor) FederatedResource(eventSource util.QualifiedName) (Fe
 
 	var namespace *unstructured.Unstructured
 	if a.targetIsNamespace {
+		if federatedName.Namespace != federatedName.Name {
+			// A FederatedNamespace is only valid for propagation
+			// if it has the same name as the containing namespace.
+			a.eventRecorder.Eventf(
+				resource, corev1.EventTypeWarning, "InConsistentNameNamespace", "Name and namespace are not consistent")
```
marun: `s/InConistentNameNamespace/InvalidName/`

`s/Name and namespace are not consistent/The name of a federated namespace must match the name of its containing namespace./`

marun: @font Maybe federated resource validation, in addition to dry-run, could have special handling for federated namespaces to prevent creation when the name doesn't match that of the containing namespace?

qiujian16: done, thanks for the comment.

pkg/controller/sync/accessor.go
```diff
@@ -240,6 +236,14 @@ func (a *resourceAccessor) FederatedResource(eventSource util.QualifiedName) (Fe
 
 	var namespace *unstructured.Unstructured
 	if a.targetIsNamespace {
+		if federatedName.Namespace != federatedName.Name {
+			// A FederatedNamespace is only valid for propagation
+			// if it has the same name as the containing namespace.
+			a.eventRecorder.Eventf(
+				resource, corev1.EventTypeWarning, "InvalidName", ""+
```
marun: Is it necessary to include `""+`?

qiujian16: right...fixed

marun:
Thank you!

/approve
/lgtm

PR #1070, opened 2019-07-31
Fix panic when statusCollectoin is enabled

qiujian16:
Set default value for statusType

When statusType is not set, return targetType.

Fixes https://github.com/kubernetes-sigs/kubefed/issues/1069

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/core/v1beta1/federatedtypeconfig_types.go
+++ b/pkg/apis/core/v1beta1/federatedtypeconfig_types.go
@@ -219,7 +219,9 @@ func (f *FederatedTypeConfig) GetStatusType() *metav1.APIResource {
 }
 
 func (f *FederatedTypeConfig) GetStatusEnabled() bool {
-	return f.Spec.StatusCollection != nil && *f.Spec.StatusCollection == StatusCollectionEnabled
+	return f.Spec.StatusCollection != nil &&
+		*f.Spec.StatusCollection == StatusCollectionEnabled &&
+		f.Name == "services"
 }
 
 // TODO(font): This method should be removed from the interface i.e. remove
--- a/pkg/controller/status/controller.go
+++ b/pkg/controller/status/controller.go
@@ -95,6 +95,9 @@ func StartKubeFedStatusController(controllerConfig *util.ControllerConfig, stopC
 func newKubeFedStatusController(controllerConfig *util.ControllerConfig, typeConfig typeconfig.Interface) (*KubeFedStatusController, error) {
 	federatedAPIResource := typeConfig.GetFederatedType()
 	statusAPIResource := typeConfig.GetStatusType()
+	if statusAPIResource == nil {
+		return nil, errors.Errorf("Status collection is not supported for %q", federatedAPIResource.Kind)
+	}
 	userAgent := fmt.Sprintf("%s-controller", strings.ToLower(statusAPIResource.Kind))
 	client := genericclient.NewForConfigOrDieWithUserAgent(controllerConfig.KubeConfig, userAgent)
 


qiujian16:
ah, thanks, what i want to do is to see the status of federateddeployment. Its targetType is deployment and it seems to work...the status of federatetypeconfig shows shows that statusCollection is running and `kubectl describe federateddeployment` shows that
```
Cluster Status:
  Cluster Name:  cluster1
  Status:
    Available Replicas:  1
    Conditions:
      Last Transition Time:  2019-07-31T05:23:27Z
      Last Update Time:      2019-07-31T05:23:27Z
      Message:               Deployment has minimum availability.
      Reason:                MinimumReplicasAvailable
      Status:                True
      Type:                  Available
      Last Transition Time:  2019-07-31T05:23:18Z
      Last Update Time:      2019-07-31T05:23:27Z
      Message:               ReplicaSet "test-deployment-554b9c67f9" has successfully progressed.
      Reason:                NewReplicaSetAvailable
      Status:                True
      Type:                  Progressing
    Observed Generation:     1
    Ready Replicas:          1
    Replicas:                1
    Updated Replicas:        1
```

If it is only for services, i agree that it should just return false an ignore other types

marun:
@qiujian16 The fact that the collected status is being written to the federated deployment is likely a result of the CRD infrastructure in < Kube 1.15 not stripping unknown fields (since those fields do not appear in the validation schema).  I believe this behavior will be changing in 1.15 to default to stripping unknown fields, in which case the behavior you are seeing would not exhibit.

pkg/controller/status/controller.go
```diff
@@ -95,6 +95,9 @@ func StartKubeFedStatusController(controllerConfig *util.ControllerConfig, stopC
 func newKubeFedStatusController(controllerConfig *util.ControllerConfig, typeConfig typeconfig.Interface) (*KubeFedStatusController, error) {
 	federatedAPIResource := typeConfig.GetFederatedType()
 	statusAPIResource := typeConfig.GetStatusType()
+	if statusAPIResource == nil {
+		return nil, fmt.Errorf("spec.statusType is nil")
```
marun: -> `"Status collection is not supported for %q", federatedAPIResource.Kind`

pkg/apis/core/v1beta1/federatedtypeconfig_types.go
```diff
@@ -219,7 +219,9 @@ func (f *FederatedTypeConfig) GetStatusType() *metav1.APIResource {
 }
 
 func (f *FederatedTypeConfig) GetStatusEnabled() bool {
-	return f.Spec.StatusCollection != nil && *f.Spec.StatusCollection == StatusCollectionEnabled
+	return f.Spec.StatusCollection != nil &&
+		*f.Spec.StatusCollection == StatusCollectionEnabled &&
+		f.Spec.FederatedType.Kind == "FederatedService"
```
marun: The name of the federated type is not necessarily  `FederatedService`.  Please check instead that name of the FTC is `services`.

pkg/controller/status/controller.go
marun: Also, as per the CI failure, please prefer use of the [errors](https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package) package to ensure traceability (e.g. `errors.Errorf`).

qiujian16: done, thanks!

pkg/apis/core/v1beta1/federatedtypeconfig_types.go
qiujian16: done

marun:
Thank you!

/approve
/lgtm

PR #1074, opened 2019-08-01
Update github pr template to reflect our use of CHANGELOG.md

marun:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,7 +1,7 @@
 <!--  Thanks for sending a pull request!  Here are some tips for you:
 1. If this is your first time, read our contributor guidelines https://git.k8s.io/community/contributors/guide#your-first-contribution and developer guide https://git.k8s.io/community/contributors/devel/development.md#development-guide
 2. If you want *faster* PR reviews, read how: https://git.k8s.io/community/contributors/guide/pull-requests.md#best-practices-for-faster-reviews
-3. Follow the instructions for writing a release note: https://git.k8s.io/community/contributors/guide/release-notes.md
+3. Add an entry to CHANGELOG.md if the PR represents a user-visible change.
```
gyliu513: @marun instead of adding entry to CHANEGLOG manually, how about generate the CHANGELOG automatically based on https://github.com/github-changelog-generator/github-changelog-generator ?

marun: Given that the changelog generator only pulls the titles of issues and pull requests, I think I'd rather continue adding entries manually.  There is frequently a difference between the title (for developers) and the user-visible impact that ends up in the changelog.

gyliu513: As far as I can see, for a PR, we should have both title and commit message, and the title should be easily to be understand by both users and developers, while the commit message can add more detail from developer point of view. With this assumption, we can still use https://github.com/github-changelog-generator/github-changelog-generator . But I have no comments if all kubefed developers like the way of adding entries manually.

gyliu513: As far as I can see, for a PR, we should have both title and commit message, and the title should be easily to be understand by both users and developers, while the commit message can add more detail from developer point of view. With this assumption, we can still use https://github.com/github-changelog-generator/github-changelog-generator . But I have no comments if all kubefed developers like the way of adding entries manually.

marun: The main point of the changelog for the project is in support of coherent release notes. The title is often not sufficient for the purposes of a release, where it is important to detail the impact of a change. I don't see what the changelog generator buys us, then, given that we could as well look at the commit log since the last release.

PR #1065, opened 2019-07-30
Get curretContext when host-cluster-context is empty

qiujian16:
**What this PR does / why we need it**:

Get current context in kubeconfig when host-cluster-context
is not set during kubefedctl join

**Which issue(s) this PR fixes**

Fixes https://github.com/kubernetes-sigs/kubefed/issues/1064

**Special notes for your reviewer**:

**Release note**:

```release-note

```

PR #1076, opened 2019-08-01
kubefedctl: Set host cluster context to current context if empty

marun:
For any invocation of `kubefedctl` that does not specify a value for `--host-cluster-context`, the current config context will be
used. However, other usage of the `HostClusterContext` variable requires explicitly looking up the name of the current context to avoid using an empty string.

#1065 fixed the `join` command to lookup the current context. This change ensures the same lookup is performed for all `kubefedctl` commands.

### DIFF
--- a/pkg/kubefedctl/options/options.go
+++ b/pkg/kubefedctl/options/options.go
@@ -25,6 +25,7 @@ import (
 	apiextv1b1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 
 	fedv1b1 "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 	genericclient "sigs.k8s.io/kubefed/pkg/client/generic"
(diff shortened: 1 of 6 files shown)


marun:
cc:  @qiujian16

pkg/kubefedctl/util/util.go
```diff
@@ -128,10 +128,16 @@ func IsFederatedAPIResource(kind, group string) bool {
...
+	currentContext, err := options.CurrentContext(clientConfig)
+	if err != nil {
+		return "", err
+	}
+
+	ns, _, err := clientConfig.Namespace()
 	if err != nil {
 		return "", errors.Wrapf(err, "Failed to get ClientConfig for host cluster context %q and kubeconfig %q",
-			hostClusterContext, kubeconfig)
+			currentContext, kubeconfig)
```
xunpan: This seems only for message correctness. If that's true, do we plan to fix in `orphaning.go` https://github.com/kubernetes-sigs/kubefed/blob/0eeda018624acdbf6fc28ef0b67d2e3e8b6e614b/pkg/kubefedctl/orphaning/orphaning.go#L105

pkg/kubefedctl/options/options.go
```diff
@@ -50,6 +51,30 @@ func (o *GlobalSubcommandOptions) GlobalSubcommandBind(flags *pflag.FlagSet) {
...
+	currentContext, err := CurrentContext(config)
+	if err != nil {
+		return err
+	}
+	o.HostClusterContext = currentContext
+	return nil
+}
+
+// CurrentContext retrieves the current context from the provided
+// config.
```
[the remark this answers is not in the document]
marun: Not really, no.  The default in emacs is quite low and sometimes I forget to change it.

pkg/kubefedctl/util/util.go
marun: Done

marun:
Rebased

marun:
Rebased

PR #1081, opened 2019-08-05
Add scripts to help automate KubeFed releases

### LINKED ISSUE #813: Automate the release of new versions

font:
This PR adds a couple of scripts for building and verifying a KubeFed release; creating a PR with the release changes and creating a GitHub draft release. It also updates the release instructions for executing them. This helps further automate the releases of KubeFed.

Fixes #813 

Commits can be reviewed individually.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -7,7 +7,48 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-### Releasing KubeFed
+## Releasing KubeFed
+
+### Automated
+
+Creating a KubeFed release can be automated by running a couple of scripts.
+
+##### Prerequisites
+
+The scripts used to automated the release of KubeFed rely on the following
+prereqs:
+
+- [hub CLI](https://github.com/github/hub#installation)
+    - Once installed verify you can run `hub release show -f "%n" v0.1.0-rc5`
+      without prompting for GitHub authentication.
+- [jq](https://stedolan.github.io/jq/)
+- helm. See
+  https://github.com/kubernetes-sigs/kubefed/blob/master/docs/development.md#binaries
+  for details.
+- git remote naming conventions used by GitHub's `hub` CLI tool:
+    - `origin` GitHub remote name for KubeFed fork. This can be overriden with
+      `GITHUB_REMOTE_FORK_NAME=<remote>`.
+    - `upstream` GitHub remote name for upstream KubeFed repo. This can be
+      overriden with `GITHUB_REMOTE_UPSTREAM_NAME=<remote>`.
+
+##### Create Release
+
+1. `./scripts/build-release.sh <RELEASE_TAG>`
+
+    - This step builds the release artifacts, creates the annotated
+      `RELEASE_TAG` and pushes it to master at `GITHUB_REMOTE_UPSTREAM_NAME`.
+      It then verifies the Travis build starts and completes successfully
+      followed by verification that the container image built is successfully
+      pushed to Quay.
+
+2. Edit the `CHANGELOG.md` for correct wording and leave it uncommitted. Then proceed to the next
+   step.
+3. `./scripts/create-gh-release.sh <RELEASE_TAG>`
+    - This step creates a pull request with the release `RELEASE_TAG` changes
+      and creates a GitHub draft release. It outputs the URLs for each at the
+      end of the script execution for manual merging and publishing when ready.
+
+### Manual
 
 Creating a KubeFed release involves the following steps:
 
--- a/scripts/build-release-artifacts.sh
+++ b/scripts/build-release-artifacts.sh
@@ -23,9 +23,11 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+source "$(dirname "${BASH_SOURCE}")/util.sh"
+
 RELEASE_TAG="${1-}"
-if [[ ! "${RELEASE_TAG}" ]]; then
-  >&2 echo "usage: $0 <release tag of the form v[0-9]+.[0-9]+.[0-9]+>"
+if [[ ! "${RELEASE_TAG}" =~ ${RELEASE_TAG_REGEX} ]]; then
+  >&2 echo "usage: $0 <release tag of the form v[0-9]+.[0-9]+.[0-9]+(-rc[0-9]+)?>"
   exit 1
 fi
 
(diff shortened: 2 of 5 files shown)


scripts/create-gh-release.sh
```diff
@@ -0,0 +1,122 @@
...
+  # Build asset attach arguments for hub release command
+  local assetArgs
+  for asset in $(cat kubefed-${RELEASE_TAG}-asset-files.txt); do
+    assetArgs+="--attach ${asset} "
+  done
+
+  # Remove trailing whitespace
+  assetArgs="$(echo ${assetArgs})"
+
+  # TODO(font): Add draft and prerelease options to this script.
```
marun: Is this already done?

scripts/create-gh-release.sh
```diff
@@ -0,0 +1,122 @@
...
+  hub release create --draft --prerelease ${assetArgs} -F "${releaseFile}" "${RELEASE_TAG}"
+}
+
+# TODO(font): Consider performing this step BEFORE tagging and pushing a new
+# release so that the release tag contains all the necessary artifacts.
+function create-release-pr() {
+  local commitFiles="${ROOT_DIR}/CHANGELOG.md ${ROOT_DIR}/charts/index.yaml"
+
+  # Use the origin and upstream git remote convention names used by hub.
+  git checkout -b "${RELEASE_TAG}-rel" --no-track "${GITHUB_REMOTE_UPSTREAM_NAME}/master"
```
marun: What is the intent of using `--no-track` and subsequently `--set-upstream`?  I was under the impression that `--set-upstream` effectively causes the branch to track the provided upstream branch.

scripts/build-release.sh
```diff
@@ -0,0 +1,131 @@
...
+    if [[ "${status}" == "${buildStatus}" ]]; then
+      return 0
+    fi
+  done
+
+  return 1
+}
+
+function verify-continuous-integration() {
+  local passedStatus="passed"
```
marun: I find it a bit strange to see this declared as a local but then used in `verify-travis-build-completed`.  Maybe just make it a global given that its a constant anyway?

scripts/build-release.sh
```diff
@@ -0,0 +1,131 @@
...
+util::log "Verifying travis CLI command installed"
+verify-command-installed
+
+util::log "Building release artifacts first to make sure build succeeds"
+build-release-artifacts
+
+util::log "Creating local git annotated tag and pushing tag to kick off build process"
+create-and-push-tag
+
+util::log "Verifing image builds and completes successfully in Travis"
```
marun: Maybe indicate in the log message that this is going to take a really long time (e.g. ~60s)?  Also, what happens if the user accidentally cancels the script?  Could it optionally restart at the verifying step?

scripts/build-release-artifacts.sh
```diff
@@ -24,8 +24,9 @@ set -o nounset
 set -o pipefail
 
 RELEASE_TAG="${1-}"
-if [[ ! "${RELEASE_TAG}" ]]; then
-  >&2 echo "usage: $0 <release tag of the form v[0-9]+.[0-9]+.[0-9]+>"
+RELEASE_TAG_REGEX="^v[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$"
```
marun: Would it make sense for this regex and check be defined once (e.g. in `util.sh`) rather than being duplicated across scripts?

scripts/build-release-artifacts.sh
```diff
@@ -37,8 +38,12 @@ command -v helm >/dev/null 2>&1 ||
 
 ROOT_DIR="$(cd "$(dirname "$0")/.." ; pwd)"
 RELEASE_VERSION="${RELEASE_TAG:1:${#RELEASE_TAG}-1}"
+# This file name is used by the create github release script. If modifying,
+# also update it there as well.
+ASSETS_FILE="kubefed-${RELEASE_TAG}-asset-files.txt"
```
marun: Would it make sense to ensure git ignores these files?   And maybe ignore the release artifacts at the same time (and/or put both in an ignored directory e.g. release/<tag>/*)?

docs/releasing.md
```diff
@@ -7,7 +7,50 @@
...
+### Automated
+
+Creating a KubeFed release can be automated by running a couple of scripts.
+
+##### Prerequisites
+
+The scripts used to automated the release of KubeFed rely on the following
+prereqs:
+
+- [travis CLI](https://github.com/travis-ci/travis.rb#installation)
```
marun: Ideally discovery of the image build status would be accomplished without imposing ruby as a dependency (e.g. via curl). If that's not the case, please ensure the check for the image build status can be optionally disabled. Given that I can check the status of the build trivially via the web interface, I'm not going to install and maintain ruby on my dev machine for the sole purpose of automating this step.

docs/releasing.md
```diff
@@ -7,7 +7,50 @@
...
+    - Once installed verify you can run `travis branches --repo
+      kubernetes-sigs/kubefed` without prompting for interaction.
+- [hub CLI](https://github.com/github/hub#installation)
```
marun: Would it make sense to automate the installation of the hub cli?

scripts/build-release.sh
marun: Er, I meant 60m.

scripts/build-release-artifacts.sh
marun: I never do a fresh clone - I have confidence if my working tree is clean that I'm working from a good state.  What about that scenario is concerning to you?

docs/releasing.md
marun: I think it's preferable to automate the dependency installation rather than documenting it (since automation is effectively docs for a developer). I'm fine with deferring it to a subsequent PR, but I think automating dependency installation should be a near-term priority. Adding it to `download-binaries.sh` would be fine - I don't think it's necessary to create a separate script just for release dependencies.

docs/releasing.md
marun: I do not have it installed, but I definitely prefer installing a standalone binary to something requiring ruby!

marun:
LGTM

font:
Thanks for your review @marun! I acknowledge your suggested enhancements and would like to note that these automation scripts are intended to be a stepping stone to continuing to automate our release process. We can continue to refine and improve them with future iterations as the need arises.

I've squashed commits and am removing the hold.

/hold cancel

marun:
@font Are so many commits necessary?

marun:
LGTM

scripts/build-release-artifacts.sh
font: It's ensuring the working tree is clean without explicitly checking for ignored files and working outside the common `GOPATH`.

PR #1089, opened 2019-08-07
1087. Replica count of federated deployment is not updated by RSP plugin

sancyx:
**What this PR does / why we need it**:
ReplicaSchedulingPreference's plugin doesn't updates FederatedDeployment with latest overrides of replica count.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1087 

**Special notes for your reviewer**:

PR #1092, opened 2019-08-07
Add unit testing for pkg/schedulingtypes/updateOverridesMap

marun:
This PR is a follow-up to #1089 to ensure test coverage of the changed code and capture details about the fix in the changelog.


xunpan:
LGTM

`updateOverridesMap` is a general function name but it handles only `replicas`. It seems not clear enough to understand logic from existing comments to me.

marun:
@xunpan  If the function were exported and available for use outside the package you would have a point.  However, a name only has to be descriptive enough for its context, and I think this is sufficient for an unexported method.

marun:
Rebased

marun:
Rebased

Issue #189, opened 2018-07-26
Update error messages for `join` and `unjoin`


gyliu513:
This is a follow up issue based on comments at https://github.com/kubernetes-sigs/federation-v2/pull/137#discussion_r203076398

The `err` returned by some api call should already include enough information, so there is no need to log extensively.

/cc @marun 

/assign

PR #1093, opened 2019-08-07
 Refactor `kubefedctl join` to enable reuse for simulated scale testing

marun:
Previously, `kubefedctl join` assumed internally that the kubefed namespace in the joining cluster should be the same as the kubefed namespace in the host cluster. This refactor allows the kubefed system namespace in the member cluster to vary in support of simulated scale testing. This is not a user-visible change and is not intended for external consumption.

### DIFF
--- a/pkg/kubefedctl/join.go
+++ b/pkg/kubefedctl/join.go
@@ -92,7 +92,7 @@ type joinFederation struct {
 
 type joinFederationOptions struct {
 	secretName      string
-	Scope           apiextv1b1.ResourceScope
+	scope           apiextv1b1.ResourceScope
 	errorOnExisting bool
 }
 
@@ -174,7 +174,7 @@ func (j *joinFederation) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return err
 	}
 
-	j.Scope, err = options.GetScopeFromKubeFedConfig(hostConfig, j.KubeFedNamespace)
+	j.scope, err = options.GetScopeFromKubeFedConfig(hostConfig, j.KubeFedNamespace)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/kubefedctl/join.go
```diff
@@ -228,13 +242,13 @@ func JoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace,
 	}
 
 	klog.V(2).Infof("Performing preflight checks.")
-	err = performPreflightChecks(clusterClientset, joiningClusterName, hostClusterName, kubefedNamespace, errorOnExisting)
+	err = performPreflightChecks(clusterClientset, joiningClusterName, hostClusterName, joiningNamespace, errorOnExisting)
 	if err != nil {
 		return err
 	}
 
 	klog.V(2).Infof("Creating %s namespace in joining cluster", kubefedNamespace)
```
xunpan: It is trying to create`joiningNamespace` namespace in joining cluster. 
Same comments for all related parts: when we use `joiningNamespace`, we should log `joiningNamespace`.

pkg/kubefedctl/join.go
```diff
@@ -243,20 +257,22 @@ func JoinCluster(hostConfig, clusterConfig *rest.Config, kubefedNamespace,
 	}
 	klog.V(2).Infof("Created %s namespace in joining cluster", kubefedNamespace)
 
-	// Create a service account and use its credentials.
-	klog.V(2).Info("Creating cluster credentials secret")
```
xunpan: any consideration to clean SA `creating` and `created` messages?

pkg/kubefedctl/join.go
marun: Done.

pkg/kubefedctl/join.go
marun: The removed logging was duplicating logging already being done in `createAuthorizedServiceAccount. Logging in join/unjoin remains terribly inconsistent (related: #189).

PR #1094, opened 2019-08-08
Add image build for darwin

hougangliu:
In Makefile, container tag miss to handle darwin OS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,8 +81,8 @@ vet:
 fmt:
 	$(shell ./hack/update-gofmt.sh)
 
-container: $(HYPERFED_TARGET)-linux-$(HOST_ARCH)
-	cp -f $(HYPERFED_TARGET)-linux-$(HOST_ARCH) images/kubefed/hyperfed
+container: $(HYPERFED_TARGET)-$(HOST_PLATFORM)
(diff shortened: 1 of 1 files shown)


marun:
@gyliu513 Why was this desirable?

gyliu513:
@marun the build on Mac OS will be failed without the fix.

marun:
@gyliu513 Why would it ever be desirable to build a container image with a darwin binary? Containers only run on linux, so a container with a darwin binary is effectively broken.  What am I missing?

gyliu513:
@marun what do you mean containers only run on linux? The docker containers below also running on my mac.

```
liugyas-mbp:~ liugya$ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS               NAMES
833ed1833445        golang:1.12.5       "/bin/sh -c 'GOARCH=…"   About a minute ago   Up About a minute                       beautiful_perlman
liugyas-mbp:~ liugya$ uname -s
Darwin
liugyas-mbp:~ liugya$
```

marun:
@gyliu513 Docker containers on mac always have to run in a VM, because containers only run on linux.  The same is true of docker on windows - hyperv is used.

See: https://stackoverflow.com/questions/43383276/how-does-docker-run-a-linux-kernel-under-macos-host

PR #1095, opened 2019-08-09
Revert "Add image build for darwin"

marun:
Reverts kubernetes-sigs/kubefed#1094

There's a reason we don't support building a container with a darwin binary. Containers are linux-only, so a kubefed image with a darwin binary is by definition a broken container.

cc: @gyliu513 @xunpan @hougangliu


xunpan:
/lgtm

This is obvious an rollback, just merge it.

Issue #1084, opened 2019-08-06
Scheduling tests are flaking


marun:
**Which test(s) are failing**:

Scheduling tests with cluster-scoped control plane.

**Since when has it been failing**:

This is the second time I've seen this flake.  The PR that exhibited the failing CI run (#1082) doesn't appear to have any relation to the failure. 

**Reason for failure**:

```
Scheduling ReplicaSchedulingPreferences scheduling for federated deployments.apps 
  should result in replicas spread equally in clusters with no explicit per cluster preferences
  /home/travis/gopath/src/sigs.k8s.io/kubefed/test/e2e/scheduling.go:174
STEP: Creating a namespace to execute the test in
STEP: Created test namespace e2e-tests-scheduling-rnbbk
I0806 15:55:51.157350     892 controller.go:70] Starting scheduling manager
I0806 15:55:51.259704     892 controller.go:179] Starting schedulingpreference controller for ReplicaSchedulingPreference
I0806 15:55:51.594233     892 controller.go:81] Starting replicaschedulingpreferences controller
I0806 15:55:51.596011     892 controller.go:197] Starting plugin FederatedReplicaSet for ReplicaSchedulingPreference
I0806 15:55:51.622927     892 federated_informer.go:205] Cluster kube-federation-system/cluster1 is ready
I0806 15:55:51.632031     892 federated_informer.go:205] Cluster kube-federation-system/cluster2 is ready
I0806 15:55:51.868062     892 federated_informer.go:205] Cluster kube-federation-system/cluster1 is ready
I0806 15:55:51.878249     892 federated_informer.go:205] Cluster kube-federation-system/cluster2 is ready
Aug  6 15:56:51.690: INFO: Failed waiting for matching federated object: timed out waiting for the condition
STEP: Collecting events from namespace "e2e-tests-scheduling-rnbbk".
STEP: Found 0 events.
• Failure [66.092 seconds]
Scheduling
/home/travis/gopath/src/sigs.k8s.io/kubefed/test/e2e/scheduling.go:44
  ReplicaSchedulingPreferences
  /home/travis/gopath/src/sigs.k8s.io/kubefed/test/e2e/scheduling.go:132
    scheduling for federated deployments.apps
    /home/travis/gopath/src/sigs.k8s.io/kubefed/test/e2e/scheduling.go:172
      should result in replicas spread equally in clusters with no explicit per cluster preferences [It]
[...]

PR #1097, opened 2019-08-12
Fix logging statements: duration logging statement should be wrapped …

jhzhu89:
Found some error prone logging statements like below:
https://github.com/kubernetes-sigs/kubefed/blob/01b9e5c1a4f1144c6a63c182b0f760eec30db57e/pkg/controller/ingressdns/controller.go#L210-L212
The 'time.Since(...)' in the defferred statement will be immediately evaluated instead of being evaluated when this statement is executed.

**What this PR does / why we need it:**
Wrapping the logging statement in a function to solve this issue.


jhzhu89:
I signed it

jhzhu89:
retry

marun:
Good catch! Apologies for the ci flakes, we in the process of #1084. 

/approve
/lgtm

PR #1098, opened 2019-08-12
Skip writing propagated version to the API when unchanged

marun:
The existing check for unchanged propagated version status was still resulting in an unnecessary API update.

### DIFF
--- a/pkg/controller/sync/version/manager.go
+++ b/pkg/controller/sync/version/manager.go
@@ -175,8 +175,12 @@ func (m *VersionManager) Update(resource VersionedResource,
 	}
 
 	if oldStatus != nil && util.PropagatedVersionStatusEquivalent(oldStatus, status) {
+		m.Unlock()
 		klog.V(4).Infof("No update necessary for %s %q", m.adapter.TypeName(), qualifiedName)
-	} else if obj == nil {
+		return nil
+	}
+
+	if obj == nil {
 		ownerReference := ownerReferenceForUnstructured(resource.Object())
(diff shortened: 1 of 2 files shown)


pkg/controller/sync/version/manager.go
```diff
@@ -175,8 +175,12 @@ func (m *VersionManager) Update(resource VersionedResource,
 	}
 
 	if oldStatus != nil && util.PropagatedVersionStatusEquivalent(oldStatus, status) {
+		m.Unlock()
```
font: Can we use a defer for this unlock and the one on line 191 or does `writeVersion` expect the version manager to be unlocked?

marun: Holding a lock while making network calls is risky, which is why `writeVersion` is called after unlock.

xunpan: Then what about lock & use & unlock? Scattering unlock for returns seems not good for maintainance.

marun: 2 unlocks - one before exit and one before writing changes - doesn't strike me as 'scattering'.  How is that worse than adding lock/unlock pairs everywhere?

font: Agreed, let's not hold a lock across network calls.

marun: I've added a comment indicating why the lock needs to be released before `writeVersion` is called.

marun:
Rebased

marun:
Rebased

PR #1099, opened 2019-08-13
Minimize updates to propagation status

marun:
Each status update to a federated resource results in a subsequent reconcile. Previously this was creating an endless cycle of reconciliation even when there were no propagation changes since the `lastProbeTime` timestamp was always changed even if nothing else did. This induced unnecessary overhead in the controller manager, kube apiserver, and etcd.

Kubernetes saw a similar problem with `lastProbeTime` on pods: https://github.com/kubernetes/kubernetes/issues/14393

This PR limits propagation status updates to when the status has actually changed - either due to a propagation error or changes being propagated to member clusters. This suggested replacing `lastProbeTime` with `lastUpdateTime`.

### DIFF
--- a/pkg/controller/sync/dispatch/managed.go
+++ b/pkg/controller/sync/dispatch/managed.go
@@ -58,7 +58,7 @@ type ManagedDispatcher interface {
 	Create(clusterName string)
 	Update(clusterName string, clusterObj *unstructured.Unstructured)
 	VersionMap() map[string]string
-	StatusMap() status.PropagationStatusMap
+	CollectedStatus() status.CollectedPropagationStatus
 
 	RecordClusterError(propStatus status.PropagationStatus, clusterName string, err error)
 	RecordStatus(clusterName string, propStatus status.PropagationStatus)
@@ -73,6 +73,10 @@ type managedDispatcherImpl struct {
 	versionMap            map[string]string
 	statusMap             status.PropagationStatusMap
 	skipAdoptingResources bool
+
+	// Track when resource updates are performed to allow indicating
+	// when a change was last propagated to member clusters.
+	resourcesUpdated bool
 }
 
 func NewManagedDispatcher(clientAccessor clientAccessorFunc, fedResource FederatedResourceForDispatch, skipAdoptingResources bool) ManagedDispatcher {
@@ -213,6 +217,7 @@ func (d *managedDispatcherImpl) Update(clusterName string, clusterObj *unstructu
 		if err != nil {
 			return d.recordOperationError(status.UpdateFailed, clusterName, op, err)
 		}
+		d.setResourcesUpdated()
 		version = util.ObjectVersion(obj)
 		d.recordVersion(clusterName, version)
 		return util.StatusAllOK
@@ -276,12 +281,21 @@ func (d *managedDispatcherImpl) recordVersion(clusterName, version string) {
 	d.versionMap[clusterName] = version
 }
 
-func (d *managedDispatcherImpl) StatusMap() status.PropagationStatusMap {
+func (d *managedDispatcherImpl) setResourcesUpdated() {
+	d.Lock()
+	defer d.Unlock()
+	d.resourcesUpdated = true
+}
+
+func (d *managedDispatcherImpl) CollectedStatus() status.CollectedPropagationStatus {
 	d.RLock()
 	defer d.RUnlock()
 	statusMap := make(status.PropagationStatusMap)
 	for key, value := range d.statusMap {
 		statusMap[key] = value
 	}
-	return statusMap
+	return status.CollectedPropagationStatus{
+		StatusMap:        statusMap,
+		ResourcesUpdated: d.resourcesUpdated,
+	}
 }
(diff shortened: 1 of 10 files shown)


marun:
Updated userguide and changelog.

marun:
Hmm, maybe the status should include `observedGeneration`?

pmorie:
I'm +1 on having observedGeneration in status.

marun:
> I'm +1 on having observedGeneration in status.

I'll tackle in a separate PR.

pkg/controller/sync/status/status.go
```diff
@@ -104,51 +105,77 @@ type GenericFederatedStatus struct {
...
+	if reason == AggregateSuccess {
+		for _, value := range collectedStatus.StatusMap {
+			if value != ClusterPropagationOK {
+				reason = CheckClusters
+				break
+			}
+		}
+	}
+
+	clustersChanged := clusterStatusChanged(collectedStatus.StatusMap, s.Clusters)
```
font: Can we somehow indicate that `clustersChanged` is in reference to their propagation status? Maybe consider renaming it to something like `clusterStatusChanged` or similar.

pkg/controller/sync/dispatch/managed.go
```diff
@@ -73,6 +73,10 @@ type managedDispatcherImpl struct {
 	versionMap            map[string]string
 	statusMap             status.PropagationStatusMap
 	skipAdoptingResources bool
+
+	// Track when resources updates are performed to allow indicating
```
font: s/resources/resource/

pkg/controller/sync/dispatch/managed.go
```diff
@@ -276,12 +281,21 @@ func (d *managedDispatcherImpl) recordVersion(clusterName, version string) {
 	d.versionMap[clusterName] = version
 }
 
-func (d *managedDispatcherImpl) StatusMap() status.PropagationStatusMap {
+func (d *managedDispatcherImpl) setResourcesUpdated() {
+	d.Lock()
+	defer d.Unlock()
+	d.resourcesUpdated = true
```
font: This should be cleared otherwise I think we'll continually update status on every reconciliation, when we've already updated resources once and we have at least one status in the collection.

marun: A dispatcher's state is specific to the reconcile of a single federated resource, so it is safe not to clear this.

pkg/controller/sync/status/status.go
marun: The variable needs to have a different name than the function. Given that the usage of this variable is limited to 10 lines, what would you suggest?

pkg/controller/sync/status/status.go
```diff
@@ -104,51 +105,77 @@ type GenericFederatedStatus struct {
...
+	if clustersChanged {
+		s.setClusterStatus(collectedStatus.StatusMap)
+	}
+
+	// Indicate that clusters were updated if the cluster status has
+	// changed or if existing resources were updated to ensure that the
+	// last update time reflects the last time changes were propagated
```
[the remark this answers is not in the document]
marun: Done

pkg/controller/sync/status/status.go
```diff
@@ -176,17 +203,27 @@ func (s *GenericPropagationStatus) setPropagationCondition(reason AggregateReaso
...
-	transition := newCondition || propCondition.Status != newStatus
+	transition := newCondition || !(propCondition.Status == newStatus && propCondition.Reason == reason)
 	if transition {
-		propCondition.LastTransitionTime = propCondition.LastProbeTime
+		propCondition.LastTransitionTime = now
+	}
+
+	// Update the prop condition status and reason only after checking
+	// whether a transition has occurred.
+	propCondition.Status = newStatus
```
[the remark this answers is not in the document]
marun: Done

pkg/controller/sync/dispatch/managed.go
marun: Done

marun:
@font I think e2e is too heavyweight for validating the frequency of propagation status update frequency.  The unit testing validates the complicated bit - determining whether a change should be written. For all its cost e2e would only be validating that the boolean result of `setPropagationStatus` was being respected.

pkg/controller/sync/dispatch/managed.go
font: Ah ok, I thought it might be relying on the default boolean zero value on every reconciliation. Maybe just a comment to clarify?

pkg/controller/sync/status/status.go
font: Oops, that was a typo on my part as I meant to add an `s` as in `clustersStatusChanged` but that's too similar perhaps. How about `clustersPropStatusChanged` or `clustersStatusHasChanged`?

pkg/controller/sync/dispatch/managed.go
marun: It is an implicit invariant of `ManagedDispatcher` that it will not persist across reconciles (as per it only being [assigned to a local variable in a function called by reconcile](https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/controller/sync/controller.go#L305)). The collection of the rest of cluster propagation status similarly depends on this invariant. Any future maintenance that missed this detail would result in broken tests.

pkg/controller/sync/status/status.go
marun: Done

marun:
Rebased

PR #1100, opened 2019-08-13
controller: process label updates to kubefedcluster resources

pjferrell:
This PR addresses label update handling on KubeFedCluster resources.  Adding/removing labels was not triggering federated resource sync.


marun:
Nice catch!

/approve
/lgtm

PR #1103, opened 2019-08-14
kubefedctl federate: ensure that --dry-run avoids mutating api state

### LINKED ISSUE #1102: --dry-run param does not take effect for kubefedctl federate --enable-type
**What happened**:
```kubefedctl federate ns test-namespace --dry-run --kubefed-namespace=federation-system -e
customresourcedefinition.apiextensions.k8s.io/federatednamespaces.types.kubefed.io created
federatedtypeconfig.core.kubefed.io/namespaces created in namespace federation-system
I0812 16:46:39.848375    7470 federate.go:459] Resource to federate is a namespace. Given namespace will itself be the container for the federated namespace
I0812 16:46:39.848703    7470 federate.go:488] Successfully created FederatedNamespace "test-namespace/test-namespace" from Namespace
```
A dry run flag on the kubefedctl should not make any federation request to the server

Run: `kubefedctl federate ns test-namespace --dry-run --kubefed-namespace=federation-system -e`

**Environment**:
- 1.14
- KubeFed version
0.1.0-rc5
Namespaced
- Others

/kind bug

abhat:
This PR propagates the dry run flag to the `enable type` functionality
fixes #1102

### DIFF
--- a/pkg/kubefedctl/enable/enable.go
+++ b/pkg/kubefedctl/enable/enable.go
@@ -181,12 +181,7 @@ func (j *enableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return nil
 	}
 
-	if j.DryRun {
-		// Avoid mutating the API
-		return nil
-	}
-
-	return CreateResources(cmdOut, hostConfig, resources, j.KubeFedNamespace)
+	return CreateResources(cmdOut, hostConfig, resources, j.KubeFedNamespace, j.DryRun)
 }
 
 type typeResources struct {
@@ -224,7 +219,7 @@ func GetResources(config *rest.Config, enableTypeDirective *EnableTypeDirective)
 // TODO(marun) Allow updates to the configuration for a type that has
 // already been enabled for kubefed.  This would likely involve
 // updating the version of the target type and the validation of the schema.
-func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResources, namespace string) error {
+func CreateResources(cmdOut io.Writer, config *rest.Config, resources *typeResources, namespace string, dryRun bool) error {
 	write := func(data string) {
 		if cmdOut != nil {
 			if _, err := cmdOut.Write([]byte(data)); err != nil {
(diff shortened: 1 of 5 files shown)


abhat:
@marun do we not want to have --dry-run flag do what it should do in spirit which is run through as much of the code it can without mutating API state? If --dry-run is true, and --enable-type is false, we have code that will call `federate.CreateFederatedResources` with the dryRun flag. I was trying to do as much of dry-run executing as much of enable type code as possible without mutating API state.

marun:
@abhat Other than validating user input, what's the point of dry run? `CreateFederatedResources` basically does nothing when `dryRun` is true, so I'm not sure why `CreateResources` should be modified to do arbitrary operations that aren't related to validating user intput.

pkg/kubefedctl/enable/enable.go
```diff
@@ -181,12 +181,7 @@ func (j *enableType) Run(cmdOut io.Writer, config util.FedConfig) error {
 		return nil
 	}
 
-	if j.DryRun {
-		// Avoid mutating the API
-		return nil
-	}
-
-	return CreateResources(cmdOut, hostConfig, resources, j.KubeFedNamespace)
+	return CreateResources(cmdOut, hostConfig, resources, j.KubeFedNamespace, j.DryRun)
```
pmorie: As a user, I've expected to get feedback durign dry-run about what control plane mutations would happen, so I'm supportive of plumbing through dry-run here.

abhat:
@marun I am trying to follow canonical examples of how dry-run flag works on most Linux tools. In addition to validating user-input, it should IMO, run through the logic of `doing the thing` without actually `doing the thing`. From a UX stand-point, that is what I would imagine users expect. 

If someone runs just `enable` with dry-run today, there is no feedback to the user that the `enable` would have created the federated type CRD and/or said that the federated type is already enabled. One of the usability goals of any CLI should be to give a clear feedback to users in response to their action consistent with their mental model. This PR tries to make the dry-run functionality consistent with end-user's expectation for both the `federate` command and `enable` command. 

Having said that, if you think there is a technical reason why the current changes are a bad idea, I will defer to your proposed changes.

pkg/kubefedctl/enable/enable.go
marun: `resources` contains the resources that would be created and the user has the option of providing `--output=yaml` if they want to see that content. I think that should be enough to satisfy said expectations and that the changes proposed here don't provide additional information of value to the user.  

The use of dry run is inconsistent across `kubefedctl` commands, and if the goal is to make that more consistent I'm fine with that.  I think that's an orthogonal concern to fixing the bug indicated in the title of this PR, though, and I think the focus should be on maintaining consistency with the existing way `CreateResources` was being called (i.e. not calling for dry run) rather than what is currently proposed.

marun:
Is the point of this PR to fix a bug (dry run resulting in API changes)?  Or is the goal to make a dent in the inconsistent use of dry run in kubefedctl?  Those seem like orthogonal concerns to me.

abhat:
I agree those are two separate orthogonal concerns. The PR is trying to address both of them and in general, I view addressing the inconsistency in the dry-run behavior as a good thing.

> Is the point of this PR to fix a bug (dry run resulting in API changes)? Or is the goal to make a dent in the inconsistent use of dry run in kubefedctl? Those seem like orthogonal concerns to me.

PR #1107, opened 2019-08-15
WIP Set status.observedGeneration for federated resources  

marun:
**NOTE** This PR is not ready for review until #1099 merges.

Set status.observedGeneration for federated resources to provide an indication of whether the sync controller has processed the current state of resource.


marun:
Rebased and ready for review

marun:
Rebased

pkg/kubefedctl/enable/validation.go
```diff
@@ -240,6 +240,10 @@ func ValidationSchema(specProps v1beta1.JSONSchemaProps) *v1beta1.CustomResource
 								},
 							},
 						},
+						"observedGeneration": {
```
font: Does it make sense to make this required?

test/common/crudtester.go
```diff
@@ -448,34 +448,34 @@ func (c *FederatedTypeCrudTester) CheckPropagation(fedObject *unstructured.Unstr
...
 	// Retrieve the resource from the API to ensure the latest status
 	// is considered.
-	genericStatus, err := GetGenericStatus(c.client, fedObject.GroupVersionKind(), qualifiedName)
+	resource, err := GetGenericResource(c.client, fedObject.GroupVersionKind(), qualifiedName)
 	if err != nil {
 		return false, err
 	}
-	if genericStatus.Status == nil {
+	if resource.Status == nil {
 		c.tl.Logf("Propagation status is not yet available for %s %q", federatedKind, qualifiedName)
```
xunpan: `s/Propagation status/status/`

pkg/controller/sync/status/status.go
```diff
@@ -92,8 +92,9 @@ type GenericCondition struct {
 }
 
 type GenericFederatedStatus struct {
-	Conditions []*GenericCondition    `json:"conditions,omitempty"`
-	Clusters   []GenericClusterStatus `json:"clusters,omitempty"`
+	ObservedGeneration int64                  `json:"observedGeneration"`
```
xunpan: This seems also `omitempty` field. Any consideration?

pkg/kubefedctl/enable/validation.go
marun: I modeled it after `Deployments` and I think as per @xunpan's comment it should be an optional field.

test/common/crudtester.go
marun: Done

pkg/controller/sync/status/status.go
marun: I've updated the field to be `omitempty` as per your suggestion for consistency. Given that this struct is internal-only (it never touches the API), whether the field is optional or not won't have a functional impact.

marun: I've updated the field to be `omitempty` as per your suggestion for consistency. Given that this struct is internal-only (it never touches the API), whether the field is optional or not won't have a functional impact.

marun:
Rebased and updated to address comments.

pkg/kubefedctl/enable/validation.go
marun: I wonder the same about Deployments. I'm not sure if `observedGeneration` being optional is due to an explicit decision or the result of evolution (the latter would make more sense).

PR #1111, opened 2019-08-16
WIP Remove the filename option to kubefedctl

abhat:
This PR removes support for the filename option in kubefedctl.
Fixes #

As per @marun: It was a convenience for testing and expected to be used by installers that would execute `kubefedctl enable` but that use never materialized. So it's prudent to just remove the filename option altogether.

### DIFF
--- a/pkg/kubefedctl/federate/federate.go
+++ b/pkg/kubefedctl/federate/federate.go
@@ -92,9 +92,8 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
-	flags.StringVarP(&j.filename, "filename", "f", "", "If specified, the provided yaml file will be used as the input for target resources to federate. This mode will only emit federated resource yaml to standard output. Other flag options if provided will be ignored.")
 	flags.StringSliceVarP(&j.skipAPIResourceNames, "skip-api-resources", "s", []string{}, "Comma separated names of the api resources to skip when federating contents in a namespace. Name could be short name "+
-		"(e.g. 'deploy), kind (e.g. 'deployment'), plural name (e.g. 'deployments'), group qualified plural name (e.g. 'deployments.apps') or group name itself (e.g. 'apps') to skip the whole group.")
+		"(e.g. 'deploy'), kind (e.g. 'deployment'), plural name (e.g. 'deployments'), group qualified plural name (e.g. 'deployments.apps') or group name itself (e.g. 'apps') to skip the whole group.")
 }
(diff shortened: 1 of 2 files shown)


pkg/kubefedctl/federate/federate.go
```diff
@@ -92,9 +92,8 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
...
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
-	flags.StringVarP(&j.filename, "filename", "f", "", "If specified, the provided yaml file will be used as the input for target resources to federate. This mode will only emit federated resource yaml to standard output. Other flag options if provided will be ignored.")
 	flags.StringSliceVarP(&j.skipAPIResourceNames, "skip-api-resources", "s", []string{}, "Comma separated names of the api resources to skip when federating contents in a namespace. Name could be short name "+
-		"(e.g. 'deploy), kind (e.g. 'deployment'), plural name (e.g. 'deployments'), group qualified plural name (e.g. 'deployments.apps') or group name itself (e.g. 'apps') to skip the whole group.")
+		"(e.g. 'deploy'), kind (e.g. 'deployment'), plural name (e.g. 'deployments'), group qualified plural name (e.g. 'deployments.apps') or group name itself (e.g. 'apps') to skip the whole group.")
```
marun: Please do your best to separate unrelated changes into separate commits to help future maintainers understand the rational for a given change.

pkg/kubefedctl/federate/federate.go
```diff
@@ -92,9 +92,8 @@ func (j *federateResource) Bind(flags *pflag.FlagSet) {
 	flags.StringVarP(&j.output, "output", "o", "", "If provided, the resource that would be created in the API by the command is instead output to stdout in the provided format.  Valid format is ['yaml'].")
 	flags.BoolVarP(&j.enableType, "enable-type", "e", false, "If true, attempt to enable federation of the API type of the resource before creating the federated resource.")
 	flags.BoolVarP(&j.federateContents, "contents", "c", false, "Applicable only to namespaces. If provided, the command will federate all resources within the namespace after federating the namespace.")
-	flags.StringVarP(&j.filename, "filename", "f", "", "If specified, the provided yaml file will be used as the input for target resources to federate. This mode will only emit federated resource yaml to standard output. Other flag options if provided will be ignored.")
```
marun: This should not be removed. Whereas `-f` for `kubefedctl enable` is unused, it allows `federate` to accept resources for conversion from disk rather than from the API.

pkg/kubefedctl/federate/federate.go
marun: Given that this change is not actually in the same file as the intended change for this PR, please move to a separate commit.

marun:
@abhat My apologies, as per the CI failure `-f` is being used by the helm sync script to generate the enable CRDs for the helm chart.  This is a legacy of a time when enabling a type required more than a type name. At this point I think a file containing a list of type names would be a good replacement and a preferable option to having to support `EnableTypeDirective` in the future.

abhat:
@marun i will just close the PR in that case?

marun:
@abhat I still think it would be desirable to make the changes necessary to remove the option, even if more changes are required than originally anticipated.

pkg/kubefedctl/federate/federate.go
abhat: 👍

PR #1113, opened 2019-08-16
The DEFAULT_KUBEFED_SCOPE env var was only being set the first time the KubeFedConfig didn't exist

abhat:

This PR makes sure, the KubeFedConfig's Spec.Scope is updated everytime the controller-manager pod is started irrespective of whether the KubeFedConfig exists or not.

Fixes # Bugzilla: [1741478](https://bugzilla.redhat.com/show_bug.cgi?id=1741478)

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -243,20 +243,33 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
 	return fedConfig
 }
 
-func setDefaultKubeFedConfigScope(fedConfig *corev1b1.KubeFedConfig) {
-	if len(fedConfig.Spec.Scope) == 0 {
-		// TODO(sohankunkerkar) Remove when no longer necessary.
-		// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
-		// Its continued existence should not be relied upon.
-		const defaultScopeEnv = "DEFAULT_KUBEFED_SCOPE"
-		defaultScope := os.Getenv(defaultScopeEnv)
-		if len(defaultScope) != 0 {
-			if defaultScope != string(apiextv1b1.ClusterScoped) && defaultScope != string(apiextv1b1.NamespaceScoped) {
-				klog.Fatalf("%s must be Cluster or Namespaced; got %q", defaultScopeEnv, defaultScope)
-			}
+func setDefaultKubeFedConfigScope(fedConfig *corev1b1.KubeFedConfig) bool {
+	// TODO(sohankunkerkar) Remove when no longer necessary.
+	// This Environment variable is a temporary addition to support Red Hat's downstream testing efforts.
+	// Its continued existence should not be relied upon.
+	const defaultScopeEnv = "DEFAULT_KUBEFED_SCOPE"
+	defaultScope := os.Getenv(defaultScopeEnv)
+	if len(defaultScope) != 0 {
+		if defaultScope != string(apiextv1b1.ClusterScoped) && defaultScope != string(apiextv1b1.NamespaceScoped) {
+			klog.Fatalf("%s must be one of %s or %s; got %q", defaultScopeEnv,
+				string(apiextv1b1.ClusterScoped), string(apiextv1b1.NamespaceScoped), defaultScope)
+			return false
+		}
+
+		if len(fedConfig.Spec.Scope) == 0 {
 			fedConfig.Spec.Scope = apiextv1b1.ResourceScope(defaultScope)
+			klog.Infof("Setting the scope of KubeFedConfig spec to %s", defaultScope)
+			return true
+		} else {
+			if fedConfig.Spec.Scope != apiextv1b1.ResourceScope(defaultScope) {
+				klog.Infof("Setting the scope of KubeFedConfig spec from %s to %s",
+					string(fedConfig.Spec.Scope), defaultScope)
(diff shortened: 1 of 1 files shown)


marun:
Is it preferable for the operator to change the scope of an existing kubefedconfig vs removing the kubefedconfig and having the desired scope apply to the new kubefedconfig that would be created?

cmd/controller-manager/app/controller-manager.go
```diff
@@ -243,20 +243,26 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
 	return fedConfig
 }
 
-func setDefaultKubeFedConfigScope(fedConfig *corev1b1.KubeFedConfig) {
-	if len(fedConfig.Spec.Scope) == 0 {
-		// TODO(sohankunkerkar) Remove when no longer necessary.
```
font: We should keep this TODO as this should be a temporary solution.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -243,20 +243,26 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
...
-		if len(defaultScope) != 0 {
-			if defaultScope != string(apiextv1b1.ClusterScoped) && defaultScope != string(apiextv1b1.NamespaceScoped) {
-				klog.Fatalf("%s must be Cluster or Namespaced; got %q", defaultScopeEnv, defaultScope)
-			}
+func setDefaultKubeFedConfigScope(fedConfig *corev1b1.KubeFedConfig) bool {
+	const defaultScopeEnv = "DEFAULT_KUBEFED_SCOPE"
+	defaultScope := os.Getenv(defaultScopeEnv)
+	if len(defaultScope) != 0 {
+		if defaultScope != string(apiextv1b1.ClusterScoped) && defaultScope != string(apiextv1b1.NamespaceScoped) {
+			klog.Fatalf("%s must be Cluster or Namespaced; got %q", defaultScopeEnv, defaultScope)
```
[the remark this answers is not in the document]
abhat: @font: That is an existing log message, but yes, I will change it. BTW, I think it's ok to catch the error early rather than wait for `ValidateKubeFedConfig` to get called.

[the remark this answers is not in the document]
font: It's not just waiting for `ValidateKubeFedConfig` (even though it's just as early IMO). This code was implemented at a time when the validation webhook did not exist. So the next function calls to either create or update the KubeFedConfig will trigger a check by the validation webhook anyway. A failure to create or update the KubeFedConfig results in a fatal message as well rendering this check repetitive.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -243,20 +243,30 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
...
+	const defaultScopeEnv = "DEFAULT_KUBEFED_SCOPE"
+	defaultScope := os.Getenv(defaultScopeEnv)
+	if len(defaultScope) != 0 {
+		if defaultScope != string(apiextv1b1.ClusterScoped) && defaultScope != string(apiextv1b1.NamespaceScoped) {
+			klog.Fatalf("%s must be one of %s or %s; got %q", defaultScopeEnv,
+				string(apiextv1b1.ClusterScoped), string(apiextv1b1.NamespaceScoped), defaultScope)
+			return false
+		}
+
+		if len(fedConfig.Spec.Scope) == 0 {
```
font: This if/else is performing the same logic i.e. assigning `defaultScope` to `fedConfig.Spec.Scope` and returning `true`. We should either log when they don't match or remove the use of the if/else. I think logging may be preferred so there's some indication of the KubeFedConfig being overwritten with the value of `DEFAULT_KUBEFED_SCOPE`.

abhat: @font, @xunpan made the changes as per the suggestion above.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -243,20 +243,33 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
...
 			fedConfig.Spec.Scope = apiextv1b1.ResourceScope(defaultScope)
+			klog.Infof("Setting the scope of fedConfig spec to %s", defaultScope)
```
font: @abhat Please use `KubeFedConfig` instead of `fedConfig`.

abhat: @font made the change. Thanks.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -243,20 +243,33 @@ func getKubeFedConfig(opts *options.Options) *corev1b1.KubeFedConfig {
...
+
+		if len(fedConfig.Spec.Scope) == 0 {
 			fedConfig.Spec.Scope = apiextv1b1.ResourceScope(defaultScope)
+			klog.Infof("Setting the scope of KubeFedConfig spec to %s", defaultScope)
+			return true
+		} else {
+			if fedConfig.Spec.Scope != apiextv1b1.ResourceScope(defaultScope) {
+				klog.Infof("Setting the scope of KubeFedConfig spec from %s to %s",
+					string(fedConfig.Spec.Scope), defaultScope)
+				fedConfig.Spec.Scope = apiextv1b1.ResourceScope(defaultScope)
```
font: Apologies for not bringing this up earlier, but we should only log within the if/else and the identical statements can be performed afterwards.

PR #1117, opened 2019-08-19
Update kubefed release automation scripts

font:
This PR adds the following updates to the release automation scripts:
- Increases the timeout to wait for Travis CI job to begin as I've seen it queued for 10-15 minutes before actually starting.
- Makes `build-release.sh` script idempotent in the event a timeout does occur or if the user accidentally cancels the script.
- Adds check to `create-gh-release.sh` to make sure assets file exists first before proceeding.

### DIFF
--- a/scripts/build-release.sh
+++ b/scripts/build-release.sh
@@ -45,6 +45,15 @@ function build-release-artifacts() {
 
 function create-and-push-tag() {
   # Use the upstream git remote convention name used by hub.
+
+  if git ls-remote --tags "${GITHUB_REMOTE_UPSTREAM_NAME}" refs/tags/"${RELEASE_TAG}" &> /dev/null; then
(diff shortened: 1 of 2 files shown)


scripts/build-release.sh
```diff
@@ -94,13 +113,13 @@ function verify-travis-build-completed() {
 }
 
 function verify-continuous-integration() {
-  PASSED_STATUS="passed"
-  util::wait-for-condition "kubefed CI build to start" "verify-travis-build-started started" 300
-  util::wait-for-condition "kubefed CI build to complete" "verify-travis-build-completed" 3600
+  util::wait-for-condition "kubefed CI build to start" "travis-build-started" 1200
```
pmorie: might want to pull the constants here into variables in the future

PR #1118, opened 2019-08-19
Fix BZ 1740052 - Federation of SAs and Secrets that have a : in them doesn't work because of CRD expression limitation

abhat:
Certain secrets and service accounts in Openshift have a ":" in their names and if we try to federate these resources in a given namespace, we get a validation error: 

```F0812 16:15:45.724853    6232 federate.go:150] Error: Error creating federated resource "test-namespace/system:deployers": FederatedRoleBinding.types.kubefed.io "system:deployers" is invalid: metadata.name: Invalid value: "system:deployers": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')```

This PR skips federating resources whose names don't conform to the DNS-1123 specification.

Fixes # BZ 1740052

### DIFF
--- a/pkg/kubefedctl/federate/util.go
+++ b/pkg/kubefedctl/federate/util.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation"
 	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
 	versionhelper "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/kubefedctl/federate/util.go [hunk 1]
```diff
@@ -168,6 +172,10 @@ func apiResourceMatchesSkipName(apiResource metav1.APIResource, skipAPIResourceN
 	return false
 }
 
+func apiResourceNameMatchesInvalidCRDResourceNames(apiResource metav1.APIResource) bool {
+	return strings.ContainsAny(apiResource.Name, forbiddenCharsInDNS1123Subdomain)
```
font: Instead of tracking which characters are forbidden in the DNS 1123 spec, I suggest we just call:
https://github.com/kubernetes-sigs/kubefed/blob/fc7a5fd38f4192f9ef802017b3fe58ef9045e5de/vendor/k8s.io/apimachinery/pkg/util/validation/validation.go#L133-L144
and return true if errors are reported.

abhat: @font That's an excellent suggestion. I did a cursory grep for such a function and had missed this somehow. Maybe didn't have the right grep search string. In either case, I have made the changes that use the above.

marun:
Given that openshift allows these names for system-created resources, would it be possible for a user of openshift to create a resource with a name that would be skipped? If so, would it make sense to log when skipping such resources to enable troubleshooting when `kubefedctl federate` unexpectedly fails for such resources?

Also, please limit references to external bug trackers to a note at the bottom of the issue text. `Fixes` is only for GH issues and referencing external trackers in the title adds noise for upstream consumers.

PR #1119, opened 2019-08-19
Add version info to KubeFed admission webhook

font:
This adds version info to the KubeFed admission webhook server similar to the `controller-manager` and `kubefedctl`

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -17,6 +17,7 @@ limitations under the License.
 package webhook
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/openshift/generic-admission-server/pkg/apiserver"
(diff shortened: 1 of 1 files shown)


xunpan:
LGTM

pkg/webhook/webhook.go
```diff
@@ -40,5 +42,16 @@ func NewWebhookCommand(stopChan <-chan struct{}) *cobra.Command {
 	cmd.Short = "Start a kubefed webhook server"
 	cmd.Long = "Start a kubefed webhook server"
 
+	versionFlag := false
+	cmd.Flags().BoolVar(&versionFlag, "version", false,
+		"Prints version information for kubefed admission webhook and quits")
+	cmd.PreRun = func(c *cobra.Command, args []string) {
```
abhat: nit: Do we need to validate if someone gives additional command line options in addition to --version?

font: We don't need to validate for additional arguments because we'd have to do that for all commands, including `-h` for help, which supersedes any option regardless if more options are specified. So this would be consistent with that behavior and any other executable with POSIX style flags. For additional options which are invalid, the `pflag` package will handle those.

abhat: 👍 . Thanks for clarifying that.

PR #1120, opened 2019-08-19
controller/util: maintain a cache of cluster clients to reduce rest discovery

pjferrell:
Cluster client creation triggers rest discovery against the target cluster.  To reduce client creations, a cache has been added to federatedInformerImpl.  The cache is populated as cluster clients are used and kubefedcluster deletes clear the cached client corresponding to that cluster.

### DIFF
--- a/pkg/controller/util/federated_informer.go
+++ b/pkg/controller/util/federated_informer.go
@@ -165,6 +165,7 @@ func NewFederatedInformer(
 		},
 		targetInformers: make(map[string]informer),
 		fedNamespace:    config.KubeFedNamespace,
+		clusterClients:  make(map[string]generic.Client),
 	}
 
 	getClusterData := func(name string) []interface{} {
@@ -278,6 +279,9 @@ type federatedInformerImpl struct {
 	// Retrieves configuration to access a cluster.
 	configFactory func(*fedv1b1.KubeFedCluster) (*restclient.Config, error)
 
+	// Caches cluster clients (reduces client discovery and secret retrieval)
+	clusterClients map[string]generic.Client
+
 	// Namespace from which to source KubeFedCluster resources
 	fedNamespace string
 }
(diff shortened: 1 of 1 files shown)


xunpan:
@pjferrell 
I restarted the test and it failed again in `configmaps`. So, please help to check if it is related to your code change. Thanks.

pjferrell:
@xunpan 
The federated configmap was being created with a cluster placement that selects all clusters.  This caused the "unlabeled" configmap to become adopted by sync controller.  When the federatedconfigmap was deleted (crud.go:213), sync controller removed the adopted configmap and the explicit delete of the configmap (crud.go:189) failed.

Test case would fail if the time between the two deletes was long enough for sync controller to complete the deletion.  With clusterClients cached, the time to propogate changes to clusters is shorter.

Original placement in test case:
`
  spec:
    placement:
      clusterSelector:
        matchLabels: {}
      clusters: []
`

New placement after test case change:
`
  spec:
    placement: {}
`

xunpan:
@pjferrell 
Thanks for you update. I think it is fine. My question is:
is there any exception/condition to make a `generic.Client` invalid? We do not have client recreating mechanism in new code. Is that a case?

pjferrell:
@xunpan Based on conversation with @marun, no controllers are currently handling secret changes for kudfedclusters.  This would be one case where generic.Client would need to be recreated.

Plan was to address this later when token rotation is handled.

xunpan:
@pjferrell 
Thanks for your explanation.

/lgtm

marun:
I'm reviewing this, I'd like to make sure I understand the test fix before merging.

marun:
> @xunpan Based on conversation with @marun, no controllers are currently handling secret changes for kudfedclusters. This would be one case where generic.Client would need to be recreated.

I've filed #1132 to ensure followup on this behavior.

PR #1121, opened 2019-08-20
Change shorthand --enable-type flag to avoid typo

font:
The `--enable-type` long option can be accidentally typed with only 1 `-` as in `-enable-type` resulting in the following error:

```bash
$ kubefedctl federate secret kubefed -enable-type
F0809 15:24:48.160303   17400 federate.go:150] Error: Error retrieving target Secret "able-type/kubefed": secrets "kubefed" not found
```

The error occurs because combined shorthand options as in `-en` are valid. If a shorthand option takes an argument, as in `-n`, it must be the last argument, so it interprets the rest of the command `able-type` as the namespace argument. This can be confusing and is only a special case because `e` and `n` are both shorthand flags while their combined use spells out the long form of an option.

This fixes that by changing the shorthand flag for `--enable-type` to `-t` instead of `-e`.

Fixes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 # Unreleased
+-  [#1121](https://github.com/kubernetes-sigs/kubefed/pull/1121) Update
+   `kubefedctl federate` shorthand option for `--enable-type` to `-t` instead
+   of `-e` to avoid confusing error message when only one dash is accidentally
+   used e.g. `-enable-type`, resulting in a valid parsing of flags but
+   erroneous use of the option.
 
 # v0.1.0-rc6
 -  [#1099](https://github.com/kubernetes-sigs/kubefed/pull/1099)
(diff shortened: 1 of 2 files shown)


font:
> The shorthand of other related options are the first charactor of the long name for easy memorized.
> What about change the long name to make it consistent with others?
> e.g. `s/enable-type/type-enabled/`

@xunpan Yes, that's the downside i.e. the shorthand letter is not the first character of the long name. Thanks for the suggestion. I think it's preferable to stick with `enable-type` instead of `type-enabled`. The alternative is to not allow a shorthand letter at all, but I thought a shorthand option might be preferable, even if it isn't the first letter of the long name.

abhat:
> > The shorthand of other related options are the first charactor of the long name for easy memorized.
> > What about change the long name to make it consistent with others?
> > e.g. `s/enable-type/type-enabled/`
> 
> @xunpan Yes, that's the downside i.e. the shorthand letter is not the first character of the long name. Thanks for the suggestion. I think it's preferable to stick with `enable-type` instead of `type-enabled`. The alternative is to not allow a shorthand letter at all, but I thought a shorthand option might be preferable, even if it isn't the first letter of the long name.

Yeah, agree with @font - the verb-noun form is more intuitive and consistent IMO. But good suggestion nevertheless @xunpan.

xunpan:
It is not a mandatory change but suggestion. I'm ok with your comments.

LGTM

PR #1129, opened 2019-08-26
WIP Ensure that an empty `clusters` field selects no clusters.

marun:
Previously an empty `spec.placement.clusters` could result in `spec.placement.clusterSelector` being used (if present). An empty `clusters` field now ensures that no clusters will be selected. `clusterSelector` will only be respected if `clusters` is nil/not provided.

This PR was prompted by a test fix proposed in #1120 by @pjferrell. The test failure resolved by that PR revealed a bug in placement handling that this PR is intended to address.


xunpan:
LGTM

marun:
Rebased

PR #1130, opened 2019-08-27
Remove unused timeout var from the sync controller

marun:


marun:
> It looks like the `SchedulingPreferenceController` contains this same unused variable. Should we remove it there too?

Good catch! Done.

PR #1134, opened 2019-08-28
Delete and Recreate the KubeFedConfig on scope change.

abhat:
In PR1113 we added an update API for KubeFedConfig. But the webhook validates the Spec.Scope change as being immutable. This PR will delete the older default KubeFedConfig and recreate a new one with the updated scope.

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -290,7 +290,7 @@ func createKubeFedConfig(config *rest.Config, fedConfig *corev1b1.KubeFedConfig)
 	}
 }
 
-func updateKubeFedConfig(config *rest.Config, fedConfig *corev1b1.KubeFedConfig) {
+func deleteKubeFedConfig(config *rest.Config, fedConfig *corev1b1.KubeFedConfig) {
 	name := fedConfig.Name
 	namespace := fedConfig.Namespace
 	qualifiedName := util.QualifiedName{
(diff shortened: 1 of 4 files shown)


marun:
I'm holding this for now.  The scope of a control plane is immutable because clusters are registered at that scope.  Changing the scope without removing all registered clusters is likely to result in unexpected behavior.

cmd/controller-manager/app/controller-manager.go
```diff
@@ -300,9 +300,9 @@ func updateKubeFedConfig(config *rest.Config, fedConfig *corev1b1.KubeFedConfig)
 
 	client := genericclient.NewForConfigOrDieWithUserAgent(config, "kubefedconfig")
 	// Update the KubeFedConfig requested by the caller
```
font: s/Update/Delete/

abhat:
@font addressed the review comment. ptal.

marun:
How is deleting+creating preferable to updating? The problem of existing clusters joined at a previous scope appears to be the  same in both cases.

font:
> Why is deleting+creating preferable to updating? The problem of existing clusters joined at a previous scope appears to be the same in both cases.

My suggestion to delete and create is only to maintain the requirement for scope immutability.

On the issue of existing clusters, the logic here is only exercised by the kubefed operator. I presume this can only happen if the operator is installing a fresh kubefed deployment (more of a reason to have the operator perform this kubefeconfig scope synchronization), but I'll leave that to @abhat to comment.

abhat:
@marun , @font: Yes, we are not inherently solving the problem of changing the scope of the control plane when there are existing clusters that are already joined with this PR. This is more so to address the fact that the KubeFedConfig was having an incorrect scope as against the scope of the control plane it is used by. We added the logic to delete and recreate the KubeFedConfig (with the right scope as the CR instance being reconciled) in order to keep the sanctity of immutability of the scope field in KubeFedConfig.

marun:
/hold cancel

@abhat Thank you for the detail. I realize now that deletion/creation will only be performed if the env var is set, which is documented as an unsupported RH-only feature.

PR #1135, opened 2019-08-28
Skip resources that don't have a valid DNS-1123 Subdomain name

abhat:
PR#1118 added an incorrect way to skip federating resources that
don't comply with the DNS-1123 subdomain naming convention.
This PR fixes the issue the correct way.

### DIFF
--- a/pkg/kubefedctl/federate/util.go
+++ b/pkg/kubefedctl/federate/util.go
@@ -31,6 +31,7 @@ import (
 	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
 	versionhelper "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/rest"
+	"k8s.io/klog"
 	"sigs.k8s.io/yaml"
 
 	"sigs.k8s.io/kubefed/pkg/apis/core/typeconfig"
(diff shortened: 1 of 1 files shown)


pkg/kubefedctl/federate/util.go
```diff
@@ -207,7 +207,11 @@ func getResourcesInNamespace(config *rest.Config, namespace string, skipAPIResou
 		targetResources := resources{apiResource: apiResource}
 		for _, item := range resourceList.Items {
 			resource := item
-			targetResources.resources = append(targetResources.resources, &resource)
+			if len(validation.IsDNS1123Subdomain(resource.GetName())) == 0 {
+				targetResources.resources = append(targetResources.resources, &resource)
+			} else {
+				klog.Warningf("Skipping resource %s of type %s", resource.GetName(), apiResource.Name)
```
font: We should add the reason it's being skipped i.e. does not conform to DNS 1123 spec. Additionally, I think it would be good to spit out the errors from the call to `IsDNS1123Subdomain` so that users are aware of what exactly does not conform.

pkg/kubefedctl/federate/util.go
```diff
@@ -207,7 +207,16 @@ func getResourcesInNamespace(config *rest.Config, namespace string, skipAPIResou
 		targetResources := resources{apiResource: apiResource}
 		for _, item := range resourceList.Items {
 			resource := item
-			targetResources.resources = append(targetResources.resources, &resource)
+			errors := validation.IsDNS1123Subdomain(resource.GetName())
+			if len(errors) == 0 {
+				targetResources.resources = append(targetResources.resources, &resource)
+			} else {
+				klog.Warningf("Skipping resource %s of type %s because it does not conform to the DNS-1123 subdomain spec.", resource.GetName(), apiResource.Name)
```
font: To simplify this into one klog invocation, you could print all the errors using `%v`.

PR #1137, opened 2019-09-05
add resource "secrets"  for kubefed-role

osswangxining:
currently kubefed-role  does not include resource "secrets"  in API group "" so that it will cause the "system:serviceaccount:kube-federation-system:kubefed-controller" cannot get resource "secrets".

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/clusterrole.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/clusterrole.yaml
@@ -51,6 +51,7 @@ rules:
   - ""
   resources:
   - namespaces
+  - secrets
   verbs:
   - get
   - watch


osswangxining:
pass CLA

marun:
While I appreciate your initiative, this change is not necessary. The controller manager only needs access to secrets in the kubefed namespace, and this access is provided by a role rather than a cluster role:

https://github.com/kubernetes-sigs/kubefed/blob/master/charts/kubefed/charts/controllermanager/templates/roles.yaml#L61

If you are experiencing a problem with access to secrets, the likely culprit is related to deployment.  Please file an issue or reach out on slack.

osswangxining:
> While I appreciate your initiative, this change is not necessary. The controller manager only needs access to secrets in the kubefed namespace, and this access is provided by a role rather than a cluster role:
> 
> https://github.com/kubernetes-sigs/kubefed/blob/master/charts/kubefed/charts/controllermanager/templates/roles.yaml#L61
> 
> If you are experiencing a problem with access to secrets, the likely culprit is related to deployment. Please file an issue or reach out on slack.

When the scope of the the kubefed control plane is specified to Cluster, it looks to use this cluster role to access the secrets.  
https://github.com/kubernetes-sigs/kubefed/blob/master/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml#L13

marun:
> When the scope of the the kubefed control plane is specified to Cluster, it looks to use this cluster role to access the secrets.
> https://github.com/kubernetes-sigs/kubefed/blob/master/charts/kubefed/charts/controllermanager/templates/clusterrolebindings.yaml#L13

For a cluster-scoped deployment of kubefed, the cluster role binding you link to is used in addition to the role binding rather than instead of it. Cluster-scoped deployment and its rbac configuration are validated on every CI job, and the job is not failing due to lack of access to secrets. 

Are you experiencing this problem in a local deployment? If so, would you be interested in sharing the steps you took to deploy?

osswangxining:
@marun Thanks, you are right, as I was missing that rolebinding during my kubefed-operator development.

PR #1146, opened 2019-09-09
#1083: use unstructured runtime.Object for more detailed reflector logging

### LINKED ISSUE #1083: Informer logging lacks type detail
KubeFed's sync controller uses informers to watch federated resources in the host cluster and target resources in member clusters. The sync controller is configured with type information via the API to support the federation of types (e.g. CRDs) for which golang structs may not be available at compilation time.  Accordingly, the sync controller configures its informers with an `ObjectType` (indicating the type of resource that will be watched) of `nil`. `ObjectType` is passed by an informer to an internal `Reflector` instance, where it is used to:

1 -  check that the type of an event's object is as expected (i.e. [1](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/reflector.go#L345))
2 -  indicate the type of resource that a log pertains to (e.g. [1](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/reflector.go#L122), [2](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/reflector.go#L201))

Passing `nil` for `ObjectType` avoids the type check to support watching resources for which a golang struct is not available, but also results in most `Reflector` log messages displaying `<nil>` instead of a type name: 

```
0806 13:49:50.180103    3856 reflector.go:131] Starting reflector <nil> (0s) from sigs.k8s.io/kubefed/pkg/controller/sync/accessor.go:161
I0806 13:49:50.180207    3856 reflector.go:169] Listing and watching <nil> from sigs.k8s.io/kubefed/pkg/controller/sync/accessor.go:161
I0806 13:49:50.180106    3856 reflector.go:169] Listing and watching <nil> from sigs.k8s.io/kubefed/pkg/controller/sync/accessor.go:156
```

Ideally `Reflector` could be configured with a `typeName` to be used in log messages. If `typeName` were not provided, it would be continue to be computed as `expectedType.String()`. A new constructor for `Reflector` and `Informer` could be added to support provision of `typeName`.

Note that KubeFed would likely prefer to supply a `typeName` of `<cluster name>/<type name>` to ensure traceability of an informer not just to a sync controller but also to the cluster it is targeting.

pjferrell:
use unstructured runtime.Object with the apiResource info when registering informers (needs backend change in client-go to log the gvk of the unstructured object)

Fixes #1083 

**Special notes for reviewer**:
Detailed logging will be dependent on merge of below PR to Kubernetes client-go:
https://github.com/kubernetes/kubernetes/pull/82388

### DIFF
--- a/pkg/controller/util/resourceinformer.go
+++ b/pkg/controller/util/resourceinformer.go
@@ -23,24 +23,32 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
 )
 
-// NewManagedResourceInformer returns an unfiltered informer.
-func NewResourceInformer(client ResourceClient, namespace string, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
-	return newResourceInformer(client, namespace, triggerFunc, "")
+// NewResourceInformer returns an unfiltered informer.
+func NewResourceInformer(client ResourceClient, namespace string, apiResource *metav1.APIResource, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
+	return newResourceInformer(client, namespace, apiResource, triggerFunc, "")
 }
 
 // NewManagedResourceInformer returns an informer limited to resources
 // managed by KubeFed as indicated by labeling.
-func NewManagedResourceInformer(client ResourceClient, namespace string, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
+func NewManagedResourceInformer(client ResourceClient, namespace string, apiResource *metav1.APIResource, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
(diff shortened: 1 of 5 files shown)


xunpan:
@pjferrell 
Please merge conflict files.

marun:
Marking WIP since it won't be mergeable until we cherry-pick or rebase the corresponding kube change.

marun:
Nevermind the WIP, this is of course mergeable as is.  

@pjferrell I'm happy to merge after rebase.

pkg/controller/util/resourceinformer.go
```diff
@@ -23,24 +23,32 @@ import (
...
 }
 
-func newResourceInformer(client ResourceClient, namespace string, triggerFunc func(pkgruntime.Object), labelSelector string) (cache.Store, cache.Controller) {
+func newResourceInformer(client ResourceClient, namespace string, apiResource *metav1.APIResource, triggerFunc func(pkgruntime.Object), labelSelector string) (cache.Store, cache.Controller) {
+	obj := &unstructured.Unstructured{}
+
+	if apiResource != nil {
+		gvk := schema.GroupVersionKind{Group: apiResource.Group, Version: apiResource.Version, Kind: apiResource.Kind}
+		obj.SetGroupVersionKind(gvk)
+		obj.SetNamespace(namespace)
```
marun: Why is it desirable to set the namespace?

hectorj2f:
We also experienced some issues to better understand the some of the current error logs, so it would be good to get these changes in.

irfanurrehman:
This looks good to me.
/retest

irfanurrehman:
@hectorj2f would you be able to please look at the failure.

jimmidyson:
This needs a rebase please @pjferrell.

Issue #1141, opened 2019-09-06
CRD validation schema is not compatible with being published in OpenAPI schema


qiyongxiao:
Dears Authors:
    I want to use 'kubectl explain ' to show FederatedSecret's substance. I don't know how to show it. While I do this command, and show below:
```
[root@master1 docs]# kubectl explain FederatedSecret
error: Couldn't find resource for "types.kubefed.io/v1beta1, Kind=FederatedSecret"
```
Could I use this kinds of command to show it's substances? Please tell me how to show it.
Thanks.

marun:
`kubectl explain` support for CRDs requires the `CustomResourcePublishOpenAPI` feature flag to be set to `true`.  In Kube 1.14 this feature was alpha and disabled by default. In Kube 1.15 the feature was beta and enabled by default. If you are still on 1.14, you can enable the feature flag as per the documentation:

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

On Kube 1.15, though, I'm getting an empty result for kubefed api types instead of an error.  It appears that our CRDS (both generated by KubeBuilder 1.x and created dynamically) do not have validation schema that is compatible with being published in the OpenAPI schema (which `kubectl explain` depends on), as per the following status condition:

```
  - lastTransitionTime: "2019-09-06T13:29:24Z"
    message: 'spec.validation.openAPIV3Schema.type: Required value: must not be empty
      at the root'
    reason: Violations
    status: "True"
    type: NonStructuralSchema
```

marun:
#1142 ensures federated types will be included in OpenAPI schema.  Addressing core type inclusion likely requires updating the vendored version of kubebuilder to 2.x.

marun:
Note that `kubectl explain` doesn't actually provide much information. Services and Spec are the most useful fields and even for core types very little detail is provided:

```
    Spec defines the behavior of a service.
     https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status

   status       <Object>
     Most recently observed status of the service. Populated by the system.
     Read-only. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status
```

The kubefed userguide is likely to be a more useful source of detail of federated types.

qiyongxiao:
> #1142 ensures federated types will be included in OpenAPI schema. Addressing core type inclusion likely requires updating the vendored version of kubebuilder to 2.x.

[...]

PR #1150, opened 2019-09-13
Update to kubebuilder 0.2.0 and kube 1.14 

marun:
Kubernetes 1.16 requires structural schema for CRDs and generating compatible schema with kubebuilder requires updating to the latest version (0.2.0). This ensures that that the validation schema of types defined in code will include top-level `type: object`. 

This change also required updating the vendored versions of kube dependencies to 1.14 (the version used by kubebuilder 0.2.0).  Updating to a more recent version of kube depends on kubebuilder updating to that newer version first.

Partial fix for #1141


xunpan:
LGTM

jimmidyson:
Thanks @marun for this! How can we get this merged? We're looking to use kubefed as a library and upgrading controller-runtime (for the dynamic client) is essential for us to use it.

marun:
@jimmidyson Just need a reviewer to do the honors, nothing blocking it. btw an update to 1.16 will likely follow as soon as [controller-runtime gets updated](https://github.com/kubernetes-sigs/controller-runtime/pull/588).

xunpan:
As there is no other reviewers has concern about this, just merge it. 
/lgtm

PR #1152, opened 2019-09-24
Allow setting custom kind image to create-servers.sh via KIND_IMAGE

marun:
Given that there is often a lag between a kube release and when official kind images become available for that release, it may be necessary for developers to build their own kind node images for testing. This change allows providing a custom image name via the `KIND_IMAGE` env var to `scripts/create-clusters.sh` script.


xunpan:
It is a clear change. Just merge it.

/lgtm

PR #1154, opened 2019-09-27
Return created kubefedcluster from JoinCluster

jimmidyson:

When using kubefed as a library, returning the created `kubefedcluster` is very useful to be able to set references, ownership, etc after successful creation.


jimmidyson:
No, thank you @font for all your hard work on this amazing project!

PR #1157, opened 2019-10-06
Code linting and other minor normalizations 

hectorj2f:

We are using kubefed and while I was checking the code, I found certain code that was non-idiomatic golang code. I also found some other minor chores that I also addressed as part of this PR:
- Flags: I normalized the descriptions in all commands to end with a dot `.` for consistency.
- golint: Ran golint and fixed major code errors. I avoided bigger changes in type names such as `type name will be used as federate.FederateArtifacts by other packages, and that stutters; consider calling this Artifacts`. Perhaps it'd be good to include golint as part of the CI pipeline at some point.
- Fixed a typo in e2e/federate.go.
- cmd/controller-manager/app/leaderelection: log errors in leader election
- gofmt two files. (doc.go)

I hope you find it helpful.

Please, let me know if you want me to create a separate PR with any of the commits. I think the current changes are high in number but easy to identify.

### DIFF
--- a/pkg/controller/kubefedcluster/controller.go
+++ b/pkg/controller/kubefedcluster/controller.go
@@ -100,7 +100,7 @@ func newClusterController(config *util.ControllerConfig, clusterHealthCheckConfi
 	client := genericclient.NewForConfigOrDieWithUserAgent(kubeConfig, "cluster-controller")
 
 	cc := &ClusterController{
-		client:                   client,
+		client: client,
 		clusterHealthCheckConfig: clusterHealthCheckConfig,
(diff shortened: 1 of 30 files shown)


pkg/controller/kubefedcluster/controller.go
```diff
@@ -100,7 +100,7 @@ func newClusterController(config *util.ControllerConfig, clusterHealthCheckConfi
 	client := genericclient.NewForConfigOrDieWithUserAgent(kubeConfig, "cluster-controller")
 
 	cc := &ClusterController{
-		client:                   client,
+		client: client,
```
yamt: why?

PR #1187, opened 2020-02-21
chore: Use distroless base image instead of alpine

jimmidyson:

This PR uses distroless static non-root Docker base image using the standard `nobody` uid (`65534`). This simplifies the build in a (very) small way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@
 # Ignore binaries built by the makefile to avoid accidentally committing them
 /images/kubefed/hyperfed
 /images/kubefed/controller-manager
+/images/kubefed/kubefedctl
+/images/kubefed/webhook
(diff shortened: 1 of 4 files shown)


irfanurrehman:
@jimmidyson Thanks for doing this. However, I am not sure if this is desirable as of now. AFAIK the `distroless` image is more secure, but it makes it difficult to debug problems.
I would recommend keeping it as it is until we have a more concrete plan for a stable release.
You might have a better reasoning for this update. Please let me know if there is one.

jimmidyson:
That is a very good point. I've got so used to using ephemeral containers for debugging on my projects, but realise that it's alpha so not enabled by default. I'll close this for now and open an issue, but I would like to revisit this for beta/GA when people are running this in production.

Out of interest, do you have a list of common issues that you regularly troubleshoot for kubefed? Would be good to get together a doc for users perhaps?

PR #1189, opened 2020-02-22
Make the chart  Helm 3 compliant while maintaining compatibility with Helm 2

gregov:
With the release of Helm3, having kubefed working seamlessly with that version will encourage adoption of this project.

Fixes #

About the XXL size:
2467 new lines are due to the copy of charts/kubefed/templates/crds.yaml and charts/kubefed/charts/controllermanager/crds.yaml to charts/kubefed-crds/crds
Only change in them is the removal of first and last line, crds does not support templating in Helm3.

Chosen solution:
The approach here is the "Separate Charts" recommended by Helm v3 documentation
https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/

Please note:
Some minor warnings seems not avoidable at the moment but the end result is working.
Documentation has been updated accordingly.

### DIFF
--- a/charts/kubefed-crds/Chart.yaml
+++ b/charts/kubefed-crds/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v2
+description: Custom resources definitions for KubeFed helm chart
+name: kubefed-crds
+version: 0.0.2
--- a/charts/kubefed-crds/crds/controllermanager-crds.yaml
+++ b/charts/kubefed-crds/crds/controllermanager-crds.yaml
@@ -1,0 +1,1170 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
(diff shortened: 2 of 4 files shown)


gregov:
CLA signed

jimmidyson:
@gregov Thank you for this! Helm 3 support is definitely welcomed. My one concern with this is upgrade paths, both for existing users using helm 2 (although not sure the upgrade path works great there either tbh) and for helm 3 users. As we develop kubefed, evolving APIs as necessary, upgrading CRDs is going to be an important thing for users. Can you help me understand if this PR eases those concerns, both for helm 2 and helm 3?

gregov:
@jimmidyson CRD upgrades are not handled by Helm3 (see https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/)
If you need an upgrade, you will have to handle it manually.

jimmidyson:
@gregov What would be the steps in handling manually? Would we be able to just deploy the latest CRD chart to upgrade?

PR #1195, opened 2020-02-24
(Updated) #1083: use unstructured runtime.Object for more detailed reflector logging

### LINKED ISSUE #1083: Informer logging lacks type detail

sreis:

Rebases #1146 with master to bring it up to date. We've used that PR successfully to track down a couple of problems and didn't found any issues while using it.

Close #1146 

Fixes #1083

### DIFF
--- a/pkg/controller/util/resourceinformer.go
+++ b/pkg/controller/util/resourceinformer.go
@@ -23,24 +23,32 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
 )
 
-// NewManagedResourceInformer returns an unfiltered informer.
-func NewResourceInformer(client ResourceClient, namespace string, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
-	return newResourceInformer(client, namespace, triggerFunc, "")
+// NewResourceInformer returns an unfiltered informer.
+func NewResourceInformer(client ResourceClient, namespace string, apiResource *metav1.APIResource, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
+	return newResourceInformer(client, namespace, apiResource, triggerFunc, "")
 }
 
 // NewManagedResourceInformer returns an informer limited to resources
 // managed by KubeFed as indicated by labeling.
-func NewManagedResourceInformer(client ResourceClient, namespace string, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
+func NewManagedResourceInformer(client ResourceClient, namespace string, apiResource *metav1.APIResource, triggerFunc func(pkgruntime.Object)) (cache.Store, cache.Controller) {
 	labelSelector := labels.Set(map[string]string{ManagedByKubeFedLabelKey: ManagedByKubeFedLabelValue}).AsSelector().String()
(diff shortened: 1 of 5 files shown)


pkg/controller/util/resourceinformer.go
```diff
@@ -52,7 +60,7 @@ func newResourceInformer(client ResourceClient, namespace string, triggerFunc fu
 				return client.Resources(namespace).Watch(options)
 			},
 		},
-		nil, // Skip checks for expected type since the type will depend on the client
+		obj, // use a unstructured type with apiVersion / Kind populated for informer logging purposes
```
jimmidyson: nit: 
```suggestion
		obj, // use an unstructured type with apiVersion / kind populated for informer logging purposes
```

pkg/controller/util/resourceinformer.go
```diff
@@ -23,24 +23,32 @@ import (
...
 }
 
-func newResourceInformer(client ResourceClient, namespace string, triggerFunc func(pkgruntime.Object), labelSelector string) (cache.Store, cache.Controller) {
+func newResourceInformer(client ResourceClient, namespace string, apiResource *metav1.APIResource, triggerFunc func(pkgruntime.Object), labelSelector string) (cache.Store, cache.Controller) {
+	obj := &unstructured.Unstructured{}
+
+	if apiResource != nil {
+		gvk := schema.GroupVersionKind{Group: apiResource.Group, Version: apiResource.Version, Kind: apiResource.Kind}
+		obj.SetGroupVersionKind(gvk)
+		obj.SetNamespace(namespace)
```
jimmidyson: Is this used by anything?

sreis: Good question. I don't think it does but I'll do some manual test to confirm and report back.

xunpan: Could you explain more about this `SetNamespace()`?  If it does not needed for current logic, keep it here makes less sense.

jimmidyson: I've read through the code and this is unused afaict. It doesn't affect the logging so removing it is fine.

PR #1196, opened 2020-02-25
feat: add custom kubefed metrics

hectorj2f:

It adds kubefed custom metrics to the list of metrics exposed and serve by https://github.com/kubernetes-sigs/kubefed/pull/1193. It is a follow-up PR that allows to have a better view of the duration of some of the complex kubefed operations. 

Fixes #

related to: https://github.com/kubernetes-sigs/kubefed/issues/1194

Screen Shot 2020-03-15 at 4 00 44 PM

I created a home made grafana dashboard to testthese metrics:
Screen Shot 2020-03-15 at 4 39 43 PM

### DIFF
--- a/docs/keps/20200302-kubefed-metrics.md
+++ b/docs/keps/20200302-kubefed-metrics.md
@@ -44,7 +44,7 @@ and consumed from Kubefed to create dashboards and better understand this engine
 ## Motivation
 
 We aim to define a generic strategy on how to identify, consume and expose
-metrics from our core components in Kommander.
+custom Kubefed metrics.
 
 
 ### Goals
@@ -115,7 +115,7 @@ of the different propagated resources. The label `action` will hold the `create`
 
 Regarding cluster join/unjoin operations, these metrics are also convenient to register:
 
-* `joinedcluster_total`: a gauge metric that holds the number joined clusters.
+* `joined_cluster_total`: a gauge metric that holds the number joined clusters.
 
 * `join_cluster_duration_seconds`: this `histogram` metric holds the duration in seconds of the join cluster action.
 
--- a/docs/keps/20200302-kubefed-metrics.md
+++ b/docs/keps/20200302-kubefed-metrics.md
@@ -1,0 +1,148 @@
+---
+kep-number: 0
+short-desc: Kubefed Custom Metrics
+title: Kubefed Custom Metrics
+authors:
+  - "@hectorj2f"
+reviewers:
+  - "@jimmidyson"
+  - "@pmorie"
+  - "@xunpan"
+approvers:
+  - "@jimmidyson"
+  - "@pmorie"
+  - "@xunpan"
+editor: TBD
+creation-date: 2020-03-02
+last-updated: 2020-03-02
+status: provisional
+---
+
+# Kubefed Custom Metrics
+
+## Table of Contents
+
+* [Kubefed Custom Metrics](#kubefed-custom-metrics)
+  * [Table of Contents](#table-of-contents)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non\-Goals](#non-goals)
+  * [Proposals](#proposals)
+    * [Metrics](#metrics)
+    * [Risks and Mitigations](#risks-and-mitigations)
+  * [Graduation Criteria](#graduation-criteria)
+  * [Implementation History](#implementation-history)
+  * [Drawbacks](#drawbacks)
+  * [Infrastructure Needed](#infrastructure-needed)
+
+## Summary
+
+This document describes the different metrics and valuable data that could be exposed
+and consumed from Kubefed to create dashboards and better understand this engine.
+
+## Motivation
+
+We aim to define a generic strategy on how to identify, consume and expose
+metrics from our core components in Kommander.
+
+
+### Goals
+
+* Identify which metrics should be exposed from Kubefed if possible.
+* Define a set of Kubefed metrics that could be consumed by Prometheus tools.
+* Specify the type of each metric (e.g histogram, gauge, counter, summary).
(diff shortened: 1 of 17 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	kubefedClustersNotReadyCount = prometheus.NewCounter(
```
jimmidyson: These are counters so only monotonically increasing - don't we want to use a gauge for these metrics?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,118 @@
...
+	kubefedClustersOfflineCount = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "kubefedclusters_offline_total",
+			Help: "Number of offline kubefed clusters.",
+		},
+	)
+
+	functionDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "function_duration_seconds",
```
jimmidyson: This needs a better name - function is too generic and can apply to many different things, not necessarily just control loops. Maybe we can take inspiration from controller-runtime for this?

hectorj2f: That is why I pass the `function` as a way to distinguish between operations. Do you have any suggestion of a better name ?

jimmidyson: Take a look at https://github.com/kubernetes-sigs/controller-runtime/blob/c0438568a706ec61de31b92f4d76e7fb7e1007b9/pkg/internal/controller/metrics/metrics.go for inspiration perhaps?

hectorj2f: I'll update my PR this week. Thanks for the pointers

pkg/metrics/metrics.go
hectorj2f: Yeah, that sounds like a better approach. I will change it

hectorj2f:
@jimmidyson could you review again this PR ? I used a grafana dashboard to validate the metrics on a running cluster to which I joined two clusters

docs/keps/20200302-kubefed-metrics.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Summary
+
+This document describes the different metrics and valuable data that could be exposed
+and consumed from Kubefed to create dashboards and better understand this engine.
+
+## Motivation
+
+We aim to define a generic strategy on how to identify, consume and expose
+metrics from our core components in Kommander.
```
sreis: This should be worded not to reference a specific product.

pkg/controller/kubefedcluster/controller.go
```diff
@@ -257,6 +260,9 @@ func (cc *ClusterController) updateIndividualClusterStatus(cluster *fedv1b1.Kube
 	if err := cc.client.UpdateStatus(context.TODO(), cluster); err != nil {
 		klog.Warningf("Failed to update the status of cluster %q: %v", cluster.Name, err)
 	}
+
+	metrics.ClusterHealthStatusDurationFromStart(clusterHealthStatusStart)
```
sreis: Why didn't you use a defer like you did [here](https://github.com/kubernetes-sigs/kubefed/pull/1196/files#diff-0dfac9e46f796a7e5d47adce78da8e0eR133)?

pkg/controller/util/federated_informer.go
```diff
@@ -324,6 +325,8 @@ func (f *federatedInformerImpl) GetClientForCluster(clusterName string) (generic
 	f.Lock()
 	defer f.Unlock()
 
+	clientConnectionStart := time.Now()
```
sreis: Should we start timing before the lock is acquired?

docs/keps/20200302-kubefed-metrics.md
```diff
@@ -0,0 +1,148 @@
...
+
+For this metric, we could choose a single metric that will include additional labels
+to distinguish the different operations:
+
+* `dispatch_operation_duration_seconds`: this `histogram` metric holds the duration in seconds of the creation/update/deletion
+of the different propagated resources. The label `action` will hold the `create`, `update` and `delete` operations.
+
+Regarding cluster join/unjoin operations, these metrics are also convenient to register:
+
+* `joinedcluster_total`: a gauge metric that holds the number joined clusters.
```
sreis: To follow the same pattern as `join_cluster` and `unjoin_cluster`.
```suggestion
* `joined_cluster_total`: a gauge metric that holds the number joined clusters.
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,212 @@
...
+	kubefedClusterTotal = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "kubefedcluster_total",
+			Help: "Number of total kubefed cluster in a specific state.",
+		}, []string{"state", "cluster"},
+	)
+
+	joinedClusterTotal = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "joinedcluster_total",
```
sreis:
```suggestion
			Name: "joined_cluster_total",
```

pkg/controller/kubefedcluster/controller.go
hectorj2f: It wasn't intentional, I started using them in a way. I now adapted to run as a `defer` call

pkg/controller/util/federated_informer.go
hectorj2f: yes

hectorj2f:
Yes, @sreis we could easily adapt the code to add metrics to register the different type of errors.

PR #1200, opened 2020-03-08
Bump Helm version to 2.16.3 in download-binaries script

tcordeu:
The current version installed in the script (2.14.0) produces the following error when executing `helm init`:  **_Error: error installing: the server could not find the requested resource_**. This is due to an incompatibility with clusters with versions 1.16.0+, found on https://github.com/helm/helm/issues/6374.
To fix this, the Helm version is bumped to `2.16.3`.


xunpan:
Thank your for fixing it.

/lgtm
/approve

PR #1210, opened 2020-04-08
Use k8s-vendored extension as replacement of openapi-standard arbitrary type 

yue9944882:
kubernetes' CRD only support a subset of openapi specification which is (at least currently) excluding the openapi's standard form of expressing arbitrary types. we're supposed to use `x-kubernetes-preserve-unknown-fields` extension as a replacement otherwise the generated CRD from typeconfigs will always be falling back to structureless CRD which basically leaves the CRs unprotected by openapi's schema validation:

```
  - lastTransitionTime: "2020-04-08T12:19:52Z"
    message: '[spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[overrides].items.properties[clusterOverrides].items.properties[value].anyOf[0].type:
      Forbidden: must be empty to be structural, spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[overrides].items.properties[clusterOverrides].items.properties[value].anyOf[1].type:
      Forbidden: must be empty to be structural, spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[overrides].items.properties[clusterOverrides].items.properties[value].anyOf[2].type:
      Forbidden: must be empty to be structural, spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[overrides].items.properties[clusterOverrides].items.properties[value].anyOf[3].type:
      Forbidden: must be empty to be structural, spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[overrides].items.properties[clusterOverrides].items.properties[value].anyOf[4].type:
      Forbidden: must be empty to be structural, spec.versions[0].schema.openAPIV3Schema.properties[spec].properties[overrides].items.properties[clusterOverrides].items.properties[value].type:
      Required value: must not be empty for specified object fields, spec.versions[0].schema.openAPIV3Schema.type:
      Required value: must not be empty at the root]'
    reason: Violations
    status: "True"
    type: NonStructuralSchema
```

### DIFF
--- a/charts/kubefed/templates/crds.yaml
+++ b/charts/kubefed/templates/crds.yaml
@@ -39,12 +39,7 @@ spec:
                         path:
                           type: string
                         value:
-                          anyOf:
-                          - type: string
-                          - type: integer
-                          - type: boolean
-                          - type: object
-                          - type: array
+                          x-kubernetes-preserve-unknown-fields: true
                       required:
                       - path
                       type: object
@@ -128,6 +123,7 @@ spec:
           type: object
       required:
       - spec
+      type: object
   version: v1beta1
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
@@ -171,12 +167,7 @@ spec:
                         path:
                           type: string
                         value:
-                          anyOf:
-                          - type: string
-                          - type: integer
-                          - type: boolean
-                          - type: object
-                          - type: array
+                          x-kubernetes-preserve-unknown-fields: true
                       required:
                       - path
                       type: object
@@ -260,6 +251,7 @@ spec:
           type: object
       required:
       - spec
+      type: object
   version: v1beta1
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
@@ -303,12 +295,7 @@ spec:
                         path:
                           type: string
                         value:
-                          anyOf:
-                          - type: string
-                          - type: integer
-                          - type: boolean
-                          - type: object
-                          - type: array
+                          x-kubernetes-preserve-unknown-fields: true
                       required:
                       - path
                       type: object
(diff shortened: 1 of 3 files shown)


hectorj2f:
@yue9944882 could you check the tests? they are failing due to these new changes. Also, could you share an example where this would solve a current conversion ?

yue9944882:
> Also, could you share an example where this would solve a current conversion ?

@hectorj2f thx for reminding. [CRD pruning](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#pruning-versus-preserving-unknown-fields.) is the exact/recommended equivalence of representing arbitrary types in CRD's openapi schema.  i tried to find an example elsewhere but it didn't work probably b/c specifically defining arbitrary types in CRD is a rare case. but i did found a unit-test can be explaining the point here.. the current way of defining a member field as "arbitrary" isn't supported by kubernetes CRD, and it actually makes the whole generated CRD (e.g. FederatedDeployment) structureless. strutureless CRD at lease loses the following functionalities:

- Any existing openapi schema definition in the CRD is actually ingored/invalidated, hence the CRs will be merely working as a plain schema-free document
- Breaks [OpenApi publishing](https://github.com/kubernetes/kubernetes/blob/f52cbea102bb6ba9fe951cbf81e7dd4b48f95d9e/staging/src/k8s.io/apiextensions-apiserver/test/integration/objectmeta_test.go#L314)


https://github.com/kubernetes/kubernetes/blob/f52cbea102bb6ba9fe951cbf81e7dd4b48f95d9e/staging/src/k8s.io/apiextensions-apiserver/test/integration/objectmeta_test.go#L314

--- Thanks, Min

yue9944882:
/hold

until #1213

yue9944882:
@hectorj2f PTAL, thanks

xunpan:
#1230 depends on this fix

/lgtm

PR #1212, opened 2020-04-13
test: add test to validate the controller actions to keep the cluster data

hectorj2f:

It is a continuation of https://github.com/kubernetes-sigs/kubefed/pull/1201 which has been stale for a long period. I cherry picked their changes and added some tests. 

This test also cover other functions of the kubefedcluster controller.

Fixes #

### DIFF
--- a/pkg/controller/kubefedcluster/controller_test.go
+++ b/pkg/controller/kubefedcluster/controller_test.go
@@ -17,102 +17,19 @@ limitations under the License.
 package kubefedcluster
 
 import (
-	"context"
-	"path/filepath"
+	"reflect"
 	"testing"
 	"time"
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/ginkgo/extensions/table"
-	. "github.com/onsi/gomega"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	//schema "k8s.io/apimachinery/pkg/runtime/schema"
-	//"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/envtest"
-	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	fedapis "sigs.k8s.io/kubefed/pkg/apis"
 	"sigs.k8s.io/kubefed/pkg/apis/core/common"
 	fedv1b1 "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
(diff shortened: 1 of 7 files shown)


pkg/controller/kubefedcluster/controller_test.go
```diff
@@ -17,19 +17,102 @@ limitations under the License.
...
 	"time"
 
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	//schema "k8s.io/apimachinery/pkg/runtime/schema"
```
irfanurrehman, line 32: Please remove the commented lines here and at other places below.

pkg/controller/kubefedcluster/controller_test.go
```diff
@@ -38,71 +121,124 @@ func TestThresholdCheckedClusterStatus(t *testing.T) {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		fedCluster.Spec.APIEndpoint = "https://my.example.com:80/path/to/newendpoint"
+		err = k8sClient.Update(ctx, fedCluster)
+		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() bool {
+			_, found = cc.clusterDataMap[kc.Name]
+			return found
+		}, 30*time.Second).Should(BeTrue())
```
irfanurrehman, line 179: Recommend: Verify that the data updated (API Endpoint) equals what is updated above?

irfanurrehman:
@hectorj2f Thanks for doing this. The test itself looks quite allright. A few nits (see comments).

Additionally little bit of history for your reference:
We had the integration tests (the kind that your wrote here) earlier, where a local etcd and api server were used for individual tests; however we saw that the overall time taken to complete all tests against the PRs was crossing 1 hour. At some point we did fold all those integration tests into e2e to save time (currently the test run still takes close to 40+ mins).
The difference I am talking about is:
Test run with etcd and api-server in this PR: [7.507s](https://travis-ci.org/github/kubernetes-sigs/kubefed/builds/674631544#L3042) 
Pure unit test from an older build: [.021s](https://travis-ci.org/github/kubernetes-sigs/kubefed/builds/623483573#L2040)

Having said that, I do agree that the current test is not verifying the cluster state change and I think its ok to introduce the new test you have added, covering the missing piece as here (It would be worse in terms of time consumed to verify this scenario in e2e).
Can I also recommend that the current unit test be left behind as it is, and add the test that you have introduced separately under test/integration (If you think its easy enough to do)?
Again, really appreciate you doing this.

pkg/controller/kubefedcluster/controller_test.go
hectorj2f: 👍

hectorj2f:
@irfanurrehman I'd check how this can all be introduced without causing a lot of confusion or messing up the code structure. I certainly realized that there are many pieces of the kubefed's code that are not tested either because it escapes from the e2e radar or it is too complex to create mocks for all these tess.

hectorj2f:
@irfanurrehman I addressed your comments. I didn't move the new test to `test/integration` due to the amount of additional changes that will imply.

irfanurrehman:
Thanks @hectorj2f 
/lgtm

PR #1221, opened 2020-04-30
chore: make cluster creation work with kind v0.7.0

### LINKED ISSUE #1220: Cluster creation script doesn't work with recent kind version
**What happened**:
Tried to run the `/scripts/create-clusters.sh` script with kind v0.7.0, causing the script to fail after the first cluster has been created due to usage of a deprecated kind CLI command.

The clusters are up and running and the scripts returns success.

Run `./scripts/create-clusters.sh` using kind v0.7.0.

**Environment**:
- KubeFed version: `master`
- Others

/kind bug

makkes:
The command 'kind get kubeconfig-path' has been deprecated in kind
v0.6.0 and removed in v0.7.0 so the cluster creation script doesn't
work with the latter version. To account for this fact and other
changes (such as kind now respecting `KUBECONFIG` and falling back to
the same defaults that `kubectl` uses), I reworked some parts of the
script.

pre-commit now uses the kind imave v1.17.5 for tests.

Note: This change makes the scripts fail to work with
      kind version < v0.7.0.

Fixes #1220

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -27,12 +27,10 @@ CONFIGURE_INSECURE_REGISTRY_HOST="${CONFIGURE_INSECURE_REGISTRY_HOST:-}"
 CONFIGURE_INSECURE_REGISTRY_CLUSTER="${CONFIGURE_INSECURE_REGISTRY_CLUSTER-y}"
 CONTAINER_REGISTRY_HOST="${CONTAINER_REGISTRY_HOST:-172.17.0.1:5000}"
 NUM_CLUSTERS="${NUM_CLUSTERS:-2}"
-OVERWRITE_KUBECONFIG="${OVERWRITE_KUBECONFIG:-}"
 KIND_IMAGE="${KIND_IMAGE:-}"
 KIND_TAG="${KIND_TAG:-}"
 docker_daemon_config="/etc/docker/daemon.json"
 containerd_config="/etc/containerd/config.toml"
-kubeconfig="${HOME}/.kube/config"
 OS=`uname`
 
 function create-insecure-registry() {
@@ -92,7 +90,7 @@ EOF
 "
       ;;
     $containerd_config)
-     echo "sed -i '/\[plugins.cri.registry.mirrors\]/a [plugins.cri.registry.mirrors."\"${CONTAINER_REGISTRY_HOST}\""]\nendpoint = ["\"http://${CONTAINER_REGISTRY_HOST}\""]' ${containerd_config}"
+     echo 'echo -e "[plugins.\"io.containerd.grpc.v1.cri\".registry.mirrors.\"'${CONTAINER_REGISTRY_HOST}'\"]\n  endpoint = [\"http://'${CONTAINER_REGISTRY_HOST}'\"]\n" >> '${containerd_config}
      ;;
     *)
      echo "Sorry, config insecure registy is not supported for $config_file"
@@ -123,13 +121,6 @@ function create-clusters() {
   echo "Waiting for clusters to be ready"
   check-clusters-ready ${num_clusters}
 
-  # TODO(font): kind will create separate kubeconfig files for each cluster.
-  # Remove once https://github.com/kubernetes-sigs/kind/issues/113 is resolved.
-  if [[ "${OVERWRITE_KUBECONFIG}" ]]; then
-    kubectl config view --flatten > ${kubeconfig}
-    unset KUBECONFIG
-  fi
-
   if [[ "${CONFIGURE_INSECURE_REGISTRY_CLUSTER}" ]]; then
     # TODO(font): Configure insecure registry on kind host cluster. Remove once
     # https://github.com/kubernetes-sigs/kind/issues/110 is resolved.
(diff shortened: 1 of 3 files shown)


makkes:
tests are failing for a reason that's not totally obvious to me. I'll have to have a further look.

irfanurrehman:
@makkes Absolutely appreciate you doing this. Does the same run fine locally for you ?

makkes:
> @makkes Absolutely appreciate you doing this. Does the same run fine locally for you ?

@irfanurrehmann, thanks for the feedback. 

To answer your question: Not yet, as I had to figure out which tests are being run by Travis. Now I got everything set up and I can reproduce the failure. Fix incoming. This is slightly larger than I thought initially. 😌

makkes:
@irfanurrehman, all tests pass locally now. 🤦‍♂️ 🤣

hectorj2f:
@makkes nice, I will give it a try myself.

scripts/create-clusters.sh
```diff
@@ -141,27 +132,23 @@ function create-clusters() {
...
-  local kubeconfig_path="$(kind get kubeconfig-path --name cluster${i})"
-  export KUBECONFIG="${KUBECONFIG:-}:${kubeconfig_path}"
-
   if [ "$OS" != "Darwin" ];then
     # Set container IP address as kube API endpoint in order for clusters to reach kube API servers in other clusters.
-    kind get kubeconfig --name "cluster${i}" --internal >${kubeconfig_path}
+    OLD_KUBECONFIG=${KUBECONFIG:-}
+    export KUBECONFIG=$(mktemp)
+    kind get kubeconfig --name "cluster${i}" --internal > ${KUBECONFIG}
+    SERVER=$(kubectl config view -o jsonpath='{.clusters[0].cluster.server}')
```
hectorj2f, line 140: @makkes should not this be `[${i}]` instead of `[0]` ?

scripts/download-e2e-binaries.sh
```diff
@@ -35,7 +35,7 @@ dest_dir="${root_dir}/bin"
 mkdir -p "${dest_dir}"
 
 # kind
-kind_version="v0.4.0"
+kind_version="v0.7.0"
```
hectorj2f, line 38: perhaps we could try using https://github.com/kubernetes-sigs/kind/releases/tag/v0.8.1

scripts/pre-commit.sh
```diff
@@ -150,8 +150,8 @@ run-unit-tests
 echo "Downloading e2e test dependencies"
 ./scripts/download-e2e-binaries.sh
 
-CREATE_INSECURE_REGISTRY=y CONFIGURE_INSECURE_REGISTRY_HOST=y OVERWRITE_KUBECONFIG=y \
-    KIND_TAG="v1.15.0" ./scripts/create-clusters.sh
+CREATE_INSECURE_REGISTRY=y CONFIGURE_INSECURE_REGISTRY_HOST=y \
+    KIND_TAG="v1.17.0" ./scripts/create-clusters.sh
```
hectorj2f, line 154: nit: `KIND_TAG="v1.17.5`

scripts/create-clusters.sh
makkes: @hectorj2f, good question. The `[0]` is actually intentional since `kind get kubeconfig` will get one and only one kubeconfig (for cluster `cluster${i}`, thus it will only contain one entry in the `clusters` array and this is the one that we want to get the server address from. We're doing all of this in the temporary kubeconfig file created two lines above which is deleted as soon as we have the internal server address.

scripts/download-e2e-binaries.sh
makkes: I was thinking about that, too, but then refrained from it because I guess that the adoption rate of 0.8.x is still pretty low. wdyt?

The best solution would be to iterate through multiple versions, like 0.7.x and 0.8.x but I'd prefer to do that in another PR. :-)

makkes: Plus: v0.8.0 introduced some breaking changes that I'd have to adapt to additional to the changes I'm already tackling in this PR. So it seems to be a good idea to go for 0.7 for now.

makkes:
I created #1223 to track kind v0.8 support.

makkes:
@hectorj2f, I squashed all commits into one.

irfanurrehman:
@makkes @hectorj2f. Apologies for not getting to this earlier. I had a glance at this over the weekend and was supposed to spend time on this today. Anywhichways, I am curious if you did verify that the scripts work on mac?

hectorj2f:
@irfanurrehman I tested on Linux where I have my setup. But I am not sure if it ever worked on OSX out of the box.

irfanurrehman:
> @irfanurrehman I tested on Linux where I have my setup. But I am not sure if it ever worked on OSX out of the box.

@hectorj2f you are right, it did not work out of the box for OSX. I was hoping that with the updated kind version, it might.

hectorj2f:
@irfanurrehman that might be a simple bug fix. I will take a look at it.

makkes:
@irfanurrehman, @hectorj2f, making `create-clusters.sh` work with an insecure registry on macOS might be a slightly bigger effort as Docker for Mac runs inside of a VM so you can't easily change the daemon cfg.

irfanurrehman:
> @irfanurrehman, @hectorj2f, making `create-clusters.sh` work with an insecure registry on macOS might be a slightly bigger effort as Docker for Mac runs inside of a VM so you can't easily change the daemon cfg.

@makkes we use `docker exec` to update the config and I think it behaves the same on OSX irrespective of where the container is actually running (Am I missing something?). 

The tricky part to take care however is that the apiserver endpoint differs depending on where you are accessing the cluster from. If you access one kind cluster from within another, you need to use the container internal ip (which cannot be reached from local machine)  whereas when you access it from local machine (for example the scripts and what is stored in the kubeconfig) it is `127.0.0.1:<some unique port>`.  This is different from linux and workaround(s) becomes a little hacky. 

Having said that, I think we can live with what we currently have.. :)

makkes:
> we use docker exec to update the config

@irfanurrehman, this is true for the kind cluster but `create-clusters.sh` also changes the host's docker daemon config if `CONFIGURE_INSECURE_REGISTRY_HOST` is set to true:
https://github.com/kubernetes-sigs/kubefed/blob/ab94a922ef079329f86fa80b1ae220cb5c3cb15d/scripts/create-clusters.sh#L70
which is the case in `pre-commit.sh`:
https://github.com/kubernetes-sigs/kubefed/blob/ab94a922ef079329f86fa80b1ae220cb5c3cb15d/scripts/pre-commit.sh#L153

irfanurrehman:
Aah yes! I was missing that part. As I said earlier, I think we can live with what we currently have.. :)

irfanurrehman:
Meanwhile thanks @makkes for doing this!

PR #1230, opened 2020-05-14
WIP setting spec.preserveUnknownFields: false

### LINKED ISSUE #986: Generated federated types should set preserveUnknownFields to false
An issue like #984 is the result of Kube not stripping unknown fields when accepting a request to add or update a custom resource. Kube 1.15 will add support for stripping unknown fields by setting `spec.preserveUnknownFields: false`, and this should be supported for generated federated types when running on kube 1.15.

xunpan:

Fixes #986

### DIFF
--- a/charts/kubefed/templates/crds.yaml
+++ b/charts/kubefed/templates/crds.yaml
@@ -11,6 +11,7 @@ spec:
   names:
     kind: FederatedClusterRole
     plural: federatedclusterroles
+  preserveUnknownFields: false
   scope: Cluster
   subresources:
     status: {}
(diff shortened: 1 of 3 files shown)


jimmidyson:
Won't this break stuff because `spec.template` of all federated resources are just marked as `object` so nested fields will be dropped?

jimmidyson:
Also we should migrate to `apiextensions.k8s.io/v1` which doesn't have this field iirc and requires a different way to specify this.

xunpan:
Agreed. This should be handled when we upgrade `CRD` to `apiextensions.k8s.io/v1`. 

/close

PR #1232, opened 2020-05-15
docs: adding documentation to add an array element using overrides

### LINKED ISSUE #1224: overrides `op: add` seems not to work as expected for array values
**What happened**:

Let say I have:
```
spec:
  template:
    spec:
      template:
        spec:
          initContainers:
            - name: myinitcontainer
              args:
                - --logtostderr
```

Adding an override on an array key using `op: add`:

```
spec:
  overrides:
    - clusterName: mycluster
      clusterOverrides:
        - path: "/spec/template/spec/initContainers/0/args"
          op: add
          value:
            - --init
  template:
    spec:
      template:
        spec:
          initContainers:
            - name: myinitcontainer
              args:
                - --logtostderr
```

Created this object in `mycluster`:
```
spec:
    initContainers:
      - name: myinitcontainer
         args:
          - --init
```

Merging the two arguments in the array, thus creating this object in `mycluster` instead:
```
spec:
    initContainers:
      - name: myinitcontainer
         args:
          - --init
          - --logtostderr
```

Apply the earlier described manifests.

I don't think so, if anything missing, please comment :).

**Environment**:
- KubeFed version: `v0.3.0`

/kind bug

ArchangelX360:

Adding a value to an array could be confusing for users, an example in the docs could alleviate the confusion.

Fixes #1224

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -688,6 +688,11 @@ spec:
         # Ensure an annotation with key "foo" does not exist
         - path: "/metadata/annotations/foo"
           op: "remove"
+        # Adds an argument `-q` at index 0 of the args list
+        # this will obviously shift the existing arguments, if any
+        - path: "/spec/template/spec/containers/0/args/0"
+          op: "add"
+          value: "-q"
 ```
 
 ### Overriding retained fields


xunpan:
@ArchangelX360 
It is better to have a valid option for the case to me. `bar` seems not valid for `nginx`.

ArchangelX360:
@xunpan agreed good catch, [`-q`](http://nginx.org/en/docs/switches.html) should be.

Issue #1205, opened 2020-03-18
Delete select federated resources upon unjoin


makkes:
**What would you like to be added**:

I'd like to be able to configure certain federated resources in such a way that kubefed deletes them from the managed cluster when the cluster is unjoined.

**Why is this needed**:

It might be desirable in certain circumstances to leave a clean cluster behind after it is unjoined from the management cluster, clearing it of all the resources created when it was joined to the management cluster.

/kind feature

makkes:
This is basically a duplicate of #583 which was closed. However, in my opinion it makes sense, to re-assess and discuss this proposal again.

hectorj2f:
Perhaps it could be possible by adding a specific annotation similar to `kubefed.io/orphan` to achieve this goal.

PR #1233, opened 2020-05-20
kep: cleanup fed resources

jr0d:
This PR introduces a proposal for cleaning up federated resources.

Fixes #

We've been using kubefed for some time and have found that removal of federated resources on unjoin is a very common feature request. This KEP provides one possible implementation, but there are certainly other ways to achieve our goals.

### DIFF
--- a/docs/keps/20200504-cleanup-fed-resources.md
+++ b/docs/keps/20200504-cleanup-fed-resources.md
@@ -1,0 +1,113 @@
+---
+kep-number: 0
+short-desc: Cleanup federated resource on unjoin
+title: Cleanup Federated Resource
+authors:
+  - "@jr0d"
+reviewers:
+  - "@jimmidyson"
+  - "@hectorj2f"
+approvers:
+  - "@jimmidyson"
+editor: TBD
+creation-date: 2020-05-18
+last-updated: 2020-05-04
+status: provisional
+---
+
+# Cleanup federated resources
+
+## Table of Contents
+
+* [Cleanup Federated Resources](#cleanup-federated-resources)
+  * [Table of Contents](#table-of-contents)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [User Story](#user-stroy)
+    * [Goals](#goals)
+  * [Proposals](#proposal)
+    * [Implementation Details](#implementation-details)
+      * [Best Effort Strategy](#best-effort-strategy)
+      * [Required Strategy](#required-strategy)
+    * [Addition ClusterConditionTypes (optional)](#additional-clusterconfigtypes-optional)
+      * [Unjoining ClusterConfigurationType](#unjoining-clusterconfigurationtype)
+      * [Failed ClusterConfigurationType](#failed-clusterconfigurationtype)
+  * [Risks](#risks)
+    * [Adding Cluster Conditions](#adding-cluster-conditions)
+
+This document proposes a mechanism to specify that a federated resource should be removed from a managed cluster when leaving federation.
+
+## Motivation
+
+In practice, federation is often used to join clusters to some higher level centralized management plane, identity management for example. To achieve this, federation is used to propagate a configuration state to clusters which overrides an existing configuration state. When a cluster is unjoined, the federated configuration becomes invalid requiring the operator to manually remove the conflicting resources.
+
+### User Story
+
+As a cluster operator using kubefed, I would like to create resources which will automatically be removed from managed clusters during unjoin.
+
+### Goals
+
+* Define how users can indicate that a federated resource should be removed when a cluster leaves federation
+* Define removal strategies
+
+## Proposals
+
+The APIResource struct should be updated to contain a `RemoveStrategy` member. `RemoveStrategy` will initially support two values, `BestEffort` and `Required`.
+
+  * The `BestEffort` strategy will **not** prevent a cluster from unjoining when there are errors removing the resource.
+  * The `Required` strategy will halt the unjoin process and set the KubefedCluster resource in a failed state when encountering errors. Unjoining is blocked until all resources with the `Required` `RemoveStrategy` have been removed successfully.
+
+By default, `RemoveStrategy` will be nil on all federated types. A user must explicitly set a removal strategy on the resource they are creating. Types without a removal strategy will not be removed.
+
+
+### Implementation details
+
+* Update APIResource to contain a new member:
+  ```
+    // RemoveStrategy indicates what strategy should be employed when removing
+    // federated resources during nnjoin
+    RemoveStrategy *RemoveStrategy `json:"removeStrategy,omitempty"`
+  ```
+
+  ```
+    type RemoveStrategy string
+
+    const (
+      RemoveStrategyBestEffort RemoveStrategy = "BestEffort"
+      RemoveStrategyRequired   RemoveStrategy = "Required"
+    )
+  ```
+* When labels are applied to federated resources (ApplyOverrides), add the following label if RemoveStrategy is !nil:
+  * `kubefed.io/remove: best-effort || required`
+* During unjoin, query for managed (`kubefed.io/managed: true`) resources with the `kubefed.io/remove` label: `[{ key: "kubefed.io/remove", operator: "Exists" }]`
+
+#### Best Effort Strategy
+
+Errors or timeouts that occur while removing resources with the best effort removal strategy should be logged but should not halt the unjoining process.
+
+#### Required Strategy
+
+Resources with the "required" removal strategy must be removed successfully on unjoin. After removing the resource, the client must verify that the resource is no longer present. If a resource cannot be removed:
+
+* UnjoinCluster should return a formatted error containing a list of resources which could not be removed
+* A ClusterCondition should be added to the KubefedCluster status with a message indicating which resources could not be removed
+* (optional) Set the ConditionType to `Failed`
+
+
+### Additional ClusterConditionTypes (optional)
+
+Presently, there are only two ClusterConditionTypes, `Ready` and `Offline`. Given the `required` remove strategy, a `KubefedCluster` may end up in a state where the cluster is `Ready` but is in the process of "Unjoining". Adding additional conditions will make it easier to communicate federation life cycle status to the cluster operator.
+
+#### Unjoining ClusterConditionType
+
+In continuation of the existing nomenclature, the `Unjoining` condition should be set on a cluster when the operated initiates `UnjoinCluster`.
+
+#### Failed ClusterConditionType
+
+In the event of an error, such as failure to remove a resource with a `required` remove strategy, the cluster should enter a `Failed` condition.
+
+## Risks
+
+### Adding Cluster Conditions
+
+Adding additional cluster states (conditions) may be outside the scope of this proposal. It is not necessary to achieve the goals of the proposal, but does aid in operator experience. At the same time, it also increases implementation complexity.


docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+### Goals
+
+* Define how users can indicate that a federated resource should be removed when a cluster leaves federation
+* Define removal strategies
+
+## Proposals
+
+The APIResource struct should be updated to contain a `RemoveStrategy` member. `RemoveStrategy` will initially support two values, `BestEffort` and `Required`.
+
+  * The `BestEffort` strategy will **not** prevent a cluster from unjoining when there are errors removing the resource.
```
hectorj2f, line 57: I'd rather remove the resources and fail if something went wrong than a halfway solution. After all, I expect users to choose between two strategies: no cleanup (current impl) and the required one.

hectorj2f: Can you justify the need for this new deletion behavior ?

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+  * The `Required` strategy will halt the unjoin process and set the KubefedCluster resource in a failed state when encountering errors. Unjoining is blocked until all resources with the `Required` `RemoveStrategy` have been removed successfully.
+
+By default, `RemoveStrategy` will be nil on all federated types. A user must explicitly set a removal strategy on the resource they are creating. Types without a removal strategy will not be removed.
+
+
+### Implementation details
+
+* Update APIResource to contain a new member:
```
hectorj2f, line 65: where would this new property be used ?

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Best Effort Strategy
+
+Errors or timeouts that occur while removing resources with the best effort removal strategy should be logged but should not halt the unjoining process.
+
+#### Required Strategy
+
+Resources with the "required" removal strategy must be removed successfully on unjoin. After removing the resource, the client must verify that the resource is no longer present. If a resource cannot be removed:
+
+* UnjoinCluster should return a formatted error containing a list of resources which could not be removed
```
hectorj2f, line 92: I believe the unjoincluster operation should be blocked until all the resources are deleted as until now.
Can you detail where you will call the function in charge of the deletion in `func UnjoinCluster...` of kubefedctl ?

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+* A ClusterCondition should be added to the KubefedCluster status with a message indicating which resources could not be removed
```
hectorj2f, line 93: I assume you would need to use the `DeletionTimedOut` and `DeletionFailed` as status of the kubefedcluster if the unjoin failed.

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+* (optional) Set the ConditionType to `Failed`
+
+
+### Additional ClusterConditionTypes (optional)
+
+Presently, there are only two ClusterConditionTypes, `Ready` and `Offline`. Given the `required` remove strategy, a `KubefedCluster` may end up in a state where the cluster is `Ready` but is in the process of "Unjoining". Adding additional conditions will make it easier to communicate federation life cycle status to the cluster operator.
+
+#### Unjoining ClusterConditionType
+
+In continuation of the existing nomenclature, the `Unjoining` condition should be set on a cluster when the operated initiates `UnjoinCluster`.
```
hectorj2f, line 103: Please, explain how the process would work. It would then exist a condition `UnjoinFailed`.

docs/keps/20200504-cleanup-fed-resources.md
hectorj2f: I just see two options.

docs/keps/20200504-cleanup-fed-resources.md
hectorj2f: I assume this property to be set to the kubefedcluster or per federated resource. I think I'd go with a simple approach and apply per kubefed cluster so you ensure to delete all the related objects.

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+    type RemoveStrategy string
+
+    const (
+      RemoveStrategyBestEffort RemoveStrategy = "BestEffort"
+      RemoveStrategyRequired   RemoveStrategy = "Required"
+    )
+  ```
+* When labels are applied to federated resources (ApplyOverrides), add the following label if RemoveStrategy is !nil:
+  * `kubefed.io/remove: best-effort || required`
+* During unjoin, query for managed (`kubefed.io/managed: true`) resources with the `kubefed.io/remove` label: `[{ key: "kubefed.io/remove", operator: "Exists" }]`
```
hectorj2f, line 82: 👍

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
+---
+kep-number: 0
+short-desc: Cleanup federated resource on unjoin
+title: Cleanup Federated Resource
```
makkes, line 4:
```suggestion
title: Cleanup Federated Resources
```

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+
+* Define how users can indicate that a federated resource should be removed when a cluster leaves federation
+* Define removal strategies
+
+## Proposals
+
+The APIResource struct should be updated to contain a `RemoveStrategy` member. `RemoveStrategy` will initially support two values, `BestEffort` and `Required`.
+
+  * The `BestEffort` strategy will **not** prevent a cluster from unjoining when there are errors removing the resource.
+  * The `Required` strategy will halt the unjoin process and set the KubefedCluster resource in a failed state when encountering errors. Unjoining is blocked until all resources with the `Required` `RemoveStrategy` have been removed successfully.
```
makkes, line 58: "failed state" doesn't sound like the intended wording here, esp. given the new condition types outlined below. From a user's perspective "failed" sounds like something definite. However, here "failed" is only a temporary condition.

```suggestion
  * The `Required` strategy will halt the unjoin process and set the KubefedCluster resource in a temporary failed state when encountering errors. Unjoining is blocked until all resources with the `Required` `RemoveStrategy` have been removed successfully.
```

docs/keps/20200504-cleanup-fed-resources.md
makkes: I agree. It would probably best to list the possible scenarios:

Ready => Unjoining => Offline
Ready => Unjoining => UnjoinFailed

In the latter case we should also specify that an appropriate reason/message is set on the condition.

RainbowMango:
It's a great proposal, thanks @jr0d.
/cc

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
+---
+kep-number: 0
+short-desc: Cleanup federated resource on unjoin
```
irfanurrehman, line 3: s/resource/resources

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+authors:
+  - "@jr0d"
+reviewers:
+  - "@jimmidyson"
+  - "@hectorj2f"
+approvers:
+  - "@jimmidyson"
+editor: TBD
+creation-date: 2020-05-18
+last-updated: 2020-05-04
```
irfanurrehman, line 14: these dates seem to be reversed.

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+    * [Implementation Details](#implementation-details)
+      * [Best Effort Strategy](#best-effort-strategy)
+      * [Required Strategy](#required-strategy)
+    * [Addition ClusterConditionTypes (optional)](#additional-clusterconfigtypes-optional)
+      * [Unjoining ClusterConfigurationType](#unjoining-clusterconfigurationtype)
+      * [Failed ClusterConfigurationType](#failed-clusterconfigurationtype)
+  * [Risks](#risks)
+    * [Adding Cluster Conditions](#adding-cluster-conditions)
+
+This document proposes a mechanism to specify that a federated resource should be removed from a managed cluster when leaving federation.
```
irfanurrehman, line 38: s/that a/how a/

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Motivation
+
+In practice, federation is often used to join clusters to some higher level centralized management plane, identity management for example. To achieve this, federation is used to propagate a configuration state to clusters which overrides an existing configuration state. When a cluster is unjoined, the federated configuration becomes invalid requiring the operator to manually remove the conflicting resources.
```
irfanurrehman, line 42: I think you should use `kubefed` as the disambiguated name and not `federation`. We did update the term `federation` to `kubernetes cluster federation` and thats what is reflected across the documentation and code in this repo.
So it would be a `federated resource` or `federated cluster` but `kubefed control plane` or `join kubefed`, etc.
This applies to the whole document.

docs/keps/20200504-cleanup-fed-resources.md
irfanurrehman: Halting the unjoin process (or blocking the unjoin command) might not be desirable in many scenarios; for the obvious reason of it taking too long in some cases. I also understand that this is the simpler implementation and we can start with that. However, please leave room for an option where this cleanup could be asynchronous too, via the kubefed control plane as garbage cleanup (for example making it clear in the command help/documentation when this is implemented).

docs/keps/20200504-cleanup-fed-resources.md
irfanurrehman: I agree that it makes sense to have this as a global default also. This could be a property on both the federated cluster and the federated resource, with some strategy like clusters property would be used if available, else look for the one on the resource.

docs/keps/20200504-cleanup-fed-resources.md
```diff
@@ -0,0 +1,113 @@
...
+
+* UnjoinCluster should return a formatted error containing a list of resources which could not be removed
+* A ClusterCondition should be added to the KubefedCluster status with a message indicating which resources could not be removed
+* (optional) Set the ConditionType to `Failed`
+
+
+### Additional ClusterConditionTypes (optional)
+
+Presently, there are only two ClusterConditionTypes, `Ready` and `Offline`. Given the `required` remove strategy, a `KubefedCluster` may end up in a state where the cluster is `Ready` but is in the process of "Unjoining". Adding additional conditions will make it easier to communicate federation life cycle status to the cluster operator.
+
```
irfanurrehman, line 100: This is not only for the cluster operator but also for the sync controller to stop reconciling the resources for this cluster.
I think this (or the above portion) needs detail as to how the whole operation would be performed and different branches it can take (failure/partial failure/pass).

This is also partly why I did mention the possibility of an asyncronous mechanism by a reconcile loop in the control plane. I think the asynchronous mechanism can be better managed and can have retries (next reconcile) on failures.

makkes:
Related issue: #1205

PR #1237, opened 2020-06-19
docs: add kep about the status of the federated resources

hectorj2f:

It proposes a Kep to improve the visibility of the health of the federated resources using the current status field of the federated resources.

Nowadays users have to connect to the kubefed clusters to be aware if the federated
resources (deployments) are healthy or not across all the kubefed clusters.

Fixes #

### DIFF
--- a/docs/keps/20200619-federated-resource-status.md
+++ b/docs/keps/20200619-federated-resource-status.md
@@ -1,0 +1,408 @@
+---
+kep-number: 20200619
+short-desc: Kubefed -- Federated Resources Status
+title: Kubefed -- Federated Resources Status
+authors:
+  - "@hectorj2f"
+reviewers:
+  - "@irfan"
+  - "@hectorj2f"
+  - "@jimmidyson"
+  - "@pmorie"
+approvers:
+- "@irfan"
+- "@jimmidyson"
+- "@pmorie"
+editor: TBD
+creation-date: 2020-06-19
+last-updated: 2020-06-19
+status: provisional
+---
+
+# Kubefed -- Federated Resources Status
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+<!-- /toc -->
+
+## Summary
+
+Kubefed needs to improve its definition of status for the federated resources.
+Users lack of proper visibility over the status of the federated
+resources. For instance, if you federated a deployment the federated status should
+report if the deployment failed or error at any time.
+
+A federated resource only reflects the status of propagation actions, but it doesn't
+reflect the status if whether the resource is running or failed.
+
+## Motivation
+
+Nowadays users have to connect to the kubefed clusters to be aware if the federated
+resources are healthy or not across clusters.
+
+### Goals
+
+* Quickly identify unhealthy federated resources by relying on the status of the federated resources.
+* Improve the troubleshooting of failures when propagating resources.
+
+
+## Proposal
+
+Kubefed reports a limited set of states for the federation of resources.
+
+
+```go
+CreationTimedOut     PropagationStatus = "CreationTimedOut"
+UpdateTimedOut       PropagationStatus = "UpdateTimedOut"
+DeletionTimedOut     PropagationStatus = "DeletionTimedOut"
+LabelRemovalTimedOut PropagationStatus = "LabelRemovalTimedOut"
+
+AggregateSuccess       AggregateReason = ""
+ComputePlacementFailed AggregateReason = "ComputePlacementFailed"
+NamespaceNotFederated  AggregateReason = "NamespaceNotFederated"
+
+PropagationConditionType ConditionType = "Propagation"
+```
+
+However the current federated resource properties don't help to track the status of the deployed resources in the kubefed clusters.
+
+The idea is to extend current GenericFederatedStatus with the Status of the resources:
+
+```go
+
+type GenericFederatedStatus struct {
+  	ObservedGeneration  int64                  `json:"observedGeneration,omitempty"`
+  	Conditions          []*GenericCondition    `json:"conditions,omitempty"`
+  	Clusters            []GenericClusterStatus `json:"clusters,omitempty"`
+}
+
+type GenericFederatedResource struct {
+	metav1.TypeMeta                `json:",inline"`
+	metav1.ObjectMeta              `json:"metadata,omitempty"`
+
+	Status *GenericFederatedStatus `json:"status,omitempty"`
+}
+
+```
+
+Nowadays `Conditions` hold the status of the federated actions (aka propagation status).
+In other words, it defines the conditions of the propagation status for a resource.
+
+```yaml
+- apiVersion: types.kubefed.io/v1beta1
+  kind: FederatedDeployment
+  metadata:
+    finalizers:
+    - kubefed.io/sync-controller
+    generation: 1
+    name: asystem
+    namespace: asystem
+    resourceVersion: "70174497"
+  spec:
+    placement:
+      clusters:
+      - name: cluster3
+      - name: cluster2
+      - name: cluster1
+    template:
+      metadata:
+        labels:
+          app: nginx
+      spec:
+        replicas: 3
+        selector:
+          matchLabels:
+            app: nginx
+        template:
+          metadata:
+            labels:
+              app: nginx
+          spec:
+            containers:
+            - image: nginx
+              name: nginx
+  status:
+    conditions:
+    - lastTransitionTime: "2020-05-25T19:47:59Z"
+      lastUpdateTime: "2020-05-25T19:47:59Z"
+      status: "True"
+      type: Propagation
+```
+
+`status.conditions` reports the latest status which defines the state of the propagation.
+Obviously it is not necessary to report all the clusters for which the propagation when
+successful.
+
+The intention in this proposal is to extend the current available `Conditions` to
+hold the status of the federated resources, e.g Ready, NotReady.
+
+The status of the federated resources should determine whether the resources satisfy
+a `Ready` condition, and otherwise report their error status.
+To do so, this property reports the status of the federated resources in their
+target clusters whenever a `ReadyCondition` is not satisfied.
+This condition would need to be identified per type or by the usage of an interface
+`IsReady` that determines this value per type of resource.
+By doing so, we ensure the `Conditions` property shows the status of only unhealthy
+resources.
+
+If we re-use the example from above and imagine a scenario where this `FederatedDeployment` resource remained `Ready=True` in two clusters, but crashed in `cluster3`.
+The value of `status.conditions` should reflect the new state for that specific cluster.
+
+```yaml
+- apiVersion: types.kubefed.io/v1beta1
+  kind: FederatedDeployment
+  metadata:
+    finalizers:
+    - kubefed.io/sync-controller
+    generation: 1
+    name: asystem
+    namespace: asystem
+    resourceVersion: "70174497"
+  spec:
+    placement:
+      clusters:
+      - name: cluster2
+      - name: cluster1
+    template:
+      metadata:
+        labels:
+          app: nginx
+      spec:
+        replicas: 3
+        selector:
+          matchLabels:
+            app: nginx
+        template:
+          metadata:
+            labels:
+              app: nginx
+          spec:
+            containers:
+            - image: nginx
+              name: nginx
+  status:
+    clusters:
+    - name: "cluster3"
+      status: "ReplicaFailure"
+    conditions:
+    - lastTransitionTime: "2020-05-25T20:23:59Z"
+      lastUpdateTime: "2020-05-25T20:23:59Z"
+      status: "True"
+      type: "NotReady"
+```
+
+The value of `spec.conditions` contains a `NotReady` condition type, that is the
+result of checking the status of that remote `Deployment` in the target clusters.
+The value of `status.clusters.status` can be extracted from the `"ReplicaFailure"`
+status of the `Deployment` and so reused for visibility.
+
+Likewise, the status of the `FederatedDeployment` remains `Ready=True` in the rest
+of clusters: `cluster1` and `cluster2`.
+
+If a federated resource does not have a status field, a successful creation/update would
+reflect its readiness. Then the `ReadyCondition` would be satisfied by its creation.
+For these resources the value of `Conditions` would rely on the value of the propagation state of that resource.
+An example could be a `ClusterRole` resource that doesn't have a status property, but
+kubefed nowadays reports if the propagation of that resource worked.
+
+```yaml
+apiVersion: types.kubefed.io/v1beta1
+kind: FederatedClusterRole
+metadata:
+  name: test-clusterrole
+spec:
+  template:
+    rules:
+    - apiGroups:
+      - '*'
+      resources:
+      - '*'
+      verbs:
+      - '*'
+  placement:
+    clusters:
+    - name: cluster2
+    - name: cluster1
+status:
+  conditions:
+  - lastTransitionTime: "2020-05-25T19:47:59Z"
+    lastUpdateTime: "2020-05-25T19:47:59Z"
+    status: "True"
+    type: Propagation
+```
+
+However, there is a problem with this approach, the status schema varies based on the custom
+resource. Unfortunately that brings a problem when determining if a federated resource
+is ready or not.
+
+In the following, there is a list of Status objects of different custom resource definitions:
+
+```go
+type AddonStatus struct {
+	Ready bool          `json:"ready" yaml:"ready"`
+	Stage status.Status `json:"stage,omitempty" yaml:"stage,omitempty"`
+}
+
+type PodStatus struct {
+  Phase PodPhase
+ ...
+ }
+
+ type ServiceStatus struct {
+  LoadBalancer LoadBalancerStatus
+ }
+
+// KonvoyClusterStatus defines the observed state of KonvoyCluster
+type KonvoyClusterStatus struct {
+	// Phase represents the current phase of Konvoy cluster actuation.
+	// E.g. Pending, Provisioning, Provisioned, Deleting, Failed, etc.
+	// +optional
+	Phase KonvoyClusterPhase `json:"phase,omitempty"`
+
+  ...
+}
+```
+
+As mentioned, their Status schema is quite different from one to another.
+
+Consequently, the intention is to **enforce** in the federated resources this approved [recommendation](https://github.com/kubernetes/enhancements/pull/1624/files).
+to expect a common schema for `.status.conditions` and share golang logic for common Get, Set, Is for `.status.conditions`.
+
+1. For all new APIs, have a common type for `.status.conditions`.
+2. Provide common utility methods for `HasCondition`, `IsConditionTrue`, `SetCondition`, etc.
+3. Provide recommended defaulting functions that set required fields and can be embedded into conversion/default functions.
+
+By following this approach, kubefed would be able to properly consume and report the status
+of any federated resource by checking the `status.conditions` (e.g `Ready=True`) fields.
+
+
+### User Stories
+
+#### Story 1
+
+Users create federated resources and want to be aware of their status without having
+to access to the remote clusters.
+
+In the following example, we have a `FederatedAddon`, named `reloader`, deployed across ten `kubefedclusters`.
+An `Addon` is a custom resource definition that abstract the creation of apps composed of
+one or multiple Helm charts.
+
+```yaml
+---
+apiVersion: types.kubefed.io/v1beta1
+kind: FederatedAddon
+metadata:
+  name: reloader
+  namespace: kubeaddons
+spec:
+  placement:
+    clusters:
+    - name: cluster10
+    - name: cluster9
+    - name: cluster8
+    - name: cluster7
+    - name: cluster6
+    - name: cluster5
+    - name: cluster4
+    - name: cluster3
+    - name: cluster2
+    - name: cluster1
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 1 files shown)


RainbowMango:
/cc @deads2k 
Could you please help to take a look at this? I guess you will be interested in this.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,408 @@
...
+            app: nginx
+        template:
+          metadata:
+            labels:
+              app: nginx
+          spec:
+            containers:
+            - image: nginx
+              name: nginx
+  status:
```
RainbowMango, line 130:
```suggestion
  status:
    clusters:
      - name: cluster1
      - name: cluster2
      - name: cluster3
```

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,408 @@
...
+hold the status of the federated resources, e.g Ready, NotReady.
+
+The status of the federated resources should determine whether the resources satisfy
+a `Ready` condition, and otherwise report their error status.
+To do so, this property reports the status of the federated resources in their
+target clusters whenever a `ReadyCondition` is not satisfied.
+This condition would need to be identified per type or by the usage of an interface
+`IsReady` that determines this value per type of resource.
+By doing so, we ensure the `Conditions` property shows the status of only unhealthy
+resources.
```
RainbowMango, line 152: Maybe it's better to show status in all clusters, not only the unhealthy ones. e.g.
```
status:  
  clusters:
  - name: cluster1
    status："Ready"
  - name: cluster2
    status: "NotReady"
  conditions:
    ...
```

The `.status.clusters` should give cleary info.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,408 @@
...
+
+Consequently, the intention is to **enforce** in the federated resources this approved [recommendation](https://github.com/kubernetes/enhancements/pull/1624/files).
+to expect a common schema for `.status.conditions` and share golang logic for common Get, Set, Is for `.status.conditions`.
+
+1. For all new APIs, have a common type for `.status.conditions`.
+2. Provide common utility methods for `HasCondition`, `IsConditionTrue`, `SetCondition`, etc.
+3. Provide recommended defaulting functions that set required fields and can be embedded into conversion/default functions.
+
+By following this approach, kubefed would be able to properly consume and report the status
+of any federated resource by checking the `status.conditions` (e.g `Ready=True`) fields.
```
RainbowMango, line 282: Do you mean `type GenericCondition struct` should be replaced by the standard condition?
I think it's totally doable given that we are still in beta state.

docs/keps/20200619-federated-resource-status.md
RainbowMango: I guess it's better to show all fields of `status` letting people know the current situations and get what will be changed.
Same as below.

docs/keps/20200619-federated-resource-status.md
hectorj2f: Imagine you have 100 clusters, that won't scale :/, that is why I understand they only saw the clusters for those whose propagation failed

RainbowMango: Yeah, I also thought about it. But, given users may use `spec.placement.clusterSelector` to specific clusters they want to deploy, It's better to clearly show the status, which means users can clearly know which clusters were matched.

There may be hundreds of items in `status.clusters`, it's a little bit not good for human reading, but I think it is sustainable.
Here is an example, that is `Node.Status.Images` which could contain a whole list of images in a specific node.

pacoxu:
For deployment, the available replica number is also important, especially when we use rebalance strategy. 

See https://github.com/kubernetes-sigs/kubefed/issues/1240#issuecomment-651514461

docs/keps/20200619-federated-resource-status.md
hectorj2f: I think having hundreds or thousands values in the status when a resource got propagated, it is hard to read/maintain :/. I'd like to hear why this was not done in the beginning by kubefed owners although I imagine this won't scale in these situations.

docs/keps/20200619-federated-resource-status.md
hectorj2f: This is an example of a current status as you can see it today. I show the clusters and conditions field, what else you would like to see.

docs/keps/20200619-federated-resource-status.md
irfanurrehman: The choice of storing only failure status compared to all is simply to reduce the load on etcd. We infact did play with an option to collect each resource status from each cluster as is into the federated resource using the controller [here](https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/controller/status/controller.go). We did then implement the current propagation status mechanism to be able to:
- easily decipher the propagation failures.
- reduce load on etcd, which goes directly proportional to all resources in all clusters.

Currently all resources status is collected [only for services](https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/apis/core/v1beta1/federatedtypeconfig_types.go#L218) if the service discovery feature is enabled. We could as well update this to make it a feature to users who understand the risk and be able to collect raw status side by side with the propagation status.

irfanurrehman:
@hectorj2f the proposal here seems ok to me. I would however like it more if propagation status and the individual cluster resource status could be segregated in in some way. For example as a user I can see both the propagation status and individual resource status as two separate sub status fields in the status resource. wdyt?
Additionally, I find it a lot more valuable to be able to have an option (for example via the typeconfig) to control what status shows up and what not and probably at some point add  an option to get the raw resource status (for any/all resources) as mentioned in [this comment](https://github.com/kubernetes-sigs/kubefed/pull/1237#discussion_r453461938).

docs/keps/20200619-federated-resource-status.md
hectorj2f: > easily decipher the propagation failures.
   reduce load on etcd, which goes directly proportional to all resources in all clusters.

This makes a lot of sense to me. Honestly I would only like to see StatusReady or AnyFailed status for those resources that aren't. If I federate 1k resources showing them as Ready or propagated will be a killer in my opinion. 
Besides, I don't think anyone else requested to be able to report the success for all the federated resources yet.

I am trying to limit the scope of this KEP, so adding this feature to show the status of all the resources sounds to me like an additional feature or iteration. Can we agree on keeping the status reporting as done until now based on the initial approach ?

* easily decipher the propagation failures.
* reduce load on etcd, which goes directly proportional to all resources in all clusters.

docs/keps/20200619-federated-resource-status.md
hectorj2f: Yes, that is the intention, or at least, to extend it.

docs/keps/20200619-federated-resource-status.md
irfanurrehman: Sound good to me!

hectorj2f:
> @hectorj2f the proposal here seems ok to me. I would however like it more if propagation status and the individual cluster resource status could be segregated in in some way. For example as a user I can see both the propagation status and individual resource status as two separate sub status fields in the status resource. wdyt?

Initially I started following this approach but then I found that logically the goal of a resource is to be ready in a remote cluster. This might differ from a resource to another, for instance:
- Secret: ready would mean creation, so the propagation status itself would determine the readiness by the creation of the resource.
- Deployment: this special case requires the checking of its status condition, therefore we need the propagation + the ready condition.

With that said, I see the status of a federated resource as a single field where the potential state transitions will go from (as examples): `PropagationOK` + `IsReady (optional)`.  

Therefore I can only envision one field as now.

> Additionally, I find it a lot more valuable to be able to have an option (for example via the typeconfig) to control what status shows up and what not and probably at some point add an option to get the raw resource status (for any/all resources) as mentioned in this comment.

I don't disagree but I think that should be tackled in a different kep or iteration. We are just trying to improve the content of the status field focusing on the failed resources (as until now).

hectorj2f:
As @pmorie raised in one of the meetings, we might need to add to this feature a `qps` flag to be able to limit the amount of queries and so avoid blowing the control-plane at a high scale.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,408 @@
...
+As mentioned, their Status schema is quite different from one to another.
+
+Consequently, the intention is to **enforce** in the federated resources this approved [recommendation](https://github.com/kubernetes/enhancements/pull/1624/files).
+to expect a common schema for `.status.conditions` and share golang logic for common Get, Set, Is for `.status.conditions`.
+
+1. For all new APIs, have a common type for `.status.conditions`.
+2. Provide common utility methods for `HasCondition`, `IsConditionTrue`, `SetCondition`, etc.
+3. Provide recommended defaulting functions that set required fields and can be embedded into conversion/default functions.
+
+By following this approach, kubefed would be able to properly consume and report the status
```
jieyu, line 281: Does that means for some core API types like Deployment, kubefed would need to maintain a mapping to the corresponding condition type that it'll look into? And for custom resources, the crd needs to follow some convention on conditions to leverage this feature?

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,408 @@
...
+            labels:
+              app: nginx
+          spec:
+            containers:
+            - image: nginx
+              name: nginx
+  status:
+    clusters:
+    - name: "cluster3"
+      status: "ReplicaFailure"
```
jieyu, line 192: I don't quite like the idea of coupling propagation status with object status on the same field.

For backwards compatibility, maybe still keep `clusters[x].status` as the propagation status, and introduce another `clusters[x].conditions` to capture the abnormal conditions for each clusters (e.g., `NotReady` type). In that way, we could also track other information like the message for the condition for users to triage problems from the mgmt cluster.

docs/keps/20200619-federated-resource-status.md
hectorj2f: Yes, certain core API types provide conditions that can facilitate that job. Regarding the Deployment, I would say we should mirror what these types provide to be considered as `Ready`.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,514 @@
...
+hold the status of the federated resources, e.g Ready, NotReady.
+
+The status of the federated resources should determine whether the resources satisfy
+a `Ready` condition, and otherwise report their error status.
+To do so, this property reports the status of the federated resources in their
+target clusters whenever a `ReadyCondition` is not satisfied.
+This condition would need to be identified per type or by the usage of an interface
+`IsReady` that determines this value per type of resource.
+By doing so, we ensure the `Clusters/Conditions` property shows the status of only unhealthy
+resources for the respective clusters.
```
jieyu, line 220: I feel that for each cluster, we should also include the `Propagation` condition as well. If for any reason, the propagation of the federated resource failed, we should also report that at per cluster basis.

irfanurrehman:
@hectorj2f Is there something open on this? This looks ok for me in its current shape.
I would be happier if we can capture the gist of having the alternative status representation in a small section in the document itself for future reference (and not having it lost).

hectorj2f:
@irfanurrehman No. I'll revisit minor details and add the alternative section to the KEP. Other than that, it is completed.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -50,14 +51,38 @@ resources are healthy or not across clusters.
 
 * Quickly identify unhealthy federated resources by relying on the status of the federated resources.
 * Improve the troubleshooting of failures when propagating resources.
+* Define a new api version potentially `v1beta2` to include these new fields.
 
+### Non Goals
+
+* Set the resource status for all the federated resources and clusters whenever a federated resource satisfies the condition `Ready=True`.
+This goal would imply adding new flags to the kubefed control-plane components to limit the value and the number of concurrent requests that can be processed by the API server. So the control-plane is not blown by the amount of requests based on the metric of existing federated resources and kubefed clusters.
```
irfanurrehman, line 59: This is quite unclear. Do you mean usage of a flag as a controller config? Or a flag on the federated type config. Also how would using a flag solve the problem of not blowing up the control-plane? I mean just the flag/switch does not solve the amount of requests. Some adaptive mechanism, which can reduce the amount of requests can only do that.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -89,9 +114,49 @@ type GenericFederatedResource struct {
...
+	// +required
+	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
+	// A human readable message indicating details about the transition.
+	// This field may be empty.
+	// +required
+	Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
+}
+```
+
+Nowadays `Conditions` field of a federated resource holds the status of the federated actions (aka propagation status).
```
irfanurrehman, line 159: `s/Nowadays/As of now/g`

docs/keps/20200619-federated-resource-status.md
```diff
@@ -148,11 +217,12 @@ To do so, this property reports the status of the federated resources in their
...
 This condition would need to be identified per type or by the usage of an interface
 `IsReady` that determines this value per type of resource.
-By doing so, we ensure the `Conditions` property shows the status of only unhealthy
-resources.
+By doing so, we ensure the `Clusters/Conditions` property shows the status of only unhealthy
+resources for the respective clusters.
 
 If we re-use the example from above and imagine a scenario where this `FederatedDeployment` resource remained `Ready=True` in two clusters, but crashed in `cluster3`.
-The value of `status.conditions` should reflect the new state for that specific cluster.
+The value of `status.clusters.conditions` should reflect the new state for that specific cluster.
```
irfanurrehman, line 224: Shouldn't this be `status.clusters[<index>].conditions`

docs/keps/20200619-federated-resource-status.md
```diff
@@ -189,21 +260,35 @@ The value of `status.conditions` should reflect the new state for that specific
...
+      status: "False"
+      type: "Propagation"
 ```
 
-The value of `spec.conditions` contains a `NotReady` condition type, that is the
-result of checking the status of that remote `Deployment` in the target clusters.
-The value of `status.clusters.status` can be extracted from the `"ReplicaFailure"`
-status of the `Deployment` and so reused for visibility.
+The value of `status.conditions` describes the propagation error that happened to `cluster3`
+while `clusters["cluster3"].conditions` contains a `NotReady` condition type with a detailed
```
irfanurrehman, line 279: Looking at the below type definition above, I was under the impression that this is a slice. Is that not so.
This representation is of a map. If its not a map, please update it to use an index.

```
type GenericClusterStatus struct {
	Name   string            `json:"name"`
	Status PropagationStatus `json:"status,omitempty"`
  Conditions []*metav1.Condition `json:"conditions,omitempty"`
}
```

docs/keps/20200619-federated-resource-status.md
```diff
@@ -189,21 +260,35 @@ The value of `status.conditions` should reflect the new state for that specific
...
+description of the state of the federated resource in `cluster3`.
+The value of `status.clusters["cluster3"].status` reuses the current set of available status types, e.g. `"CreationFailed"`, which identifies the current problem during the propagation.
+
 
-Likewise, the status of the `FederatedDeployment` remains `Ready=True` in the rest
+The status of the `FederatedDeployment` remains `Ready=True` in the rest
 of clusters: `cluster1` and `cluster2`.
+However as you can see, the `clusters[*].conditions` does not contain values with the `IsReady` condition
```
irfanurrehman, line 286: Are we going to support regular expressions?

docs/keps/20200619-federated-resource-status.md
```diff
@@ -282,6 +367,22 @@ By following this approach, kubefed would be able to properly consume and report
...
+components to avoid blowing out the control-plane due to frequent and concurrent API calls to
+update the status of the federated resources.
+
+Obviously users might want to be able to enable/disable the definition of:
+
+* Which condition to look for each federated resource to determine its readiness, e.g. `Ready=True`, `Deployed=True`.
+This might be especially useful for custom resource types without a `IsReady` standard condition.
+
+* When to show up the raw status of the resources or just the failed status (as today).
+By raw status, the system understand to show the status of all the federated resources `Ready` and not `Ready`.
```
irfanurrehman, line 380: 1. By raw status, I assume its the copy of the status that appears under the status subresource of an arbitraty k8s resource. Atleast that is what I meant when I did refer to it earlier. Do you mean something else here?
2. To compute the `Ready/NotReady` status of the federated resource you would anyways need to fetch and go through the status of each individual resource. If any additional detail, for example the `message` that you have alluded in the example above, you might need to go through/parse all conditions that appear in that subresource. IMHO, this itself incurs a performance overhead. Do you yet have any insight/idea on what/how much this might be?

irfanurrehman:
@hectorj2f I have few more comments and questions, after your changes. Overall this looks good!

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,537 @@
...
+## Motivation
+
+Nowadays users have to connect to the kubefed clusters to be aware if the federated
+resources are healthy or not across clusters.
+
+### Goals
+
+* Quickly identify unhealthy federated resources by relying on the status of the federated resources.
+* Improve the troubleshooting of failures when propagating resources.
+* Define a new api version potentially `v1beta2` to include these new fields.
```
RainbowMango, line 54: Which API will be promoting to `v1beta2`?

docs/keps/20200619-federated-resource-status.md
```diff
@@ -0,0 +1,537 @@
...
+The value of `status.clusters["cluster3"].status` reuses the current set of available status types, e.g. `"CreationFailed"`, which identifies the current problem during the propagation.
+
+
+The status of the `FederatedDeployment` remains `Ready=True` in the rest
+of clusters: `cluster1` and `cluster2`.
+However as you can see, the `clusters[*].conditions` does not contain values with the `IsReady` condition
+for `cluster2` and `cluster1`.
+The system omits the reporting of these `Ready=True` status due two reasons:
+
+* easily decipher the propagation and other failures.
```
RainbowMango, line 290: My main concern is how do users know the control plane has set status correctly and completely?

If a  `status.condations` shows `Propagation = True`(as below), means both propagation and running status is ready, right?
```
status
  conditions:
  - status: "True"
    type: "Propagation"
```

hectorj2f: Yes, that means that the creation of the resource is considered to be ready. That could be something you might indicate with resources such as cluster roles,etc...

docs/keps/20200619-federated-resource-status.md
hectorj2f: The new properties added to the FederatedTypeConfig should bump the api version.

docs/keps/20200619-federated-resource-status.md
hectorj2f: No, I'll replace `[*]` by `[<index>]`.

docs/keps/20200619-federated-resource-status.md
hectorj2f: @irfanurrehman Yes, the approach includes a `flag` in the control plane, if you wanna know more about this kind of flag, it'd suggest to check the `kube-api-qps` or `kube-api-burst` flag used by the controller-manager, kube-proxy, ..., to talk to the Kubernetes API Server. I didn't want to throw too many details and leave it a bit open (we can always update this KEP). 



More details:

- Push reconciler: the kubefed controller plane talks to the Kubernetes API server of the host cluster, as well as the e Kubernetes API servers of the kubefed clusters. Therefore I considered we should define this flag (qps,burst), thus when creating the clients to the remote Kubernetes API server we can limit the amount of requests per seconds etc. This becomes more important when enabling the collection of the status from the resources when the amount of requests would increase.



- Pull reconciler: In this scenario, the flag values would be used to talk to the kubefed shim API server from the kubefed daemons, but it follows the same logic. The client connection would use these values to create the Kubernetes client and limit the clients.

docs/keps/20200619-federated-resource-status.md
hectorj2f: 1. Yes, I understood the same by raw status.
2. I was planning to use the ObservedGeneration to exclude out of date conditions. I didn't run an estimation yet, I am waiting to define this kep into implementable stage to proceed defining some solutions and/or limitations. We all agree this approach might be benefitial but there are some performance challenges.

hectorj2f: @irfanurrehman Do you have any opinions or additional questions here?

docs/keps/20200619-federated-resource-status.md
hectorj2f: @irfanurrehman wdyt?

irfanurrehman: Suggestion: s/This goal would imply adding new flags to the kubefed control-plane components to limit the value and the number of concurrent requests that can be processed by the API server. So the control-plane is not blown by the amount of requests based on the metric of existing federated resources and kubefed clusters./Addition of feature flags to control plane components to limit or throttle requests made to kubernetes API servers akin to  `kube-api-qps` or `kube-api-burst` in kubelet./

docs/keps/20200619-federated-resource-status.md
irfanurrehman: Fine with me.

irfanurrehman:
This is quite fine as provisional for me. It can certainly be refined and updated later.
I will leave it to another set of eyes to check or merge this.

jimmidyson:
I agree with @irfanurrehman [above](https://github.com/kubernetes-sigs/kubefed/pull/1237#issuecomment-671871335) that this is in a good state as a provisional KEP which we can refine in a few more PRs which I think are already on the way.

/lgtm

PR #1244, opened 2020-07-02
Kubernetes 1.18.5 dependency upgrade (plus other upgrades)

jimmidyson:

Upgraded dependencies for Kubernetes 1.18.5 as well as various other small upgrades. Needed to keep up to date with Kubernetes changes and improvements, especially for those projects using kubefed as a library.


mrbobbytables:
The repo-infra commit will need to be updated to use the version without nico in the owners file. 👍

RainbowMango:
Thanks Bob, I'm just curious, seems `clarketm` already the member of org, why robot still saying not.

mrbobbytables:
They have to accept the org membership request first before it will return valid

RainbowMango:
Oh, right, that makes sense.

RainbowMango:
Another question comes to my mind, why we need to check `OWNER` files even in a vendor path? 

In this case, we vendors a `Kubernetes`  repo, that's ok we add the guys to Kubernetes org. What if we vendors another repo which not belongs to `Kubernetes` but have an `OWNER` file in it?

@mrbobbytables How do you think? Is there anything I didn't know?

nikhita:
> What if we vendors another repo which not belongs to Kubernetes but have an OWNER file in it?

@RainbowMango It should ignore vendor directories and that was the behavior before. Looks like a bug that it doesn't ignore it anymore, I'll check if there were some things changed recently that could affect this behavior. For reference:

https://github.com/kubernetes/test-infra/blob/bb65d07c9ece6311ea82e58925b59ebd24f28d22/prow/config/config.go#L394

nikhita:
As a side note, should the `repo-infra` directory be under the path `vendor/k8s.io/repo-infra` instead of `vendor/repo-infra`?

RainbowMango:
> As a side note, should the repo-infra directory be under the path vendor/k8s.io/repo-infra instead of vendor/repo-infra?

Awesome! Nikhita.  I doubt if the vendor is the right choice, maybe a third-part path?

jimmidyson:
@RainbowMango I agree that `vendor` is probably not the right choice, but `vendor` used to be ignored for `OWNERS` check as @nikhita said above, hence placing it in there.

nikhita:
Fwiw I've now manually removed the `do-not-merge/invalid-owners-file` label since it appears to be a bug in prow. Will investigate but don't want it to be a blocker for y'all :)

RainbowMango:
@jimmidyson Thanks for your explanation, I didn't know the background before.

> Will investigate but don't want it to be a blocker for y'all 

@nikhita I guess Prow will check this for every PR. That means prow will set the `do-not-merge/invalid-owners-file` label before it gets fixed, right?
No worries, I just curious about the mechanisms of Prow.

nikhita:
> That means prow will set the do-not-merge/invalid-owners-file label before it gets fixed, right?

Yes, but it's problematic only for the PRs that would touch `vendor/`. Works as intended in other cases. :+1:

jimmidyson:
From @nikhita's [hint](https://github.com/kubernetes-sigs/kubefed/pull/1244#issuecomment-653400884), I've moved `repo-infra` from `vendor/repo-infra` to `vendor/k8s.io/repo-infra` which should match with the ignore list in invalid owners check.

irfanurrehman:
Thanks for doing this @jimmidyson 
/approve

Issue #1252, opened 2020-07-16
Extend federated type config status umbrella issue


RainbowMango:
## Background
If a target type in a `FederatedTypeConfig` not present in one or more member clusters, that will cause the sync controller totally not working because the federated informer hanging in sync. 
For more details discussion and reproduce, you can get from [here](https://github.com/kubernetes-sigs/kubefed/issues/1241). 

## Solutions
The community has discussed this issue years ago and proposed a solution at [here](https://github.com/kubernetes-sigs/kubefed/issues/314). The solution looks pretty good, so the main solution will follow this solution.

This umbrella issue tries to split the solution to some small pieces of iteration items. So that we can keep each PRs small and easy to review.
Any comments are welcome and feel free to take tasks you are interested in(just send a comment to this issue).

## Iteration Items/tasks
- [ ] API extension on board. (@RainbowMango)
This is the first PR that focus on API extension and deal with relative auto-generated things. 

- [ ] SyncController maintaining the new fields of status
- The sync controller will determine the status of the target type in each FederatedCluster in the namespace - present or not
- The overall status is driven off the status in each cluster - if the target type is not present in one or more clusters, the overall status is NotPresentInAllClusters

Maybe prevent `not present` cluster from federated informer too?

- [ ] SyncController use the new fields in reconciliation loop
Not sync to and not delete from the `not present` clusters.

- [ ] Add/Update E2E test

New iteration items could be added in the future, and it is certain that each item could not break the current functionalities.

/kind feature
/help

PR #1253, opened 2020-07-17
Extend federated type config status API

RainbowMango:
This PR focus on take the API change onboard.

Refer #1252


hectorj2f:
@RainbowMango is there any update here ? are you still working on this change ?

RainbowMango:
Yes, but I'm busy with preparing [KubeCon](https://cncf.lfasiallc.cn/) recently.

As I mentioned before, the whole task is a little bit more complicated, so I suggested run it by iteration, so other guys could collaborate together.

hectorj2f:
@RainbowMango In my opinion, I would suggest to add iterations that can be useful or self contained. If you add this new type without tests and without any connection to the rest of the code, it could end into an unused state if there is not more progress into this direction.

PR #1261, opened 2020-07-29
charts: split properties for webhook and manager

### LINKED ISSUE #1259: chart: expose the `env` property for the kubefed components in `values.yaml`
We should expose the `env` properties of the deployment specs for the kubefed pods in case we need to add ENV variables such as HTTP_NOPROXY, ...

**Why is this needed**:

/kind feature

hectorj2f:

We should split the webhook properties from the manager properties.
Also we should expose the env properties for the chart as well.

Fixes #

Closes #1259 

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/deployments.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/deployments.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     kubefed-control-plane: controller-manager
 spec:
-  replicas: {{ .Values.replicaCount }}
+  replicas: {{ .Values.controller.replicaCount }}
   selector:
     matchLabels:
       kubefed-control-plane: controller-manager
@@ -22,10 +22,17 @@ spec:
       containers:
       - command:
         - /hyperfed/controller-manager
-        - "--v={{ .Values.logLevel }}"
-        image: "{{ .Values.repository }}/{{ .Values.image }}:{{ .Values.tag }}"
-        imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+        - "--v={{ .Values.controller.logLevel }}"
+        image: "{{ .Values.controller.repository }}/{{ .Values.controller.image }}:{{ .Values.controller.tag }}"
+        imagePullPolicy: "{{ .Values.controller.imagePullPolicy }}"
         name: controller-manager
+{{- if .Values.controller.env }}
+        env:
+{{- range $key, $value := .Values.controller.env }}
+        - name: "{{ $key }}"
+          value: "{{ $value }}"
+{{- end }}
+{{- end }}
         ports:
         - containerPort: 9090
           name: metrics
@@ -37,8 +44,8 @@ spec:
           periodSeconds: 3
           timeoutSeconds: 3
         resources:
-{{- if .Values.resources }}
-{{ toYaml .Values.resources | indent 12 }}
+{{- if .Values.controller.resources }}
+{{ toYaml .Values.controller.resources | indent 12 }}
 {{- end }}
       terminationGracePeriodSeconds: 10
 ---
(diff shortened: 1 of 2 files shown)


charts/kubefed/values.yaml
```diff
@@ -6,19 +6,36 @@
...
+    logLevel: 2
+    env: {}
+    resources:
+      limits:
+        cpu: 500m
+        memory: 512Mi
+      requests:
+        cpu: 100m
+        memory: 64Mi
+  webhook:
```
RainbowMango, line 24: Why we put `webhook`'s properties under `controllermanager`?

charts/kubefed/values.yaml
```diff
@@ -6,19 +6,36 @@
 ##
 controllermanager:
```
RainbowMango, line 7: (Not sure if it works.) Just a guess, would it be better to rename `controllermanager` to other common names such as `component`?

RainbowMango:
Just a nit, otherwise LGTM.

charts/kubefed/values.yaml
```diff
@@ -6,19 +6,36 @@
...
-  imagePullPolicy: IfNotPresent
-  logLevel: 2
-  resources:
-    limits:
-      cpu: 500m
-      memory: 512Mi
-    requests:
-      cpu: 100m
-      memory: 64Mi
+  manager:
```
irfanurrehman, line 9: suggestion: I would rather name this `controller`.

charts/kubefed/values.yaml
```diff
@@ -6,19 +6,36 @@
...
+    env: {}
+    resources:
+      limits:
+        cpu: 100m
+        memory: 256Mi
+      requests:
+        cpu: 100m
+        memory: 64Mi
+
   clusterAvailableDelay:
```
irfanurrehman, line 39: Aren't these properties also applicable only to `controller`?

charts/kubefed/values.yaml
hectorj2f: They are part of the chart components named controller manager.

charts/kubefed/values.yaml
hectorj2f: That might imply to change the chart name. I don't know if we really want to do that now.

charts/kubefed/values.yaml
hectorj2f: Yes, we could use `controller` too.

charts/kubefed/values.yaml
irfanurrehman: The webhook can actually be running external to the application, however this was for the sake of simplicity in the first place, probably not to define many new additional settings.

charts/kubefed/values.yaml
hectorj2f: These are part of `controllermanager` main chart, perhaps I can move them up to easily identify the scope for manager` and `webhook`.

charts/kubefed/values.yaml
hectorj2f: I didn't want to reorganize the whole chart before we get this potential PR with Helmv3 support. But I could do it if you want it.

irfanurrehman: I did not imply to move it out. I am ok until there is a need for it. :) (YAGNI)

charts/kubefed/values.yaml
hectorj2f: I moved them up so it is more clear the purpose of the properties manager and webhook. To me, properties such as `clusterAvailableDelay` are defining a more generic behavior to the whole kubefed control-plane.

charts/kubefed/charts/controllermanager/templates/deployments.yaml
```diff
@@ -22,10 +22,17 @@ spec:
       containers:
       - command:
         - /hyperfed/controller-manager
-        - "--v={{ .Values.logLevel }}"
-        image: "{{ .Values.repository }}/{{ .Values.image }}:{{ .Values.tag }}"
-        imagePullPolicy: "{{ .Values.imagePullPolicy }}"
+        - "--v={{ .Values.manager.logLevel }}"
```
irfanurrehman, line 25: Wouldn't you need to change the to `controller`? As you have updated in `values.yaml` below.

hectorj2f: yes, I did now pushed the changes.

RainbowMango:
LGTM. I'd like to give a chance to @irfanurrehman for final approval.

irfanurrehman:
Thanks for doing this @hectorj2f.
/lgtm
/approve

hectorj2f:
Thanks for the reviews.

PR #1265, opened 2020-08-05
fix: verify deployed image

makkes:
This adds code to verify that the image used in the kubefed
deployments is actually the one that was expected and prevents
regression for the bug fixed in
2ba430ee9f5e68c05a237cc006d938bf5296431e.

no issue


scripts/deploy-kubefed.sh
```diff
@@ -108,6 +111,15 @@ function kubefed-admission-webhook-ready() {
   [[ "${readyReplicas}" -ge "1" ]]
 }
 
+function deployment-image-as-expected() {
+  local namespace="${1}"
+  local deployment="${2}"
+  local expected_image="${3}"
+
+  local deployed_image="$(kubectl -n kube-federation-system get deploy kubefed-admission-webhook -o jsonpath={.spec.template.spec.containers[0].image})"
```
jimmidyson, line 119:
```suggestion
  local deployed_image="$(kubectl -n ${namespace} get deploymment ${deployment} -o jsonpath={.spec.template.spec.containers[0].image})"
```

scripts/deploy-kubefed.sh
```diff
@@ -108,6 +111,15 @@ function kubefed-admission-webhook-ready() {
   [[ "${readyReplicas}" -ge "1" ]]
 }
 
+function deployment-image-as-expected() {
+  local namespace="${1}"
+  local deployment="${2}"
+  local expected_image="${3}"
+
+  local deployed_image="$(kubectl -n ${namespace} get deployment ${deployment} -o jsonpath={.spec.template.spec.containers[0].image})"
```
jimmidyson, line 119: How about using the container name rather than index in case stuff moves around in future?

```suggestion
  local deployment="${2}"
  local container="${3}"
  local expected_image="${4}"

  local deployed_image="$(kubectl -n ${namespace} get deployment ${deployment} -o jsonpath='{.spec.template.spec.containers[?(@.name=="'"${container}"'")].image}')"
```

PR #1269, opened 2020-08-06
Move k8s.io/repo-infra from vendor to third-party

jimmidyson:
This PR moves `k8s.io/repo-infra` code from `vendor` to `third-party`. With go modules, if the vendor directory exists, go mod will run in vendor mode, which should cause the build to fail. For some reason, it doesn't, but I was hit by this earlier today so moving to be safe.


RainbowMango:
@jimmidyson 
Frankly speaking, I don't think it's the ideal way to track tools like `repo-infra` to a `third-party` directory. 
It's hard to track the tools' version and it's dependencies.

For this kind of case, the Golang community provides a [solution post here](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module). 

And, also there is an example from [Kuberenetes](https://github.com/kubernetes/kubernetes/tree/master/hack/tools).

jimmidyson:
@RainbowMango thanks for the pointer. I'm aware of using a tools module for specifying versions of required build tools and love the technique. However, I don't think it applies here because we are not building any binaries provided by the tool repo in the normal way, by running `go install`. The example you share is actually used in kubernetes to install a command called `kazel` from repo infra. We need the source directory to call sheet scripts provided in that repo. Calling `go mod download` would do that, caching the source repo, but then afaik there is no easy way to reference that cached path. I'd be happy to learn a way though.

This is also the recommended way to use his subtree for this repo as documented in the repo-infra readme.

RainbowMango:
> This is also the recommended way to use his subtree for this repo as documented in the repo-infra readme.

After going through the [repo-infra README](https://github.com/kubernetes/repo-infra/blob/master/README.md), I'd say it's a doable way to using it.

But, I also have done some interesting tests locally, I'm not sure if it really full fit our requirement.
Since we only need `k8s.io/repo-infra/hack/verify_boilerplate.py` file, so we can simply depend on `k8s.io/repo-infra/hack` package like [this commit does](https://github.com/RainbowMango/kubefed/commit/fc27d7e264a88b653300ea7504e6d342a8329f61).
In that case, we can easily `track the repo-infra version` and `vendor it`, in addition, only `k8s.io/repo-infra/hack` directory will be vendors like [this commit does](https://github.com/RainbowMango/kubefed/commit/1be9b51544ed9dbb4e544102f9d4a47c0a736030).

How do you think? By the way, if you insist on using the subtree way, it's totally fine for me.

jimmidyson:
@RainbowMango I have tried these approaches and hit a few problems, namely vendoring `k8s.io/repo-infra/hack` works (it contains go source files), but vendoring `k8s.io/repo-infra/verify/boilerplate` does not work (no go source files). I think we've spent enough time on trying to tweak something that, while I agree is not perfect, is good enough for what we need.

hectorj2f:
Let's give it a try, and see if anyone faces issues or finding it problematic.

PR #1263, opened 2020-08-03
feat: remove openshift/generic-admission-server

makkes:
**What this PR does / why we need it**:
This PR removes the dependency on the now unmaintained github.com/openshift/generic-admission-server library by migrating the webhook server to sigs.k8s.io/controller-runtime.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
no issue

**Special notes for your reviewer**:

PR #1270, opened 2020-08-06
Dependency and golang version updates

jimmidyson:
Upgrades to Go module dependencies, especially controller-runtime, and golang version to 1.14.7.

Includes commits from #1269 to fix go mod running in vendor mode - will rebase once that is merged.


Makefile
```diff
@@ -43,7 +43,7 @@ endif
 BUILDMNT = /go/src/$(GOTARGET)
 # The version here should match the version of go configured in
 # .travis.yml
-BUILD_IMAGE ?= golang:1.14.4
+BUILD_IMAGE ?= golang:1.14.6
```
makkes, line 46: maybe it's worth it using `golang:1.14` to always use the latest patch release?

jimmidyson: That wouldn't use the latest release if we already have golang:1.14 locally. I'd prefer to keep this explicit.

RainbowMango:
Hi @jimmidyson , why split to so many commits for a dependency update thing? Can you squash them?

/label tide/merge-method-squash

RainbowMango:
How about updating the version to the latest [Go 1.14.7](https://github.com/golang/go/releases/tag/go1.14.7)?

Makefile
RainbowMango: > I'd prefer to keep this explicit

+1

jimmidyson:
I've squashed dependency commits and upgraded to golang 1.14.7.

/remove-label tide/merge-method-squash

irfanurrehman:
@jimmidyson thanks for doing this. I am however curious about the need for this update?
IMHO a good strategy to update dependencies and/or go version is when the major  releases happen, or some functionality change warrants the update.

jimmidyson:
@irfanurrehman The golang update includes security fixes to packages such as `crypto/x509` and `net/http`, that while I haven't reviewed in depth whether we're affected by them, the upgrade doesn't introduce any risk so I don't see why we wouldn't do the update. The update to controller-runtime includes fixes but also metrics for webhooks (important since #1263 migrated to the controller-runtime webhook server).

> IMHO a good strategy to update dependencies and/or go version is when the major releases happen, or some functionality change warrants the update.

hectorj2f:
This PR just bumps certain dep versions.

irfanurrehman:
> @irfanurrehman The golang update includes security fixes to packages such as `crypto/x509` and `net/http`, that while I haven't reviewed in depth whether we're affected by them, the upgrade doesn't introduce any risk so I don't see why we wouldn't do the update. The update to controller-runtime includes fixes but also metrics for webhooks (important since #1263 migrated to the controller-runtime webhook server).
> 
> > IMHO a good strategy to update dependencies and/or go version is when the major releases happen, or some functionality change warrants the update.
> 
> I agree if we're doing major version upgrades of dependencies, but otherwise updating to minor or patch releases where backwards compatibility applies, I would prefer to upgrade adhoc. That way the project takes advantage of latest bug fixes, security fixes, and improvements in dependencies as we go, while also reducing the need for a bigger PR to do upgrades later (smaller updates are always preferred imo).

I did not mean to indicate that this change is not welcome, if that is what it did sound like. I wanted to point out my inclination on having a good strategy for such changes. I think we did lack this strategy earlier and right now is as good a time as any.

Additionally I think it would also be a good call to put a review strategy in place (given we still have active reviewers from multiple organisations). This is important to maintain the opinion diversity/sense of community alive in the project, lest the project becomes further opinionated. For an example we had a review strategy earlier which indicated that the same organisation contributors refrained from the final lgtm on the PR and atleast 2 reviewers needed to be happy for the PR to be merged. I am not saying that we force something like this stringently, but again IMO  this is a good strategy as any, unless somebody has a better one.
@hectorj2f @jimmidyson @RainbowMango @makkes

jimmidyson:
@irfanurrehman Thank you for your comments and I really do apologise for what has happened with these last 2 PRs - we will make sure this doesn't happen again. I completely understand what you are saying and agree completely with you about the review strategy. @makkes (a colleague of mine) made the same comments to me when he saw what was happening and has volunteered to write up a review strategy for us to review :) to work better together. Again, apologies that this has happened but let's use it as a catalyst for improvement.

PR #1274, opened 2020-08-17
Update linters and gofmt script

jimmidyson:
Update golangci-lint configuration to explicitly list all enabled linters and add a few more to the existing set. The original change was driven by a desire to  remove gofmt scripts and relying on golangci-lint in combination with goimports to provide formatting and validation.

Second commit fixes all lint failures.

Fixes #

Sorry for PRs like this but as we're working on the codebase more, I would like to ensure we have as good tooling for keeping a consistent codebase as much as possible.


hectorj2f:
@irfanurrehman @RainbowMango do you have any comment regarding this change ? Otherwise I'd say, we can merge it.

RainbowMango:
Just quickly went through the PR. LGTM. Thanks.

Would you like to take an eye on it? @irfanurrehman

RainbowMango:
I can imagine it becomes harder to make the robot happy:)  Hope people/contributor won't get crazy with so many linters :)
Maybe we should consider improve the performance of CI job later.

hectorj2f:
I'd wait for @irfanurrehman before giving it the `#/lgtm`

jimmidyson:
@RainbowMango the majority of these linters were already applied by virtue of being on by default. This pr just ecologically specifies then to reduce chance of surprise on upgrade of golangci-lint.

Makefile
```diff
@@ -67,7 +67,7 @@ TEST = $(TEST_CMD) $(TEST_PKGS)
 DOCKER_BUILD ?= $(DOCKER) run -it --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
 
 # TODO (irfanurrehman): can add local compile, and auto-generate targets also if needed
-.PHONY: all container push clean hyperfed controller kubefedctl test local-test vet fmt build bindir generate webhook e2e deploy.kind
+.PHONY: all container push clean hyperfed controller kubefedctl test local-test vet build bindir generate webhook e2e deploy.kind
```
irfanurrehman, line 70: I think it will be useful to have an option like `make lint` for a developer to run before submitting a PR. I mean one can run it explicitly also, but having it as part of `make`, makes it intuitive. WDYT?

irfanurrehman:
Thanks for doing this @jimmidyson.
I have just one suggestion, pending which the PR looks good to me.

Makefile
jimmidyson: Sure, can do. I was going to add [pre-commit](https://pre-commit.com) config to make it easier to run but can absolutely add `lint` target to makefile.

jimmidyson: Added lint target.

PR #1282, opened 2020-09-22
chore: webhook functions now return AdmissionResponse

### LINKED ISSUE #1267: pkg/controller/webhook.Validate et al should return an `AdmissionResponse`
The functions in [pkg/controller/webhook/util.go](https://github.com/kubernetes-sigs/kubefed/blob/bb3cd73f60ec8053c7fb991c94efe6d4a99aaced/pkg/controller/webhook/util.go) should not mutate the `admission.Response` parameter but rather return a new one.

**Why is this needed**:
This came out of [this PR review](https://github.com/kubernetes-sigs/kubefed/pull/1263#discussion_r465677103).

/kind feature

makkes:

Having these functions modify the AdmissionResponse passed in as
parameter makes it hard to reason about the code.

Fixes #1267 

### DIFF
--- a/pkg/controller/webhook/util.go
+++ b/pkg/controller/webhook/util.go
@@ -21,13 +21,12 @@ import (
 	"fmt"
 	"net/http"
 
-	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
-
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
 )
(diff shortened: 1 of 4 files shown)


pkg/controller/webhook/util.go
```diff
@@ -56,43 +56,52 @@ func NewValidatingResource(resourcePluralName string) schema.GroupVersionResourc
...
-		status.Allowed = true
-		return true
+		return &admission.Response{
+			AdmissionResponse: admissionv1beta1.AdmissionResponse{
+				Allowed: true,
+			},
+		}
 	}
-	return false
+	return nil
```
RainbowMango, line 72: I guess return `nil` equals to `not allowed` would bring ambiguous, and return a `admission.Response` will break the function as well.

The function name `Allowed` designed to check whether a specific request is allowed. So the return type should be `bool`.

A simple way maybe just remove `status` from the original function signature.

makkes: Thanks for the feedback, @RainbowMango, I'll look to improve the signature of this func.

makkes:
I'll have to have a look at the failing tests as well.

makkes:
Ok, failing test was just caused by the linter.

pkg/controller/webhook/util.go
makkes: @RainbowMango I addressed your concern and improved the utility functions a bit. PTAL.

RainbowMango:
I guess commit should be squashed, otherwise, LGTM.

makkes:
commits are squashed now.

RainbowMango:
/lgtm
Thanks.

jimmidyson:
/approved

Thanks @makkes!

jimmidyson:
/approve

Thanks @makkes!

PR #1292, opened 2020-10-02
feat: collect remote resource status when enabled

hectorj2f:

This PR aims to implement this feature https://github.com/kubernetes-sigs/kubefed/blob/ba8f9a44e85951e460454ab7875cb0a551750124/docs/keps/20200619-federated-resource-status.md. This work is a first iteration towards improving the troubleshooting of the status of federated resources.

In this branch, we add a new property to the FederatedResources `remoteStatus` which stores the status of the created object on each federated cluster. An example of a FederatedDeployment and its status is shown below:

```
- apiVersion: types.kubefed.io/v1beta1
  kind: FederatedDeployment
  metadata:
    finalizers:
    - kubefed.io/sync-controller
    generation: 1
    name: asystem
    namespace: asystem
    resourceVersion: "70174497"
  spec:
    placement:
      clusters:
      - name: cluster1
    template:
      metadata:
        labels:
          app: nginx
      spec:
        replicas: 3
        selector:
          matchLabels:
            app: nginx
        template:
          metadata:
            labels:
              app: nginx
          spec:
            containers:
            - image: nginx
              name: nginx
  status:
    clusters:
    - name: "cluster1"
      remoteStatus:
        availableReplicas: 3
        conditions:
        - lastTransitionTime: "2020-09-18T11:07:55Z"
          lastUpdateTime: "2020-09-18T11:18:31Z"
          message: ReplicaSet "asystem-f89759699" has successfully progressed.
          reason: NewReplicaSetAvailable
          status: "True"
          type: Progressing
        - lastTransitionTime: "2020-09-24T05:42:11Z"
          lastUpdateTime: "2020-09-24T05:42:11Z"
          message: Deployment has minimum availability.
          reason: MinimumReplicasAvailable
          status: "True"
          type: Available
        observedGeneration: 3
        readyReplicas: 3
        replicas: 3
        updatedReplicas: 3
    conditions:
    - lastTransitionTime: "2020-05-25T20:23:59Z"
      lastUpdateTime: "2020-05-25T20:23:59Z"
      status: "True"
      type: "Propagation"
```

To enable this feature, you need to add the feature gate in the kubefedconfig spec:

```
  featureGates:
  - name: RawResourceStatusCollection
    configuration: "Enabled"
```

In addition to that step, you will need to enable the `statusCollection` for the FederatedTypeConfig resources you wish to collect the `remoteStatus`. To do so, you set the property `statusCollection` to `Enabled`.

```
spec:
  federatedType:
    ...
  propagation: Enabled
  statusCollection: Enabled
  targetType:
    ...
```

Fixes #

I am testing some changes in the test suite so additional polishing actions are expected. 

Obviously this feature could potentially cause some performance degradation if our scenario federates many resources and we enable the status collection for all them. We are working on measuring and improving this behavior.


hectorj2f:
I'll have a look at the tests.

hectorj2f:
It seems the in-memory controllers test is failing 🔍 .

pkg/controller/sync/controller.go
```diff
@@ -405,19 +416,20 @@ func (s *KubeFedSyncController) setFederatedStatus(fedResource FederatedResource
...
+		if updateRequired, err := status.SetFederatedStatus(obj, reason, *collectedStatus, *collectedResourceStatus, resourceStatusCollection); err != nil {
+			klog.Infof("Failed to set the status for %s %q", kind, name)
 			return false, errors.Wrapf(err, "failed to set the status")
 		} else if !updateRequired {
-			klog.V(4).Infof("No status update necessary for %s %q", kind, name)
+			klog.Infof("No status update necessary for %s %q", kind, name)
 			return true, nil
 		}
-
+		klog.Infof("Updating status of name '%v' and kind '%v'", name, kind)
```
makkes, line 426: I'd suggest aligning the order of name/kind so it's always the same. Same goes for the type, either `%s` or `%v`.

test/e2e/defaulting.go
```diff
@@ -88,6 +88,10 @@ var _ = Describe("Default", func() {
 							Name:          string(features.PushReconciler),
 							Configuration: v1beta1.ConfigurationDisabled,
 						},
+						{
+							Name:          string(features.RawResourceStatusCollection),
+							Configuration: v1beta1.ConfigurationEnabled,
```
makkes, line 93: so by default this feature is enabled, right? But still one has to enable it per type, is that correct?

test/e2e/framework/controller.go
```diff
@@ -50,7 +50,13 @@ func NewSyncControllerFixture(tl common.TestLogger, controllerConfig *util.Contr
 	if err != nil {
 		tl.Fatalf("Error starting sync controller: %v", err)
 	}
-	if typeConfig.GetStatusEnabled() {
+	// The status controller should be only enabled for services whenver the statusAPIResource is defined
```
makkes, line 53:
```suggestion
	// The status controller should be only enabled for services whenever the statusAPIResource is defined
```

makkes (approved):
just some questions.

pkg/controller/sync/status/status.go
```diff
@@ -71,8 +73,10 @@ const (
 )
 
 type GenericClusterStatus struct {
-	Name   string            `json:"name"`
-	Status PropagationStatus `json:"status,omitempty"`
+	Name         string            `json:"name"`
+	Status       PropagationStatus `json:"status,omitempty"`
+	RemoteStatus interface{}       `json:"remoteStatus,omitempty"`
```
kensipe, line 78: not a fan of unstructured status... this answers my earlier question but needs to be defined clearly in the documentation.

pkg/features/features.go
```diff
@@ -69,4 +75,5 @@ var DefaultKubeFedFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec
 	PushReconciler:               {Default: true, PreRelease: featuregate.Beta},
 	CrossClusterServiceDiscovery: {Default: true, PreRelease: featuregate.Alpha},
 	FederatedIngress:             {Default: true, PreRelease: featuregate.Alpha},
+	RawResourceStatusCollection:  {Default: false, PreRelease: featuregate.Beta},
```
kensipe, line 78: shouldn't this be Alpha?

test/common/crudtester.go
```diff
@@ -714,27 +715,133 @@ func (c *FederatedTypeCrudTester) versionForCluster(version *fedv1a1.PropagatedV
...
+			c.waitInterval, 30*time.Second)
+
+		util.AddManagedLabel(targetObject)
+		labeledObj, err := CreateResource(clusterConfig, c.typeConfig.GetTargetType(), targetObject)
+		if err != nil {
+			c.tl.Fatalf("Failed to create labeled resource in cluster %q: %v", clusterName, err)
+		}
+
+		clusterClient := genericclient.NewForConfigOrDie(clusterConfig)
+		defer func() {
```
kensipe, line 733: questioning a `defer` in a for loop

test/common/crudtester.go
```diff
@@ -714,27 +715,133 @@ func (c *FederatedTypeCrudTester) versionForCluster(version *fedv1a1.PropagatedV
...
+			obj.SetGroupVersionKind(labeledObj.GroupVersionKind())
+			err := clusterClient.Get(context.TODO(), obj, labeledObj.GetNamespace(), labeledObj.GetName())
+			if err != nil {
+				c.tl.Errorf("Error retrieving kubefed cluster object resource: %v", err)
+				return false, nil
+			}
+
+			objStatus = obj.Object[util.StatusField]
+			c.tl.Logf("Show kubefed cluster object status: %v", objStatus)
+			return (obj.Object[util.StatusField] != nil), nil
```
kensipe, line 753:
```suggestion
			return obj.Object[util.StatusField] != nil, nil
```
parens are not needed

test/common/crudtester.go
```diff
@@ -714,27 +715,133 @@ func (c *FederatedTypeCrudTester) versionForCluster(version *fedv1a1.PropagatedV
...
+	// The default is normally 30 seconds
+	waitTimeout := 6 * wait.ForeverTestTimeout
+	err := wait.PollImmediate(c.waitInterval, waitTimeout, func() (bool, error) {
+		fedObj, err := client.Resources(qualifiedName.Namespace).Get(context.Background(), qualifiedName.Name, metav1.GetOptions{})
+		if err != nil {
+			c.tl.Errorf("An unexpected error occurred while polling for remote status: %v", err)
+			return false, nil
+		}
+
+		if err == nil {
```
kensipe, line 787: this is always true and is unnecessary.

if err != nil the function returns... at this point... err is nil

kensipe: which will remove 1 of the returns and should just be `return false, nil`

test/common/crudtester.go
```diff
@@ -714,27 +715,133 @@ func (c *FederatedTypeCrudTester) versionForCluster(version *fedv1a1.PropagatedV
...
-		}
-		return (err == nil), nil
-	})
+		client := c.resourceClient(*statusAPIResource)
+		err := wait.PollImmediate(c.waitInterval, wait.ForeverTestTimeout, func() (bool, error) {
+			_, err := client.Resources(qualifiedName.Namespace).Get(context.Background(), qualifiedName.Name, metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				c.tl.Errorf("An unexpected error occurred while polling for desired status: %v", err)
+			}
+			return (err == nil), nil
```
kensipe, line 839:
```suggestion
			return err == nil, nil
```
parens are not needed

kensipe (commented):
a few nits and questions along the way

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -433,3 +439,17 @@ func (c *Controller) reconcileOnNamespaceFTCUpdate() {
 		}
 	}
 }
+
+func (c *Controller) isEnabledFederatedServiceStatusCollection(tc *corev1b1.FederatedTypeConfig) bool {
+	if tc.GetStatusEnabled() && tc.Name == "services" {
```
alenkacz, line 444: what does the services mean here? 🤔 is this some leftover from debugging? or is the old status collection only for services? I am a bit lost here...

pkg/controller/sync/status/status.go
```diff
@@ -157,54 +168,66 @@ func (s *GenericFederatedStatus) update(generation int64, reason AggregateReason
...
 		})
 	}
 	return true
 }
 
 // clustersDiffers checks whether `status.clusters` differs from the
 // given status map.
-func (s *GenericFederatedStatus) clustersDiffers(statusMap PropagationStatusMap) bool {
-	if len(s.Clusters) != len(statusMap) {
+func (s *GenericFederatedStatus) clustersDiffers(statusMap PropagationStatusMap, resourceStatusMap map[string]interface{}) bool {
```
alenkacz, line 218: nit: clustersDiffer

pkg/controller/sync/status/status.go
```diff
@@ -157,54 +168,66 @@ func (s *GenericFederatedStatus) update(generation int64, reason AggregateReason
...
+	if len(s.Clusters) != len(statusMap) || len(s.Clusters) != len(resourceStatusMap) {
+		klog.Info("Clusters differs from the size")
 		return true
 	}
 	for _, status := range s.Clusters {
 		if statusMap[status.Name] != status.Status {
 			return true
 		}
+		if !reflect.DeepEqual(resourceStatusMap[status.Name], status.RemoteStatus) {
+			klog.Infof("Clusters differs resource status: %v VS %v", resourceStatusMap[status.Name], status.RemoteStatus)
```
alenkacz, line 228: nit: Clusters resource status differ

pkg/controller/sync/status/status.go
```diff
@@ -157,54 +168,66 @@ func (s *GenericFederatedStatus) update(generation int64, reason AggregateReason
...
 
 	propStatusUpdated := s.setPropagationCondition(reason, changesPropagated)
 
 	statusUpdated := generationUpdated || propStatusUpdated
+
+	klog.Infof("Values of propStatusUpdated '%v' statusUpdated '%v' changesPropagated '%v'", propStatusUpdated, statusUpdated, changesPropagated)
 	return statusUpdated
 }
 
 // setClusters sets the status.clusters slice from a propagation status
```
alenkacz, line 197: from propagation and resource status

alenkacz (commented):
I've done a first pass, it's mostly nits. On Monday I'll do another one as I think I understand the flow of the feature much better now :)

test/e2e/defaulting.go
hectorj2f: This is part of the tests, we want to enable it in the e2e tests for testing purposes.

pkg/features/features.go
hectorj2f: yep, you are right.

pkg/controller/sync/status/status.go
hectorj2f: Due to the uncertain nature of the status of the different resources, we have to use this unstructured status. That is explained in the KEP.

pkg/controller/federatedtypeconfig/controller.go
hectorj2f: Yes, this is the behavior of an existing feature that collects the status of Services, this function aims to check the conditions to determine if the old functionality is enabled.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -379,8 +374,98 @@ This might be especially useful for custom resource types without a `IsReady` st
...
+  targetType:
+    group: apps
+    kind: Deployment
+    pluralName: deployments
+    scope: Namespaced
+    version: v1
+```
+
+An example of a `FederatedDeployment` federated in a `cluster1` where the feature
+collected the resource remote status `remoteStatus`.
```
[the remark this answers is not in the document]
hectorj2f: `status: "True"` is part of the status of the [DeploymentStatus](https://godoc.org/k8s.io/api/apps/v1#DeploymentCondition) core resources.

test/common/crudtester.go
hectorj2f: good catch! This was part of an old change, it should not be a defer func.

docs/keps/20200619-federated-resource-status.md
```diff
@@ -379,8 +374,98 @@ This might be especially useful for custom resource types without a `IsReady` st
 By raw status, the system understand to show the status of all the federated resources `Ready` and not `Ready`.
 This can have an impact in the performance, so this should be configured with precautions.
 
-To do so, the system exposes properties as part of each `FederatedTypeConfig` to define
-the desired behavior at federated resource type.
+**Note that**, this proposed solution includes additional flags to the kubefed control-plane
+components to avoid blowing out the control-plane due to frequent and concurrent API calls to
```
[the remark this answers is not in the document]
hectorj2f: I re-phrased it!

pkg/controller/sync/dispatch/managed.go
```diff
@@ -299,15 +315,23 @@ func (d *managedDispatcherImpl) setResourcesUpdated() {
...
-	return status.CollectedPropagationStatus{
-		StatusMap:        statusMap,
-		ResourcesUpdated: d.resourcesUpdated,
+
+	for key, value := range d.resourceStatusMap {
+		resourceStatusMap[key] = value
 	}
+	return status.CollectedPropagationStatus{
+			StatusMap:        statusMap,
+			ResourcesUpdated: d.resourcesUpdated,
```
alenkacz, line 332: should this rather be a field of `CollectedPropagationStatus`?

alenkacz (changes requested):
Alright, did my second pass. I think the general direction makes sense (doing this in the sync controller where propagation status is already updated) - so 👍 .

I have couple of questions/concerns though and because of that not ready to approve.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -148,7 +148,14 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	corev1b1.SetFederatedTypeConfigDefaults(typeConfig)
 
 	syncEnabled := typeConfig.GetPropagationEnabled()
-	statusEnabled := typeConfig.GetStatusEnabled()
+	// TODO (hectorj2f): RawResourceStatusCollection is a new feature and is
```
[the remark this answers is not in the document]
hectorj2f: This is a todo because we hope to clean it in a short term future. So we don't need to keep the old status collection feature.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -433,3 +440,17 @@ func (c *Controller) reconcileOnNamespaceFTCUpdate() {
 		}
 	}
 }
+
+func (c *Controller) isEnabledFederatedServiceStatusCollection(tc *corev1b1.FederatedTypeConfig) bool {
+	if tc.GetStatusEnabled() && tc.Name == "services" {
+		federatedAPIResource := tc.GetFederatedType()
+		statusAPIResource := tc.GetStatusType()
+		if statusAPIResource == nil {
+			klog.Infof("Skipping status collection, status API resource is not defined for %q", federatedAPIResource.Kind)
```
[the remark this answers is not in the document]
hectorj2f: It is not an error or similar, it is just an additional information. Therefore it is not considered to be a warning.

pkg/controller/sync/dispatch/managed.go
hectorj2f: Can you elaborate more the reasoning behind this decision ?

pkg/controller/sync/status/status.go
```diff
@@ -71,8 +73,10 @@ const (
 )
 
 type GenericClusterStatus struct {
-	Name   string            `json:"name"`
-	Status PropagationStatus `json:"status,omitempty"`
+	Name         string            `json:"name"`
+	Status       PropagationStatus `json:"status,omitempty"`
+	RemoteStatus interface{}       `json:"remoteStatus,omitempty"`
+	//	Conditions []*metav1.Condition `json:"conditions,omitempty"`
```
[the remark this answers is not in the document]
hectorj2f: Yes, it is part of the KEP, but how stated in the PR description `This work is a first iteration`. Anyway I could remove it.

pkg/controller/sync/dispatch/managed.go
```diff
@@ -74,19 +75,24 @@ type managedDispatcherImpl struct {
 	fedResource           FederatedResourceForDispatch
 	versionMap            map[string]string
 	statusMap             status.PropagationStatusMap
```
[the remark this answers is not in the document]
hectorj2f: I didn't want to touch too much the current source code, and follow more an incremental approach. But I could rename it if you insist.

pkg/controller/sync/status/status.go
```diff
@@ -157,54 +168,66 @@ func (s *GenericFederatedStatus) update(generation int64, reason AggregateReason
...
+	if len(s.Clusters) != len(statusMap) || len(s.Clusters) != len(resourceStatusMap) {
+		klog.Info("Clusters differs from the size")
 		return true
 	}
 	for _, status := range s.Clusters {
 		if statusMap[status.Name] != status.Status {
 			return true
 		}
+		if !reflect.DeepEqual(resourceStatusMap[status.Name], status.RemoteStatus) {
+			klog.Infof("Clusters resource status differ: %v VS %v", resourceStatusMap[status.Name], status.RemoteStatus)
```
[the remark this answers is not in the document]
hectorj2f: I will decrease the verbosity but it'd be good to debug potential issues in this early stage of the feature. Once it is considered stable, we could remove it.

pkg/apis/core/v1beta1/federatedtypeconfig_types.go
```diff
@@ -207,18 +207,20 @@ func (f *FederatedTypeConfig) GetFederatedType() metav1.APIResource {
 	return apiResourceToMeta(f.Spec.FederatedType, f.GetFederatedNamespaced())
```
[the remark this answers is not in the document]
hectorj2f: Sorry, you propose to mark statusType as deprecated, where we agreed on that. I don't think we mentioned in the KEP to remove the code from the old mechanism.

pkg/controller/sync/dispatch/managed.go
```diff
@@ -235,31 +246,36 @@ func (d *managedDispatcherImpl) Update(clusterName string, clusterObj *unstructu
...
-	d.RecordStatus(clusterName, status.LabelRemovalTimedOut)
+	d.RecordStatus(clusterName, status.LabelRemovalTimedOut, clusterObj.Object[util.StatusField])
 
 	d.unmanagedDispatcher.RemoveManagedLabel(clusterName, clusterObj)
 }
 
 func (d *managedDispatcherImpl) RecordClusterError(propStatus status.PropagationStatus, clusterName string, err error) {
 	d.fedResource.RecordError(string(propStatus), err)
-	d.RecordStatus(clusterName, propStatus)
+	d.RecordStatus(clusterName, propStatus, "UnknownClusterError")
```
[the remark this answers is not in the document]
hectorj2f: The resource status for the federated resource does not have a known structure.

pkg/controller/sync/dispatch/managed.go
```diff
@@ -235,31 +246,36 @@ func (d *managedDispatcherImpl) Update(clusterName string, clusterObj *unstructu
...
+	if d.rawResourceStatusCollection && resourceStatus != nil {
+		klog.Infof("Recording resource status %v", resourceStatus)
+		d.resourceStatusMap[clusterName] = resourceStatus
+	}
 }
 
 func (d *managedDispatcherImpl) recordOperationError(propStatus status.PropagationStatus, clusterName, operation string, err error) util.ReconciliationStatus {
 	d.recordError(clusterName, operation, err)
-	d.RecordStatus(clusterName, propStatus)
+	d.RecordStatus(clusterName, propStatus, "UnknownOperationError")
```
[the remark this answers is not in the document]
hectorj2f: I am open to suggestions.

jimmidyson:
@hectorj2f I noticed that the default federated CRDs don't have `RemoteStatus` in their CRD, plus this is also not present when you `kubefedctl enable` a new resource for federation. This will also affect the migration to `apiextensions.k8s.io/v1` (cc @makkes for #1296).

pkg/controller/sync/dispatch/managed.go
alenkacz: well it's a field with the same value. Actually it looks like `ResourcesUpdated` on `CollectedResourceStatus` is never used. Not sure if it's bug or if it's just redundant. Might be redundant because it always has the same value as the other one 🤷

pkg/controller/sync/status/status.go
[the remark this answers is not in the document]
alenkacz: okay, my second part of the question is whether they are still needed now that we federate the whole status. The original idea was to federate only conditions but we stepped away from that.

pkg/controller/sync/dispatch/managed.go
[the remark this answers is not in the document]
alenkacz: it does not have explicit schema but as a consumer of the Federated api you would expect the schema to be the same as schema of the status of the underlying resource, right?

pkg/controller/sync/dispatch/managed.go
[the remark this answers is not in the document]
alenkacz: leaving it empty? :-) why is propagation status not enough in this case?

pkg/controller/sync/dispatch/managed.go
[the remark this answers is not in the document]
alenkacz: yeah, that makes sense. I am fine with doing this as a follow up :)

pkg/controller/sync/status/status.go
```diff
@@ -157,54 +168,66 @@ func (s *GenericFederatedStatus) update(generation int64, reason AggregateReason
...
+// maps. Returns a boolean indication of whether the status.clusters was
 // modified.
-func (s *GenericFederatedStatus) setClusters(statusMap PropagationStatusMap) bool {
-	if !s.clustersDiffers(statusMap) {
+func (s *GenericFederatedStatus) setClusters(statusMap PropagationStatusMap, resourceStatusMap map[string]interface{}) bool {
+	if !s.clustersDiffer(statusMap, resourceStatusMap) {
 		return false
 	}
 	s.Clusters = []GenericClusterStatus{}
 	for clusterName, status := range statusMap {
```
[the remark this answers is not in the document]
alenkacz: @hectorj2f ? :)

pkg/controller/federatedtypeconfig/controller.go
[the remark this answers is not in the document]
alenkacz: well in my head it's a configuration error, ideally such config will be rejected by admissionwebhook, so I thought that we should highlight that a bit more in the logs, WDYT?

pkg/apis/core/v1beta1/federatedtypeconfig_types.go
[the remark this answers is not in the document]
alenkacz: yes, I just wanted for this PR to enhance the docs above that field. Not remove anything

pkg/controller/sync/dispatch/managed.go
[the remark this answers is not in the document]
irfanurrehman: nil :-)
and the error should surface in `conditions`

pkg/controller/sync/status/status.go
irfanurrehman: I would recommend using `map[string]interface{}`, which is the format of the unstructured schema.

irfanurrehman (changes requested):

docs/keps/20200619-federated-resource-status.md
```diff
@@ -115,9 +115,9 @@ type GenericFederatedResource struct {
 }
 
 type GenericClusterStatus struct {
-	Name   string            `json:"name"`
-	Status PropagationStatus `json:"status,omitempty"`
-
+  Name   string                  `json:"name"`
+  Status PropagationStatus       `json:"status,omitempty"`
+  RemoteStatus interface{}       `json:"remoteStatus,omitempty"`
```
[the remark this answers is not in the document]
alenkacz: That said, it will be collected only where you wanna collect it - you still have to turn it on per type. That said, maybe this perf problems will be partially solved by a solution we were also discussing and that's not federating whole status but just a selected subset via jsonpath. That at least makes it possible to lower the performance impact.

But overall I feel your concerns regarding performance even though this is opt-in so current installations won't be affected.

pkg/controller/sync/dispatch/managed.go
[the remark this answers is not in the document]
alenkacz: yeah I think empty is much better than string from consumer standpoint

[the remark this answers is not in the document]
alenkacz: oh I see, yeah. I like the fact that conditions will be used for such cases! So +1

docs/keps/20200619-federated-resource-status.md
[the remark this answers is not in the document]
hectorj2f: Thanks for the review @irfanurrehman 

> What you have added in this PR can already be done via the status controller as I did mention in this comment on your KEP, but as a separate status resource.

@irfanurrehman Yes, we've deviated from the original KEP due to some recent requirements, but we also updated the original KEP as part of this PR. The reason is to store the status of the federated resources in the target clusters instead of using the original approach `FederatedServiceStatus`. As mentioned in the description, this PR is a first iteration of the original proposal, fetching the status from the federated resources does not determine its readiness, so there is some work to be done.

@irfanurrehman The main reason to exclude the current approach is the inefficiency of having to store/manage the status of each resource as an additional resource type and resource in the datastore.  I will share our thoughts about your concerns in the following:

> Performance: This is more of a question then anything else (as I don't have concrete numbers myself). I guess the CRUD on individual resources will be much more expensive, given all the additional statuses stored in it. This can be a problem as the sync reconciler already does a lot and is a very regular process.

Firstly, the feature is disabled by default for all the resources. When enabled, it'd be done for those specific resources. In addition to that, I decided to omit the status controller to carry on this new task, and instead rely on the dispatch controller which is already fetching the resource. We are already periodically fetching the resource, so we could simply reuse the same logic to consume its status.  

> Another problem, that I do remember with the raw status collection is user reports of cluster etcd being filled up to its limits, the reason why we did put this condition in place.  

This sounds more like a problem we had with etcd v2 where bolt stored data locally on each node. Etcdv3 has improved this behavior and the size of the database is relatively improved with new boltdb improvements.
 In any case we are planning to evaluate the performance cost of this change.

> Manually unreadable: Think getting a federated resource list and then trying to figure out federated fields among all the statuses.  

I agree but I assume this is a decision of the user to desire having the status of the resources as part of the federated resources. 

> Confusing statuses/so many statuses at the same place: As a user what I am really interested in is:- if my resource was propagated fine to all clusters and then, is my resource in ready/not ready state in my clusters, with some place I can go an refer to (for example the separate resourceStatus) if I see a problematic state.  

Yes, I agree with the terms here. But I see this feature as an addition to have a better visibility of the status of the federated resources to be aware of consume their status in centralized manner. I can see value out of this feature when having the status as part of the FederatedResource.

irfanurrehman:
@hectorj2f I see that you have pushed new changes. Have you completed your review rework?

hectorj2f:
@irfanurrehman Yes. But I want to share some graphs and numbers from a test we performed to compare existing vs current approach. I'll ping you this week, thanks for keeping one eye on the PR.

hectorj2f:
As promised, I am sharing the results of some local tests about the resource usage of the current kubefed status collection versus the additional remote status collection provided in this PR.

We run incremental tests over two different clusters running kubefed v0.5.0 and a build `kubefed_fed_remote_status` with the code of this PR. These tests created respectively N number of federated namespaces for which 2 federated services and 2 federated deployments were created.

`N FederatedNamespace => 2*N FederatedService and 2*N FederatedDeployments`

The clusters were both instrumented with Prometheus as monitoring tool to collect metrics, and at the same time cpu and heap profiles were captured.

I am sharing couple of charts that extracted from the global resource usage focusing on the SetFederatedStatus action whose logic slightly changed here.

In the first graph,  I show the CPU usage time (%) of the FederatedStatusAction for each one of the 4 tests:

Screen Shot 2020-11-27 at 3 09 05 PM

A flamegraph of the current kubefed version v0.5.0 using the cpu pprof file with 1000 federated resources:
Screen Shot 2020-11-27 at 3 26 58 PM

And a flamegraph of the proposed solution using the cpu pprof file with 1000 federated resources:

Screen Shot 2020-11-27 at 3 27 08 PM


The second chart highlights the percentage of memory allocation to the FederatedStatusAction. There is a small and obvious increment due to the memory allocation to store the remote status from the resources. This increment is overall extra 60MB of memory consumption when handling 3000 federated resources with their status.


Screen Shot 2020-11-27 at 3 08 59 PM

Both flamegraphs about the allocatted memory space based on the heap pprof file:

Screen Shot 2020-11-27 at 1 22 59 PM

and

Screen Shot 2020-11-27 at 1 23 06 PM



The following chart shows the memory in-use consumed by the json.DecodeState action. This value has to be higher because the size of the decoding data is bigger as well. However the growth of both approach remains parallel and following a similar increment.

Screen Shot 2020-11-27 at 3 08 45 PM

Note that, a cooling down period was reserved after each test to give some time to the system to stabilize.

jimmidyson:
@hectorj2f Thank you so much for that in-depth performance benchmarking. I don't see anything of concern there. One thing that I don't see covered though is when resources are propagated to multiple clusters. It would be good to see what happens when propagating resources to 10, 50, 100, 200 clusters. Even without that, I am happy with this approach and would like to see it merged.

hectorj2f:
@jimmidyson I agree those tests are missing. But the degradation would be proportional to the amount of clusters in a linear manner to how the system behaves today under these circumstances. 
Of course, this feature won't avoid ending into a situation where a FederatedResource placed across 1k clusters will face similar issues to those the core API type Endpoints faced (a consequence to the introduction of EndpointSlice). In other words, the scale of this feature might be affected by the amount of clusters to which you federate the resource and the size of the remote status property.

hectorj2f:
I will look at the tests, since I rebased last time it is failing 🤔.

irfanurrehman:
Thanks @hectorj2f for doing this and apologies for missing your update earlier.
Its satisfactory to know that the overhead increase while collecting status has a linear corelation with the number of resources. The corelation should ideally be similar when collecting status from multiple clusters.
I think this is good to go. I can do a final pass after you confirm that you have finished your changes.

hectorj2f:
@irfanurrehman Yes, it is done. Please, feel free to re-review it. Thanks

alenkacz (approved):
LGTM! I would love to do some follow-ups around naming stuff but otherwise this looks great. I also tested it locally and it's working 👏

docs/development.md
```diff
@@ -98,6 +98,8 @@ types enabled by default:
 1. a label update is reflected in the objects stored in the target
    clusters.
 1. a placement update for the object is reflected in the target clusters.
+1. optionally if `RawResourceStatusCollection` feature is enabled, tests check the value
```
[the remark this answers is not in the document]
hectorj2f: That is mentioned in the doc. A feature is enabled in the kubefedconfig file as done for other features. But I also mention that the status need to be enabled at the specific target resource.

pkg/controller/federatedtypeconfig/controller.go
```diff
@@ -148,7 +148,10 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	corev1b1.SetFederatedTypeConfigDefaults(typeConfig)
 
 	syncEnabled := typeConfig.GetPropagationEnabled()
-	statusEnabled := typeConfig.GetStatusEnabled()
+	// NOTE (hectorj2f): RawResourceStatusCollection is a new feature and is
+	// Disabled by default. When RawResourceStatusCollection is enabled,
```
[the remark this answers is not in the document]
hectorj2f: It is disabled by default, check features.go.

pkg/controller/sync/controller.go
```diff
@@ -405,19 +416,20 @@ func (s *KubeFedSyncController) setFederatedStatus(fedResource FederatedResource
 	// If the underlying resource has changed, attempt to retrieve and
 	// update it repeatedly.
 	err := wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
-		if updateRequired, err := status.SetFederatedStatus(obj, reason, *collectedStatus); err != nil {
+		if updateRequired, err := status.SetFederatedStatus(obj, reason, *collectedStatus, *collectedResourceStatus, resourceStatusCollection); err != nil {
+			klog.Infof("Failed to set the status for %s %q", kind, name)
 			return false, errors.Wrapf(err, "failed to set the status")
 		} else if !updateRequired {
-			klog.V(4).Infof("No status update necessary for %s %q", kind, name)
+			klog.Infof("No status update necessary for %s %q", kind, name)
```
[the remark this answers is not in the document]
hectorj2f: I'll change that log. 👍🏻

pkg/controller/sync/status/status.go
```diff
@@ -139,6 +147,8 @@ func SetFederatedStatus(fedObject *unstructured.Unstructured, reason AggregateRe
 	if err != nil {
 		return false, errors.Wrapf(err, "Failed to marshall generic resource json to unstructured")
 	}
+
+	klog.Infof("Setting the status of federated object '%v' and resource object '%v'", fedObject.GetName(), resourceObj.GetName())
```
irfanurrehman, line 151: I still see many `per resource` messages at loglevel 0. As mentioned in [this comment](https://github.com/kubernetes-sigs/kubefed/pull/1292#discussion_r504540891) it does not make sense to have per resource messages at loglevel zero. This has the potential  of flooding the logs for no reason. There would be no way for the user to reduce the logs generated if debug messages as above are set at this level. These should be level 4 or above.

Only once in a lifecycle message, for example controller being initiated/shutdown can be at this level.

Please update across this PR.

irfanurrehman: If there are some errors in collecting status or setting that, that you want  reported, I think that can be reported as level  2 info. The actual error should eventually surface in the status conditions of the resource.

irfanurrehman (changes requested):
@hectorj2f Thanks for doing this. I have one observation which remains open. Please update that and then we can merge.

pkg/controller/sync/status/status.go
hectorj2f: @irfanurrehman I changed the log verbosity accordingly. Mainly, I set all to V(4).

irfanurrehman:
Thanks @hectorj2f for doing this.
/lgtm

irfanurrehman (approved):

PR #1294, opened 2020-10-09
fix: prefer core resources when enabling a type

### LINKED ISSUE #1078: kubedctl enable services has conflict with knative services
**What happened**:

run `$ ./bin/kubefedctl enable services --kubefed-namespace default`
returns `
F0802 15:16:29.293248   61760 enable.go:110] Error: Multiple resources are matched by "services": services, services.serving.knative.dev. A group-qualified plural name must be provided.`

The reason is that i installed knative services crd before. However, we cannot enable kube services using command line, since it has no group.

install knative services and run kubefedctl enable services

**Environment**:
- KubeFed version
- Others

/kind bug

makkes:
When a resource type such as `pods` is `services` is passed to
`kubefedctl enable` and there's a match for this name in the `core`
group as well as another group, there was no way to tell `kubefedctl`
that the `core` resource should be enabled.

Now, whenever an API resource from the `core` group matches the
desired type to be enabled, this resource type is preferred.

Fixes #1078 

Here's the steps to test this:

-  `kind create cluster`
- `kubectl apply --filename https://github.com/knative/serving/releases/download/v0.18.0/serving-crds.yaml`
- `make kubefedctl`
- `./bin/kubefedctl enable services --kubefed-namespace="default" -o yaml`

Before this change:
```
$ ./bin/kubefedctl enable services --kubefed-namespace="default" -o yaml
F1012 22:22:47.217201   66506 enable.go:111] Error: Multiple resources are matched by "services": services, services.serving.knative.dev. A group-qualified plural name must be provided.
```

With this change:
```
$ ./bin/kubefedctl enable services --kubefed-namespace="default" -o yaml
---
apiVersion: core.kubefed.io/v1beta1
kind: FederatedTypeConfig
[...]
```

### DIFF
--- a/docs/userguide.md
+++ b/docs/userguide.md
@@ -128,6 +128,12 @@ The `<target kubernetes API type>` can be any of the following
 
 for the intended target API type.
 
+**NOTE:** To specifically target a type of a certain API group, you can append the group's name as in `deployment.apps`.
+For core Kubernetes types such as `Pod` or `Service` this is not possible because they have no specific API group assigned
+to them. Therefore, core types always take precendence over other types of the same name from another API group. Please
+see the [Enabling an API type with a non-default API group](#enabling-an-api-type-with-a-non-default-api-group) section
+for more information on the impact of possible naming conflicts.
+
 The `kubefedctl` command will create
  - a CRD for the federated type named `Federated<Kind>`
  - a `FederatedTypeConfig` in the KubeFed system namespace with the group-qualified plural name of the target type.
--- a/pkg/kubefedctl/enable/util.go
+++ b/pkg/kubefedctl/enable/util.go
@@ -106,6 +106,13 @@ func LookupAPIResource(config *rest.Config, key, targetVersion string) (*metav1.
 					targetResource = resource.DeepCopy()
 					targetResource.Group = group
 					targetResource.Version = gv.Version
+					// a resource from the 'core' group matches the 'key', so we prefer
+					// that one over any other because otherwise there would be no way
+					// for the user to signify that he wants to enable federation of
+					// a specific core group resource.
+					if targetResource.Group == "" {
+						return targetResource, nil
+					}
 				}
 				matchedResources = append(matchedResources, groupQualifiedName(resource.Name, gv.Group))
 			}


irfanurrehman:
@makkes thanks for doing this.
The implicitly selection of `core` when no group is specified and the core group is in the matches, seems fine to me.
However, it should be documented simply because its implicit. I think a good place to put it would be [here](https://github.com/kubernetes-sigs/kubefed/blob/master/docs/userguide.md#enabling-federation-of-an-api-type) in someway referring to details mentioned [here](https://github.com/kubernetes-sigs/kubefed/blob/master/docs/userguide.md#enabling-an-api-type-with-a-non-default-api-group).
Can you please do that.

jimmidyson:
I wonder if we should do the same as kubectl does and follow API group/version priority when short names are used?

makkes:
@jimmidyson do you have a hint where I would find out how kubectl does it?

jimmidyson:
@makkes Check out `k8s.io/client-go/discovery` package, which allows you to list server resources.

makkes:
> I wonder if we should do the same as kubectl does and follow API group/version priority when short names (or unqualified as in this case) are used?

We're already using `ServerPreferredResources` to fetch resources to consider so already following `kubectl`'s approach with regards to order. The only difference being that we error out when multiple resources match. In the service of keeping backwards-incompatible changes to a minimum, I would prefer to leave this PR as it is and follow-up with a PR that changes semantics to align with `kubectl`'s approach more. 

@jimmidyson wdyt?

jimmidyson:
This PR effectively puts core v1 API as higher priority than anything else, which is reasonable and fixes the reported bug, so I'm happy to approve this PR as-is and follow up with a PR that aligns with kubectl semantics. I agree with @irfanurrehman  that this should be documented and then we can get this merged.

I think we can clean up the code quite a bit by using discovery client smarter, something like:

```go
restMapper := restmapper.NewDeferredDiscoveryRESTMapper(discoveryClient)
shortcutRestMapper := restmapper.NewShortcutExpander(restMapper, discoveryClient)

gvks, err := shortcutRestMapper.KindsFor(schema.GroupVersionResource{Group: optionalGroup, Version: optionalVersion, Resource: resource})
```

makkes:
I agree and will add documentation soon!

makkes:
@irfanurrehman @jimmidyson I added a section to the user guide.

PR #1298, opened 2020-10-12
chore: disable CrossClusterDiscovery feature by default

hectorj2f:

Fixes #

By the reactions in GH https://github.com/kubernetes-sigs/kubefed/issues/1283, this feature does not seem to be used. It is an alpha feature that is handling certain features that should not be tackled by kubefed, but instead use other tools such as Istio.
I am disabling it by default for now. Note that, this feature is in an alpha stage so it could be deleted in the future reduce the maintenance burden.

related to https://github.com/kubernetes-sigs/kubefed/issues/1283


hectorj2f:
The in-memory controller errors are failing. I'll check how to enable the feature only for those tests.

irfanurrehman:
Thanks @hectorj2f 
This seems like a valid thing to do as of now.
/lgtm

PR #1299, opened 2020-10-13
fix issue #1285

### LINKED ISSUE #1285: service dns controller ignores err while reconciling
**What happened**:
```pkg/controller/servicedns/controller.go```
```
	if err != nil {
		runtime.HandleError(errors.Wrap(err, "Failed to get ready cluster list"))
		return util.StatusError
	}

	fedDNSStatus := make([]dnsv1a1.ClusterDNS, 0, len(clusters))
	// Iterate through all ready clusters and aggregate the service status for the key
	for _, cluster := range clusters {
		if cluster.Status.Region == nil || *cluster.Status.Region == "" || len(cluster.Status.Zones) == 0 {
			runtime.HandleError(errors.Wrapf(err, "Cluster %q does not have Region or Zones Attributes", cluster.Name))
			return util.StatusError
		}
		clusterDNS := dnsv1a1.ClusterDNS{
			Cluster: cluster.Name,
			Region:  *cluster.Status.Region,
			Zones:   cluster.Status.Zones,
		}
```
Because err is nil, so runtime.HandleError function will not handle the err.
It should print the err.
If one of the cluster does not have region or zone, the err happens and will be ignored.

**Environment**:
v1.17.9
- KubeFed version
v0.4.0
- Others

/kind bug

yuswift:
Display more clear logs.
Fixes #1285 

### DIFF
--- a/pkg/controller/servicedns/controller.go
+++ b/pkg/controller/servicedns/controller.go
@@ -308,7 +308,7 @@ func (c *Controller) reconcile(qualifiedName util.QualifiedName) util.Reconcilia
 	// Iterate through all ready clusters and aggregate the service status for the key
 	for _, cluster := range clusters {
 		if cluster.Status.Region == nil || *cluster.Status.Region == "" || len(cluster.Status.Zones) == 0 {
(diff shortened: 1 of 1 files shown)


jimmidyson:
/lgtm
/approve

Thanks @yuswift!

PR #1301, opened 2020-10-14
chore: upgrade kubernetes dependencies

sagikazarmark:
Signed-off-by: Mark Sagi-Kazar <mark.sagikazar@gmail.com>

Upgrades the Kubernetes libraries to 1.19.2
Keeping libraries up to date is generally a good idea. I'm not aware of any major changes that might break backwards compatibility.

Fixes #

The logr package needed a manual update because of container-runtime. Once 0.7.0 is out, we don't need to manually pin the logr library version.

Some other libraries have also been updated because they are also transitive dependencies.


sagikazarmark:
Hm, can't see any obvious reason why the tests are breaking...

scripts/pre-commit.sh
```diff
@@ -149,7 +149,7 @@ echo "Downloading e2e test dependencies"
 ./scripts/download-e2e-binaries.sh
 
 CREATE_INSECURE_REGISTRY=y CONFIGURE_INSECURE_REGISTRY_HOST=y \
-    KIND_TAG="v1.18.4" ./scripts/create-clusters.sh
+    KIND_TAG="v1.19.2" ./scripts/create-clusters.sh
```
jimmidyson, line 152: There is only a [v1.19.1 tag available](https://hub.docker.com/r/kindest/node/tags?page=1&name=v1.19) here so please use that.
```suggestion
    KIND_TAG="v1.19.1" ./scripts/create-clusters.sh
```

go.mod
```diff
@@ -3,26 +3,28 @@ module sigs.k8s.io/kubefed
...
+	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
-	k8s.io/api v0.18.6
-	k8s.io/apiextensions-apiserver v0.18.6
-	k8s.io/apimachinery v0.18.6
-	k8s.io/apiserver v0.18.6
-	k8s.io/client-go v0.18.6
-	k8s.io/component-base v0.18.6
+	k8s.io/api v0.19.2
```
jimmidyson, line 19: Now that `v1.19.3` is out we could update to `v0.19.3` here and other dependencies. Would you mind taking care of that?

jimmidyson (changes requested):
Thank you for this. Please fix the kind image and, if you don't mind, update to v0.19.3 dependencies, as well as running `go mod tidy` please to try to keep `go.sum` in a reasonable state.

scripts/pre-commit.sh
sagikazarmark: Done

go.mod
sagikazarmark: 1.19.3 is not available in all components yet (eg. in k8s.io/api). I guess we can hold off merging until it's available.

sagikazarmark:
@jimmidyson thanks for the review. I addressed/replied to your comments.

I actually ran `go mod tidy`, can you see anything unusual?

go.mod
jimmidyson: Ah yeah good point - thanks for checking that out.

jimmidyson:
@sagikazarmark I was surprised there weren't a lot more removals from `go.sum`. It's not a blocker though.

jimmidyson (approved):

sagikazarmark:
I'll take a look at it. My guess is there is a transitive dependency on an earlier Kubernetes version.

sagikazarmark:
Controller runtime for instance still relies on 1.18.6

jimmidyson:
@sagikazarmark Ah ok that makes sense

makkes (approved):

makkes (approved):

jimmidyson:
@sagikazarmark Sorry for the bother but would you mind rebasing this against master and dropping the last two commits in this branch (only need the updated kubernetes dependencies)? Hopefully that should sort out the CI issues.

sagikazarmark:
@jimmidyson I've just rebased it, looks like at the same time you did. 😄

jimmidyson:
Oh I'm sorry @sagikazarmark, I was just trying to help out to get it in, hope that's ok. I'm going to push another commit just to bump the dependencies to latest now they're published, then we can get this merged. Thank you for kicking this upgrade off.

sagikazarmark:
Sure, no worries. 😄 Let me know if I can help. Thanks!

sagikazarmark:
Looks like it fails with the same error 😕

jimmidyson:
@sagikazarmark My dodgy rebase - I knew I should have waited for you :wink:

sagikazarmark:
Hm, I'm not sure, the rebase looks fine to me. Only `go.mod` and `go.sum` have changes.

jimmidyson:
Yeah it seems to be a consistent test failure.

jimmidyson:
To test the one failing test you can run (after spinning up some clusters and deploying kubefed - `scripts/create-clusters.sh` and `make deploy.kind` should help I think):

```shell
$ make bin/e2e
$ bin/e2e -kubeconfig=${HOME}/.kube/config -ginkgo.v -single-call-timeout=1m -ginkgo.trace -ginkgo.randomizeAllSpecs -ginkgo.focus="deployments.apps.* should be created, read, updated and deleted successfully"
```

sagikazarmark:
I can see a bunch of `ERROR: Cluster object is not as expected` errors caused by an extra `"deployment.kubernetes.io/revision": "1"` annotation in the object.

jimmidyson:
Yeah it looks like a bug in the jsonpatch upgrade (test below).

```go
package a

import (
	"encoding/json"
	"fmt"
	"testing"

	jsonpatch "github.com/evanphx/json-patch"

	"sigs.k8s.io/kubefed/pkg/controller/util"
)

func TestPatch(t *testing.T) {
	orig := &unstructured.Unstructured{
		Object: map[string]interface{}{
			"metadata": map[string]interface{}{
				"annotations": map[string]interface{}{
					"deployment.kubernetes.io/revision": "1",
				},
			},
		},
	}
	origJSON, _ := orig.MarshalJSON()

	overrides := util.ClusterOverrides{{
		Op:   "add",
		Path: "/metadata/annotations",
		Value: map[string]string{
			"foo": "bar",
		},
	}, {
		Path: "/metadata/annotations/foo",
		Op:   "remove",
	}}
	jsonPatchBytes, _ := json.Marshal(overrides)
	patch, _ := jsonpatch.DecodePatch(jsonPatchBytes)

	patchedObjectJSONBytes, _ := patch.Apply(origJSON)
	fmt.Println(string(patchedObjectJSONBytes))
}
```

jimmidyson:
Or potentially a fix in the JSONPatch library... I think this is it tbh.

jimmidyson:
Ah yes this is because the override is wrong... fixing now.

jimmidyson:
OK fixed. The override was actually now invalid...

jimmidyson:
Thanks @sagikazarmark!

sagikazarmark:
Thanks for the help @jimmidyson

PR #1303, opened 2020-10-19
chore: Use Kubernetes v1.19.1 and kind v0.9.0 for testing

jimmidyson:

Upgrades to use Kind v0.9.0 and Kubernetes v1.19.1. This is to unblock #1301 as a separate PR.

Fixes #

I haven't tested this on a Mac so help there would be appreciated. @makkes I see you've previously touched these files and I'm pretty sure you're a Mac user - would you be able to help our here? This can be tested with

```shell
$ DOWNLOAD_BINARIES=y bash -x ./scripts/pre-commit.sh`
```

and cleaned up with

```shell
DELETE_INSECURE_REGISTRY=y ./scripts/delete-clusters.sh
```

### DIFF
--- a/scripts/create-clusters.sh
+++ b/scripts/create-clusters.sh
@@ -90,11 +90,15 @@ EOF
 "
       ;;
     $containerd_config)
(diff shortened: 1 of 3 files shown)


makkes:
The script doesn't work on macOS even on the current `master` branch. @jimmidyson and I talked quickly and I'll try to come up with a fix.

PR #1307, opened 2020-10-21
Fix up CRD generation

jimmidyson:

CRD generation was broken because the script did not copy across the generated CRDs to the correct final location. At the same time as fixing that, I've taken the opportunity to ensure that the specified version of controller-gen is used to ensure we do not have issues caused by different versions of controller-gen causing hard to diagnose conflicts.

### DIFF
--- a/charts/kubefed/charts/controllermanager/crds/crds.yaml
+++ b/charts/kubefed/charts/controllermanager/crds/crds.yaml
@@ -1,3 +1,5 @@
+
+
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
@@ -78,7 +80,6 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
----
 
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
@@ -161,7 +162,6 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
----
 
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
@@ -335,7 +335,6 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
----
 
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
@@ -475,7 +474,6 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
----
 
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
@@ -606,7 +604,6 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
----
 
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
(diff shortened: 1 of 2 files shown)


scripts/sync-up-helm-chart.sh
```diff
@@ -47,17 +47,18 @@ fi
...
-echo "---" > ${TEMP_CRDS_YAML}
+echo "" > ${TEMP_CRDS_YAML}
 for filename in ./config/crds/*.yaml; do
   # Remove unwanted kubebuilder annotation
-   sed '/controller-gen.kubebuilder.io/d; /annotations:/d' $filename >> ${TEMP_CRDS_YAML}
-  echo "---" >> ${TEMP_CRDS_YAML}
+   ${SED} '/controller-gen.kubebuilder.io/d; /annotations:/d' $filename >> ${TEMP_CRDS_YAML}
 done
 
+mv ${TEMP_CRDS_YAML} ./charts/kubefed/charts/controllermanager/crds/crds.yaml
```
jimmidyson, line 60: This is the line that fixes the generation issue (other changes in this file are either cosmetic or to ensure specific version of controller-gen is used in all environments).

charts/kubefed/charts/controllermanager/crds/crds.yaml
```diff
@@ -1075,9 +1067,9 @@ spec:
     kind: ReplicaSchedulingPreference
     listKind: ReplicaSchedulingPreferenceList
     plural: replicaschedulingpreferences
-    singular: replicaschedulingpreference
     shortNames:
     - rsp
+    singular: replicaschedulingpreference
```
jimmidyson, line 1072: This proves that generation is working correctly now. This was manually added incorrectly (bad key ordering) in https://github.com/kubernetes-sigs/kubefed/pull/1279 and because the generated CRDs were never moved to the correct location, validation didn't pick up on this.

makkes (approved):

scripts/sync-up-helm-chart.sh
```diff
@@ -47,17 +47,18 @@ fi
 rm -f ./config/crds/*.yaml
 
 # Generate CRD manifest files
-command -v controller-gen &> /dev/null || (cd ${ROOT_DIR}/tools && go install sigs.k8s.io/controller-tools/cmd/controller-gen)
-controller-gen crd:trivialVersions=true paths="./pkg/apis/..." output:crd:artifacts:config=config/crds
+(cd ${ROOT_DIR}/tools && GOBIN=${ROOT_DIR}/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen)
```
alenkacz, line 50: so a dumb question, how do we ensure the version? 🤔

jimmidyson: Using a separate tools module (https://github.com/kubernetes-sigs/kubefed/blob/master/tools/go.mod#L7).

hectorj2f (commented):
/approve

hectorj2f (commented):
/lgtm

Issue #1296, opened 2020-10-12
Transition from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1


makkes:
**What happened**:
When installing the Kubefed CRDs into a cluster, the API server will issue a warning:
```
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
```

**What you expected to happen**:
No warnings

**How to reproduce it (as minimally and precisely as possible)**:
```
kubectl create -f charts/kubefed/crds/crds.yaml
```

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.19.1
- KubeFed version: `master`
- Scope of installation (namespaced or cluster)
- Others

/kind bug

PR #1308, opened 2020-10-22
Upgrade tools

jimmidyson:

Upgrades tools (helm, controller-gen, golangci-lint) to latest, as well as go 1.15.3.

By virtue of upgrading `controller-gen`, the generated CRDs have converted to `apiextensions.k8s.io/v1` which is in line with the intention of #1296 (although doesn't fix it because that also requires the default and generated federated type CRDs to be `apiextensions.k8s.io/v1`.


hectorj2f (approved):
/lgtm

PR #1306, opened 2020-10-20
chore: Remove need for insecure registry

jimmidyson:
**What this PR does / why we need it**:

Removes the need for the insecure registry deployment and configuration from running e2e tests locally with kind.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #496.

**Special notes for your reviewer**:

PR #1309, opened 2020-10-22
fix: `make deploy.kind` didn't work, anymore.

makkes:
After #1306 has been merged, `make deploy.kind` broke because it doesn't provide the`KIND_CLUSTER_NAME` variable. An empty default provided to `kind`'s `--name` parameter doesn't make much sense because `kind` will simply error out in that case so using a default of `kind` for the name is the solution that should work most of the time, especially when following the quick start guide in the README.md.

Fixes #


jimmidyson:
Oops! Thanks @makkes!

/approve
/lgtm

PR #1315, opened 2020-11-07
fix: klog verbosity detection

### LINKED ISSUE #1314: No log output
**What happened**:
add --v=X, X is log level. There is no log output for any level.

option --v=X works well

**Environment**:
- KubeFed version
master
- Others

/kind bug

hectorj2f:
Signed-off-by: Hector Fernandez <hfernandez@mesosphere.com>

The verbosity level was not initialized to klog anymore, so the level of verbosity was the default one instead of the specified in the flags.

Fixes #1314 

### DIFF
--- a/cmd/controller-manager/app/controller-manager.go
+++ b/cmd/controller-manager/app/controller-manager.go
@@ -93,16 +93,18 @@ member clusters and do the necessary reconciliation`,
 		},
 	}
 
-	// Add the command line flags from other dependencies(klog, kubebuilder, etc.)
-	cmd.Flags().AddGoFlagSet(flag.CommandLine)
+	flags := cmd.Flags()
+	opts.AddFlags(flags)
+	flags.StringVar(&healthzAddr, "healthz-addr", healthzDefaultBindAddress, "The address the healthz endpoint binds to.")
+	flags.StringVar(&metricsAddr, "metrics-addr", metricsDefaultBindAddress, "The address the metric endpoint binds to.")
+	flags.BoolVar(&verFlag, "version", false, "Prints the Version info of controller-manager.")
+	flags.StringVar(&kubeFedConfig, "kubefed-config", "", "Path to a KubeFedConfig yaml file. Test only.")
+	flags.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	flags.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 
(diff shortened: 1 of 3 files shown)


cmd/controller-manager/app/controller-manager.go
```diff
@@ -93,16 +93,18 @@ member clusters and do the necessary reconciliation`,
...
+	flags.StringVar(&healthzAddr, "healthz-addr", healthzDefaultBindAddress, "The address the healthz endpoint binds to.")
+	flags.StringVar(&metricsAddr, "metrics-addr", metricsDefaultBindAddress, "The address the metric endpoint binds to.")
+	flags.BoolVar(&verFlag, "version", false, "Prints the Version info of controller-manager.")
+	flags.StringVar(&kubeFedConfig, "kubefed-config", "", "Path to a KubeFedConfig yaml file. Test only.")
+	flags.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	flags.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+
+	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	klog.InitFlags(local)
+	flags.AddGoFlagSet(local)
```
jimmidyson, line 107: The pattern I've seen adopted more in https://github.com/kubernetes/kubernetes repo is like:

```
# First line of main
klog.InitFlags(nil)
...
# Do other flag stuff
...
cmd.AddGoFlagSet(flag.CommandLine)
```

hectorj2f: The pattern shared in this PR comes from the Kubernetes repo. Check how the initialize klog flags.

hectorj2f: Could you detail what you see wrong in the current pattern?

mYmNeo:
@hectorj2f I've tried, it works well. :+1:

cmd/controller-manager/app/controller-manager.go
makkes: Stupid question: Is there a difference between first initializing klog flags and then others as opposed to doing it the other way around? If not, I'm fine either way.

makkes (approved):

cmd/controller-manager/app/controller-manager.go
jimmidyson: Yeah either work, I was just pointing the more common approach (see https://github.com/kubernetes/kubernetes/search?q=%22klog.InitFlags%22) but it's not a biggie.

PR #1318, opened 2020-11-13
chore: make certain cert-manager properties configurable

hectorj2f:
Signed-off-by: Hector Fernandez <hfernandez@mesosphere.com>

While migrating from older cert-manager versions to the latest, we found that older cert-manager version set a default organization if the organization was empty. However the new cert-manager version ignores that default organization if it is not specified as part of the certificate fields. Therefore we should make these properties configurable for the ca generated certificate via cert-manager.

Fixes #


charts/kubefed/README.md
```diff
@@ -121,6 +121,10 @@ chart and their default values.
...
+| controllermanager.certManager.enabled             | Specifies whether to enable the usage of the cert-manager for the certificates generation.                                                                                      | false                           |
+| controllermanager.certManager.organizations       | Specifies the list of organizations to include in the cert-manager generated ca certificate.                                                                                    | []                              |
+| controllermanager.certManager.commonName     | Specifies the CN value for the cert-manager generated ca certificate.                                                                                                                | ca.webhook.kubefed              |
+| controllermanager.certManager.dnsNames       | Specifies the list of subject alternative names for the cert-manager generated ca certificate.                                                                                       | ["ca.webhook.kubefed"]          |
```
mhrabovcin, line 127: Can we somehow incorporte into these names the fact that these will only get applied to single Issuer configuration - `{{ .Release.Name}}-root-certificate`

hectorj2f: I added it under the rootCertificate element to be more clear about it.

mhrabovcin (approved):

armandgrillet (approved):

Issue #1278, opened 2020-09-02
retainReplicas failed


lmmmml:
**What happened**: 
When retainReplicas is set to true, the replicas of the member cluster deployment is still overrided by federatred sync controller.

**What you expected to happen**: 
When retainReplicas is set to true,  the replicas of the member cluster deployment cannot be change by modify the rsp.

**How to reproduce it (as minimally and precisely as possible)**: 
set the retainReplicas to true, and then change the totalReplicas of rsp.

**Anything else we need to know?**:
the order of these two functions mybe wrong, the  RetainClusterFields should be affer ApplyOverrides.
https://github.com/kubernetes-sigs/kubefed/blob/57a1852ce9db67cdca440687998c026bf878f142/pkg/controller/sync/dispatch/managed.go#L203
https://github.com/kubernetes-sigs/kubefed/blob/57a1852ce9db67cdca440687998c026bf878f142/pkg/controller/sync/dispatch/managed.go#L209

**Environment**:
- Kubernetes version (use `kubectl version`)
- KubeFed version
- Scope of installation (namespaced or cluster)
- Others

/kind bug

makkes:
@lmmmml thanks for filing this issue.  I was trying to reproduce your issue but wasn't available to so let's try to get on the same page:

I created a `FederatedDeployment` with `replicas: 2`. This causes a `Deployment` with `replicas: 2` to be created on the attached cluster, leading to two pods. Then scaling down the `Deployment` to 1 replica on the attached cluster will lead to only one pod and kubefed rapidly changing the `Deployment` again to 2 replicas.

Now adding the field `retainReplicas: true` to the `FederatedDeployment's` `spec` and then scaling down the `Deployment` on the attached cluster to 1 replica leads to 1 pod left and kubefed will not change that.

This is how it worked for me so I'd love to hear what went wrong on your side.

makkes:
Interestingly enough I was able to reproduce an error when setting `retainReplicas: false` again which should lead to the `Deployment` scaled up again but that didn't happen in my scenario. I'll look into this.

lmmmml:
> @lmmmml thanks for filing this issue. I was trying to reproduce your issue but wasn't available to so let's try to get on the same page:
> 
[...]

PR #1326, opened 2020-12-08
fix: value of retainReplicas isn't always respected

makkes:

When creating a `FederatedDeployment` with `retainReplicas` set to
`true`, the `spec.replicas` value of the target Deployment isn't
touched by kubefed. However, when changing that value on the
`FederatedDeployment` back to `false`, the
`spec.template.spec.replicas` value isn't propagated to the target
resource.

Therefore, the `util.ObjectNeedsUpdate` func now compares the `spec`
field of federated resources so that any changes in it are actually
taken into account when deciding whether to update the target resource
or not.

Refs #1278

I must admit I'm not sure whether this is the best way to solve the problem as it might impact performance given kubefed now compares the complete `spec` field.

### DIFF
--- a/pkg/controller/util/meta.go
+++ b/pkg/controller/util/meta.go
@@ -26,6 +26,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


hectorj2f:
@makkes did you manage to find the reason why the tests are failing ?

makkes:
> @makkes did you manage to find the reason why the tests are failing ?

I haven't had the time, yet. I'll try to track it down today or over the weekend.

PR #1332, opened 2020-12-18
Fix error on exiting

huiwq1990:

When use kubefedctl join cluster setting `error-on-existing=false` and `secret-name` have specify value, it will case secret exist error.  Join command demo:

```shell
kubefedctl join newdev --cluster-context newdev --host-cluster-context minikube --kubefed-namespace tpaas-federation --secret-name=minikube-newdev --error-on-existing=false -v 2
```

Fixes #

### DIFF
--- a/pkg/kubefedctl/join.go
+++ b/pkg/kubefedctl/join.go
@@ -261,7 +261,7 @@ func joinClusterForNamespace(hostConfig, clusterConfig *rest.Config, kubefedName
 	}
 
 	secret, caBundle, err := populateSecretInHostCluster(clusterClientset, hostClientset,
-		saName, kubefedNamespace, joiningNamespace, joiningClusterName, secretName, dryRun)
+		saName, kubefedNamespace, joiningNamespace, joiningClusterName, secretName, dryRun, errorOnExisting)
 	if err != nil {
 		klog.V(2).Infof("Error creating secret in host cluster: %s due to: %v", hostClusterName, err)
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/kubefedctl/join.go
```diff
@@ -157,9 +159,9 @@ func (j *joinFederation) Complete(args []string) error {
 		klog.Fatal("host-cluster-name must be set if the name of the host cluster context contains one of \":\" or \"/\"")
 	}
 
-	klog.V(2).Infof("Args and flags: name %s, host: %s, host-system-namespace: %s, kubeconfig: %s, cluster-context: %s, secret-name: %s, dry-run: %v",
+	klog.V(2).Infof("Args and flags: name %s, host: %s, host-system-namespace: %s, kubeconfig: %s, cluster-context: %s, secret-name: %s, error-on-existing: %v,  dry-run: %v",
```
makkes, line 162:
```suggestion
	klog.V(2).Infof("Args and flags: name %s, host: %s, host-system-namespace: %s, kubeconfig: %s, cluster-context: %s, secret-name: %s, error-on-existing: %t,  dry-run: %v",
```

makkes (commented):
I don't fully agree with the changes proposed herein as I believe the semantics and the implementation should be simpler:

* If `--error-on-existing` is set to `true` and the secret exists, return an error.
* If `--error-on-existing` is set to `false` and the secret exists, just update it.

My main concern is around the update logic that you're introducing here which in my opinion makes the interface more complicated for users to reason about (adding annotations) and error-prone, basically something that might not work as expected.

So in my humble opinion just updating the secret in any case if `error-on-existing` is set to `false` is just simpler and less error-prone.

huiwq1990:
@makkes if we don't auto trigger clustercontroller reconcile, the joinedcluster will use old token until it restart or somethingelse trigger reconcile, it will cause user find real reason if the token not work .

makkes:
maybe @hectorj2f or @jimmidyson might want to chime in here as well.

makkes (approved):

hectorj2f (commented):
/lgtm

PR #1335, opened 2020-12-30
structed Logs

wyyxd2017:
what I submmit
[Introducing Structured Logs](https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/)，replaced the Klog.infof to Klog.infoS

PR #1337, opened 2020-12-30
structed Logs

wyyxd2017:
what I submmit
[Introducing Structured Logs](https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/)，replaced the Klog.infof to Klog.infoS


xunpan:
Let's make things done in #1335

makkes:
I agree with @xunpan, let's do it all in #1335.

PR #1341, opened 2020-12-31
Migrate pkg/controller logs to structured logging

ariesliuwei:
Migrate logs of pkg/controller/schedulingmanager to structured logging


hectorj2f:
@ariesliuwei @xunpan Can you explain why there are three PRs to achieve the same goal ?

xunpan:
@hectorj2f 
I don't know the reason that why there are so many PRs for the same issue. Please see my comments below. My concern is that we only need 1 PR even we need to do this kind of things.

https://github.com/kubernetes-sigs/kubefed/pull/1335#issuecomment-752401931

hectorj2f:
Closing due to inactivity. Feel free to reopen if needed.

Issue #1313, opened 2020-10-31
Add e2e upgrade tests


makkes:
**What would you like to be added**: End-to-end test verifying upgrade behaviour between two different kubefed versions.

**Why is this needed**: 

* Follow-up to #1296 
* These tests could (should?) be run on each PR to ensure existing kubefed clusters don't break with a certain change.

/kind feature

hectorj2f:
I might take care of this test ;).

hectorj2f:
@makkes  We added a test to upgrade from a previous image to the latest in master.

makkes:
@hectorj2f would you mind linking it here and closing the issue then, please? Thanks!

hectorj2f:
@makkes it was added here https://github.com/kubernetes-sigs/kubefed/pull/1346/commits/8a435bc9eb59adf6712a812868077491527c12b4.

PR #1346, opened 2021-01-12
fix: broken upgrade path from previous versions

hectorj2f:
Signed-off-by: Hector Fernandez <hectorj@gmail.com>

The upgrade path from v0.5.0 to v0.6.0 was failing due to the following error:

```
Error: UPGRADE FAILED: cannot patch "kubefed" with kind KubeFedConfig: admission webhook "kubefedconfigs.core.kubefed.io" denied the request: spec.featureGates.name: Unsupported value: "RawResourceStatusCollection":
```

The running admissionWebhook was not upgraded by the time the new kubefedconfig was updated. Therefore we needed to add certain hooks to the created resources. Once I added those, I managed to upgrade from v0.3.0 --> v0.4.0 --> v0.5.0 --> v0.6.0 without any issue.

Fixes https://github.com/kubernetes-sigs/kubefed/issues/1333

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/kubefedconfig.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/kubefedconfig.yaml
@@ -24,8 +24,6 @@ spec:
 {{- if .Values.featureGates }}
   - name: PushReconciler
     configuration: {{ .Values.featureGates.PushReconciler | default "Enabled" | quote }}
-  - name: RawResourceStatusCollection
-    configuration: {{ .Values.featureGates.RawResourceStatusCollection | default "Disabled" | quote }}
   - name: SchedulerPreferences
     configuration: {{ .Values.featureGates.SchedulerPreferences | default "Enabled" | quote }}
   - name: CrossClusterServiceDiscovery
--- a/charts/kubefed/charts/controllermanager/templates/kubefedconfig.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/kubefedconfig.yaml
@@ -30,4 +30,7 @@ spec:
     configuration: {{ .Values.featureGates.CrossClusterServiceDiscovery | default "Disabled" | quote }}
   - name: FederatedIngress
     configuration: {{ .Values.featureGates.FederatedIngress | default "Disabled" | quote }}
+  # NOTE: Commented feature gate to fix https://github.com/kubernetes-sigs/kubefed/issues/1333
+  #- name: RawResourceStatusCollection
+  #  configuration: {{ .Values.featureGates.RawResourceStatusCollection | default "Disabled" | quote }}
 {{- end }}
--- a/charts/kubefed/charts/controllermanager/templates/post-install-job.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/post-install-job.yaml
@@ -1,0 +1,103 @@
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: "{{ .Release.Name }}-kubefed-config-hook"
+  namespace: "{{ .Release.Namespace }}"
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+data:
+  setup.sh: |-
+    #!/bin/bash
+    set -euo pipefail
+
+    kubectl patch kubefedconfig -n {{ .Release.Namespace }} kubefed --type='json' -p='[{"op": "add", "path": "/spec/featureGates", "value":[{"configuration": {{ .Values.featureGates.PushReconciler | default "Enabled" | quote }},"name":"PushReconciler"},{"configuration": {{ .Values.featureGates.CrossClusterServiceDiscovery | default "Disabled" | quote }},"name":"CrossClusterServiceDiscovery"},{"configuration": {{ .Values.featureGates.RawResourceStatusCollection | default "Disabled" | quote }},"name":"RawResourceStatusCollection"},{"configuration": {{ .Values.featureGates.FederatedIngress | default "Disabled" | quote }},"name":"FederatedIngress"},{"configuration": {{ .Values.featureGates.SchedulerPreferences | default "Enabled" | quote }},"name":"SchedulerPreferences"}]}]'
+
+    echo "Kubefedconfig patched successfully!"
+
+    kubectl rollout restart deployment/kubefed-controller-manager -n {{ .Release.Namespace }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ .Release.Name }}-{{ randAlphaNum 10 | lower }}"
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "-4"
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+spec:
+  template:
+    metadata:
+      name: "{{ .Release.Name }}"
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never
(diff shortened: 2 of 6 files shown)


charts/kubefed/charts/controllermanager/templates/deployments.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     kubefed-control-plane: controller-manager
 {{- if .Values.controller.annotations }}
   annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
```
jimmidyson, line 12: Won't `hook-succeeded` delete the hook after hook succeeds in being created?

charts/kubefed/charts/controllermanager/templates/deployments.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     kubefed-control-plane: controller-manager
 {{- if .Values.controller.annotations }}
   annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
```
jimmidyson, line 10: Marking the deployment as a hook doesn't feel right to me. This will not be cleaned up on helm uninstall.

hectorj2f: I just ran helm uninstall, and the resources were all cleaned up 🤔.

hectorj2f: Okay, I see what is happening.

charts/kubefed/charts/controllermanager/templates/deployments.yaml
hectorj2f: Reverting these changes!

charts/kubefed/charts/controllermanager/templates/kubefedconfig.yaml
```diff
@@ -24,12 +24,13 @@ spec:
...
-    configuration: {{ .Values.featureGates.RawResourceStatusCollection | default "Disabled" | quote }}
   - name: SchedulerPreferences
     configuration: {{ .Values.featureGates.SchedulerPreferences | default "Enabled" | quote }}
   - name: CrossClusterServiceDiscovery
     configuration: {{ .Values.featureGates.CrossClusterServiceDiscovery | default "Disabled" | quote }}
   - name: FederatedIngress
     configuration: {{ .Values.featureGates.FederatedIngress | default "Disabled" | quote }}
+  # NOTE: Commented feature gate to fix https://github.com/kubernetes-sigs/kubefed/issues/1333
+  #- name: RawResourceStatusCollection
+  #  configuration: {{ .Values.featureGates.RawResourceStatusCollection | default "Disabled" | quote }}
```
makkes, line 35: * Does this completely remove the effects of setting this feature gate?
* When can we enable this again?

hectorj2f: Generally, this won't disable the feature. With the current code, if you set it in values to true, the hook will enable it. I had to explicitly remove it here to avoid older webhook to fail and so the helm upgrade. But anyone could now upgrade from v0.5.0 to v0.6.0 enabling it.
It is more of a hidden it, rather than disabling it.

hectorj2f: Wdyt?

hectorj2f:
Ping @jimmidyson

charts/kubefed/charts/controllermanager/templates/post-install-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation
+data:
+  setup.sh: |-
+    #!/bin/bash
+    set -euo pipefail
+
+    kubectl patch kubefedconfig -n {{ .Release.Namespace }} kubefed --type='json' -p='[{"op": "add", "path": "/spec/featureGates", "value":[{"configuration": {{ .Values.featureGates.PushReconciler | default "Enabled" | quote }},"name":"PushReconciler"},{"configuration": {{ .Values.featureGates.CrossClusterServiceDiscovery | default "Disabled" | quote }},"name":"CrossClusterServiceDiscovery"},{"configuration": {{ .Values.featureGates.RawResourceStatusCollection | default "Disabled" | quote }},"name":"RawResourceStatusCollection"},{"configuration": {{ .Values.featureGates.FederatedIngress | default "Disabled" | quote }},"name":"FederatedIngress"},{"configuration": {{ .Values.featureGates.SchedulerPreferences | default "Enabled" | quote }},"name":"SchedulerPreferences"}]}]'
```
jimmidyson, line 16: I'm a bit concerned about this patch. The `op` is `add` for `path` `/spec/featureGates` but that is already present in https://github.com/kubernetes-sigs/kubefed/pull/1346/files#diff-9c32952a289abd6043f1083b7296ab530b31dbbcf0e7ad35f36d61e13393372aR23 if feature gates are specified. What is the behaviour here with `add` operation? I expect that to fail.

jimmidyson: Ah [JSON patch spec](https://tools.ietf.org/html/rfc6902#section-4.1) details behaviour if field already exists:

> If the target location specifies an object member that does exist, that member's value is replaced.

jimmidyson:
@hectorj2f I'm really not confident with this approach. We need an upgrade test to confirm - can you add one please?

charts/kubefed/charts/controllermanager/templates/post-install-job.yaml
hectorj2f: No, that won't fail. In this case, we set the value of `/spec/featureGates`. 

```
If the object targetted by route exists is replaced, if it does not exist it creates it and if it is an Array it creates it in the indicated position in the Array, and the element “-” is used to add new elements to an Array.
```
We re-create the value of featureGates here to my understanding.
```
$ kubectl logs -f kubefed-7qzkgwlapd-b5rt8 -n kube-federation-system2
kubefedconfig.core.kubefed.io/kubefed patched
Kubefedconfig patched successfully!
```

hectorj2f:
@jimmidyson I add an initial test to validate the upgrade. Is that enough for you or you are looking for a specific piece of code ?

scripts/pre-commit.sh
```diff
@@ -38,6 +38,9 @@ E2E_TEST_CMD="${TEMP_DIR}/e2e-${PLATFORM} ${COMMON_TEST_ARGS}"
 # given control plane scope.
 IN_MEMORY_E2E_TEST_CMD="go test -v -timeout 900s -race ./test/e2e -args ${COMMON_TEST_ARGS} -in-memory-controllers=true -limited-scope-in-memory-controllers=false"
 
+KUBEFED_UPGRADE_TEST_NS="upgrade-test"
+KUBEFED_UPGRADE_TEST_VERSION="v0.5.0"
```
jimmidyson, line 42: This will have to be bumped for every release, right? Can we update documentation for this or better still automate this, either in the test (retrieve latest release here) or as part of release script?

hectorj2f: I automated to get the previous version to the latest stable in the remote chart.

hectorj2f:
The build failed due to a long job that was stopped during the cleanup step. I simplified the upgrade test to take less time.

hectorj2f:
@makkes @jimmidyson please, take a look. I think I addressed all your comments.

scripts/pre-commit.sh
```diff
@@ -62,6 +64,43 @@ function run-e2e-tests() {
...
 
+function run-e2e-upgrade-test() {
+  HOST_CLUSTER="$(kubectl config current-context)"
+
+  echo "Adding a repo to install an older kubefed version"
+  helm repo add kubefed-charts https://raw.githubusercontent.com/kubernetes-sigs/kubefed/master/charts
+  helm repo  update
+
+  # Get the previous version prior to our latest stable version
+  KUBEFED_UPGRADE_TEST_VERSION=$(helm search repo kubefed-charts/kubefed  --versions | awk '{print $2}' | head -3 | tail -1)
```
makkes, line 75: For PRs this would currently test an upgrade from 0.5.1 to what's in the PR. However, I believe that for PRs we would like to test upgrades from the latest version (i.e. 0.6.0 atm). wdyt?

jimmidyson: This was testing from the old version 0.5.1 that showed the issue that this PR fixes. Generally, we should run upgrades from the most recent release, but the upgrade from 0.5.1 -> 0.6.0 is broken right now, hence this test.

I guess we should actually remove 0.6.0 once 0.6.1 is released as it's a broken release (at least for upgrades)? Thoughts?

hectorj2f: Yes, we should delete v0.6.0 once v0.6.1 is released.

jimmidyson: :+1:

makkes:
closes #1313

PR #1347, opened 2021-01-14
retain healthCheckNodePort for service when updating

mars1024:

For services which type==LoadBalancer and externalTrafficPolicy==Local, their field `spec.healthCheckNodePort` will be allocated by APIServer, and after that it should be unchangeable, so we should retain this field during update, just like `spec.clusterIP` and `nodePort` in ports.

Here we don't need to check whether the desired service need `healthCheckNodePort` but just pass it on, because if the passed `healthCheckNodePort` is not needed any more, it will be recycled by APIServer.

Codes in kubernetes about the validation on `spec.healthCheckNodePort`, [https://github.com/kubernetes/kubernetes/blob/68108c70e29a74bb455ab63adeb5725a37e94e4f/pkg/registry/core/service/storage/rest.go#L369](https://github.com/kubernetes/kubernetes/blob/68108c70e29a74bb455ab63adeb5725a37e94e4f/pkg/registry/core/service/storage/rest.go#L369)

### DIFF
--- a/pkg/controller/sync/dispatch/retain.go
+++ b/pkg/controller/sync/dispatch/retain.go
@@ -46,6 +46,17 @@ func RetainClusterFields(targetKind string, desiredObj, clusterObj, fedObj *unst
 }
 
 func retainServiceFields(desiredObj, clusterObj *unstructured.Unstructured) error {
+	// healthCheckNodePort is allocated by APIServer and unchangeable, so it should be retained while updating
+	healthCheckNodePort, ok, err := unstructured.NestedInt64(clusterObj.Object, "spec", "healthCheckNodePort")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


makkes:
CI is failing with
> docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.

makkes:
I just restarted the build. Let's hope for the best.

hectorj2f:
@mars1024 Could you add one unit test for this change ?

mars1024:
> @mars1024 Could you add one unit test for this change ?

Of course! I added some unit tests for this.

hectorj2f (commented):
/lgtm

PR #1361, opened 2021-02-22
kubefedcluster use cadata in kubeconfig file

huiwq1990:

Kubefedcluster's cadata build from member cluster's serviceaccount, it work well when kubefed directlly conntect to member cluster. In my case, kubefed connect to a proxyserver (it proxy all apiserver's traffic), and the proxyserver has a independent cabundle.

Fixes #

### DIFF
--- a/docs/cluster-registration.md
+++ b/docs/cluster-registration.md
@@ -25,6 +25,7 @@ Repeat this step to join any additional clusters.
 
 **NOTE:** `cluster-context` will default to use the joining cluster name if not
 specified.
+**NOTE:** Before the [PR](https://github.com/kubernetes-sigs/kubefed/pull/1361), `kubefed` automatically fetches apiserver's `certificate-authority-data` from member cluster, after that kubefed will use `certificate-authority-data` in joining cluster's kubeconfig file.
 
 # Checking status of joined clusters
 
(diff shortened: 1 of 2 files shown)


RainbowMango:
It's an interesting use case.

@huiwq1990 I guess you joining cluster with a `kubeconfig` file which contains `proxyserver`'s CA.  The joining process is perfectly fine, but the following request to a member cluster will be disallowed as `proxyserver` failed to verify the certificate. Is that true?

huiwq1990:
@RainbowMango Yes, kubefedcluster only use kubeconfig's apiserfver endpoint, it should use kubeconfig's cabundle too.

irfanurrehman:
Looks legit to me. LGTM. @RainbowMango is this change allright with you?

makkes (changes requested):
This needs to be documented somewhere. I would suggest [this page](https://github.com/kubernetes-sigs/kubefed/blob/master/docs/cluster-registration.md) is the right place for it.

huiwq1990:
@makkes @irfanurrehman I write a note specify the change.

RainbowMango:
@irfanurrehman Yeah, I think the feature is reasonable. If there is CA data present in KUBECONFIG file, it is the preferred choice.
Thanks @huiwq1990 
/retest

hectorj2f:
@makkes could you have a look again ? It looks good to me.

hectorj2f (approved):
/approve

PR #1378, opened 2021-03-16
Infinite reconciliation loop rawstatuscollection

### LINKED ISSUE #1373: Intermittent huge selective delay in propagation of federated resources and infinite reconciliation loop
**What happened**:
We're running `kubefed` 0.6.0 on a eks controlplane with 2 target clusters joined successfully.

We're intermittently (but quite often) observing a huge delay (between 7mins and 50minutes) in propagation of newly created resources, of selective types (FederatedConfigmap and FederatedSecret so far, and one custom CRD that we can exclude from the picture for sake of simplicity)

FederatedConfigmap and FederatedSecret might be affected at the same time or just one type at time.

When this happens, `k get federatedconfigmaps -A -w` returns a constant stream of updates and `kubefed-controller-manager` fills up with 

```
I0312 11:05:54.304084       1 controller.go:388] Setting the federated status '{map[] false}'
I0312 11:05:54.304708       1 status.go:219] Clusters differs from the size
```

Bouncing the controller will make it immediately pick the new resources up for propagation at startup, otherwise they eventually are propagated (from ~10 to 50minutes later)

any pointers on how to debug this issue would be greatly appreciated.

I was expecting any newly created federated resource would be picked up and propagated in few seconds.

We managed to reproduce it in our environments, with our provisioning process, ~but unfortunately we're still unable to reproduce it locally~

Install `kubefed` on a `kind` cluster as specified here: https://github.com/kubernetes-sigs/kubefed#quickstart

(here a few steps)
```
kind create cluster --name cluster1 --image=kindest/node:v1.17.5

helm --namespace kube-federation-system upgrade -i kubefed kubefed-charts/kubefed --version=0.6.0 --set controllermanager.featureGates.RawResourceStatusCollection=Enabled --create-namespace

# join the cluster with itself
kubefedctl join cluster1 --cluster-context cluster1 --host-cluster-context cluster1 --v=2

kubectl create ns test

cat <<EOF > federatedns.yaml
apiVersion: types.kubefed.io/v1beta1
kind: FederatedNamespace
metadata:
  name: test
  namespace: test
spec:
  placement:
    clusterSelector:
      matchLabels: {}
  template: {}
EOF

kubectl apply -f federatedns.yaml -n test
```

then if we apply 7 `FederatedConfigmaps` the infinite loop is triggered (this doesn't happen with 6 FederatedConfgmaps).

```
cat <<EOF > federatedcms.yaml
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap1
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
---
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap2
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
---
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap3
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
---
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap4
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
---
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap5
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
---
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap6
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
---
apiVersion: types.kubefed.io/v1beta1
kind: FederatedConfigMap
metadata:
  name: configmap7
  namespace: test
spec:
  placement:
    clusterSelector: {}
  template:
    data:
      values.yaml: test
EOF
```

now, if you watch all the federatedconfigmaps:
```
k get federatedconfigmaps -A -w
```

and you apply with

```
k apply -f federatedcms.yaml -n test
```

you can see the continuous streams of changes

We have `RawResourceStatusCollection` enabled

**Investigation so far:**

watching a specific FederatedConfigMap, for example, we're able to observe it being updated, and see the cluster list changing order continuously:

```
status:
  clusters:
  - name: cluster1
  - name: cluster2
  conditions:
  - lastTransitionTime: "2021-03-11T22:15:33Z"
    lastUpdateTime: "2021-03-12T15:32:20Z"
    status: "True"
    type: Propagation
  observedGeneration: 1
```
```
status:
  clusters:
  - name: cluster2
  - name: cluster1
  conditions:
  - lastTransitionTime: "2021-03-11T22:15:33Z"
    lastUpdateTime: "2021-03-12T15:32:20Z"
    status: "True"
    type: Propagation
  observedGeneration: 1
```

which combined with 
```
I0312 11:05:54.304084       1 controller.go:388] Setting the federated status '{map[] false}'
I0312 11:05:54.304708       1 status.go:219] Clusters differs from the size
```

makes me think that the status has being updated to an empty map,  which triggers another reconciliation loop where it's updated with a new cluster list, in a different order.

The other thing that I could notice is that, on a local kind environment, after the initial startup period, logs settle onto
```
Cluster {go representation of a kubefedcluster} not updated to {go representation of a kubefedcluster} as ready status and specs are identical
```
while this doesn't happen on our remote environment. 

We could see that the recurrent error message is generated here: https://github.com/kubernetes-sigs/kubefed/blob/v0.6.0/pkg/controller/sync/status/status.go#L219

and that `resourceStatusMap` is recorded here:
https://github.com/kubernetes-sigs/kubefed/blob/v0.6.0/pkg/controller/sync/dispatch/managed.go#L270

and resourceStatus is often passed as `nil` or `obj.Object[util.StatusField]` and I was wondering if this could be somehow related.

As specified above, any pointers would be highly appreciated.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.17.12-eks-7684af
- KubeFed version: 0.6.0
- Scope of installation (namespaced or cluster): cluster
- Others

/kind bug

cebernardi:
Federated resources that have target resources without `status` field are constantly evaluated as "changed", if RawStatusCollection is Enabled. This triggers an infinite loop when multiple FederatedResources are created at the same time in the cluster

Fixes #1373

Draft PR, still working on the tests

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -293,6 +293,7 @@ func (s *KubeFedSyncController) syncToClusters(fedResource FederatedResource) ut
 	// Enable raw resource status collection if the statusCollection is enabled for that type
 	// and the feature is also enabled.
 	enableRawResourceStatusCollection := s.typeConfig.GetStatusEnabled() && s.rawResourceStatusCollection
+	klog.V(4).Infof("typeconfigstatus for %v enabled: %v, rawresstatus: %v", fedResource.FederatedName(), s.typeConfig.GetStatusEnabled(), enableRawResourceStatusCollection)
 
 	clusters, err := s.informer.GetClusters()
 	if err != nil {
@@ -387,7 +388,8 @@ func (s *KubeFedSyncController) syncToClusters(fedResource FederatedResource) ut
 	}
 
 	collectedStatus, collectedResourceStatus := dispatcher.CollectedStatus()
-	klog.V(4).Infof("Setting the federated status '%v'", collectedResourceStatus)
+	klog.V(4).Infof("Setting the federated status '%v' for %s %q", collectedResourceStatus, kind, key)
+	klog.V(4).Infof("Collected status '%v' for %s %q", collectedStatus, kind, key)
 	return s.setFederatedStatus(fedResource, status.AggregateSuccess, &collectedStatus, &collectedResourceStatus, enableRawResourceStatusCollection)
 }
 
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/logr v0.3.0 // indirect
 	github.com/go-logr/zapr v0.2.0 // indirect
+	github.com/google/go-cmp v0.5.2
 	github.com/json-iterator/go v1.1.10
 	github.com/onsi/ginkgo v1.14.2
 	github.com/onsi/gomega v1.10.3
--- a/pkg/controller/sync/status/status.go
+++ b/pkg/controller/sync/status/status.go
@@ -177,7 +177,7 @@ func (s *GenericFederatedStatus) update(generation int64, reason AggregateReason
 		}
 	}
 
-	clustersChanged := s.setClusters(collectedStatus.StatusMap, collectedResourceStatus.StatusMap)
+	clustersChanged := s.setClusters(collectedStatus.StatusMap, collectedResourceStatus.StatusMap, resourceStatusCollection)
 
 	// Indicate that changes were propagated if either status.clusters
 	// was changed or if existing resources were updated (which could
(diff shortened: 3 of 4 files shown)


pkg/controller/sync/controller.go
```diff
@@ -293,6 +293,7 @@ func (s *KubeFedSyncController) syncToClusters(fedResource FederatedResource) ut
 	// Enable raw resource status collection if the statusCollection is enabled for that type
 	// and the feature is also enabled.
 	enableRawResourceStatusCollection := s.typeConfig.GetStatusEnabled() && s.rawResourceStatusCollection
+	klog.V(4).Infof("typeconfigstatus for %v enabled: %v, rawresstatus: %v", fedResource.FederatedName(), s.typeConfig.GetStatusEnabled(), enableRawResourceStatusCollection)
```
hectorj2f, line 296: I'll rephrase this entry to `"federated status for '%q' enabled '%b' with resource status collection enabled: '%b'"`

cebernardi:
closing in favor off #1380

PR #1379, opened 2021-03-17
make objects in workqueue "comparable" to avoid multiple reconciliations on one object at same time

### LINKED ISSUE #1375: repeated sync/status controller initialization
**What happened**:
When one `federatedTypeConfig` is created with expectation of enabled sync and status controllers, I found sync/status controller will be initialized several times.
```
I0315 08:41:51.911705       1 controller.go:106] Starting sync controller for "FederatedService"
I0315 08:41:52.102628       1 controller.go:106] Starting sync controller for "FederatedService"
I0315 08:41:52.307827       1 controller.go:96] Starting status controller for "FederatedService"
I0315 08:41:52.331943       1 controller.go:106] Starting sync controller for "FederatedService"
I0315 08:41:52.420020       1 controller.go:96] Starting status controller for "FederatedService"
I0315 08:41:52.438640       1 controller.go:96] Starting status controller for "FederatedService"
```

Each sync/status controller should be initialized only once.

First creating one `federatedTypeConfig` with enabled sync/status controllers.

/kind bug

mars1024:
As we know, `workqueue` is using a map [dirty map[interface{}]struct{}](https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/util/workqueue/queue.go#L78) to deduplicate objects and prevent same object(with same namespace and name) to be reconciled by multiple workers at same time.
This means that the objects enqueued must be *comparable*.
But kubefed is just enqueuing objects of type `*DelayingDelivererItem` which will never be equal (pointer values with different addresses), [https://github.com/kubernetes-sigs/kubefed/blob/16b2f5cc4f05c9e9b696970295f353c56d33aa3a/pkg/controller/util/worker.go#L119-L121](https://github.com/kubernetes-sigs/kubefed/blob/16b2f5cc4f05c9e9b696970295f353c56d33aa3a/pkg/controller/util/worker.go#L119-L121).
So, even `federatedTypeConfig` controller has [a global locker](https://github.com/kubernetes-sigs/kubefed/blob/16b2f5cc4f05c9e9b696970295f353c56d33aa3a/pkg/controller/federatedtypeconfig/controller.go#L54-L55) for sync/status controllers, sync/status controllers will still be initialized several times.

Here is a timeline,
1. `federatedService` reconciliation starts in worker1
2. add finalizer to `federatedService` brings an update event
3. `federatedService` reconciliation starts in worker2 (workqueue regards this as a new object because they are not equal)
4. worker1 checks sync controller map(not found) and create one
5. worker2 checks sync controller map(not found) and create another
6. worker1 fills the stop channel to map
7. worker2 fills another stop channel to map(the first created sync controller will lose control)
....

Fixes #1375 

1. "comparable" here means objects should be identical and equal with same namespace and name
2. precheck (type cast) object type before enqueuing and after getting
3. move `reconcile logic` into a new function to call `Done` in defer for more returns

### DIFF
--- a/pkg/controller/util/worker.go
+++ b/pkg/controller/util/worker.go
@@ -117,7 +117,10 @@ func (w *asyncWorker) EnqueueWithDelay(qualifiedName QualifiedName, delay time.D
 func (w *asyncWorker) Run(stopChan <-chan struct{}) {
 	StartBackoffGC(w.backoff, stopChan)
 	w.deliverer.StartWithHandler(func(item *DelayingDelivererItem) {
-		w.queue.Add(item)
+		qualifiedName, ok := item.Value.(*QualifiedName)
+		if ok {
+			w.queue.Add(*qualifiedName)
+		}
 	})
 	go wait.Until(w.worker, w.timing.Interval, stopChan)
 
@@ -148,26 +151,32 @@ func (w *asyncWorker) deliver(qualifiedName QualifiedName, delay time.Duration, 
 }
 
 func (w *asyncWorker) worker() {
-	for {
-		obj, quit := w.queue.Get()
-		if quit {
-			return
-		}
-
-		item := obj.(*DelayingDelivererItem)
-		qualifiedName := item.Value.(*QualifiedName)
-		status := w.reconcile(*qualifiedName)
-		w.queue.Done(item)
-
-		switch status {
-		case StatusAllOK:
-			break
-		case StatusError:
-			w.EnqueueForError(*qualifiedName)
-		case StatusNeedsRecheck:
-			w.EnqueueForRetry(*qualifiedName)
-		case StatusNotSynced:
-			w.EnqueueForClusterSync(*qualifiedName)
-		}
+	for w.reconcileOnce() {
 	}
+}
+
+func (w *asyncWorker) reconcileOnce() bool {
+	obj, quit := w.queue.Get()
+	if quit {
+		return false
+	}
+	defer w.queue.Done(obj)
+
+	qualifiedName, ok := obj.(QualifiedName)
+	if !ok {
+		return true
+	}
+
+	status := w.reconcile(qualifiedName)
+	switch status {
+	case StatusAllOK:
+		break
+	case StatusError:
+		w.EnqueueForError(qualifiedName)
+	case StatusNeedsRecheck:
+		w.EnqueueForRetry(qualifiedName)
+	case StatusNotSynced:
+		w.EnqueueForClusterSync(qualifiedName)
+	}
+	return true
 }
--- a/pkg/controller/util/worker_test.go
+++ b/pkg/controller/util/worker_test.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


pkg/controller/util/worker.go
```diff
@@ -148,26 +151,32 @@ func (w *asyncWorker) deliver(qualifiedName QualifiedName, delay time.Duration,
 }
 
 func (w *asyncWorker) worker() {
-	for {
-		obj, quit := w.queue.Get()
-		if quit {
-			return
-		}
+	for w.reconcileOnce() {
```
RainbowMango, line 154: I'd prefer `reconcileNext` or `processNext` ?  But it's not big deal.

pkg/controller/util/worker.go
```diff
@@ -148,26 +151,32 @@ func (w *asyncWorker) deliver(qualifiedName QualifiedName, delay time.Duration,
...
+	status := w.reconcile(qualifiedName)
+	switch status {
+	case StatusAllOK:
+		break
+	case StatusError:
+		w.EnqueueForError(qualifiedName)
+	case StatusNeedsRecheck:
+		w.EnqueueForRetry(qualifiedName)
+	case StatusNotSynced:
+		w.EnqueueForClusterSync(qualifiedName)
```
RainbowMango, line 179: Should we add a default case here?

RainbowMango (commented):
Nice catch! @mars1024

pkg/controller/util/worker.go
mars1024: Default case to handle no-op? Maybe we can add this when it is **truly** needed. :)

hectorj2f (approved):
@mars1024 Could we add some tests here ? I'd like to know Otherwise it looks good to me.

mars1024:
> @mars1024 Could we add some tests here ? I'd like to know Otherwise it looks good to me.

OK, I'll try my best to add some tests for this.

hectorj2f:
@mars1024 Awesome. Thanks.

mars1024:
Sorry for the delay, and I have added some unit tests for this, could we move on? @hectorj2f

hectorj2f:
Thanks @mars1024. I'll have another look as soon as possible.

mars1024:
> Thanks @mars1024. I'll have another look as soon as possible.

friendly ping @hectorj2f , is there something wrong with this?

hectorj2f:
@mars1024 The tests are failing. Otherwise nothing is wrong, I just need to find time to look and test these changes.

mars1024:
> @mars1024 The tests are failing. Otherwise nothing is wrong, I just need to find time to look and test these changes.

Copy something for failing CI, 
```
Unable to find image 'golang:1.15.3' locally
1.15.3: Pulling from library/golang
 library/golang
docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
See 'docker run --help'.
Makefile:103: recipe for target 'bin/kubefedctl-linux-amd64' failed
make: *** [bin/kubefedctl-linux-amd64] Error 125
make: Leaving directory '/home/travis/gopath/src/sigs.k8s.io/kubefed'
The command "test $DOC_ONLY && echo "Skipped!" || DOWNLOAD_BINARIES=y bash -x ./scripts/pre-commit.sh" exited with 2.
```
And I think this fail (building kubefedctl) has nothing to do with code change. Could you help to trigger a retest? @hectorj2f

mars1024:
force-push for a retest, all tests passed

hectorj2f (commented):
/lgtm

hectorj2f (commented):
I've tested and it works for me 👍🏻

PR #1381, opened 2021-03-18
support retain custom fields for federate resource

shiyan2016:
Define a list of fields for resource that if they're updated on some cluster won't trigger federation controller to sync and overwrite them.

Fixes # https://github.com/kubernetes-sigs/kubefed/issues/1370

### DIFF
--- a/charts/kubefed/crds/crds.yaml
+++ b/charts/kubefed/crds/crds.yaml
@@ -78,6 +78,10 @@ spec:
                       type: object
                     type: array
                 type: object
+              retainCustomFields:
+                items:
+                  type: string
+                type: array
               template:
                 type: object
                 x-kubernetes-preserve-unknown-fields: true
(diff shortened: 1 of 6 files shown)


shiyan2016:
@gyliu513 @makkes Can you help me to review this PR? Thanks.

hectorj2f (changes requested):
@shiyan2016 Thanks for opening this PR and proposing changes to kubefed. 

However I am missing the context here. I'd like to see a KEP document where you explain the motivation/examples/user stories to add this feature. Also I think we need this kind of behaviors to be enabled by a Feature Gate that is how we added some other changes here that affect to the reconciliation loops of the kubefed core.
Generally, we should initially discuss it in the KEP and approve it before going and merging into the code.
Another detail we might also want to see, it could be how it affects to the current performance of kubefed.

In particular for this PR, we are missing e2e tests here, so we cannot evaluate how does work for common resources.

shiyan2016:
@hectorj2f There are some discussions for this PR: https://github.com/kubernetes-sigs/kubefed/issues/1370. And I will add the KEP doc and improve the code with feature gate and add e2e tests. Thanks.

hectorj2f:
Thanks @shiyan2016. I missed your last comment.

PR #1382, opened 2021-03-18
bugfix: register workqueue metrics in controller-runtime firstly

### LINKED ISSUE #1372: workqueue metrics do not work.
Due to the do.Once in this: 
https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go#L217-L227

This causes these 2 items to go first.

```
vendor/k8s.io/apiserver/pkg/endpoints/filters/storageversion.go:	_ "k8s.io/component-base/metrics/prometheus/workqueue" // for workqueue metric registration
vendor/k8s.io/apiserver/pkg/storageversion/manager.go:	_ "k8s.io/component-base/metrics/prometheus/workqueue" // for workqueue metric registration

```
Then finally 

https://github.com/kubernetes-sigs/kubefed/blob/master/cmd/controller-manager/app/controller-manager.go#L117

which has the doOnce flag already set from the other two and will not register. 

/kind bug

mars1024:
Ref to #1372 , because of importing race as [https://github.com/kubernetes-sigs/kubefed/issues/1372#issuecomment-801732030](https://github.com/kubernetes-sigs/kubefed/issues/1372#issuecomment-801732030), workqueue metrics do not work now.
And this PR try to use an import with highest order to fix this.

Fixes #1372 

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"os"
 
+	_ "sigs.k8s.io/controller-runtime/pkg/metrics" // for workqueue metrics registration
+
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	_ "k8s.io/client-go/plugin/pkg/client/auth" // Load all client auth plugins for GCP, Azure, Openstack, etc
 	"k8s.io/component-base/logs"


hectorj2f:
@mars1024 Could you share an example in this PR of what you are getting solved by importing this package ? What were we missing before ?

mars1024:
Yeah, let's start with the precondition that kubefed is exposing the metrics registry defined in `controller-runtime/pkg/metrics`.
[https://github.com/kubernetes-sigs/kubefed/blob/16b2f5cc4f05c9e9b696970295f353c56d33aa3a/cmd/controller-manager/app/controller-manager.go#L402-L404](https://github.com/kubernetes-sigs/kubefed/blob/16b2f5cc4f05c9e9b696970295f353c56d33aa3a/cmd/controller-manager/app/controller-manager.go#L402-L404)

So we must make sure that [the metrics collectors](https://github.com/kubernetes-sigs/controller-runtime/blob/b5065bd85190e92864522fcc85aa4f6a3cce4f82/pkg/metrics/workqueue.go#L40-L88) defined in `controller-runtime/pkg/metrics` will be used by `client-go/util/workqueue` through calling [SetProvider](https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/util/workqueue/metrics.go#L257-L261) function to get itself registered, and the `SetProvider` function has a `sync.Once` to admit only the first call.

`SetProvider` is being called in init functions of two packages, `k8s.io/component-base/metrics/prometheus/workqueue` and `sigs.k8s.io/controller-runtime/pkg/metrics`, and both of them are imported by `kubefed`, here is the dependency paths,
```
kubefed-controller-manager -> k8s.io/apiserver/pkg/server -> k8s.io/apiserver/pkg/storageversion -> k8s.io/component-base/metrics/prometheus/workqueue
kubefed-controller-manager -> sigs.k8s.io/kubefed/cmd/controller-manager/app -> sigs.k8s.io/controller-runtime/pkg/metrics
```

From [the specification](https://golang.org/ref/spec#Package_initialization) of golang, we will know that `k8s.io/component-base/metrics/prometheus/workqueue` wins, and `client-go/util/workqueue` will never use the metrics collectors(defined in `sigs.k8s.io/controller-runtime/pkg/metrics`) which are being exposed.

So this PR is trying to import `sigs.k8s.io/controller-runtime/pkg/metrics` before `k8s.io/component-base/metrics/prometheus/workqueue` to make sure calling `SetProvider` firstly.

mars1024:
In our product environments, we are using `kubfed` on dependencies of kubernetes 1.17 which has no `k8s.io/apiserver/pkg/storageversion`.
So the race condition of `SetProvider` was missed before.

hectorj2f (approved):
/lgtm

Issue #1369, opened 2021-03-04
Bug Report envolving ReplicaSchedulingPreference


marinoborges:
**What happened**:
I've set up kubefed with 2 clusters and I was testing RSP with targetKind FederatedDeployment and totalReplicas=1 with greater weight on the downstream cluster. 
The pod is scheduled and gets ready on downstream cluster accordingly. My attempt was to check whether the host cluster will schedule the pod when cordoning all downstream clusters' workers.
What I found after cordoning nodes is that the pod was evicted and then put in "Pending" state but in the didn't get moved even waiting for hours for the expected move.
Interestingly, if the kubefed-controller-manager is manually deleted, the new kubefed-controller-manager actually does the move accordingly, including moving back to downstream cluster after uncordoning the nodes. But the same issue happens again if I cordon the downstream cluster nodes one more time.
I raised the controller-manager logs to --v=5 as kindly suggested by fellow 'irfanurrehman' (slack channel #sig-multicluster) and didn't see any evidence that the controller noticed the pod unavailability. Logs below.

**What you expected to happen**:
I expected the pod to move to the host cluster when downstream cluster got the only node cordoned and pods in pending state.

**How to reproduce it (as minimally and precisely as possible)**:
- Create FederatedNamespace resource
- Create FederatedDeployment resource
- Create ReplicaSchedulingPreference resource
- Cordon nodes on downstream-cluster
- Wait for pods to move

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18.9-eks-d1db3c (host-cluster) and v1.19.7 (downstream cluster)
- KubeFed version: 0.6.1
- Scope of installation (namespaced or cluster): cluster
- Others

**RSP**
```
apiVersion: scheduling.kubefed.io/v1alpha1
kind: ReplicaSchedulingPreference
metadata:
  name: deployment-100-108
  namespace: video-ingest
spec:
  targetKind: FederatedDeployment
  totalReplicas: 1
  rebalance: true
  clusters:
    "host-cluster":
      weight: 1
    "downstream-cluster":
      weight: 10

```

**Logs**
Timeline:
- I cordoned downstream cluster only node at 2:00:01
- I collected the logs at 02:04:42
Notes:
- There were no lines in the log between 02:00:11 and 02:04:42
```
[...]

PR #1385, opened 2021-03-22
Replica rebalance reconciliation fix

irfanurrehman:

Fixes #
Fixes https://github.com/kubernetes-sigs/kubefed/issues/1369
~~I will probably update relevant unit tests also, in a short while.~~
Updated unit tests.

### DIFF
--- a/pkg/controller/util/podanalyzer/pod_helper.go
+++ b/pkg/controller/util/podanalyzer/pod_helper.go
@@ -20,6 +20,8 @@ import (
 	"time"
 
 	api_v1 "k8s.io/api/core/v1"
+
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
 )
 
 type PodAnalysisResult struct {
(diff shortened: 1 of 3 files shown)


irfanurrehman:
cc @marinoborges

hectorj2f:
@irfanurrehman Have you manage to replica the same issue https://github.com/kubernetes-sigs/kubefed/issues/1369 ? From you last comment I understand you asked for those files.

hectorj2f (commented):
@irfanurrehman Could we get an e2e that exercise this function a part from this unit test ?

hectorj2f (approved):

irfanurrehman:
@hectorj2f Apologies, for not noticing this earlier. Was on PTO.

> @irfanurrehman Have you manage to replica the same issue #1369 ? From you last comment I understand you asked for those files.
Yes, I got the scenario and the yamls from @marinoborges offline.

> @irfanurrehman Could we get an e2e that exercise this function a part from this unit test ?
 Adding an e2e for this did not appear to be very straight forward. Would involve changing atleast one of the kind cluster runs with an additional node, managing labels (or taints) on that node wrt the scheduled pod, while also ensuring that this does not affect the other tests using the same cluster. I could not justify the additional value wrt to what the unit test already does.

Thanks for reviewing it anyways.

PR #1386, opened 2021-03-22
fix metrics workqueues

tehlers320:
The metrics workqueues seem to appear randomly for various builds/users. This PR ensures the correct package always loads first. 

Fixes # https://github.com/kubernetes-sigs/kubefed/issues/1372

ExpediaInc is in the process of getting the CLA signed ASAP.

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -42,6 +42,4 @@ func main() {
 	}
 }
 
-// Without init, the _ imports do not initialize before app package
-func init() {
-}
+func init() {}
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -41,3 +41,7 @@ func main() {
 		os.Exit(1) //nolint:gocritic
 	}
 }
+
+// Without init, the _ imports do not initialize before app package
+func init() {
+}


cmd/controller-manager/main.go
```diff
@@ -41,3 +41,7 @@ func main() {
 		os.Exit(1) //nolint:gocritic
 	}
 }
+
+// Without init, the _ imports do not initialize before app package
```
mars1024, line 45: Could you help to link the related golang specification about this behavior? Very curious about this.

tehlers320: @mars1024 i couldnt find a good doc on this, this is the only thing i could find https://golang.org/doc/effective_go#blank_import

```
For example, during its init function, the net/http/pprof package registers HTTP handlers that provide debugging information.
```

This implies that init is needed for a side effect import, though it does not clearly state it.

I've tested it extensively and it is fixing it.

tehlers320: Should i re-phrase the comment 
```
// init for _ imports 
```
instead ?

mars1024:
Hi, @tehlers320 , I tested this locally but ran into an unexpected result, 

First run,
```
package main

import (
	_ "bruce/dep2"

	"bruce/dep1"
)

func init() {}

func main() {
	dep1.NoOp()
}
```
Result is 
```
dep2 init
dep1 init
```

Second run,
```
package main

import (
	_ "bruce/dep2"

	"bruce/dep1"
)

func main() {
	dep1.NoOp()
}
```

Result is still, 
```
dep2 init
dep1 init
```

What did I miss?

tehlers320:
> What did I miss?

Sorry i didn't have time to dig into this and wont probably until the weekend, i did re-run your example test and i'm seeing the same thing. This is loading as expected with or without the func init() using example code.  I started drafting an e2e test to get to the bottom of this and will try and finish that ASAP to prove/disprove the issue.

hectorj2f:
@mars1024 @tehlers320 What is the state of this PR? Is it working for you both?

mars1024:
> @mars1024 @tehlers320 What is the state of this PR? Is it working for you both?

Sorry, all I know is already here, let's wait for more input from @tehlers320

tehlers320:
> @mars1024 @tehlers320 What is the state of this PR? Is it working for you both?

It works, but it seems my co-worker has the proper fix. closing.

PR #1388, opened 2021-03-29
added blank import to ensure correct workqueue metric registration

cebernardi:
we're facing (as described here: #1386 and #1372) a racing condition on workqueue metrics registration, between 

* k8s.io/component-base/metrics/prometheus/workqueue
* sigs.k8s.io/controller-runtime/pkg/metrics

what we want is `sigs.k8s.io/controller-runtime/pkg/metrics` to get registered first, but both of:
- cmd/controller-manager/main.go
- cmd/hyperfed/main.go

have a dependency from `k8s.io/apiserver/pkg/server`, that is bringing `k8s.io/component-base/metrics/prometheus/workqueue` into the picture (and it gets initialized before `sigs.k8s.io/controller-runtime/pkg/metrics`

this PR (#1386) solved the issue for the `controller-manager` executable produced with `make build`, but it doesn't solve it for the single `hyperfed` executable produced with `make container` as the compilation entrypoint is different (`cmd/controller-manager/main.go` vs `cmd/hyperfed/main.go`)

Fixes #1386

I'm not sure if worth it to uniform the output between `make container` and `make build`, so that we produce the same executable (`hyperfed`) on both targets.. please let me know your thoughts

### DIFF
--- a/cmd/hyperfed/main.go
+++ b/cmd/hyperfed/main.go
@@ -31,6 +31,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	_ "sigs.k8s.io/controller-runtime/pkg/metrics" // for workqueue metrics registration
+
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	_ "k8s.io/client-go/plugin/pkg/client/auth" // Load all client auth plugins for GCP, Azure, Openstack, etc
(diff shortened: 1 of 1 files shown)


hectorj2f (approved):

mars1024 (approved):
/lgtm, thanks!

makkes (approved):
/lgtm

PR #1389, opened 2021-03-31
use merge-patch on finalizer operations to resolve racing conflicts

mars1024:
`Patch` is more efficient than `Update` to handle racing conditions on high-frequency changes of objects, which is also recommended by kubernetes spec.

Try to avoid some errors, like [https://github.com/kubernetes-sigs/kubefed/blob/4dbd63b07de4b2bcebf3bd2632e0ec08e254f4c3/pkg/controller/sync/controller.go#L283](https://github.com/kubernetes-sigs/kubefed/blob/4dbd63b07de4b2bcebf3bd2632e0ec08e254f4c3/pkg/controller/sync/controller.go#L283)

### DIFF
--- a/charts/kubefed/charts/controllermanager/templates/aggregate_clusterroles.yaml
+++ b/charts/kubefed/charts/controllermanager/templates/aggregate_clusterroles.yaml
@@ -51,7 +51,7 @@ rules:
   - list
   - update
   - create
-  - update
+  - patch
   - delete
 - apiGroups:
   - multiclusterdns.kubefed.io
@@ -63,6 +63,7 @@ rules:
   - list
   - create
   - update
+  - patch
   - delete
 - apiGroups:
   - core.kubefed.io
@@ -74,6 +75,7 @@ rules:
   - list
   - create
   - update
+  - patch
   - delete
 - apiGroups:
   - types.kubefed.io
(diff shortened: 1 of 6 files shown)


yuswift:
There are many `Update` operations. Could you tell me why do you only change the `finalizer` update operation?

mars1024:
> There are many `Update` operations. Could you tell me why do you only change the `finalizer` update operation?

Actually, this is only the first step. I'll try to use more `Patch` after this merged.

yuswift:
> > There are many `Update` operations. Could you tell me why do you only change the `finalizer` update operation?
> 
> Actually, this is only the first step. I'll try to use more `Patch` after this merged.

@hectorj2f What's your suggestion?

hectorj2f (commented):
I'd like to see a more consistent and global change of Update to Patch operations across the whole code. I agree with @mars1024  here.

mars1024:
> I'd like to see a more consistent and global change of Update to Patch operations across the whole code. I agree with @mars1024 here.

Thanks for your agreement, @hectorj2f , and I think this PR is already ready as the first step.

makkes (approved):

mars1024:
friendly ping @hectorj2f for approval or more comments

hectorj2f (commented):
@mars1024 What did it happen with the rest of changes to use Patch instead of Update ?

mars1024:
> @mars1024 What did it happen with the rest of changes to use Patch instead of Update ?

Recently I didn't have much time on this, so I opened an help-wanted UMBRELLA issue to track the rest changes.

PR #1393, opened 2021-04-05
add intersection behavior for RSP

### LINKED ISSUE #981: clusterSelector is ignored with ReplicaSchedulingPreference
Following FederatedDeployment schedules correctly only to clusters with label `region=eu-west1`

```
apiVersion: types.kubefed.k8s.io/v1beta1
kind: FederatedDeployment
metadata:
  name: nginx
  namespace: testing
spec:
  placement:
    clusterSelector:
      matchLabels:
        region: eu-west1
  template:
    metadata:
      labels:
        app: nginx
    spec:
      replicas: 5
      terminationGracePeriodSeconds: 0
      selector:
        matchLabels:
          app: nginx
      template:
        metadata:
          labels:
            app: nginx
        spec:
          containers:
          - image: nginx
            name: nginx
```

When I add RSP to limit the maximum replicas it schedules to all clusters, ignoring the label:
```
apiVersion: scheduling.kubefed.k8s.io/v1alpha1
kind: ReplicaSchedulingPreference
metadata:
  name: nginx
  namespace: testing
spec:
  targetKind: FederatedDeployment
  totalReplicas: 15
  clusters:
    "*":
      maxReplicas: 3
      weight: 1
```

xunpan:
We add intersection behavior because it makes sense to result placement for resource propagation.

Fixes #981

- placement computing rule is get from current federated resource calculation. That means the related rule is reused in `RSP` for placement calculation 
- I'm trying to add test and add test cases. So, please review when related tasks are done.


xunpan:
@irfanurrehman 
It is ready to review.

irfanurrehman:
Thanks for doing this @xunpan. I will positively take a look at this tomorrow.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -190,6 +191,34 @@ func (s *ReplicaScheduler) Reconcile(obj pkgruntime.Object, qualifiedName ctluti
 	}
 
 	key := qualifiedName.String()
+
+	if rsp.Spec.PlacementIntersection {
+		klog.V(3).Infof("Computing placement of resource %q", qualifiedName)
+
+		fedClusters, err := s.podInformer.GetReadyClusters()
```
irfanurrehman, line 198: Is this not the same names that you already have at line 167 above (`clusterNames, err := s.clusterNames()`)?

charts/kubefed/charts/controllermanager/crds/crds.yaml
```diff
@@ -860,6 +860,13 @@ spec:
                   provided. If omitted, clusters without explicit preferences should
                   not have any replicas scheduled.
                 type: object
+              placementIntersection:
```
irfanurrehman, line 863: I think `IntersectWithClusterSelector` will be a better name.

pkg/controller/util/placement.go
```diff
@@ -86,3 +89,92 @@ func SetClusterNames(obj *unstructured.Unstructured, clusterNames []string) erro
...
+	placement, err := UnmarshalGenericPlacement(resource)
+	if err != nil {
+		return nil, err
+	}
+
+	selectedNames := sets.String{}
+	clusterNames := placement.ClusterNames()
+	// Only use selector if clusters are nil. An empty list of
+	// clusters implies no clusters are selected.
+	if clusterNames == nil {
```
irfanurrehman, line 155: The problem with this approach is that RSP uses the clusternames itself to specify its decision, so the intersection of the cluster selector with the RSP decision will not work if both the cluster selector and the clusternames are specified on the target. 
Also, for the scenario where only the clusterselector is specified, the first reconcile result will be the intersection of the clusterselector and the RSP decision, but next time onwards the clusterselector will be ignored becaue RSP would have written clusternames on the target resource. So if the clusterselector is updated after that the update will be ignored.

pkg/schedulingtypes/plugin.go
```diff
@@ -133,6 +143,24 @@ func (p *Plugin) FederatedTypeExists(key string) bool {
...
+	// get FederatedNamespace with namespace name of the object
+	fedNsObject, err := p.fedNsClient.Resources(qualifiedName.Namespace).Get(context.Background(), qualifiedName.Namespace, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	if p.typeConfig.GetNamespaced() {
+		return util.ComputeNamespacedPlacement(fedObject, fedNsObject, clusters, p.limitedScope)
+	}
+	return util.ComputePlacement(fedObject, clusters)
```
irfanurrehman, line 161: Please see my comment above about using this as is here. I think we need a separate function which gets only the clusterselector from the target resource. We are trying to intersect with the clusterselector only, not the clusternames.

pkg/schedulingtypes/replicascheduler.go
```diff
@@ -190,6 +191,34 @@ func (s *ReplicaScheduler) Reconcile(obj pkgruntime.Object, qualifiedName ctluti
...
+			return ctlutil.StatusError
+		}
+
+		resultClusters, err := plugin.(*Plugin).GetResourceClusters(qualifiedName, fedClusters)
+		if err != nil {
+			runtime.HandleError(errors.Wrapf(err, "Failed to get prefrerred clusters while reconciling RSP named %q", key))
+			return ctlutil.StatusError
+		}
+
+		preferredClusters := []string{}
```
irfanurrehman, line 210: Could not understand the need of `preferredClusters` it seems to be a copy of `resultClusters`

test/e2e/scheduling.go
```diff
@@ -202,10 +230,21 @@ func createTestObjs(tl common.TestLogger, client genericclient.Client, typeConfi
...
+
+	clusters := []string{}
+	fedObject, err := common.NewTestObject(typeConfig, namespace, clusters, fixture)
 	if err != nil {
 		return "", err
 	}
+
+	if rspSpec.PlacementIntersection {
+		clusters = append(clusters, fedClusters[0])
+		err = util.SetClusterNames(fedObject, clusters)
```
irfanurrehman, line 242: I think this needs to be tested with a clusterselector rather then a cluster name.

irfanurrehman (changes requested):
@xunpan Many thanks for spending time on this.
However, I think there are some rework that would be needed on this PR. Please take a look at my comments.

charts/kubefed/charts/controllermanager/crds/crds.yaml
```diff
@@ -860,6 +860,13 @@ spec:
...
+                description: If set to true then the target kind (FederatedDeployment
+                  or FederatedReplicaset) placement will be intersected with Clusters
+                  of the RSP to determine the result clusters of the federated resource.
+                  if set false or not defined, Clusters in the RSP overriding the
+                  placment in the federated resource.
+                type: boolean
```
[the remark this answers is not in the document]
xunpan: OK.

pkg/apis/scheduling/v1alpha1/replicaschedulingpreference_types.go
```diff
@@ -45,6 +45,14 @@ type ReplicaSchedulingPreferenceSpec struct {
 	// +optional
 	Rebalance bool `json:"rebalance,omitempty"`
 
+	// If set to true then the target kind (FederatedDeployment or FederatedReplicaset) placement
+	// will be intersected with Clusters of the RSP to determine the result clusters of the
+	// federated resource.
+	// if set false or not defined, Clusters in the RSP overriding the placment in the federated
+	// resource.
+	// +optional
+	PlacementIntersection bool `json:"placementIntersection,omitempty"`
```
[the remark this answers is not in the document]
xunpan: ClusterSelector is a little misleading I guess. For a federated resource, we have 
```
spec:
  placement:
    clusters:
        ...
    clusterSelector:
        ...
```
I use Placement here because it includes both `clusters` and `clusterSelector`.

Any comments?

pkg/schedulingtypes/replicascheduler.go
xunpan: you're right. I will update it.

hectorj2f (commented):
Could you add some documentation with an example as part of our current docs ?

hectorj2f:
@xunpan Could you address the requested changes by @irfanurrehman ?

xunpan:
Yes. The code was done and I'm testing it locally. The updated diff will be committed by this week.

irfanurrehman:
> Yes. The code was done and I'm testing it locally. The updated diff will be committed by this week.

@xunpan I missed checking on this. I see that you did some updates. Were they finalised by you? Should I check them again?

xunpan:
@irfanurrehman I forgot to send you a message after update the PR.

I think I addressed your comments above. One does not changed:  
[Q] `preferredClusters` it seems to be a copy of `resultClusters`
[A] It is used to transform `sets.String` to `string` slice

pkg/schedulingtypes/replicascheduler.go
irfanurrehman: nit: @xunpan this is still unresolved but I will approve as this does not seem to harm the logic at all.

irfanurrehman:
Thanks for doing this @xunpan.
/lgtm

irfanurrehman (approved):

PR #1399, opened 2021-04-11
correct command line example for test

xunpan:
development document is good for reference during development work. Keeping it correct helps contributors.
1. upgrade kubebuilder version
2. remove `--v` `-v` because it is not the valid arguments.

This is found during development. No need an extra issue.


hectorj2f (commented):
/lgtm

PR #1400, opened 2021-04-11
feat: make concurrency of the worker configurable

### LINKED ISSUE #1392: Make worker count configurable for different controllers
Since #1379 has already fixed the reconciliation conflicts of same object, a worker count option should be imported to `ReconcileWorker` for better performance on different reconciliations.

**Why is this needed**:
As the comments above, a controller-friendly option for better performance.

/kind feature

zqzten:
As to gain better performance on different reconciliations, the concurrency of worker should be made configurable for different controllers.

In short, this PR does two things below:
* Introduce a new `MaxConcurrentReconciles` option to worker for inner controller concurrency settings.
* Introduce a new `Concurrency` section to KubeFedConfg to expose the concurrency settings to users.

Fixes #1392


zqzten:
@hectorj2f Mind take a look? Advice on further design is needed.

yuswift:
I'd prefer the first piece of advice. How do we define the `kubefedconfig` custom resource if we choose the second advice？Cause we don't know what sync controllers we will run.

zqzten:
> I'd prefer the first piece of advice. How do we define the `kubefedconfig` custom resource if we choose the second advice？Cause we don't know what sync controllers we will run.

I agree that configuring sync controller for each type in `kubefed` is not applicable, but what if we can configure that in FTC? In addition, there's still a choice between one config for all of the core controllers (sync, status, etc.) and seperated config for each core controller and higher level controller (like RSP) of kubefed.

zqzten:
To keep it simple as well as extendable, currently I exposed the `MaxConcurrentReconciles` option of the two core controllers' workers to user, because they do the most work of KubeFed and their performance is subject to the scale of clusters KubeFed managed. 

Other concurrency settings can alse be exposed in the `Concurrency` section of KubeFedConfg in the future when needed.

zqzten:
@jimmidyson @makkes This PR is ready for review.

hectorj2f:
@zqzten Could you highlight the performance benefits of this solution ? and why would you propose these changes ?

zqzten:
@hectorj2f We are using KubeFed in some large-scale cluster management cases, such as a public cloud service. It's common that operations on some clusters may take a long time (or even timeout) and block the controller processing resources belong to other users' clusters. Increasing the concurrency of sync controller mitigated this problem because now it has chance to process other resources on the fly.

In addition, it's apparent that properly increase the concurrency of a controller can improve its performance because it can simply process more events at the same time. The `MaxConcurrentReconciles` is a commonly used option in controller-runtime (as its [name](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L38) comes from the controller-runtime library too) to improve the controller's performance.

hectorj2f (commented):
@zqzten Thanks for the explanation. @irfanurrehman @jimmidyson Do you have any concerns about these properties ?

pkg/apis/core/v1beta1/kubefedconfig_types.go
```diff
@@ -36,6 +36,8 @@ type KubeFedConfigSpec struct {
 	// +optional
 	ClusterHealthCheck *ClusterHealthCheckConfig `json:"clusterHealthCheck,omitempty"`
 	// +optional
+	Concurrency *ConcurrencyConfig `json:"concurrency,omitempty"`
```
jimmidyson, line 39: How about we move the concurrency settings under the respective controller config, e.g. add sync controller concurrency config in `SyncControllerConfig`?

zqzten: Thanks for the advice, it seems to be more clear indeed. I'll make this change.

jimmidyson: Thanks @zqzten!

charts/kubefed/values.yaml
```diff
@@ -18,7 +18,10 @@ controllermanager:
   ## Supported options are `configmaps` and `endpoints`
   leaderElectResourceLock:
   syncController:
+    maxConcurrentReconciles:
     adoptResources:
```
jimmidyson, line 22: nit (not a blocker): would be good to sort these alphabetically I think but I don't mind leaving as is
```suggestion
    adoptResources:
    maxConcurrentReconciles:
```

config/kubefedconfig.yaml
```diff
@@ -26,4 +26,7 @@ spec:
     successThreshold: 1
     timeout: 3s
   syncController:
+    maxConcurrentReconciles: 1
     adoptResources: Enabled
```
jimmidyson, line 30: nit: again personally prefer this to be sorted alphabetically but not a blocker.
```suggestion
    adoptResources: Enabled
    maxConcurrentReconciles: 1
```

pkg/apis/core/v1beta1/defaults/defaults.go
```diff
@@ -87,10 +90,20 @@ func SetDefaultKubeFedConfig(fedConfig *v1beta1.KubeFedConfig) {
 		spec.SyncController = &v1beta1.SyncControllerConfig{}
 	}
 
+	syncController := spec.SyncController
+	setInt64(&syncController.MaxConcurrentReconciles, DefaultSyncControllerMaxConcurrentReconciles)
```
jimmidyson, line 94: nit: no need to introduce variable here:

```suggestion
	setInt64(&spec.SyncController.MaxConcurrentReconciles, DefaultSyncControllerMaxConcurrentReconciles)
```

pkg/apis/core/v1beta1/defaults/defaults.go
```diff
@@ -87,10 +90,20 @@ func SetDefaultKubeFedConfig(fedConfig *v1beta1.KubeFedConfig) {
...
 		spec.SyncController.AdoptResources = new(v1beta1.ResourceAdoption)
 		*spec.SyncController.AdoptResources = v1beta1.AdoptResourcesEnabled
 	}
+
+	if spec.StatusController == nil {
+		spec.StatusController = &v1beta1.StatusControllerConfig{}
+	}
+
+	statusController := spec.StatusController
+	setInt64(&statusController.MaxConcurrentReconciles, DefaultStatusControllerMaxConcurrentReconciles)
```
jimmidyson, line 106: nit: no need to introduce variable here
```suggestion
	setInt64(&spec.StatusController.MaxConcurrentReconciles, DefaultStatusControllerMaxConcurrentReconciles)
```

pkg/apis/core/v1beta1/validation/validation.go
```diff
@@ -356,10 +356,19 @@ func ValidateKubeFedConfig(kubeFedConfig, oldKubeFedConfig *v1beta1.KubeFedConfi
 	case sync.AdoptResources == nil:
 		allErrs = append(allErrs, field.Required(adoptPath, ""))
 	default:
+		allErrs = append(allErrs, validateIntPtrGreaterThan0(syncPath.Child("maxConcurrentReconciles"), sync.MaxConcurrentReconciles)...)
 		allErrs = append(allErrs, validateEnumStrings(adoptPath, string(*sync.AdoptResources),
 			[]string{string(v1beta1.AdoptResourcesEnabled), string(v1beta1.AdoptResourcesDisabled)})...)
 	}
 
+	status := spec.StatusController
```
jimmidyson, line 364: Naming this `status` when it's a field in `spec` is a bit confusing. Perhaps
```suggestion
	statusController := spec.StatusController
```

charts/kubefed/values.yaml
zqzten: Just want to put the general options above the specific options, in accordance of the type defination of KubeFedConfig. And other options in this file are also not sorted alphabetically, so I'm going to leave it as is.

config/kubefedconfig.yaml
zqzten: Same as above.

charts/kubefed/values.yaml
jimmidyson: :+1: I'm down with that :)

jimmidyson (approved):

irfanurrehman (approved):
Thanks @zqzten for doing this.
/lgtm

PR #1420, opened 2021-06-08
build: Switch to Github Actions from Travis

### LINKED ISSUE #1419: Migrate from travis-ci.org to travis-ci.com
**What happened**:
We are using travis-ci.org for PR checks but that's going to shut down (has shut down rather): https://blog.travis-ci.com/2021-05-07-orgshutdown

We need to migrate to travis-ci.com or another service (Appveyor, GitHub Actions or similar)

/kind bug

jimmidyson:

This PR switches from Travis to Github Actions for PR builds - travis-ci.org is shutting down and migrating to travis-ci.com is equivalent effort to switching to Github Actions, which supposedly members have had a better experience with.

**Which issue(s) this PR fixes**
Fixes #1419 

Need to test this PR somehow - hope that pushing it triggers Github Actions.


.travis.yml
```diff
@@ -1,42 +0,0 @@
...
-# There are two environment variables configured in travis that allow the travis
-# build job to use this robot account:
-#
-# - QUAY_USERNAME
-# - QUAY_PASSWORD
-#
-# These are statically configured in the travis settings and are set by travis
-# in the build job's execution environment.
-after_success:
-  - test $DOC_ONLY && echo "Skipped!" || make push
```
makkes, line 38: This path isn't reflect in the GitHub Actions workflow above. Given the Makefile reads the TRAVIS_PULL_REQUEST environment variable in the push target, I take it we need to rework that a bit. My preference would be to just push anytime the target is called and only call that `on: push.tags`. wdyt?

.github/workflows/pr-workflow.yml
```diff
@@ -0,0 +1,25 @@
+name: PR build
+
+on: [pull_request]
```
makkes, line 3: This might pose a problem as GitHub Actions workflows triggered from forks do not have access to secrets: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks
Solution:
```suggestion
on: [pull_request_target]
```

jimmidyson: I don't think this will work because this runs in the context of the base branch - it's only used for labelling, commenting, etc. not running code from an untrusted fork.

jimmidyson: `pull_request_target` doesn't use the pull request code though from what i've read, so it doesn't open up security issues. i'm hoping that github actions run on servers that have no pull rate limiting :crossed_fingers:

.travis.yml
jimmidyson: Done in 77a49debaedde724862dbe6157c60b982664c93a (i hope!)

.github/workflows/pr-workflow.yml
jimmidyson: https://github.com/actions/virtual-environments/issues/1445 explains that GitHub actions have no pull limit (agreement with docker hub) so no need to login.

makkes (approved):

makkes:
/lgtm

I believe there's no other way than to try it out. 🙂

PR #1428, opened 2021-06-17
fix: retry on recoverable propagation failure

zqzten:
Currently sync controller won't retry propagting a fed resource on any propagation failure, it's not resonable for some recoverable failures such as quota exceed, lack of permissions, etc. For the quota case, after we expand the quota for the failed cluster, we expect KubeFed to automatically succeed in propagating rather than manually edit the fed resource (which is meaninglessly) to trigger a reconciliation.

This PR enable sync controller to retry with back off on recoverable propagation failure.

### DIFF
--- a/pkg/controller/sync/controller.go
+++ b/pkg/controller/sync/controller.go
@@ -450,6 +450,15 @@ func (s *KubeFedSyncController) setFederatedStatus(fedResource FederatedResource
 		return util.StatusError
 	}
 
+	// return Error to trigger a retry with back off on recoverable propagation failure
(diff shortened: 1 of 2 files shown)


zqzten:
The test result is confusing... seems to have nothing to do with this PR.

zqzten:
friendly ping @font @xunpan for a review

xunpan:
Thanks!

/lgtm

irfanurrehman:
Thanks @zqzten for doing this.
/approve

PR #1432, opened 2021-06-23
build(deps): Upgrade and align k8s.io dependency versions

jimmidyson:
This aligns all k8s dependencies to the same version to reduce
potential dependency conflicts in future.


hectorj2f (commented):
/lgtm

PR #1460, opened 2021-10-15
feat: introduce informer cache sync timeout

### LINKED ISSUE #1459: informer cache sync needs a timeout
A timeout should be introduced to stop waiting informer cache sync forever, even it will never succeed.

**Why is this needed**:
[https://github.com/kubernetes-sigs/kubefed/blob/5b1f868c77383ada5c19e0f3aff1a5d6f2075941/pkg/controller/sync/controller.go#L206-L229](https://github.com/kubernetes-sigs/kubefed/blob/5b1f868c77383ada5c19e0f3aff1a5d6f2075941/pkg/controller/sync/controller.go#L206-L229)

Now `isSynced` function will prevent reconciliations, but it only returns true or false, which means we will wait forever, even it may not succeed.

cc @zqzten 

/kind feature

zqzten:
According to [this issue](https://github.com/kubernetes-sigs/controller-runtime/issues/1219) of controller-runtime, an informer cache timeout is needed to prevent controllers from blocking indefinitely. In KubeFed, we have far more informers than a simple controller and any of the informer cache sync problems will block the whole reconcile loop (which has been encountered in our prod env serveral times).

This PR introduces a configurable informer cache timeout to the core controllers of KubeFed. It can let KubeFed controller error out if its informers are unable to sync their caches within this timeout. With this behavior, one KubeFed controller will never be kept running without working which can be useful for users to discover watch/list problems in time and can also give chance to other working replicas to run.

This PR also adds logs to the `ClustersSynced` check to help find out which member cluster's informer cache sync is blocking.

Fixes #1459


zqzten:
friendly ping @xunpan for approval or further opinions

PR #1461, opened 2021-10-19
Increase the default QPS and Burst value of the controller manager

iawia002:

I found that the default QPS and Burst values for the rest-client of the kubefed-controller-manager component were configured very low(5 and 10), causing the component to start very slowly:

```
Waited for due to client-side throttling, not priority and fairness
```

So I set it to a suitable value.

Fixes #


xunpan:
There is no other comments. Let me merge it.
/approve

PR #1476, opened 2021-11-12
Introduce cluster config malformed condition to KubeFedCluster

### LINKED ISSUE #1475: KubeFed controller stop working when there are KubeFedClusters with malformed configs whose previous status is ready
**What happened**:
We accidentally deleted a Secret of one KubeFedCluster which is ready previously. After that, we found that KubeFed controller stop working due to an informer sync check error `The number of target informers mismatch with given clusters`. This is because the KubeFedCluster with malformed config is still in **Ready** state but the controller cannot create a REST client for it.

KubeFedClusters with malformed config should turn into a non-ready condition in order not to block the whole controller.

Delete any ready KubeFedCluster's Secret then restart KubeFed controller.

/kind bug

zqzten:
Currently when a ready KubeFedCluster's config become malformed (e.g. by deleting its Secret or some other necessary fields), there's no way to change its status to not ready. This can lead to some misbehavior of KubeFed controller and can also confuse maintainers.

This PR introduce a `ConfigMalformed` condition to KubeFedCluster to correctly reflect the unready status of a KubeFedCluster with malformed configuration.
It also mark `ClientRetrievalFailed` as a recoverable error since it's recoverable when some KubeFedCluster's malformed Secret get fixed.

Fixes #1475

### DIFF
--- a/pkg/controller/kubefedcluster/controller.go
+++ b/pkg/controller/kubefedcluster/controller.go
@@ -175,7 +175,7 @@ func (cc *ClusterController) addToClusterSet(obj *fedv1b1.KubeFedCluster) {
 	cc.mu.Lock()
 	defer cc.mu.Unlock()
 	clusterData := cc.clusterDataMap[obj.Name]
-	if clusterData != nil && clusterData.clusterKubeClient != nil {
+	if clusterData != nil && clusterData.clusterKubeClient.kubeClient != nil {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/kubefedcluster/controller.go
```diff
@@ -186,7 +186,6 @@ func (cc *ClusterController) addToClusterSet(obj *fedv1b1.KubeFedCluster) {
 	if err != nil || restClient == nil {
 		cc.RecordError(obj, "MalformedClusterConfig", errors.Wrap(err, "The configuration for this cluster may be malformed"))
 		klog.Errorf("The configuration for cluster %s may be malformed", obj.Name)
-		return
```
hectorj2f, line 189: question: pls, could you explain the reason to remove the return statement here ?

zqzten: The KubeFedCluster controller only does health check (and update status) for those in the `clusterDataMap`. Previously KubeFedClusters with malformed config never have a chance to be added in the `clusterDataMap` just because the `addToClusterSet` function return here, so I remove the return statement in order to let the `updateIndividualClusterStatus` func detect them (by checking whether its `clusterKubeClient` is nil) and update their statuses correctly.

zqzten:
/assign @hectorj2f
PTAL, thanks

zqzten:
friendly ping @hectorj2f for approval or further opinion

PR #1486, opened 2022-01-21
Enabled to specify namespace in FederatedTypeConfig.

ryutoyasugi:
I want to be able to specify a namespace with FederatedTypeConfig.
This is because the namespace is not given to FederatedTypeConfig even if the following command is executed.
```sh
$ helm template kubefed charts/kubefed --namespace kube-federation-system
```

### DIFF
--- a/charts/kubefed/templates/federatedtypeconfig.yaml
+++ b/charts/kubefed/templates/federatedtypeconfig.yaml
@@ -3,6 +3,7 @@ apiVersion: core.kubefed.io/v1beta1
 kind: FederatedTypeConfig
 metadata:
   name: clusterroles.rbac.authorization.k8s.io
+  namespace: {{ .Release.Namespace }}
 spec:
   federatedType:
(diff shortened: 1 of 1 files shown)


irfanurrehman:
@ryutoyasugi can you please explain your usage and the reason for this PR a bit more. Also can you please sign the CLA.

ryutoyasugi:
@irfanurrehman Sorry for the lack of explanation.
I want to apply Kubefed to private k8s cluster.
There are some restrictions on the private k8s cluster, so some modifications are needed when applying Kubefed.
So, I rendered the template with `helm template` command.
However, since FederatedTypeConfig does not have a namespace attached to it, I apply it without a namespace, it will apply in the default namespace.

ryutoyasugi:
I signed it

ryutoyasugi:
"I signed it"

makkes (approved):

PR #1497, opened 2022-03-09
fix: FailureThreshold and SuccessThreshold not take effect

### LINKED ISSUE #1496: clusterHealthCheckConfig FailureThreshold and SuccessThreshold not take effect
**What happened**:
about below struct
type ClusterHealthCheckConfig struct {
	Period           time.Duration
	FailureThreshold int64
	SuccessThreshold int64
	Timeout          time.Duration
}
we set Period 10s，FailureThreshold 100，SuccessThreshold 1，Timeout 60s。

but when kubefed check child k8s failed just only 1 time, the status of child k8s becomes notReady.
the status of child k8s in kubefed becomes notReady after kubefed checks child k8s failed 10 times continuously.

version informations:
1)kubefed version: lastest
2) k8s version: maybe all versions, we use 1.11
3) kubefed contract child k8s use domain(kubefed->dns->lvs->nginx->child k8s apiserver)

just down(use reboot) one lvs node

**Environment**:
- KubeFed version
- Others

/kind bug

FengXingYuXin:
cluster health check's threshold config not take effect. 
Fixes  #1496 

### DIFF
--- a/pkg/controller/kubefedcluster/controller_test.go
+++ b/pkg/controller/kubefedcluster/controller_test.go
@@ -52,39 +52,51 @@ func TestThresholdCheckedClusterStatus(t *testing.T) {
 	}{
 		"ClusterReadyAtBegining": {
 			clusterStatus:         clusterStatus(corev1.ConditionTrue, t1, t1),
-			storedClusterData:     &ClusterData{clusterStatus: nil, resultRun: 0},
+			storedClusterData:     &ClusterData{clusterStatus: nil, lastProbeStatus: nil, resultRun: 0},
 			expectedClusterStatus: clusterStatus(corev1.ConditionTrue, t1, t1),
 			expectedResultRun:     1,
 		},
 		"ClusterNotReadyAtBegining": {
 			clusterStatus:         clusterStatus(corev1.ConditionFalse, t1, t1),
-			storedClusterData:     &ClusterData{clusterStatus: nil, resultRun: 0},
+			storedClusterData:     &ClusterData{clusterStatus: nil, lastProbeStatus: nil, resultRun: 0},
 			expectedClusterStatus: clusterStatus(corev1.ConditionFalse, t1, t1),
+			expectedResultRun:     1,
+		},
+		"ClusterFirstProbeNotReady": {
+			clusterStatus: clusterStatus(corev1.ConditionFalse, t3, t3),
+			storedClusterData: &ClusterData{
+				clusterStatus:   clusterStatus(corev1.ConditionTrue, t2, t1),
+				lastProbeStatus: clusterStatus(corev1.ConditionTrue, t2, t1),
+				resultRun:       1000},
+			expectedClusterStatus: clusterStatus(corev1.ConditionTrue, t3, t1),
 			expectedResultRun:     1,
 		},
 		"ClusterNotReadyButWithinFailureThreshold": {
 			clusterStatus: clusterStatus(corev1.ConditionFalse, t3, t3),
 			storedClusterData: &ClusterData{
-				clusterStatus: clusterStatus(corev1.ConditionFalse, t2, t1),
-				resultRun:     1},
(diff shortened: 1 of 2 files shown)


RainbowMango:
cc @irfanurrehman @hectorj2f 
Could you please take a look?

irfanurrehman:
@FengXingYuXin Thanks for doing this. Will it be possible for you to add some kind of a test for this change?

irfanurrehman:
> @FengXingYuXin Thanks for doing this. Will it be possible for you to add some kind of a test for this change?

@FengXingYuXin nevermind, there are unit tests for this and they seem to fail with your change. Please take a look.

FengXingYuXin:
> > @FengXingYuXin Thanks for doing this. Will it be possible for you to add some kind of a test for this change?
> 
> @FengXingYuXin nevermind, there are unit tests for this and they seem to fail with your change. Please take a look.

@irfanurrehman  thanks for your reply and remind, and I have fix the ut cases, please check it in your own good time.

pkg/controller/kubefedcluster/controller_test.go
```diff
@@ -65,17 +65,17 @@ func TestThresholdCheckedClusterStatus(t *testing.T) {
 		"ClusterNotReadyButWithinFailureThreshold": {
 			clusterStatus: clusterStatus(corev1.ConditionFalse, t3, t3),
 			storedClusterData: &ClusterData{
-				clusterStatus: clusterStatus(corev1.ConditionTrue, t2, t1),
-				resultRun:     2},
-			expectedClusterStatus: clusterStatus(corev1.ConditionTrue, t3, t1),
-			expectedResultRun:     3,
+				clusterStatus: clusterStatus(corev1.ConditionFalse, t2, t1),
```
irfanurrehman, line 68: This is a valid test that you are changing. The test checks that the cluster is currently ready, but cluster status check is saying that there is a problem with the cluster. It additionally states that the check has happened two times before and this is the third time we are checking. The result is as expected with existing code and existing test. The cluster status should still remain `ready=true`.

With your code change this test is failing.

pkg/controller/kubefedcluster/controller_test.go
```diff
@@ -65,17 +65,17 @@ func TestThresholdCheckedClusterStatus(t *testing.T) {
...
+			expectedClusterStatus: clusterStatus(corev1.ConditionFalse, t3, t1),
+			expectedResultRun:     2,
 		},
 		"ClusterNotReadyAndCrossedFailureThreshold": {
 			clusterStatus: clusterStatus(corev1.ConditionFalse, t4, t4),
 			storedClusterData: &ClusterData{
 				clusterStatus: clusterStatus(corev1.ConditionTrue, t3, t1),
 				resultRun:     3},
-			expectedClusterStatus: clusterStatus(corev1.ConditionFalse, t4, t4),
+			expectedClusterStatus: clusterStatus(corev1.ConditionTrue, t4, t1),
```
irfanurrehman, line 78: Here also you are changing a valid test. Extending to previous test and it explaination above. When the check is done 4th time (with the failureThreshold as 3) The cluster state changes from `ready=true` to `ready=false`. Also the transition time is also updated to t4 because that is the time of the cluster state transition.

irfanurrehman (changes requested):
@FengXingYuXin can you please ensure that the existing tests pass, rather then changing their meaning. I think the existing tests are valid tests.
You can add new tests for the scenario you have mentioned in your bug.

pkg/controller/kubefedcluster/controller_test.go
FengXingYuXin: @irfanurrehman  thanks a lot for your detailed explanation and I think your suggestion is very good.

FengXingYuXin:
@irfanurrehman  about the moment of status translates, for example, if fail threshold is 3, as I understand it, if it's better for translating status from ready to notReady when probe failed status on the third time continuously, now the existing code translates status on the 4th time. If you agree with it, I can adjust it later.

PR #1501, opened 2022-04-08
feat: introduce Kubernetes version to status of KubeFedCluster

zqzten:
There're cases where higher level systems or administrators of KubeFed need to know the Kubernetes version of managed clusters. For example we are managing some clusters with a variety of Kubernetes version and need to use different API versions for some kind of objects (such as Ingress) as template for federated objects. This PR introduces Kubernetes version information to KubeFedCluster which would be helpful for those cases.

This PR also includes some other handy tweaks:
* record the wating for removal information as event to help user easily find out which clusters' objects are pending deleting without resorting to controller logs
*  utilize field constants in code


zqzten:
The golangci-lint in test is breaking on Go 1.18, maybe we should pin the test version to Go 1.16 according to the project Go version

makkes (commented):
In general I'd like to see this change being included in kubefed, thanks for coming up with it. Some questions/requests remain after reviewing the PR:

1. How and when are K8s upgrades reflected on the KubefedCluster object? I'd like to see this documented in [the user guide](docs/userguide.md) and maybe in a [KEP](docs/keps).
2. Please make sure to add tests for the new behaviour.

zqzten:
@makkes Thanks for the comment. I have added the required description in the cluster-registration part of user guide, this behavior is trivial so I think it doesn't deserve a full KEP for now. Test has been added with some other tweak and fix BTW, please take another look.

.github/workflows/build-and-test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
-          go-version: '^1.16.6'
+          go-version: '~1.16'
```
makkes, line 14: not sure if we want the Go version to be a moving target. Is this necessary for any of the other changes in this PR to work?

/cc @hectorj2f

charts/kubefed/charts/controllermanager/crds/crds.yaml
```diff
@@ -457,6 +457,8 @@ spec:
     kind: KubeFedCluster
     listKind: KubeFedClusterList
     plural: kubefedclusters
+    shortNames:
+    - kfc
```
makkes, line 461: I'd prefer this PR to focus on the single goal of adding K8s version to the CRD's status.

cmd/controller-manager/app/leaderelection/leaderelection.go
```diff
@@ -36,8 +36,9 @@ import (
 
 func NewKubeFedLeaderElector(opts *options.Options, fnStartControllers func(*options.Options, <-chan struct{})) (*leaderelection.LeaderElector, error) {
 	const component = "kubefed-controller-manager"
-	restclient.AddUserAgent(opts.Config.KubeConfig, "kubefed-leader-election")
-	leaderElectionClient := kubeclient.NewForConfigOrDie(opts.Config.KubeConfig)
+	kubeConfig := restclient.CopyConfig(opts.Config.KubeConfig)
```
makkes, line 39: Why is it necessary to create a copy of this object?

pkg/controller/sync/controller.go
```diff
@@ -113,9 +114,10 @@ func newKubeFedSyncController(controllerConfig *util.ControllerConfig, typeConfi
 	federatedTypeAPIResource := typeConfig.GetFederatedType()
 	userAgent := fmt.Sprintf("%s-controller", strings.ToLower(federatedTypeAPIResource.Kind))
 
-	// Initialize non-dynamic clients first to avoid polluting config
-	client := genericclient.NewForConfigOrDieWithUserAgent(controllerConfig.KubeConfig, userAgent)
-	kubeClient := kubeclient.NewForConfigOrDie(controllerConfig.KubeConfig)
+	kubeConfig := restclient.CopyConfig(controllerConfig.KubeConfig)
```
makkes, line 117: same question as above: Why do we need to create the copy here?

makkes (commented):
some small adjustments, otherwise I'm fine with merging it.

.github/workflows/build-and-test.yml
zqzten: `^1.16.6` means `< 2.0`, `~1.16` means `< 1.17`, so actually I'm limiting the range rather then expand it. I made this change because the current latest Go 1.x release (1.18) is breaking the golangci-lint in test.

charts/kubefed/charts/controllermanager/crds/crds.yaml
zqzten: Just found the `kubefedcluster` is a little verbose in everyday use lol. I prefer to keep it but if you don't like it I can remove it or seperate to another PR, no big deal.

pkg/controller/sync/controller.go
zqzten: ditto

irfanurrehman:
@zqzten nice work and a set of welcome changes.. :-) Appreciate it.
/lgtm

PR #1506, opened 2022-05-25
Introduce standard controller runtime metrics

zqzten:
Controller runtime provides some [useful metrics for controllers](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/metrics/metrics.go). However, since KubeFed does not directly use controller runtime to build controllers (for some "federated" reasons), it lacks these standard metrics. This PR introduces these metrics to KubeFed and deprecates some old non-standard ones.


zqzten:
Friendly ping @irfanurrehman @jimmidyson for review

irfanurrehman:
/lgtm
/approve
Thanks @zqzten

PR #1515, opened 2022-08-05
fix: Use specific service account token secret for join

jimmidyson:
With `LegacyServiceAccountTokenNoAutoGeneration` feature graduating to beta in
k8s v1.24, a token is not automatically generated when a service account is created.
This commit fixes this by requesting a specific service token account secret which
is populated by the token controller. This is backwards compatible with previous
versions of Kubernetes.

### DIFF
--- a/pkg/kubefedctl/join.go
+++ b/pkg/kubefedctl/join.go
@@ -37,6 +37,7 @@ import (
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	fedv1b1 "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
@@ -417,8 +418,8 @@ func createKubeFedNamespace(clusterClientset kubeclient.Interface, kubefedNamesp
 	return fedNamespace, nil
 }
 
-// createAuthorizedServiceAccount creates a service account and grants
-// the privileges required by the KubeFed control plane to manage
+// createAuthorizedServiceAccount creates a service account and service account token secret
+// and grants the privileges required by the KubeFed control plane to manage
 // resources in the joining cluster.  The name of the created service
 // account is returned on success.
 func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface,
@@ -435,6 +436,16 @@ func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface
 	}
 
 	klog.V(2).Infof("Created service account: %s in joining cluster: %s", saName, joiningClusterName)
+
+	secretName, err := createServiceAccountTokenSecret(saName, joiningClusterClientset, namespace,
+		joiningClusterName, hostClusterName, dryRun, errorOnExisting)
+	if err != nil {
+		klog.V(2).Infof("Error creating service account: %s in joining cluster: %s due to: %v",
+			saName, joiningClusterName, err)
+		return "", err
+	}
+
+	klog.V(2).Infof("Created service account token secret: %s in joining cluster: %s", secretName, joiningClusterName)
 
 	if scope == apiextv1.NamespaceScoped {
 		klog.V(2).Infof("Creating role and binding for service account: %s in joining cluster: %s", saName, joiningClusterName)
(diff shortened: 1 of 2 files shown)


pkg/kubefedctl/join.go
```diff
@@ -508,6 +523,43 @@ func createServiceAccount(clusterClientset kubeclient.Interface, namespace,
 	default:
 		return saName, nil
 	}
+} // createServiceAccount creates a service account in the cluster associated
```
makkes, line 526: this doesn't look right.
```suggestion
}

// createServiceAccount creates a service account in the cluster associated
```

pkg/kubefedctl/util/util.go
```diff
@@ -107,6 +107,12 @@ func ClusterServiceAccountName(joiningClusterName, hostClusterName string) strin
 	return fmt.Sprintf("%s-%s", joiningClusterName, hostClusterName)
 }
 
+// ClusterServiceAccountTokenName returns the name of a service account token secret whose
```
makkes, line 110:
```suggestion
// ClusterServiceAccountTokenSecretName returns the name of a service account token secret whose
```

pkg/kubefedctl/join.go
```diff
@@ -436,6 +437,16 @@ func createAuthorizedServiceAccount(joiningClusterClientset kubeclient.Interface
 
 	klog.V(2).Infof("Created service account: %s in joining cluster: %s", saName, joiningClusterName)
 
+	secretName, err := createServiceAccountTokenSecret(saName, joiningClusterClientset, namespace,
+		joiningClusterName, hostClusterName, dryRun, errorOnExisting)
+	if err != nil {
+		klog.V(2).Infof("Error creating service account: %s in joining cluster: %s due to: %v",
```
makkes, line 443:
```suggestion
		klog.V(2).Infof("Error creating service account token secret: %s in joining cluster: %s due to: %v",
```

pkg/kubefedctl/join.go
```diff
@@ -841,32 +893,20 @@ func populateSecretInHostCluster(clusterClientset, hostClientset kubeclient.Inte
 	// Get the secret from the joining cluster.
 	var secret *corev1.Secret
 	err := wait.PollImmediate(1*time.Second, serviceAccountSecretTimeout, func() (bool, error) {
-		sa, err := clusterClientset.CoreV1().ServiceAccounts(joiningNamespace).Get(
+		joiningClusterSASecret, err := clusterClientset.CoreV1().Secrets(joiningNamespace).Get(
 			context.Background(), saName, metav1.GetOptions{},
```
makkes, line 897: Isn't the Secret's name derived by calling `util.ClusterServiceAccountTokenSecretName`?

```suggestion
			context.Background(),  util.ClusterServiceAccountTokenSecretName(joiningClusterName, hostClusterName), metav1.GetOptions{},
```

makkes (changes requested):
It feels like the PR is doing more than what the title implies in that it generally aligns kubefed with K8s 1.24.

I'd prefer for the PR title and description to clarify what this is about.

Otherwise lgtm except some smaller remarks.

jimmidyson:
Thanks for the review @makkes!

> It feels like the PR is doing more than what the title implies in that it generally aligns kubefed with K8s 1.24.

You're right, I did that to test, but I will separate that out into a different PR and instead try just enabling the `LegacyServiceAccountTokenNoAutoGeneration` feature gate in KinD tests to ensure that the functionality works and we don't get false positives in our integration tests. Does that sound OK to you?

jimmidyson:
Ah I can't enable `LegacyServiceAccountTokenNoAutoGeneration` in previous versions because it was introduced in v1.24 in beta and therefore enabled by default. I suppose what I could do is still take out the 1.24 upgrade stuff in this PR and run with existing tests to check there's no regression - that's what I'll do now.

makkes:
> Ah I can't enable `LegacyServiceAccountTokenNoAutoGeneration` in previous versions because it was introduced in v1.24 in beta and therefore enabled by default. I suppose what I could do is still take out the 1.24 upgrade stuff in this PR and run with existing tests to check there's no regression - that's what I'll do now.

That sounds perfect. Thanks so much!

jimmidyson:
@makkes Done as discussed.

pkg/kubefedctl/join.go
```diff
@@ -510,6 +526,45 @@ func createServiceAccount(clusterClientset kubeclient.Interface, namespace,
 	}
 }
 
+// createServiceAccount creates a service account in the cluster associated
```
makkes, line 529:
```suggestion
// createServiceAccountTokenSecret creates a service account in the cluster associated
```

makkes (commented):
lgtm except the doc change

jimmidyson:
@makkes Done for reals this time :joy:

makkes (approved):
🥳 Thank you!!

PR #1516, opened 2022-08-08
build: Use k8s v1.24 for tests

jimmidyson:


charts/kubefed/templates/federatedtypeconfig.yaml
```diff
@@ -58,7 +58,7 @@ spec:
 apiVersion: core.kubefed.io/v1beta1
 kind: FederatedTypeConfig
 metadata:
-  name: ingresses.extensions
```
jimmidyson, line 61: `extensions/v1beta1.Ingress` was deprecated in k8s v1.14 and `networking.k8s.io/v1.Ingress` is available from k8s v1.19 so kubefed will still support ingresses back to v1.19 - this is an acceptable support window considering v1.24 is now out.

jimmidyson:
Splitting into multiple PRs for easier review.

charts/kubefed/templates/federatedtypeconfig.yaml
makkes: We should make sure to properly communicate this breaking change.

scripts/download-binaries.sh
```diff
@@ -43,14 +43,11 @@ mkdir -p "${dest_dir}"
...
-kb_url="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb_version}/${kb_tgz}"
-curl "${curl_args}" "${kb_url}" \
-  | tar xzP -C "${dest_dir}" --strip-components=2
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
-export KUBEBUILDER_ASSETS="${dest_dir}"
-echo "Setting to KUBEBUILDER_ASSETS ${dest_dir}"
+source <(setup-envtest use -p env 1.24.x)
+
+echo "KUBEBUILDER_ASSETS is set to KUBEBUILDER_ASSETS"
```
makkes: Is this missing a `$` sign somewhere?

scripts/download-binaries.sh
```diff
@@ -43,14 +43,11 @@ mkdir -p "${dest_dir}"
...
-kb_version="2.3.1"
-kb_tgz="kubebuilder_${kb_version}_${platform}_amd64.tar.gz"
-kb_url="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb_version}/${kb_tgz}"
-curl "${curl_args}" "${kb_url}" \
-  | tar xzP -C "${dest_dir}" --strip-components=2
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
-export KUBEBUILDER_ASSETS="${dest_dir}"
-echo "Setting to KUBEBUILDER_ASSETS ${dest_dir}"
+source <(setup-envtest use -p env 1.24.x)
```
makkes, line 48: Why do we need to setup envtest here?

charts/kubefed/templates/federatedtypeconfig.yaml
jimmidyson: I'm not sure if this is a breaking change :thinking: Anyone using a newer version of Kubernetes will not be able to propagate ingress without this change... the breaking change is in the deletion of the API in Kubernetes, this is actually a fix!

jimmidyson: Pulled this change out into #1521 to keep the PRs small and focused.

jimmidyson:
/hold

Blocking until #1521 is merged.

charts/kubefed/templates/federatedtypeconfig.yaml
irfanurrehman: I tend to agree with @jimmidyson. This is not a breaking change, but may be worth documenting anyways. @jimmidyson you also mentioned that you will cut a new release post this. It should ideally be sufficient to explicitly point this out in the release notes.

test/common/fixtures/ingresses.networking.k8s.io.yaml
```diff
@@ -0,0 +1,8 @@
+kind: fixture
```
irfanurrehman, line 1: perhaps a rebase needed given https://github.com/kubernetes-sigs/kubefed/pull/1521 is merged.

jimmidyson: :+1: that's why i blocked it, thanks for reminding me.

jimmidyson: Done.

makkes (approved):

charts/kubefed/templates/federatedtypeconfig.yaml
jimmidyson: @irfanurrehman I'll make sure to do that.

PR #1518, opened 2022-08-09
build: Use KinD for syncing helm chart resources

jimmidyson:
Previously this was using kubebuilder provided API server and etcd, but these were very out of date (k8s v1.16), leading to potentially out of date resources and difficulties upgrading to later APIs that were not available in v1.16 (e.g. `networking.k8s.io/v1`).

This commit fixes this by using KinD as the cluster to discover resources to
sync to the Helm chart.


makkes (commented):
I'm not sure if any of the documentation is affected by this and I don't have the time to verify it. If you would mind checking this @jimmidyson, please? Otherwise lgtm.

jimmidyson:
Hi @makkes. This doesn't affect the docs but it reminds me that the next PR does so I'll update docs in there.

makkes (approved):

PR #1519, opened 2022-08-09
build: Use setup-envtest to download later version of envtest

jimmidyson:
`controller-runtime` provides a convenient tool to download various versions of k8s. This commit switches to using that to ensure we use a single version for testing in unit tests and in e2e tests, namely v1.21.x right now so to keep the version stable while just changing the installation mechanism.

### DIFF
--- a/scripts/download-binaries.sh
+++ b/scripts/download-binaries.sh
@@ -49,14 +49,28 @@ kb_url="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb_v
 curl "${curl_args}" "${kb_url}" \
   | tar xzP -C "${dest_dir}" --strip-components=2
 
-export KUBEBUILDER_ASSETS="${dest_dir}"
-echo "Setting to KUBEBUILDER_ASSETS ${dest_dir}"
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+source <(setup-envtest use -p env 1.21.x)
+
+echo "KUBEBUILDER_ASSETS is set to ${KUBEBUILDER_ASSETS}"
 
 helm_version="3.6.0"
 helm_tgz="helm-v${helm_version}-${platform}-amd64.tar.gz"
 helm_url="https://get.helm.sh/$helm_tgz"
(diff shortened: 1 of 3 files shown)


scripts/download-binaries.sh
```diff
@@ -49,15 +49,29 @@ kb_url="https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb_v
 curl "${curl_args}" "${kb_url}" \
   | tar xzP -C "${dest_dir}" --strip-components=2
 
-export KUBEBUILDER_ASSETS="${dest_dir}"
-echo "Setting to KUBEBUILDER_ASSETS ${dest_dir}"
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+source <(setup-envtest use -p env 1.21.x)
```
makkes, line 53: same question as in #1516: 

Why do we need to setup envtest here?

makkes (commented):
one question, otherwise lgtm

scripts/download-binaries.sh
jimmidyson: kubebuilder releases used to bundle `kubebuilder` binary along with `kube-apiserver`, `etcd`, `kubectl`, etc. These stopped happening a long time ago, with version 2.3.2 iirc of kubebuilder (the version used in this project) bundling k8s v1.16 - too old for us now. Recent versions of kubebuilder no longer do that, instead distributing k8s binaries per version. Although you can define the URL to download directly, using `setup-envtest` is a convenience that allows you to not /discover know the URL and change the minor version to test against as this PR now does. It's just to make things easier tbh...

jimmidyson: oh you mean why here! Well strictly we don't I guess could remove it but it is just replacing what was here before downloading kubebuilder provided kube-apiserver, etc. I don't mind removing it if you think it's best to though.

makkes: Ah, I see. Yeah, with this line the behaviour would be retained. I'm fine either way.

makkes (approved):

PR #1521, opened 2022-08-09
fix: Move ingress from extensions/v1beta1 to networking.k8s.io/v1

jimmidyson:
`extensions/v1beta1` was deprecated in v1.14,
`networking.k8s.io/v1beta1` was deprecated in v1.19 so this commit moves
straight to `networking.k8s.io/v1` which has been available since v1.19.
Considering k8s is now at v1.24.3 this is a sufficient compatibility
window.


irfanurrehman:
Thanks for doing this @jimmidyson 
/lgtm
/approve

